PR #2216, opened 2017-05-17
Replace ">" with "-" for [LOCATION]

gitdevmod:
When using [LOCATION] for directory it does work with sub directory because ">" is not allowed in directory name in Windows.
(example: file:///X:/APP/[LOCATION]/[NAME])

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | ?
| Deprecations? | ?
| Tests pass?   | yes
| Fixed tickets |

### DIFF
--- a/inc/link.class.php
+++ b/inc/link.class.php
@@ -299,6 +299,7 @@ class Link extends CommonDBTM {
             $link = str_replace("[LOCATION]",
                                 Dropdown::getDropdownName("glpi_locations",
                                                           $item->getField('locations_id')), $link);
+            $link = str_replace(">", "-", $link);
       }
       if (strstr($link, "[NETWORK]")
           && $item->isField('networks_id')) {


orthagh:
Hello @gitdevmod 

LOCATION tag in link class aim to represent real location (places) not windows directories.
It seems you are misappropriating the original purpose of Location object and your fix adresses a workaround for that "bad" behavior.

I can't accept this pr so.

Could you explain why you store directories in locations object ?

gitdevmod:
Hi @orthagh,

I understand, I use this for network switches. I store information (documentation, backup, etc.) in a windows network directory for each switch ie: x:\bureau-1\switch42\, it works only with location without child.
I can attach document in glpi but I think it's more easy to maintain document in a directory and open the correct directory from the switch glpi page link tab.

trasher:
I'm closing this one since this is not the usage of locations; cannot be accepted

Issue #1086, opened 2016-09-28
Search engine for ticket actors


yllen:
http://forum.glpi-project.org/viewtopic.php?id=155498

yllen:
The issue is for all actors dropdown (requester, observer or technician)

vinz38500:
The issue is the same for 9.1 ...

orthagh:
Regression in e04db09
I'm currently investigating

vihuarar:
I already made the changes in 0f50552 with my user (super user) works fine.. but for some users that have "Technical" profile dont work...

PR #2241, opened 2017-05-26
Fix dropdown users when use entity. Get from branch 9.1

ddurieux:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | What is BC???
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #number


orthagh:
The pr content is OK but why 0.90 ? we won't release a version for this branch ?

orthagh:
See #1086 and 0689cd672db62b320802baa745efbc6d71e984d8

orthagh:
Also BC : Breaking change

PR #2250, opened 2017-05-30
split token for api

### LINKED ISSUE #2129: api, use dedicated token

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes (api user's tokens will need to be reset)
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2129


inc/user.class.php
```diff
@@ -497,11 +497,16 @@ function isEmail($email) {
...
     * @param $token user token
+    * @param string $field the field storing the token
     *
     * @return true if succeed else false
    **/
-   function getFromDBbyToken($token) {
-      return $this->getFromDBByQuery("WHERE `".$this->getTable()."`.`personal_token` = '$token'");
+   function getFromDBbyToken($token, $field = 'personal_token') {
+      if (!in_array($field, ['personal_token', 'api_token'])) {
+         return false;
```
trasher: Maybe add something in the logs in this case? This sounds like an error to request something that does not exists.

Issue #2262, opened 2017-06-02
Warning when  geenrating a user token


btry:
when GLPI generates a user token, the code uses an undefined variable **$field_date**. This afects the date of generation of the token.

btry:
fixed

PR #2263, opened 2017-06-02
fix warning when generating a user token

btry:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | wait a few minutes
| Fixed tickets | #2262 

see teh ticket.

### DIFF
--- a/inc/user.class.php
+++ b/inc/user.class.php
@@ -4275,9 +4275,9 @@ class User extends CommonDBTM {
             return $user->fields[$field];
          }
          $token = self::getUniqueToken($field);
-         $user->update(array('id'          => $user->getID(),
-                             $field        => $token,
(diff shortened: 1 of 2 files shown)


btry:
good idea to add unit tests by the way.

trasher:
Some (late) comments:
- you should not use hardcoded user id (but get TU_SER instead),
- this test can be run only once since ganerated values are not reset.

I've fixed this in my atoum workbranch, see https://github.com/trasher/glpi/commit/9bd7cbc37651f70fb26a77e529fd85f22d6c06d9 ;)

btry:
No problem

Issue #2284, opened 2017-06-12
PHP notice


yllen:
*** PHP Notice(8): Undefined index: notifications_mailing

yllen:
Resolved after two cliupdate

PR #2287, opened 2017-06-12
fix PHP notice

yllen:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2284


trasher:
Hello,

I do not agree with the fix. If this indeed fix the notice; your should have this entry in the configuration, or there is something wrong (this is present in the install file and in the update file).

Are you sure your database is up-to-date?

yllen:
I force a new cliupdate 2 times.
After the second time errors deseapeared

Issue #918, opened 2016-09-02
Operating system informations in DB


trasher:
Hello,

Some informations on operating systems have been added in the FusionInventory plugin, I guess that should be done in core. It also rework the database structure, exporting OS informations into several tables.

I've reported the issue on the plugin side:
https://github.com/fusioninventory/fusioninventory-for-glpi/issues/1920

Since it will be required to change the database structure, it can't reach 9.1; but maybe one of the future releases.

remicollet:
How many new fields are needed to store the information retrieved by FusionInventory ?

If only 1-2 new dropdown, IMHO, seems something which can be fixed in 9.1 to avoid to much change and facilitate coordination between both projects.

ddurieux:
No, we are in RC, so not modify DB structure or add features now...

wawax:
Hi David.

Could you explain what is the need for these new fields ? What are the uses cases ? What does it brings ?
Do we need to take it in account in rules engines and reports ?

Thank you :)

ddurieux:
See specs here: https://github.com/fusioninventory/fusioninventory-for-glpi/issues/1295#issuecomment-222272332

wawax:
@tsmr what do you think about this discussion ? 
Are these new informations relevant in OCS case ?

tsmr:
In XML from OCS, for example : 
```
* OSNAME : Microsoft Windows 10 Entreprise
* OSVERSION : 10.0.10240
* ARCH : x86 64 bit
```
So OCSNAME contains already the two fields (OS name / OS edition) & so i need to explode the original data for use two fields if this is the goal of this thread.

Kernel name & version are not used by OCS, i think.

PR #2298, opened 2017-06-15
Feature/os

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes (kind of)
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #918


trasher:
Current PR may have missed stats (and it certainly miss unit tests).

Some plugins may been impacted:
- PDF,
- reports,
- mreporting.

Issues wll be opened on those ones once BC will be confirmed. If you know other parts or plugins that can be impacted, please let me know!

wawax:
@tsmr take care of this PR, as you may need to do some modifications on your plugins !

inc/computer.class.php
```diff
@@ -72,7 +72,7 @@ static function getTypeName($nb=0) {
     *
     * @since version 9.1
    **/
-   function getTabNameForItem(CommonGLPI $item, $withtemplate=0) {
```
orthagh: Could we remove entirely this part ?

inc/item_operatingsystem.class.php
```diff
@@ -0,0 +1,379 @@
...
+ * along with GLPI. If not, see <http://www.gnu.org/licenses/>.
+ * ---------------------------------------------------------------------
+**/
+class Item_OperatingSystem extends CommonDBRelation {
+
+   static public $itemtype_1 = 'OperatingSystem';
+   static public $items_id_1 = 'operatingsystems_id';
+   static public $itemtype_2 = 'itemtype';
+   static public $items_id_2 = 'items_id';
+   static public $checkItem_2_Rights = self::DONT_CHECK_ITEM_RIGHTS;
```
orthagh: I'm not sure about this right flag.
I think revert to the other side the right could be better and make more sense. aka:
``` static public $checkItem_1_Rights = self::DONT_CHECK_ITEM_RIGHTS;```

I think, when you add an os to a computer, you want to check if you have right on the computer, not the right on the os.

File: inc/item_operatingsystem.class.php [hunk 3]
```diff
@@ -0,0 +1,379 @@
...
+
+   //static protected $notable = false;
```
orthagh: remove this line

inc/item_operatingsystem.class.php
```diff
@@ -0,0 +1,379 @@
...
+
+   static function getTypeName($nb=0) {
+      return _n('Item operating system', 'Item operating systems', $nb);
+   }
+
+
+   static function countForItem(CommonDBTM $item) {
```
orthagh: Missing docblock (i didn't find declaration in parents)

inc/item_operatingsystem.class.php
```diff
@@ -0,0 +1,379 @@
...
+                        "<a href='javascript:reloadTab(\"sort=$key&amp;order=".
+                          (($order == "ASC") ?"DESC":"ASC")."&amp;start=0\");'>$val</a></th>";
+      }
+
+      $header_end .= "</tr>";
+      echo $header_begin.$header_top.$header_end;
+
+      if ($number) {
+         foreach ($os as $data) {
+            $link         = NOT_AVAILABLE;
```
orthagh: overrided l249

inc/item_operatingsystem.class.php
```diff
@@ -0,0 +1,379 @@
...
+            $assocID      = $data["assocID"];
```
orthagh: ``$assocID`` is not used

inc/item_operatingsystem.class.php
```diff
@@ -0,0 +1,379 @@
...
+      echo "<tr class='headerRow'><th colspan='".$colspan."'>";
+      echo __('Operating system');
+      echo Html::hidden('itemtype', ['value' => $this->fields['itemtype']]);
+      echo Html::hidden('items_id', ['value' => $this->fields['items_id']]);
+      echo "</th></tr>";
+
+      echo "<tr class='tab_bg_1'>";
+      echo "<td><label for='dropdown_operatingsystems_id$rand'>".__('Name')."</label></td>";
+      echo "<td>";
+      OperatingSystem::dropdown(array('value' => $this->fields["operatingsystems_id"], 'rand' => $rand));
```
orthagh: Here and the following dropdowns, could we use the short notation (```[...]```) for arrays like the other further in this function ?

install/update_91_92.php
```diff
@@ -866,6 +872,146 @@ function update91to92() {
...
+                  OR os_kernel_version IS NOT NULL
+                  OR os_licenseid IS NOT NULL";
+      $DB->queryOrDie($query, "9.2 migrate main operating system informations");
+
+      //migrate kernel versions.
+      $query = "SELECT id, os_kernel_version FROM glpi_computers WHERE os_kernel_version IS NOT NULL";
+      if ($result = $DB->query($query)) {
+         if ($DB->numrows($result)>0) {
+            $mapping = [];
+            $kver = new \OperatingSystemKernelVersion();
```
orthagh: namespace

inc/item_operatingsystem.class.php
```diff
@@ -0,0 +1,379 @@
...
+      }
+
+      $canedit = $item->canEdit($item->getID());
+
+      if ($canedit && $numrows >= 1
+          && !(!empty($withtemplate) && ($withtemplate == 2))) {
+         echo "<div class='center firstbloc'>".
+            "<a class='vsubmit' href='" . Toolbox::getItemTypeFormURL(self::getType()) . "?items_id=" . $item->getID() .
+            "&amp;itemtype=" . $item->getType() . "&amp;withtemplate=" . $withtemplate."'>";
+         echo __('Add an operating system');
```
trasher: I've added the button Since the new database schema permit that, but this is not something I've planned to implement in the FI plugin. We can hide the button if you prefer, no problem.
Multiboot is not the norm anyways.

inc/define.php
```diff
@@ -266,7 +266,9 @@
                                                   'OperatingSystemVersion', 'PeripheralModel',
                                                   'PeripheralType', 'PhoneModel', 'PhoneType',
                                                   'Printer', 'PrinterModel', 'PrinterType',
-                                                  'Software', 'OperatingSystemArchitecture');
+                                                  'Software', 'OperatingSystemArchitecture',
+                                                  'OperatingSystemKernel', 'OperatingSystemKernelVersion',
+                                                  'OperatingSystemEdition');
```
trasher: 👍 for a new parameter.

trasher:
Item_OperatingSystem class is ready to be linked with other itemtypes; but is linked with computers only right now

inc/define.php
orthagh: same opinion as @trasher, tabs don't need any aditionnal var

PR #2200, opened 2017-05-11
see #2198

yllen:
Add new hook

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2198

*Please update this template with something that matches your PR*

PR #2304, opened 2017-06-16
Add new hooks to see computers of my group

trasher:
Backport for #2200 into master


inc/commondbtm.class.php
```diff
@@ -2581,13 +2581,13 @@ function can($ID, $right, array &$input=NULL) {
 
       /* Hook to restrict user right on current item @since 9.2 */
       $this->right = $right;
+      plugin::doHook('can', $this);
```
trasher: I still wonder if this new hook is OK (see https://github.com/glpi-project/docdev/pull/50)... If not, it should be renamed in the 9.1 branch.

PR #1538, opened 2017-01-06
Add form hooks on helpdesk.public.php

trasher:
Follows #1444

PR #2305, opened 2017-06-16
Rename new hook

trasher:
Hook already exists with another name in the master branch. Since this one is new, it has to be renamed. See https://github.com/glpi-project/glpi/pull/2304/commits/4958e8be79f5deef6bb8e55b54095e6284cc920d 

@yllen take care if you already used 'can' hook ;)

### DIFF
--- a/inc/commondbtm.class.php
+++ b/inc/commondbtm.class.php
@@ -2484,7 +2484,7 @@ class CommonDBTM extends CommonGLPI {
       }
     
       $this->right = $right;
-      plugin::doHook('can', $this);
+      plugin::doHook('item_can', $this);
       switch ($this->right) {
          case READ :
             // Personnal item


yllen:
Johan,

I don't saw Remi change also in master.
But with 
https://github.com/glpi-project/glpi/commit/cf911a8b286f34c17b2a3bcf681b7107d0130a17#diff-7f569484c496f62a6ccf54647fed2b1b
the name is can in the 9.1/bugfixes, so it also must be can in master 
branch.

Regards,

trasher:
> I don't saw Remi change also in master.

Well, that's why new features should only reach master :)

>  the name is can in the 9.1/bugfixes, so it also must be can in master

I do not agree; the one in master exists for a while and is documented; while the one in 9.1/bf is really new, (and is not released). The impact of the change would be more important on master.

Also, all similar hook contains `item` in their name; it sounds more logical this one is named consistently.

yllen:
I need this hook in 9.1.x

I don't understand why this hook must only be in master.
This new hook don't change the database structure, or the glpi display.

Only a teclib agent can add new hook in current correct version? like 
https://github.com/glpi-project/glpi/pull/1538 ?

Regards,

trasher:
I've never said new hook cannot be added in 9.1. Your hook has been added, no?

What I say, is new features should be on master only, the initial hook was not needed in 9.1 when it's been added, and have been added to master, as a "standard" new feature...

Please don't move the debate; we're talking about the hook name here. That did not depend of who is employing me or not.

yllen:
Le 19/06/2017 à 09:55, Johan Cwiklinski a écrit :
>
> I've never said new hook cannot be added in 9.1. Your hook has been 
> added, no?
>
No it was not added in 9.1.4 release.
>
> What I say, is new features should be on master only, the initial hook 
> was not needed in 9.1 when it's been added, and have been added to 
> master, as a "standard" new feature...
>
Remi done this commit forhelp me because i was not at home at this 
moment and do it to be in the next major version if release. So he 
respects rules of GLPI.

But when i saw that you can add a hook in a corrective version like for 
9.1.2 and this hook was needed by ministries, i add it and don't saw 
Remi already done it.
Like i said in MIMO this hook will be in 9.1.x version. I will said them 
to stay in 0.85.x still and wait for 9.2
>
> Please don't move the debate; we're talking about the hook name here. 
> That did not depend of who is employing me or not.
>
I don't move the debate, i simply notice that a hook can be done in 
9.1.2 by a teclib agent but not in 9.1.4 by me.

trasher:
As discussed by phone; OK for the rename :)

Issue #2329, opened 2017-06-20
fix gettext, CS and W3C


yllen:
with https://github.com/glpi-project/glpi/issues/2316 check also in master

PR #2330, opened 2017-06-20
fix gettext, CS and W3C

yllen:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  |no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2329

### DIFF
--- a/inc/savedsearch.class.php
+++ b/inc/savedsearch.class.php
@@ -36,6 +36,8 @@ if (!defined('GLPI_ROOT')) {
 
 /**
  * Saved searches class
+ *
+ * @since version 9.2
 **/
 class SavedSearch extends CommonDBTM {
 
@@ -49,15 +51,19 @@ class SavedSearch extends CommonDBTM {
    const COUNT_YES = 1;
    const COUNT_AUTO = 2;
 
+
    static function getForbiddenActionsForMenu() {
       return array('add');
    }
+
 
    public static function getTypeName($nb = 0) {
       return _n('Saved search', 'Saved searches', $nb);
    }
 
+
    function getForbiddenStandardMassiveAction() {
+
       $forbidden   = parent::getForbiddenStandardMassiveAction();
       $forbidden[] = 'update';
       return $forbidden;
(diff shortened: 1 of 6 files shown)


inc/savedsearch.class.php
```diff
@@ -334,8 +332,8 @@ function cleanDBonPurge() {
    /**
     * Print the saved search form
     *
-    * @param integer $ID      ID of the item
-    * @param array   $options Options:
+    * @param $ID         integer    ID of the item
```
trasher: This does not follow phpdoc rules; see http://glpi-developer-documentation.readthedocs.io/en/master/codingstandards.html#parameters-documentation

inc/savedsearch.class.php
```diff
@@ -595,33 +584,19 @@ function getParameters($ID) {
...
-   /**
-    * Mark saved search as default view for the current user
-    *
-    * @param integer $ID ID of the saved search
-    *
-    * @deprecated since version 9.2; @see SavedSearch::markDefault()
-    *
-    * @return void
-   **/
-   function mark_default($ID) {
```
trasher: Deprecated methods should not be dropped in current version. But the ones in the current class are specific since the class itself has been renamed.

I guess keeping them may help to migrate anyways (and all methods deprecated in 9.1.2 should be removes as of 9.1..3)

inc/savedsearch.class.php
yllen: I follow GLPI CS rules.
If we change it, wez must change also for all comment of functions.

inc/savedsearch.class.php
yllen: It's not a deprecated method because this method doesn't exist in current version.
I think you create this function and after change this name. Savedsearch.class doesn't exist in 9.1.5

yllen: This function only exist in bookmark.class in 9.1.5.
If you change class name, function can't be used even deprecated

inc/savedsearch.class.php
trasher: Yes, I've begin to change all comments already. The link I gave you are the GLPI CS rules.

Old rules were wrong, the API output was missing type, and this does not respect the PHPDOC guidelines.

yllen: Ok in write it in my brain

yllen: Corrected

trasher: Great, thank you ;)

Issue #2322, opened 2017-06-19
Create an item from a template


yllen:
http://forum.glpi-project.org/viewtopic.php?id=158010

This problem is only for item created from a template and tabs of the template changed before creating item.

yllen:
Need to check all tabs of all template

yllen:
Check done

PR #2331, opened 2017-06-20
Read only tabs when creating an item from a template

yllen:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2322

### DIFF
--- a/inc/contract_item.class.php
+++ b/inc/contract_item.class.php
@@ -316,7 +316,7 @@ class Contract_Item extends CommonDBRelation{
 
       switch ($item->getType()) {
          case 'Contract' :
-            self::showForContract($item);
+            self::showForContract($item, $withtemplate);
 
          default :
             if (in_array($item->getType(), $CFG_GLPI["contract_types"])) {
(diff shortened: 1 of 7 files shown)


yllen:
Need to check all tab for all template

yllen:
Computer template done

yllen:
I think i checked every object creation with a template

inc/contract_item.class.php
```diff
@@ -524,11 +523,12 @@ static function showForItem(CommonDBTM $item, $withtemplate='') {
     *
     * @since version 0.84
     *
-    * @param $contract   Contract object
+    * @param           $contract       Contract object
```
trasher: Missing type

inc/ticket.class.php
```diff
@@ -5242,11 +5242,12 @@ static function showCentralNewList() {
     *
     * Will also display tickets of linked items
     *
-    * @param $item CommonDBTM object
+    * @param           $item          CommonDBTM object
```
trasher: Missing type

Issue #2308, opened 2017-06-16
[9.2] - runnung forced upgrade from 9.2 fails


btry:
Hi

If I upgrade from 9.2 to itself , the upgrade breaks:

the upgrade backs up **glpi_notifications_notificationtemplates** before trying to re-create it. Next it populates the re-created table with data from a table supposed from GLPI 9.1.x

```sql
INSERT INTO glpi_notifications_notificationtemplates (
  notifications_id, 
  mode, 
  notificationtemplates_id
) SELECT 
  id, 
  mode, 
  notificationtemplates_id 
FROM glpi_notifications
```
Unfortunately  **glpi_notifications** does not longer contains a column **mode** in 9.2

yllen:
Coorected https://github.com/glpi-project/glpi/pull/2335

PR #2335, opened 2017-06-21
fix upgrade

yllen:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2308


trasher:
This indeed fixes the issue thank you.

I guess we should not drop the `glpi_notifications_notificationtemplates` table also; all notifications config would be lost.

yllen:
Yes, we never delete an old table, just rename it backup_glpi_oldtable

Issue #2294, opened 2017-06-13
[9.1.2] "Associate Myself" doesn't trigger ticket rule


Miminoux:
Hi all,

Thx a lot for making GLPI so usefull !

I am experiencing a kinda bug :
1. create a rule with : 
  - criteria : 
    a. Technician exists
    b. Technician group does not exist
  - action : assign a specific group of your choice
2. create a simple ticket (any account)
3. with a tech account, click "Associate myself"
  - no group added
4. unassociate yourself
5. Use the "Assign to" + sign, User, find yourself and save
  - group is added

It seems that the associate myself button does not trigger the rules very like assigning yourself manunally does. I used to use the plugin "Behaviors" do automaticly assign tech groups but having groups in multiple entities seems to break the feature, So I tried to make a rule to kinda replace it. I tried to disable the plugin but it remains the same.

Can someone reproduce ?

Bye bye.

yllen:
Your rule is played on Add or on Update ticket?

lcpj:
Issue still persists on 9.1.5

Miminoux:
Sorry, I missed the e-mail notification, the rule is applied on ticket update.

I updated GLPI to 9.1.6 and the bug is still present.

PR #2337, opened 2017-06-21
fix autoassign

yllen:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2294


yllen:
Correction tests not goods.
User default right is set to 1 to set_default_assign.
So, CanAssignToMe is always False (already assigned)

Issue #2347, opened 2017-06-22
[9.2] CS for upgrade


yllen:
Don't use SELECT
at the begging save only existing table will change (as indicated in the comment)
Many //Yllen comment to review

Issue #2328, opened 2017-06-20
Mutliple foreign keys on DBIterator


trasher:
It is actually possible to do a native join on only 2 tables using DBIterator.

We often need to to more complex joins, like:

```
SELECT * FROM table1 
   JOIN table2
      ON table1.fid = table2.id
   JOIN table3
       ON table1.f2id = table3.id
```

or:
```
SELECT * FROM table1 
   JOIN table2
      ON table1.fid = table2.id
   JOIN table3
       ON table2.fid = table3.id
```

remicollet:
Simple proposal, allow FKEY1, FKEY2, FKEY3, ...

PR #2348, opened 2017-06-22
fix upgrade

yllen:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2328 #2347


remicollet:
+1 for DBIterator part

inc/dbmysql.class.php
```diff
@@ -821,6 +822,17 @@ function __construct ($dbconnexion, $table, $crit="", $debug=false) {
...
                      unset($crit[$key]);
                      break;
+
+                  case 'INNER JOIN' :
+                     if (is_array($val)) {
+                        foreach ($val as $jointable => $joincrit) {
+                           $join .= " INNER JOIN " .  self::quoteName($jointable) . " ON (" . $this->analyseCrit($joincrit) . ")";
+                        }
+                     } else {
+                        trigger_error("BAD JOIN, value sould be [ table => criteria ]", E_USER_ERROR);
```
btry: Hi
There is a typo : sould => should

remicollet: Even "sould" => "must" ;)

install/update_91_92.php
```diff
@@ -53,67 +53,43 @@ function update91to92() {
...
-      'glpi_devicefirmwaretypes',
-      'glpi_savedsearches_alerts',
-      'glpi_items_operatingsystems',
-      'glpi_operatingsystemkernels',
-      'glpi_operatingsystemkernelversions',
-      'glpi_operatingsystemeditions'
-   );
+   $newtables     = array("glpi_bookmarks",
+                          "glpi_bookmarks_users",
+                          "glpi_queuedmails");
```
remicollet: I think there is a big misunderstanding here... :(

Digging in the code, and old migration scripts, it seems only "new table" have to be backup...  but still not clear

install/update_91_92.php
```diff
@@ -625,6 +598,7 @@ function update91to92() {
                                                 'login_remember_default' => 1));
 
    if (TableExists('glpi_bookmarks')) {
+      // yllen: you backup an inexisting table. If backup existing table it's done at the beginning
```
remicollet: Totally agree here, this seems wrong.
Once the table is rename from "glpi_bookmarks", "glpi_savedsearches", you MUST keep it.
How can both exists ?

install/update_91_92.php
trasher: Yes, only tables created in the update script should be in the array; they'll be renamed as backup on further runs.

install/update_91_92.php
trasher: Indeed, that can be dropped

install/update_91_92.php
```diff
@@ -633,29 +607,32 @@ function update91to92() {
...
-      $migration->addKey("glpi_savedsearches", 'last_execution_time');
-
       $migration->addField("glpi_savedsearches", "do_count", "tinyint(1) NOT NULL DEFAULT '2' COMMENT 'Do or do not count results on list display; see SavedSearch::COUNT_* constants'");
-      $migration->addKey("glpi_savedsearches", 'do_count');
-
       $migration->addField("glpi_savedsearches", "last_execution_date", "DATETIME NULL DEFAULT NULL");
-      $migration->addKey("glpi_savedsearches", 'last_execution_date');
-
       $migration->addField("glpi_savedsearches", "counter", "int(11) NOT NULL DEFAULT '0'");
+      $migration->migrationOneTable("glpi_savedsearches");
```
trasher: Why the `migrationOneTable`? `addPostQuery` will be executed after the migration, so, this should not be needed.

inc/dbmysql.class.php
```diff
@@ -811,7 +811,8 @@ function __construct ($dbconnexion, $table, $crit="", $debug=false) {
                      unset($crit[$key]);
                      break;
 
-                  case 'JOIN' :
+                  case 'JOIN' : // deprecated
```
trasher: Maybe add an output in the log when the deprecated case is used

tests/units/DBmysqlIterator.php
```diff
@@ -191,6 +191,19 @@ public function testJoin() {
    }
 
 
+   public function testLeftJoin() {
+
+      $it = new \DBmysqlIterator(NULL, 'foo', ['LEFT JOIN' => ['bar' => ['FKEY' => ['bar' => 'id', 'foo' => 'fk']]]]);
+      $this->string($it->getSql())->isIdenticalTo('SELECT * FROM `foo` LEFT JOIN `bar` ON (`bar`.`id` = `foo`.`fk`)');
+   }
+
+   public function testInnerJoin() {
```
trasher: The error case should be tested as well

install/update_91_92.php
```diff
@@ -800,31 +775,37 @@ function update91to92() {
       $DB->queryOrDie($query, "9.2 add saved searches alerts notification translation");
    }
 
+   // Yllen: must be done at the beginning for all tables to forget change place of the message if new table added in list
```
trasher: I guess this should stay at the bottom, this is probably more "visible" for the users.

Anyways, I propose not to keep that kind of global message here, the Migration object can be in charge.

As example:
```diff
diff --git a/inc/migration.class.php b/inc/migration.class.php
index a8278f2..e2e74f2 100644
--- a/inc/migration.class.php
+++ b/inc/migration.class.php
@@ -51,6 +51,7 @@ class Migration {
       'pre'    => [],
       'post'   => []
    ];
+   private $has_backups;
 
    const PRE_QUERY = 'pre';
    const POST_QUERY = 'post';
@@ -647,6 +648,13 @@ class Migration {
          $DB->queryOrDie($query['query'], $query['message']);
       }
 
+      if ($this->has_backups) {
+         $migration->displayWarning(
+            __("You can delete backup tables if you have no need of them."),
+            true
+         );
+      }
+
       // end of global message
       $this->displayMessage(__('Task completed.'));
    }
@@ -854,7 +862,7 @@ class Migration {
     * @return boolean
     */
    public function backupTables($tables) {
-      $backup_tables = false;
+      $this->has_backups = false;
       foreach ($tables as $table) {
          // rename new tables if exists ?
          if (TableExists($table)) {
@@ -866,10 +874,10 @@ class Migration {
                   "backup_$table"
                )
             );
-            $backup_tables = true;
+            $this->has_backups = true;
             $this->renameTable("$table", "backup_$table");
          }
       }
-      return $backup_tables;
+      return $this->has_backups;
    }
 }
diff --git a/install/update_91_92.php b/install/update_91_92.php
index 299da48..0f49234 100644
--- a/install/update_91_92.php
+++ b/install/update_91_92.php
@@ -800,11 +800,6 @@ Regards,',
       $DB->queryOrDie($query, "9.2 add saved searches alerts notification translation");
    }
 
-   if ($has_backups) {
-      $migration->displayWarning("You can delete backup tables if you have no need of them.",
-                                 true);
-   }
-
    if (TableExists('glpi_queuedmails')) {
       $migration->renameTable("glpi_queuedmails", "glpi_queuednotifications");
       $migration->addPostQuery("UPDATE glpi_crontasks SET itemtype='QueuedNotification' WHERE itemtype='QueuedMail'");
```

install/update_91_92.php
yllen: it's for adding fields before adding keys

install/update_91_92.php
yllen: It's more logical for me to add this message in backupTable function

install/update_91_92.php
trasher: This is already the case; I've just tested an upgrade from a 9.1 database against current master; all three keys have been created

trasher:
For further PR, please do not make so much "CS changes" as this make real changes difficult to see :/
Either regroup all that kind of changes into one specific commit, or on another PR to keeps things "simple" ;)

trasher (changes requested):
Globaly OK, remaining points:
* test error cases,
* add a warning on deprecated case (or leave it not deprecated)

trasher:
> test error cases

==> done. i've also moved DBMysqlIterator to its own class; and factorized a bit join conditions ('INNER' was mainly a copy/paste from 'LEFT')

trasher (approved):

orthagh (approved):

install/update_91_92.php
yllen: i've tested from master only this table and keys were not added.
addPostQuery is done after migration so after addKeys.

The better way to do is to put the addField "do_count" at the end and add the array update in it.

PR #2354, opened 2017-06-23
Extract main lang along with extracting locales

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no


trasher:
This is what I had to do manually since tx gives us en en.po, and not en_GB. Maybe there is another solution.

Point here is that en_GB should be exactly the same as the source code, but I do not find how to copy automatically strings from source :/

PR #2359, opened 2017-06-24
Feature - anonymous endpoint and extra (not itemtypes) endpoints for plugins

btry:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | PoC for now
| Fixed tickets | 

For FlyveMDM: ability to handle endpoints not matching an itemtype.

Solves the need to have endpoints useable without authentication (depends on the implementation of the hook function in the plugin), and also solves the need to have endpoints which are not handled as an itemtype.

However, this feature allows potentially very big power. Plugin is responsible for proper implementation.

One problem I wish to sholve though: the endpoints are available without authnetication. The implementation is in charge to verify the session if needed. Thus the API cannot check if the plugin is enabled. Plugins may need to implement the callee function in setup.php or use post_init hook to ensure hooks.php is included.

Moreover, to check the session, the callee needs the instance if the API class (not implemented yet)

### DIFF
--- a/inc/apirest.class.php
+++ b/inc/apirest.class.php
@@ -68,7 +68,6 @@ class APIRest extends API {
     * @return mixed json with response or error
     */
    public function call() {
-
       //parse http request and find parts
       $this->request_uri  = $_SERVER['REQUEST_URI'];
       $this->verb         = $_SERVER['REQUEST_METHOD'];
@@ -104,6 +103,8 @@ class APIRest extends API {
       if (isset($this->parameters['session_write'])) {
          $this->session_write = (bool)$this->parameters['session_write'];
       }
+
+      $this->extraEndpoints = $this->getExtraEndpoints();
 
       // inline documentation (api/)
       if ($is_inline_doc) {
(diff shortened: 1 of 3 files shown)


inc/apirest.class.php
```diff
@@ -485,7 +493,7 @@ public function parseIncomingParams($is_inline_doc = false) {
     *
     * @return void
     */
-   public function returnResponse($response, $httpcode=200, $additionalheaders=array()) {
+   protected function returnResponse($response, $httpcode=200, $additionalheaders=array()) {
```
btry: this access mode does not matches the overriden method in the class API, then fixing.

btry:
After discussing with  @orthagh it appears it is simplaer (and safer) to use current authentication mechanism. When manu accounts are needed, let's assigna acategory to accounts to distinguish them.

This also allows the users of GLPI to filter out those which are not for humans users.

Issue #2307, opened 2017-06-16
[9.2] malformed login error page


btry:
Hi

If I enter wrong credentials to login in GLPI I get the following page:
![image](https://user-images.githubusercontent.com/14139801/27220911-0522c816-5287-11e7-987f-4f97d068b011.png)

i cannot click on "log in again", because the footer is invisible ans palced over the link. This may lead the user to http://glpi-project.org

See the screenshot with highlighted text
![image](https://user-images.githubusercontent.com/14139801/27221021-62d587e6-5287-11e7-8234-71612b8752c3.png)

yllen:
I test and i can click on "Log in again" and i'm redirect to login windows of connection

btry:
As far as I see there is a CSS and / or HTML problem.

The content of the footer partially overlaps the hyperlink making it not clickable everywhere it should be.

If you don't reprocduce then I shall tell I'm using Chromium packaged for Debian SId. I'll report the exact version soon.

yllen:
I'm using Firefox on CEntOS 7

btry:
I'll compare with others versions of FF, and feedback

yllen:
My version of Firefox is 52.1.0

btry:
Chromium Version 59.0.3071.86 64-bit

Firefox 54.0 64 bits  => OK
Firefox ESR 52.2.0 64 bits => OK too

I don't have Chrome to see how ot works, but it would be interesting to see if the bug is actually in Chromium or not.

btry:
I searched where is the footer by selecting text on the page when running FF

![image](https://user-images.githubusercontent.com/14139801/27343531-84bcab6e-55e3-11e7-82a5-b8e872fb6657.png)

Now if I reduce width of my window (I'm used to use half width of my display 1920x1080)

![image](https://user-images.githubusercontent.com/14139801/27343587-a8bf9cc4-55e3-11e7-86fc-b24347e4e682.png)

There is a small overlap, but it appears to be less thant Chromium.

I think we need to fix the vertical position of the footer, or remove it, as it is supposed to be invisible.

yllen:
I don't have footer like you login failed

btry:
The footer is actually here: see the HTML code (4 last lines)
![image](https://user-images.githubusercontent.com/14139801/27348473-7978166a-55f4-11e7-8c28-42b572115a0a.png)

[...]

PR #2363, opened 2017-06-26
fix css for login error page, upgrade page

btry:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2307 

footer is verticaly positionned at the same place as buttons. With windows less than 1000px width, the footer overlaps the button. see screenshots in the issue

### DIFF
--- a/css/styles.css
+++ b/css/styles.css
@@ -2434,7 +2434,6 @@ a.icon_nav_move:hover img {
 #footer-login {
    position: absolute;
    right: 10px;
-   bottom: 0px;
    height: 40px;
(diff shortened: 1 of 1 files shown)


trasher:
If this "need cherry-pick"; it should have been opened against 9.1 branch :)

btry:
I noticed afterwards a fix is needed in 9.1/bugfix. And anyway we decided to remove the footer completely, then this PR is obsolete.

PR #2368, opened 2017-06-27
Improve speed of getEntitiesRestrictRequest when entity list is huge

### LINKED ISSUE #2367: getEntitiesRestrictRequest is very slow when entity list is huge (more than 1400 entities)
Hello,
When the entity list is huge (more than 1400 entities), the getEntitiesRestrictRequest function is very slow as it will call getAncestorsOf function for each entity in the list.

And when using the fields plugin at ticket save the getEntitiesRestrictRequest will be called two times during the update hook, and then getAncestorsOf will be called 2800 times and such doing 2800 MySQL queries. At the end the ticket update may take up to 20s (depending on the load of the MySQL server).
Here is an xDebug profiling report:
image
As you can see the getEntitiesRestrictRequest takes 22% of the complete ticket update call.

And here is a MySQL profiling session, when doing a save of a ticket:
image

I'm going to propose a quicker version of getEntitiesRestrictRequest function.

Regards,
Tomolimo

tomolimo:
Added function getAncestorsOfArray, to retrieve ancestors for an array of entities
Fixes #2367

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2367

### DIFF
--- a/inc/db.function.php
+++ b/inc/db.function.php
@@ -644,11 +644,11 @@ function getTreeValueName($table, $ID, $wholename="", $level=0) {
  * Get the ancestors of an item in a tree dropdown
  *
  * @param $table     string   table name
- * @param $items_id  array  The IDs of the items
+ * @param $items_id  mixed (array or string) The IDs of the items
  *
  * @return array of IDs of the ancestors
 **/
-function getAncestorsOfArray($table, $items_id) {
+function getAncestorsOf($table, $items_id) {
    global $DB;
 
    // IDs to be present in the final array
@@ -656,8 +656,11 @@ function getAncestorsOfArray($table, $items_id) {
    $parentIDfield = getForeignKeyFieldForTable($table);
    $use_cache     = FieldExists($table, "ancestors_cache");
 
-   if ($use_cache
-       && is_array($items_id)) {
+   if(!is_array($items_id)){
+      $items_id = array($items_id);
+   }
+
+   if ($use_cache) {
 
       $query = "SELECT `id`, `ancestors_cache`, `$parentIDfield`
                 FROM `$table`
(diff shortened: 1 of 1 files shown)


inc/db.function.php
```diff
@@ -644,6 +644,96 @@ function getTreeValueName($table, $ID, $wholename="", $level=0) {
  * Get the ancestors of an item in a tree dropdown
  *
  * @param $table     string   table name
+ * @param $items_id  array  The IDs of the items
+ *
+ * @return array of IDs of the ancestors
+**/
+function getAncestorsOfArray($table, $items_id) {
```
trasher: Since this is mainly a copy/paste from the existing method, I wonder if it would not be better not to add a new one

tomolimo: what's your idea?

trasher: My idea would be to adapt `getAncestorsOf()` so it can accept a single value or an array

tomolimo: then we may break compatibility?

tomolimo: I'll look at it tomorrow

trasher: Why would compatibility broken? If the existing function may take either an integer or an array; all existing code will still working as expected.
We probably need some tests on this.

inc/db.function.php
```diff
@@ -644,6 +644,96 @@ function getTreeValueName($table, $ID, $wholename="", $level=0) {
...
+               if (($parent > 0)
+                   || ($table == 'glpi_entities')) {
+                  $loc_id_found[$parent] = $parent;
+               }
+
+               // Store cache datas in DB
+               $query = "UPDATE `$table`
+                      SET `ancestors_cache` = '".exportArrayToDB($loc_id_found)."'
+                      WHERE `id` = '".$row['id']."'";
+               $DB->query($query);
```
remicollet: Already done in getAncestorsOf

remicollet:
Benchmark please

inc/db.function.php
```diff
@@ -644,6 +644,96 @@ function getTreeValueName($table, $ID, $wholename="", $level=0) {
...
+   $id_found      = array();
+   $parentIDfield = getForeignKeyFieldForTable($table);
+   $use_cache     = FieldExists($table, "ancestors_cache");
+
+   if ($use_cache
+       && is_array($items_id)) {
+
+      $query = "SELECT `id`, `ancestors_cache`, `$parentIDfield`
+                FROM `$table`
+                WHERE `id` IN ('".implode("','", $items_id)."')";
```
remicollet: => Dbiterator

trasher:
Also, as this is an important function, we must have tests on it.

tomolimo:
@remicollet: 

> Benchmark please

See #2368

orthagh:
i globally agree with @trasher comments:
- see if factorisation is possible
- we mandatory need tests for this core function

Plus, target master for this change, next 9.1.5 will be the last and will contains only minor fixes.
We can't accept these kind of major changes for the bugfixes branch. For information, we plan 9.2 release in september and, as mentioned above, a last 9.1 release between today and 9.2 release.

BTW, it's a very good and need enhancement.

tomolimo:
New version with factorization, :)
But I really don't have time for developing the unitary tests, this fix is claimed by our servicedesk, and I'll be out-of-office for the summer (to have a medical  operation + holidays).
I made functional tests and coverage tests and all passed.

inc/db.function.php
```diff
@@ -644,7 +644,7 @@ function getTreeValueName($table, $ID, $wholename="", $level=0) {
  * Get the ancestors of an item in a tree dropdown
  *
  * @param $table     string   table name
- * @param $items_id  integer  The ID of the item
+ * @param $items_id  mixed (array or string) The IDs of the items
```
trasher: Should be `array|string`

inc/db.function.php
```diff
@@ -656,65 +656,74 @@ function getAncestorsOf($table, $items_id) {
    $parentIDfield = getForeignKeyFieldForTable($table);
    $use_cache     = FieldExists($table, "ancestors_cache");
 
-   if ($use_cache
-       && ($items_id > 0)) {
+   if(!is_array($items_id)){
+      $items_id = array($items_id);
```
trasher: You no longer check if the item is greater than 0. I guess array should be filtered to drop all those cases.

inc/db.function.php
```diff
@@ -656,65 +656,74 @@ function getAncestorsOf($table, $items_id) {
...
-      $query = "SELECT `ancestors_cache`, `$parentIDfield`
-                FROM `$table`
-                WHERE `id` = '$items_id'";
+   if ($use_cache) {
 
-      if (($result = $DB->query($query))
-          && ($DB->numrows($result) > 0)) {
-         $ancestors = trim($DB->result($result, 0, 0));
-         $parent    = $DB->result($result, 0, 1);
+      $query = "SELECT `id`, `ancestors_cache`, `$parentIDfield`
```
trasher: DBIterator should be used here

inc/db.function.php
trasher: As far as I can see, this is in getAncestorsOf :)

trasher:
Replaced by https://github.com/glpi-project/glpi/pull/2441

tomolimo:
@trasher: thank you
regards,
Tomolimo

PR #2381, opened 2017-06-29
fix "The each() function is deprecated (7.2)

remicollet:
Need to find and  fix other possible usages

### DIFF
--- a/inc/mailcollector.class.php
+++ b/inc/mailcollector.class.php
@@ -1317,9 +1317,8 @@ class MailCollector  extends CommonDBTM {
 
          if ($structure->type == 1) { /* multipart */
             $prefix = "";
-            reset($structure->parts);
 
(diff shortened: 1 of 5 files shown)


remicollet:
NOTICE: blind fix, test needed

inc/mailcollector.class.php
```diff
@@ -1399,8 +1398,7 @@ private function getDecodedFetchbody($structure, $mid, $part) {
    function getRecursiveAttached($mid, $path, $maxsize, $structure, $part = "") {
 
       if ($structure->type == 1) { // multipart
-         reset($structure->parts);
-         while (list($index, $sub) = each($structure->parts)) {
+         foreach ($structure->parts as $index => $sub_structure) {
```
trasher: I guess we're waiting for a `$sub` here

remicollet:
Tested, excepted mailcollector

Issue #2393, opened 2017-06-30
PHP 7.1 - Toolbox::getRandomString() may call a deprecated method in RandomLib


btry:
Hi

here is a backtrace, where Toolbox::getRandomString is called with $high = true.

```
+'2017-06-30 22:40:01 [2@testing-docker-f68a5bbe-cff7-489d-995e-cae2a6ff3f43]\n
  *** PHP Deprecated function(8192): Function mcrypt_module_open() is deprecated\n
  Backtrace :\n
  ...om-lib/lib/RandomLib/AbstractMcryptMixer.php:75 \n
  ...maxell/random-lib/lib/RandomLib/Factory.php:232 RandomLib\AbstractMcryptMixer->__construct()\n
  ...cmaxell/random-lib/lib/RandomLib/Factory.php:78 RandomLib\Factory->findMixer()\n
  ...maxell/random-lib/lib/RandomLib/Factory.php:122 RandomLib\Factory->getGenerator()\n
  inc/toolbox.class.php:1490                         RandomLib\Factory->getMediumStrengthGenerator()\n
  inc/user.class.php:4234                            Toolbox::getRandomString()\n
  inc/user.class.php:4277                            User::getUniqueToken()\n
  plugins/flyvemdm/inc/invitation.class.php:142      User::getToken()\n
  inc/commondbtm.class.php:891                       PluginFlyvemdmInvitation->prepareInputForAdd()\n
  inc/api.class.php:1571                             CommonDBTM->add()\n
  inc/apirest.class.php:233                          API->createItems()\n
  ...-mdm/test/src/Glpi/Test/ApiRestTestCase.php:133 APIRest->call()\n
```

https://github.com/ircmaxell/RandomLib/issues/55
https://wiki.php.net/rfc/mcrypt-viking-funeral

the later link references several pages, the first one is yours @remicollet :) about mcrypt

FYI In #1301 we already lowered the maximum strength of the generator from high to medium.

remicollet:
See https://github.com/ircmaxell/RandomLib/issues/57

So for now, "high" is simply not allowed

PR #2401, opened 2017-07-03
MediumStrenght generator rely on deprecated mcrypt extension

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | yes (maybe -- partial 9.2 only)
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2393

### DIFF
--- a/inc/toolbox.class.php
+++ b/inc/toolbox.class.php
@@ -1477,20 +1477,15 @@ class Toolbox {
     * Get a random string
     *
     * @param integer $length of the random string
-    * @param boolean $high strength of the random source (since 9.2)
     *
     * @return random string
    **/
-   static function getRandomString($length, $high = false) {
+   static function getRandomString($length) {
 
       $factory = new RandomLib\Factory();
-      if ($high) {
-         /* Notice "High" imply mcrypt extension, unwanted for now
-            See https://github.com/ircmaxell/RandomLib/issues/57 */
-         $generator = $factory->getMediumStrengthGenerator();
-      } else {
-         $generator = $factory->getLowStrengthGenerator();
-      }
+      /* "High" (and apparently "Medium") imply mcrypt extension, unwanted for now:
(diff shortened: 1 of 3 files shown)


inc/toolbox.class.php
```diff
@@ -1477,20 +1477,13 @@ static function getItemTypeTabsURL($itemtype, $full = true) {
...
    **/
-   static function getRandomString($length, $high = false) {
+   static function getRandomString($length) {
 
       $factory = new RandomLib\Factory();
-      if ($high) {
-         /* Notice "High" imply mcrypt extension, unwanted for now
-            See https://github.com/ircmaxell/RandomLib/issues/57 */
-         $generator = $factory->getMediumStrengthGenerator();
-      } else {
```
remicollet: please keep the link to upstream issue

remicollet:
Really needed ?

trasher:
On the core side, this was unit tested without any issues; but one of the plugins get a deprecated issue using the `User::getToken()` method from the master branch :/
So, I guess there really is an issue somewhere

inc/toolbox.class.php
trasher: Re-added

remicollet:
Missing openssl extension ?

trasher:
I don't know, I've admitted travis environment were the same :)

btry:
Hi

My plugin requires openssl extension and checks it in requirements. The installation of the plugin is unit tested and it passes.

trasher:
The issue is reproductible on GLPI side with debug mode activated (error_reporting(E_ALL | E_STRICT) and with mcrypt module installed.

Issue is from the RandomLib library that relies on mcrypt for "medium" and "high"...

Maybe https://stackoverflow.com/questions/4356289/php-random-string-generator/31107425#31107425 would be an alternative to the lib?

trasher:
See #2423

PR #2402, opened 2017-07-03
Add a method to check for requirements from plugin infos

### LINKED ISSUE #2379: Plugins compatibility check method
Actually, each plugin check if it is compatible with GLPI on its own side.

I've added some methods to centralize displayed messages (and translate them) already, but we could provide a complete method that would take a minial version, a maximal version and a flag to support  dev versions.

Recent change on GLPI version (set to "9.2-dev") may mark some plugins as incompatible (because 9.2-dev is lower than 9.2).

Thoughts?

trasher:
fixes #2379

| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2379

### DIFF
--- a/inc/config.class.php
+++ b/inc/config.class.php
@@ -1797,7 +1797,7 @@ class Config extends CommonDBTM {
       }
       echo "\n";
 
-      self::checkExtensions(true);
+      self::displayCheckExtensions(true);
 
       self::checkWriteAccessToDirs(true);
       toolbox::checkSELinux(true);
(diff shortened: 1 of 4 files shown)


trasher:
Example of use: https://github.com/TECLIB/fusioninventory-for-glpi/commit/a01e435b8d6ab02ad9ba4c63b30efbc290d69ae3

trasher:
We could expand this, yes; I had that in mind :D

What we could check:
- GLPI versions (done),
- PHP versions,
- PHP extensions,
- PHP parameters,
- Linked plugins.

I'll probably merge this one and continue the job :)

wawax:
I agree with you.
Also discussing the possibily to check for a file & directory existance and rights

trasher:
You're talking about plugins that may need contents? This should not be required since plugins must use GLPI data directory; ACLs are already checked and should be ok on it.

If you had something else in mind, tell me ;)

wawax:
No you're right.

trasher:
Finally, I've updated current PR; discussions are here, and it sounds logical to centralize all this stuff ;)

orthagh:
still ok for me.

Do you plan to update plugin dev doc and example plugin ?

trasher:
I must admit I've forget the example plugin :D But I will do, doc was already planned

trasher:
PR opened for doc: https://github.com/glpi-project/docdev/pull/55

Issue #1389, opened 2016-12-02
planning view - sub task of project task is displayed - but done in the past


Armin83:
I can see a sub task (of project task) which is done in past. All dates are in past so i should not see it on 02.12.2016 today, but i can see it in my home planning view.
![img_018](https://cloud.githubusercontent.com/assets/19192087/20831321/b43a3368-b884-11e6-9f1f-185146180ed4.jpg)

I think it is displayed becouse of project task date section but i should use the settings from sub task

Armin83:
@orthagh 
Can you approve this bug ?

It is worse than I have assumed. Als task, sub task, tickets, change, etc are all at Done and 100% percent but i get the task in my view. If i delete me from team member (you can see in my screenshot) the task dissappear but this can't be a solution. I should look at a past project who have done this task.... 

![img_003](https://cloud.githubusercontent.com/assets/19192087/24442264/a736b616-145d-11e7-8eff-5959d683b0a0.jpg)

pvasener:
I confirm the bug. It's been there since 9.1.0 and not fixed yet. It looks like the project task status is not checked. All tasks with statuses with 'Finished State' = Yes shouldn't appear in the planning.

orthagh:
Hello here.

Sorry for the long delay for this issue.
Here is the fix, it should come into 9.1.5

PR #2403, opened 2017-07-03
don't show finished project tasks in planning

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1389

### DIFF
--- a/inc/projecttask.class.php
+++ b/inc/projecttask.class.php
@@ -1299,11 +1299,22 @@ class ProjectTask extends CommonDBChild {
                      AND ";
       }
 
+      $DONE_EVENTS = '';
+      if (!$options['display_done_events']) {
+         $DONE_EVENTS = "`glpi_projecttasks`.`percent_done` < 100
+                         AND (glpi_projectstates.is_finished = 0
+                              OR glpi_projectstates.is_finished IS NULL)
+                         AND ";
+      }
+
       $query = "SELECT `glpi_projecttasks`.*
                 FROM `glpi_projecttaskteams`
                 INNER JOIN `glpi_projecttasks`
                   ON (`glpi_projecttasks`.`id` = `glpi_projecttaskteams`.`projecttasks_id`)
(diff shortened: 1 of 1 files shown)


inc/projecttask.class.php
```diff
@@ -1299,11 +1299,22 @@ static function populatePlanning($options=array()) {
...
+                         AND (glpi_projectstates.is_finished = 0
+                              OR glpi_projectstates.is_finished IS NULL)
+                         AND ";
+      }
+
       $query = "SELECT `glpi_projecttasks`.*
                 FROM `glpi_projecttaskteams`
                 INNER JOIN `glpi_projecttasks`
                   ON (`glpi_projecttasks`.`id` = `glpi_projecttaskteams`.`projecttasks_id`)
+                LEFT JOIN `glpi_projectstates`
```
trasher: Should not the join be conditionned as well? Well, not really important, but this could maybe save query time when the option is on.

orthagh: i think the difference must be very small with and without join (a microsecond ?)
And in fact, ``$options['display_done_events'] = false`` is the default behavior.

I kept this options (maybe plugins use it) but on glpi side, i think it's never used (and so join always required)

Krisstof:
Hi,

Thanks for your answer.
We have been using GLPI for almost 10 years and we have always seen the "closed" and "resolved" Tasks events in calendars. All our collaborative work is precisely based on these calendar data (with CalDAV sync). Remove data that is there for 10 years and indicating that it was a bug is not acceptable. I ask that we have the choice to display the different states of the tasks in the calendars. Moreover, the fact of displaying everything was not a problem since graphically, colors and icons of state are not the same!
Can you indicate which part of the code is to be modified to find the "NORMAL" behavior (the old GLPI guys will not contradict).
I thank you very much in advance.

Regards

PR #2406, opened 2017-07-04
Add display of task in state 'todo' of tickets and problems on homepage.

ddurieux:
Fix the fields in search engine, the completename for group to have 'my group' for groups in charge of the task and the name of the technician in charge of the task

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes

### DIFF
--- a/inc/commonitiltask.class.php
+++ b/inc/commonitiltask.class.php
@@ -733,7 +733,7 @@ abstract class CommonITILTask  extends CommonDBTM {
          'table'              => 'glpi_users',
          'field'              => 'name',
          'linkfield'          => 'users_id_tech',
-         'name'               => __('Technician'),
+         'name'               => __('Technician in charge of the task'),
          'datatype'           => 'itemlink',
          'right'              => 'own_ticket',
          'forcegroupby'       => true,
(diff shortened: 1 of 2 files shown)


inc/commonitiltask.class.php
```diff
@@ -1748,4 +1748,161 @@ function showFormMassiveAction() {
...
+               self::showVeryShort($ID, $itemtype);
+            }
+         }
+         echo "</table>";
+      }
+   }
+
+
+
+   static function showVeryShort($ID, $itemtype, $forcetab = '') {
```
trasher: missing phpdoc ;)

inc/commonitiltask.class.php
```diff
@@ -1748,4 +1748,161 @@ function showFormMassiveAction() {
    }
 
 
+
+   /**
+    * @param $start
```
trasher: Incorrect phpdoc, see http://glpi-developer-documentation.readthedocs.io/en/master/codingstandards.html#comments

inc/commonitiltask.class.php
```diff
@@ -1748,4 +1748,161 @@ function showFormMassiveAction() {
...
+    * @param $status             (default ''process)
+    * @param $showgrouptickets   (true by default)
+    */
+   static function showCentralList($start, $status = 'todo', $showgrouptickets = true) {
+      global $DB, $CFG_GLPI;
+
+      $query = "SELECT `".self::getTable()."`.`id` "
```
trasher: Use dbiterator?

inc/commonitiltask.class.php
ddurieux: Oh yes you're right, it's not ok, I will modify it

inc/commonitiltask.class.php
trasher: +1 for dbiterator; do not forget to remove the old query ;)

Issue #1967, opened 2017-03-24
GLPI email notification error in email follow ups with some mail providers


wanderleihuttel:
We have our business mail in a provider where the usernames configuration are a bit different. Instead of using the "@" character as the separator, they use the character "=".

Is it possible to see the tutorial or provider:
https://www.hostnet.com.br/config.email/?email=admin@hostnet.com.br&tutorial=11

I'm sending a printscreen of my configuration:
In Setup -> Notification -> Email followups configuration.
![glpi](https://cloud.githubusercontent.com/assets/16232272/24275797/ddd2db38-1010-11e7-9b42-9f496da25534.png)

I've done some tests and I've discovered how to fix it for this cases. 
Is necessary to inform 2 more parameters ($this->Sender and $this->From) for phpmailer in the file "/var/www/glpi/inc/glpimailer.class.php" after line 69. 

Before:

         if ($CFG_GLPI['smtp_username'] != '') {
            $this->SMTPAuth = true;
            $this->Username = $CFG_GLPI['smtp_username'];
            $this->Password = Toolbox::decrypt($CFG_GLPI['smtp_passwd'], GLPIKEY);
         }

After:

         if ($CFG_GLPI['smtp_username'] != '') {
            $this->SMTPAuth = true;
            $this->Username = $CFG_GLPI['smtp_username'];
            $this->Password = Toolbox::decrypt($CFG_GLPI['smtp_passwd'], GLPIKEY);
            $this->Sender = str_replace('=','@',$CFG_GLPI['smtp_username']);
            $this->From = str_replace('=','@',$CFG_GLPI['smtp_username']);
         }

I would like to suggest, if I have enabled the debug mode in GLPI, would be interesting have an option to enable debug of phpmailer and improve the way to collect mail logs for discover possible errors.

$this->SMTPDebug   = 1; // 1 = errors and messages or 2 = messages only

Best Regards
Wanderlei Hüttel

trasher:
Hello,

I do not really understand what goes wrong with your configuration.
I think using smtp username as "from" in the resulting email is a mistake. The "from" should be populated using the "Administrator email" from the preferences. And what about real email address with a "=" sign? (seems to be legible).

Also, +1 to set debug on phpmailer when glpi is in debug mode.

wanderleihuttel:
Hello Johan

[...]

PR #2414, opened 2017-07-05
Feature/improve mail config

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1967

Several changes in this proposal:
- add a global from email and names to be used on each send, instead of admin or entity admin infos,
- add a smtp_sender that seems to be required from some providers (it uses the "from" field is fnot set),
- rework debug (not sure it was OK) and log in a specific file,
- rework config form:
   - add html labels,
   - hide smtp configuration when not needed,
   - move some fields from "smtp config" that are not SMTP specific (so they're not hidden)
- drop empty and unused `$options` parameter for ` getSender()` method.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@ new whole window
 - `NotificationTemplate::getDataToSend()` signature has changed
 - `QueuedMail` has been renamed to `QueuedNotification`
 - `CommonDBTM::mailqueueonaction` has been renamed to `CommonDBTM::notificationqueueonaction`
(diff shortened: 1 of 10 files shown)


trasher:
mail_config_php
mail_config_smtp

trasher:
Note: smtp debugging is explained on https://github.com/PHPMailer/PHPMailer/wiki/SMTP-Debugging

PR #2415, opened 2017-07-05
Add ticket/task last editor; fixes #2187

### LINKED ISSUE #2187: Add edit user
When someone edit an existing followop or task; it was previously set as author. This is not correct, because the editor is not the author.

@tomolimo has proposed in #2161 to store last editor and display it (like we can see on most forums systems for example).

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2187


install/mysql/glpi-empty.sql
```diff
@@ -686,6 +686,7 @@ CREATE TABLE `glpi_changetasks` (
   `begin` datetime DEFAULT NULL,
   `end` datetime DEFAULT NULL,
   `users_id` int(11) NOT NULL DEFAULT '0',
+  `editors_id` int(11) NOT NULL DEFAULT '0',
```
wawax: it's better to call the field users_id_editor, as it's done for example for users_id_tech

install/update_91_92.php
```diff
@@ -1294,6 +1294,26 @@ function update91to92() {
       $migration->dropTable('glpi_networkequipmentfirmwares');
    }
 
+   //add editor in followupps
+   if (!fieldExists('glpi_ticketfollowups', 'editors_id')) {
```
wawax: should be in upper case FieldExists, but the underlying question is more : with this function is not called fieldExists ?

trasher: This should be called "fieldExists"... I've begin to work on a PR to fix method names; this is a huge job :D I'lle change to use current method name anyways.

Issue #1314, opened 2016-11-18
API : lost password feature


btry:
Hi

GLPi has a feature to retrieve / reset lost passwords (see glpi/front/lostpassword.php)

There is no equivalent in the API.

btry:
We need an endpoint usable without authentication.

PR #2430, opened 2017-07-07
Feature/api password recovery

btry:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | not added yet
| Fixed tickets | #1314

This PR adds and endpoint to recover an account.

By the way, fixes some links in the rest API documentation

### DIFF
--- a/apirest.md
+++ b/apirest.md
@@ -6,6 +6,7 @@
 * [Important](#important)
 * [Init session](#init-session)
 * [Kill session](#kill-session)
+* [Lost password](#lost-password)
 * [Get my profiles](#get-my-profiles)
 * [Get active profile](#get-active-profile)
 * [Change active profile](#change-active-profile)
@@ -20,9 +21,9 @@
 * [Get multiple items](#get-multiple-items)
 * [List searchOptions](#list-searchoptions)
 * [Search items](#search-items)
-* [Add item(s)](#add-items)
-* [Update item(s)](#update-items)
-* [Delete item(s)](#delete-items)
+* [Add item(s)](#add-item-s)
+* [Update item(s)](#update-item-s)
+* [Delete item(s)](#delete-item-s)
 * [Special cases](#special-cases)
 * [Errors](#errors)
 * [Servers configuration](#servers-configuration)
@@ -71,7 +72,7 @@ App(lication) token
 * you should always precise a Content-Type header in your HTTP calls.
    Currently, the api supports:
   * application/json
-  * multipart/form-data (for files upload, see [Add item(s)](#add-items) endpoint.
+  * multipart/form-data (for files upload, see [Add item(s)](#add-item-s) endpoint.
 
 * GET requests must have an empty body. You must pass all parameters in URL.
   Failing to do so will trigger an HTTP 400 response.
@@ -81,7 +82,7 @@ App(lication) token
   * [initSession](#init-session)
   * [killSession](#kill-session)
   * [changeActiveEntities](#change-active-entities)
-  * [changeActiveProfile](#change-active-profiles)
+  * [changeActiveProfile](#change-active-profile)
 
   You could pass an additional parameter "session_write=true" to bypass this default.
   This read-only mode allow to use this API with parallel calls.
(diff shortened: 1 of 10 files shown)


apirest.md
```diff
@@ -168,6 +169,26 @@ $ curl -X GET \
...
+* **Parameters**: (query string)
+  * *email*: email address of the user to recover. Mandatory.
+* **Returns**:
+  * 200 (OK).
+  * 400 (Bad Request) with a message indicating an error in input parameter.
+
+```bash
+$ curl -X GET \
+-H 'Content-Type: application/json' \
+'http://path/to/glpi/apirest.php/lostPassword'
```
orthagh: we miss the `email` parameter in the example

inc/api.class.php
```diff
@@ -1817,6 +1817,77 @@ protected function deleteItems($itemtype, $params = []) {
...
+
+      if (!isset($params['email'])) {
+         return $this->returnError(__("email parameter missing"));
+      }
+
+      if (isset($_SESSION['glpiID'])) {
+         return $this->returnError(__("A session is active"));
+      }
+
+      $user = new User();
```
orthagh: I think this part need to be factorised with User::updateForgottenPassword

inc/apirest.class.php
```diff
@@ -189,6 +189,13 @@ public function call() {
 
          return $this->returnResponse($response, $code, $additionalheaders);
 
+      } else if ($resource === "lostPassword") {
+         if ($this->verb != 'POST' && $this->verb != 'PUT') {
+            // forbid password reset with GET because it may be logged
+            return $this->returnError();
```
orthagh: maybe precise the error cause

btry: ReturnError uses 'Bad Request' by default. Do you think this is not sufficient?

orthagh: maybe precise that this endpoint couldn't be used by GET.
By the way, you have also an inconsistence with the inline documentation which precise GET verb for this endpoint

btry: OK to send a specific message.

I initially thought GET would be suffficient. I have to update the documentation.

inc/apirest.class.php
```diff
@@ -189,6 +189,13 @@ public function call() {
 
          return $this->returnResponse($response, $code, $additionalheaders);
 
+      } else if ($resource === "lostPassword") {
+         if ($this->verb != 'PUT') {
+            // forbid password reset with GET because it may be logged
```
orthagh: comment inconsistent with the relative code

inc/config.class.php
```diff
@@ -38,6 +38,20 @@
    die("Sorry. You can't access this file directly");
 }
 
+
+class ConfigPasswordTooWeak extends Exception {
```
orthagh: I'm not sure about the Exception class in config file.
Maybe we could add a new sub exceptions directory into inc to store this first one (and next).

inc/user.class.php
```diff
@@ -41,6 +41,10 @@
 
 use Sabre\VObject;
 
+class UserForgetPasswordException extends Exception {
```
orthagh: same for exception

inc/config.class.php
btry: I agree.

tests/APIBaseClass.php
```diff
@@ -1280,6 +1280,98 @@ public function testKillSession() {
    }
 
    /**
+    * @tags api
+    * @covers API::lostPassword
```
trasher: There is no `@covers` with atoum ;)

tests/APIBaseClass.php
```diff
@@ -1280,6 +1280,98 @@ public function testKillSession() {
...
+                          'ERROR');
+
+      // Test the verb GET is not alloxed
+      $res = $this->query('lostPassword',
+                          ['verb'    => 'GET',
+                          ],
+                          400,
+                          'ERROR');
+
+      // Test the verb DELETE is not alloxed
```
trasher: Typo

trasher:
Looks globally OK but updatePassword methods needs to be tested as well (success and exceptions).
Also, I'm not sure exceptions are tested from API.

btry:
I had to add in  the API an new handler for thie endpoint, and it catches the exceptions. See api.class.php line 1860.

@covers and typo fixes incoming

PR #2434, opened 2017-07-10
DBmysqlIterator implements Countable

remicollet:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | perhaps DBmysqlIterator::numrows could be deprecated (out of scope of this PR)
| Tests pass?   | yes
| Fixed tickets | N/A

DBmysqlIterator implements Countable

### DIFF
--- a/inc/dbmysqliterator.class.php
+++ b/inc/dbmysqliterator.class.php
@@ -37,7 +37,7 @@ if (!defined('GLPI_ROOT')) {
 /**
  *  Database iterator class for Mysql
 **/
-class DBmysqlIterator implements Iterator {
+class DBmysqlIterator implements Iterator, Countable {
    /**
     * DBmysql object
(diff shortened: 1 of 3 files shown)


inc/dbmysqliterator.class.php
```diff
@@ -420,4 +420,15 @@ public function valid() {
    public function numrows() {
       return ($this->res ? $this->conn->numrows($this->res) : 0);
    }
+
+   /**
+    * Number of rows on a result
+    *
+    * @since 9.2
+    *
+    * @return int
```
trasher: We usually return the full form (integer) rather than the short one (int).

remicollet: fixed

PR #2437, opened 2017-07-10
Use WHERE, fix query

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

### DIFF
--- a/inc/commonitiltask.class.php
+++ b/inc/commonitiltask.class.php
@@ -1751,6 +1751,8 @@ abstract class CommonITILTask  extends CommonDBTM {
    /**
     * Get tasks list
     *
+    * @since 9.2
+    *
     * @return DBmysqlIterator
     */
    public static function getTaskList($status, $showgrouptickets) {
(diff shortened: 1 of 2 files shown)


trasher:
Following comments on commited code

inc/commonitiltask.class.php
```diff
@@ -1748,42 +1748,50 @@ function showFormMassiveAction() {
...
-         if (count($_SESSION['glpigroups'])) {
-            $prep_req['AND']['groups_id_tech'] = $_SESSION['glpigroups'];
-         } else {
-            $prep_req['AND'][0] = 1;
-         }
+      if ($showgrouptickets && isset($_SESSION['glpigroups']) && count($_SESSION['glpigroups'])) {
+         $prep_req['WHERE']['groups_id_tech'] = $_SESSION['glpigroups'];
       } else {
-         $prep_req['AND']['users_id_tech'] = $_SESSION['glpiID'];
+         $prep_req['WHERE']['users_id_tech'] = $_SESSION['glpiID'];
```
ddurieux: It's not good, the behavior has changed.
We need:

1/ we get the task of my groups
    a/ I have group, so add where to get my groups
    b/ I'm not in a group, we must not get tasks (that's why I use 0=1)
2/ we get task personnal, the where is ok

Issue #2472, opened 2017-07-16
9.1.5 Problem with self-service perfil


glpibr:
Tickets opened by self-service users are losing categories.

GLPI 9.1.5 only.

glpibr:
Another problem is that the user is able to change the category after the error. But GLPI attributes user as a technical responsible for the call, even though he is only self-service.

orthagh:
Hello @glpibr 

Could you check the glpi logs (on fs, in glpi/files/_log/) if you have relevant entries ?

orthagh:
Finally found the problem, @glpibr, could you test the following changes: https://github.com/glpi-project/glpi/pull/2474/files

Armin83:
It works for me on my test system.

glpibr:
The test is ok here.
--
Atenciosamente






*Halexsandro de Freitas Sales*

Gerente de projeto

+55 81 3091 4252

halexsandro.sales@verdanatech.com



2017-07-17 6:01 GMT-03:00 Armin83 <notifications@github.com>:

yllen:
@orthagh you add a part deleted to fix a security part?

orthagh:
yes.
I was off yesterday, so i suggest a fast fix.
But today, i need to check to get this part working without security hole present.

orthagh:
Finally after check the original issue, i'm pretty sure that is not present anymore with the new fix

orthagh:
and for a final word, i removed the potential problematic part.

PR #2474, opened 2017-07-17
retrieve values from post in self-service

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2472

### DIFF
--- a/inc/ticket.class.php
+++ b/inc/ticket.class.php
@@ -3049,6 +3049,13 @@ class Ticket extends CommonITILObject {
                               '_filename'           => array(),
                               '_tag_filename'       => array());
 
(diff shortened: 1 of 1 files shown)


orthagh:
regression in #2458 and 6a18a32e68dc0f3b810cd57760bb28f5b39938b7

orthagh:
What are you doing here @trasher ? :D

But yes, i pushed early this fix for addressing comments about regression, but i'm currently working on a better fix. BTW, the previous security hole is not present anymore (i checked some minutes ago)

PR #2520, opened 2017-07-26
Feature/send document from api

btry:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | not tested yet
| Fixed tickets | none

Provides a way to download a document from the API

### DIFF
--- a/apirest.md
+++ b/apirest.md
@@ -23,7 +23,7 @@
 * [Add item(s)](#add-items)
 * [Update item(s)](#update-items)
 * [Delete item(s)](#delete-items)
-* [Download a document](#download-a-document)
+* [Special cases](#special-cases)
 * [Errors](#errors)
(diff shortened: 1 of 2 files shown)


btry:
I think the tests on canView() ans canViewItem() can be removed, as the API already checks this earlier.

Neds an unit test

orthagh:
without all stuff mentionned above, already lgtm.

So waiting for enhancement on documentation and tests before merge

btry:
Currently I don't see an good way to unit test this. There are calls to exit on the way.

Doc added, and removed redundant calls to canView() and canViewItem()

btry:
This pr and #2525 should use the same HTTP header **Accept** 

Please have a look on it to ensure we use the same header everywhere.

PR #2545, opened 2017-08-03
use rich file upload 'field' for user picture

btry:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | none

The main purpose of this PR is to allow update of user picture from the API by removing the need tu handle a file upload with the HTTP PUT verb.

This also allows to drag / drop a picture in GLPI's UI and unifies the interface between User and Document.

To update the picture of a user from the API, the API consumer needs to upload a file with the endpoint ajax/fileUpload.php, then send a HTTP PUT request on the user object with a additional property **_picture** containing the filename as answered in JSON after a successful upload.

```json
{
  "input": {
    "id": 9,
    "_picture": "my_picture.jpg"
  }
}
````

### DIFF
--- a/inc/user.class.php
+++ b/inc/user.class.php
@@ -684,54 +684,36 @@ class User extends CommonDBTM {
          self::dropPictureFiles($this->fields['picture']);
          $input['picture'] = 'NULL';
       } else {
+         if (isset($input["_picture"][0]) && !empty($input["_picture"][0]) == 1) {
+            $fullpath = GLPI_TMP_DIR."/".$input["_picture"][0];
+            if (toolbox::getMime($fullpath, 'image')) {
+               // Unlink old picture (clean on changing format)
+               self::dropPictureFiles($this->fields['picture']);
+               // Move uploaded file
+               $filename     = uniqid($this->fields['id'].'_');
+               $sub          = substr($filename, -2); /* 2 hex digit */
+               $tmp          = explode(".", $input["_picture"][0]);
+               $extension    = Toolbox::strtolower(array_pop($tmp));
+               @mkdir(GLPI_PICTURE_DIR . "/$sub");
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: inc/user.class.php [hunk 1]
```diff
@@ -675,54 +675,47 @@ function prepareInputForUpdate($input) {
...
-
-                     //prepare a thumbnail
-                     $thumb_path = GLPI_PICTURE_DIR . "/$sub/${filename}_min.$extension";
-                      Toolbox::resizePicture($picture_path, $thumb_path);
-                  } else {
-                     Session::addMessageAfterRedirect(__('Potential upload attack or file too large. Moving temporary file failed.'),
-                                                     false, ERROR);
-                  }
+         $newPicture = false;
+         if (isAPI()) {
```
orthagh: i think this whole part (if...else) could be factorized a little.
It seems actually a little redondant.

trasher:
ping?

btry:
pong !

I'll be busy for a while until end of november. I'll postpone further work on  this PR (december)

btry:
> One comment, in form mode, we could have lost the $_FILES error part, but maybe it's managed by the ajax part, could you confirm ?

If I look into **inc/glpiuploadhandler.class.php** it seems there are error handling, but actual test fails

Have a look if I try to upload a huge file, larger than the limit alowed by my server (image restri tion disabled for the test)

image

$_FILES is an empty array, so travefrsing it with foreach () fails.

I can use this PR to fix this issue.

btry:
@trasher : PR rebased

cedric-anne:
I made some fixes in my last commit. Just tested with API and user forms (from preferences page and from user list), and it works.
@trasher @orthagh Could you review my changes ?

Issue #2546, opened 2017-08-03
Recursion in search engine


yllen:
http://forum.glpi-project.org/viewtopic.php?id=158350

Add recursion in search options and massive actions for all recursive objects

yllen:
Done for master https://github.com/glpi-project/glpi/commit/f0d349de975a88848d6d9da576e657475704a603

nfarias1882:
Hola estimados buenas tardes, junto con saludar comento que la problemática que tengo de minuto es que no puedo filtrar por el criterio solicitante. adjunto imagen, me aparece el mensaje la carga fallo, 
favor agradecere me puedan orientar.

saludos desde chile 

![sin titulo](https://user-images.githubusercontent.com/46006378/50242437-1a78ee00-03a9-11e9-93cc-67f8ff9dfc86.png)

trasher:
Please use english only.

nfarias1882:
Hello dear good afternoon, along with health. attached image, I see the message the load failed,
please thank me you can guide me.

greetings from Chile
![ssx](https://user-images.githubusercontent.com/46006378/50243917-45654100-03ad-11e9-9a1d-e151fa2548c6.png)

PR #2547, opened 2017-08-03
Recursion in search engine

yllen:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2546

No change in database just display so can be done in 9.1.7

### DIFF
--- a/inc/commondbtm.class.php
+++ b/inc/commondbtm.class.php
@@ -2482,7 +2482,7 @@ class CommonDBTM extends CommonGLPI {
             return false;
          }
       }
-    
+
       $this->right = $right;
       plugin::doHook('item_can', $this);
       switch ($this->right) {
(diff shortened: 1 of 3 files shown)


remicollet:
Does a massive update will take care to check the **canUnrecurs** check ?

After a quick look, I think canUnrecurs is only checked in the webUI (to propose or not the dropdown), so allowing it in massive action can raise bad issue...

remicollet:
After discussion, we discovered bugs related to canUnrecus
* bug #2548
* bug #2549

So, as bugs exists, I think this PR is acceptable for 9.1, as Massive Actions already allows various bad actions (normally not allowed by WebUI).

yllen:
Done for master https://github.com/glpi-project/glpi/commit/f0d349de975a88848d6d9da576e657475704a603

PR #2550, opened 2017-08-03
Deleted not used variable $extraparam and Fixed #2031 and #2553 

### LINKED ISSUE #2031: [9.1.2] backslashes replaced in ticket description
Hi
When users gives an local url (or any url containing backslash) in a ticket description (send by mail to the mail support) (i.e. : C:\Windows\system32), Glpi replaces the backslashes with no character so for the i.e. above it gaves : C:Windowssystem32

tomolimo:
$extraparam was not used at all, so I deleted it and I deleted the line that assigned a value to it.
Also fixed #2031 and #2553
Moved for Ticket type only the $extraparamhtml mechanism to saveInput() and restoreInput()
Added a new filter for \\\\n to be converted into \n in cleanPostForTextArea
Added a test for this new filter in testCleanPostForTextArea() test method.
Deleted in ticket::showForm() the section that was doing replace of "\\r", "\\n" and "\\" by "" (is done by cleanPostForTextArea)
Changed restoreInput so that it wall call clearSavedInput.

regards,
Tomolimo

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

### DIFF
--- a/inc/commondbtm.class.php
+++ b/inc/commondbtm.class.php
@@ -760,7 +760,7 @@ class CommonDBTM extends CommonGLPI {
          $saved = Html::cleanPostForTextArea($_SESSION['saveInput'][$this->getType()]);
 
          // clear saved data when restored (only need once)
-         unset($_SESSION['saveInput'][$this->getType()]);
+         $this->clearSavedInput();
 
          return $saved;
       }
--- a/inc/commonglpi.class.php
+++ b/inc/commonglpi.class.php
@@ -685,7 +685,6 @@ class CommonGLPI {
 
       $target         = $_SERVER['PHP_SELF'];
       $extraparamhtml = "";
-      $extraparam     = "";
       $withtemplate   = "";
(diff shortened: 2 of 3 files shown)


yllen:
You don't have clean all $extraparam defined

tomolimo:
Exact, I first deleted only the one that was in the function in which I made changes. But now I have deleted all $extraparm that was not used

inc/html.class.php
```diff
@@ -282,12 +282,14 @@ static function cleanPostForTextArea($value) {
          return array_map(array(__CLASS__, __METHOD__), $value);
       }
       $order   = array('\r\n',
+                       '\\\\n',
```
trasher: I do not agree with that change.

This sounds like a workaround than a fix.

trasher:
Replaced with #2682

PR #2571, opened 2017-08-08
Database upgrade test

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | Sure! :D

Test if an upgraded database is identical to a fresh install.

### DIFF
--- a/install/update_91_92.php
+++ b/install/update_91_92.php
@@ -211,7 +211,7 @@ function update91to92() {
    foreach ($product_types as $type) {
       if (class_exists($type . 'Model')) {
          $table = getTableForItemType($type . 'Model');
-         $migration->addField($table, 'product_number', 'string');
+         $migration->addField($table, 'product_number', 'string', ['after' => 'comment']);
          $migration->migrationOneTable($table);
          $migration->addKey($table, 'product_number');
       }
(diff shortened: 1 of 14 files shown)


install/update_91_92.php
```diff
@@ -1870,6 +1870,22 @@ function update91to92() {
...
+   $migration->addKey('glpi_networknames', 'is_deleted');
+   $migration->addKey('glpi_networknames', 'is_dynamic');
+   $migration->addKey('glpi_projects', 'is_template');
+   $migration->addKey('glpi_projecttasks', 'is_template');
+   $migration->dropKey('glpi_savedsearches_users', 'bookmarks_id');
+   $migration->addKey('glpi_savedsearches_users', 'savedsearches_id');
+   //this one was not on the correct field
+   $migration->dropKey('glpi_softwarelicenses', 'is_deleted');
+   $migration->migrationOneTable('glpi_softwarelicenses');
+   $migration->addKey('glpi_softwarelicenses', 'is_deleted');
```
yllen: I don't understand this part.
You drop an index and you create the same

trasher: Because the original created index was not on the right field :/

We had sthing like `KEY is_deleted (is_visible)`

PR #2585, opened 2017-08-09
adjusted width to 140px for date in solution in timeline

### LINKED ISSUE #2579: [9.2-dev] There is a small issue with timeline date position
Hello,

I believe this issue is a CSS one, and as I don't master this subject....

Here is the issue:
In a ticket with an approved solution, the timeline is shown like this:
image

Last digit of the date is missing...

and this is only the case for the date of the solution.

regards,
Tomolimo

tomolimo:
Fixes #2579
adjusted width to 140px for date in solution in timeline

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  |no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2579

### DIFF
--- a/css/styles.css
+++ b/css/styles.css
@@ -3293,7 +3293,6 @@ a.copyright {
 }
 .timeline_history .left .h_date,
 .timeline_history .middle .h_date {
-   width: 140px;
    margin: 0 auto;
 }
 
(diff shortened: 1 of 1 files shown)


trasher:
What about just removing this width?

Issue #2598, opened 2017-08-11
Entity creation with blank name


yllen:
http://forum.glpi-project.org/viewtopic.php?id=158586

PR #2599, opened 2017-08-11
Entitycreation

yllen:
| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | /no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2598

New lang string

### DIFF
--- a/inc/entity.class.php
+++ b/inc/entity.class.php
@@ -232,6 +232,13 @@ class Entity extends CommonTreeDropdown {
    function prepareInputForAdd($input) {
       global $DB;
 
+      $input['name'] = trim($input['name']);
+      if (empty($input["name"])) {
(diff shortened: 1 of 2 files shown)


config/.htaccess
```diff
@@ -1,4 +1,4 @@
-<IfModule mod_authz_core.c>
```
trasher: Typo

yllen: ?? j'ai remis ce qui était dans 9.1/bugfixes

trasher: No, you write `<<` instead of `<`. There would be no diff here if files were the same :)

trasher: I've fixed.

PR #2600, opened 2017-08-11
Issue wddxdeprecated

remicollet:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  |o
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | don't know yet
| Fixed tickets | no

See http://php.net/manual/en/function.set-error-handler.php
The 5th parameter ($errcontext) is deprecated (7.2)
More **wddx** is unsecure, and will probably be deprecated / removed in a near future.

### DIFF
--- a/inc/toolbox.class.php
+++ b/inc/toolbox.class.php
@@ -506,9 +506,8 @@ class Toolbox {
     * @param $errmsg    string   error message.
     * @param $filename  string   filename that the error was raised in.
     * @param $linenum   integer  line number the error was raised at.
-    * @param $vars      array    that points to the active symbol table at the point the error occurred.
    **/
-   static function userErrorHandlerNormal($errno, $errmsg, $filename, $linenum, $vars) {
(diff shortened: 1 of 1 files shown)


inc/toolbox.class.php
```diff
@@ -508,7 +508,7 @@ static function logInFile($name, $text, $force = false) {
     * @param $linenum   integer  line number the error was raised at.
     * @param $vars      array    that points to the active symbol table at the point the error occurred.
    **/
-   static function userErrorHandlerNormal($errno, $errmsg, $filename, $linenum, $vars) {
+   static function userErrorHandlerNormal($errno, $errmsg, $filename, $linenum) {
```
trasher: Since this method is called from `userErrorHandlerDebug`; you should change this one too.

remicollet: good catch

Issue #1076, opened 2016-09-28
'Right' or 'Left' position for ticket items on timeline view


tomolimo:
Hello,
I would like to start a discussion about the rule which decides the position of the items (task, followups,...) on the timeline view.
### The current rule is:

> // check if curent item user is assignee or requester
> 
> > ```
> >   $user_position = 'left';
> >    if ((isset($ticket_users_keys[$item_i['users_id']])
> >          && ($ticket_users_keys[$item_i['users_id']] == CommonItilActor::ASSIGN))
> >         || ($item['type'] == 'Assign')) {
> >        $user_position = 'right';
> >    }
> > ```
> 
> if writer of the item (task, followup,...) is assigned to ticket (=ticket technician) then the position of the item will be on the right

Note: there is a condition `|| ($item['type'] == 'Assign')` which is strange to me as this value is never assigned to the array.
### Current rule problems
1. we may have technicians that write followups and they are not assigned to the ticket. They may belong to a group assigned to (or not).
2. the 'assigned to' may change during the ticket life, and so the timeline will no longer reflect the real exchanges between the requester and the technicians.
3. a watcher may write a followup. A watcher may be a technician or a requester. So where do we put the item written by a watcher, on the left or on the right? Today on the left.
4. a requester may also be a technician in a ticket, so where do we position the items? On the left or on the right?
### Proposed solutions:

Could we imagine that the position of the item is stored into the item itself in DB? Like this at creation, we may decide to set it on the left or the right depending on several rules:
1. Writer's default profile is 'Simplified interface' or not? But as this setting is not mandatory, we may have troubles!
2. Extended rights on this ticket (=read/write on most of the ticket fields) in one of writer's profiles
3. Based on a new field (='to requester': yes/no default no): when set to 'no' then item will be on the right. 
4. Instead of a new field, we may use the request source, and set a new flag in 'request source' dropdown to determine if it's from Technician or not.
5. Based also on the privacy flag: as only technicians can set followup and task as 'private'.
6. Based on the current (=at item creation) ticket assignments. At item (task, followup,...) creation is the writer also assigned to? then item will be set to the right. 
7. based on ticket business rules to be defined with the rule engine.
8. or any other ideas...
### Favourite solution:

And what about if we do the reverse? Instead of setting the position if the writer is 'assigned to', we may decide the position if the writer is 'requester' or 'observer'? In this case set it to the left. 
[...]

PR #2602, opened 2017-08-11
Added timeline_position to following tables: glpi_ticketfollowups, gl…

tomolimo:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1076


tomolimo:
Unit tests are in the last commit

inc/ticket.class.php
```diff
@@ -6707,12 +6765,13 @@ function getTimelineItems() {
...
          $timeline[$solution_date."_solution"]
             = ['type' => 'Solution',
-                    'item' => ['id'               => 0,
-                                    'content'          => Toolbox::unclean_cross_side_scripting_deep($solution_content),
-                                    'date'             => $solution_date,
-                                    'users_id'         => $users_id,
-                                    'solutiontypes_id' => $this->fields['solutiontypes_id'],
-                                    'can_edit'         => Ticket::canUpdate() && $this->canSolve()]];
+                     'item' => ['id'                => 0,
+                                      'content'           => Toolbox::unclean_cross_side_scripting_deep($solution_content),
```
orthagh: CS: align the array lines.
(not your fault originally, but in the current workflow , if we edit a part like this, we realign the arrays)

inc/ticket.class.php
```diff
@@ -6820,12 +6875,23 @@ function showTimeline($rand) {
...
+                  break;
+               case self::TIMELINE_MIDLEFT:
+                  $user_position = 'left middle';
+                  break;
+               case self::TIMELINE_MIDRIGHT:
+                  $user_position = 'right middle';
+                  break;
+               case self::TIMELINE_RIGHT:
+                  $user_position = 'right';
+                  break;
```
orthagh: maybe add a default case (TIMELINE_RIGHT), to prevent potential empty field, also, add an else statement (for the isset) to do the same thing

inc/ticket.class.php
```diff
@@ -7136,14 +7206,12 @@ function getTicketActors() {
...
-                WHERE `type` != ".CommonItilActor::OBSERVER."
-                GROUP BY `users_id`
-                ORDER BY `type` DESC";
+                ";
 
       $res               = $DB->query($query);
       $ticket_users_keys = [];
       while ($current_tu = $DB->fetch_assoc($res)) {
-         $ticket_users_keys[$current_tu['users_id']] = $current_tu['type'];
+         $ticket_users_keys[$current_tu['users_id']][] = $current_tu['type'];
```
orthagh: i don't understand this change

tests/units/Ticket.php
```diff
@@ -1077,4 +1077,174 @@ public function testUpdateFollowup() {
...
+         $this->integer(
+            (int)$val->fields['timeline_position']
+         )->isEqualTo($user['pos']);
+      }
+   }
+
+   protected function _testGetTimelinePositionSolution($tlp, $tickets_id) {
+      foreach ($tlp as $users_name => $user) {
+         $auth = new \Auth();
+         $this->boolean((boolean)$auth->login($users_name, $user['pass'], true))->isTrue();
```
orthagh: maybe factoring, DbTestCase::login (Ticket extends that class) could be enhanced to support username/password parameters

tests/units/Ticket.php
```diff
@@ -1077,4 +1077,174 @@ public function testUpdateFollowup() {
...
+         $this->integer(
+            (int)$tkt->getTimelinePosition($tickets_id, 'Solution', $uid)
+         )->isEqualTo($user['pos']);
+      }
+   }
+
+   function testGetTimelinePosition() {
+      // login glpi
+      $auth = new \Auth();
+      $this->boolean((boolean)$auth->login('glpi', 'glpi', true))->isTrue();
```
orthagh: ``$this->login()``

tests/units/Ticket.php
```diff
@@ -1077,4 +1077,174 @@ public function testUpdateFollowup() {
...
+      return $ticket->getID();
+   }
+
+   function testGetTimelineItems() {
+
+      $tkt_id = $this->testGetTimelinePosition();
+
+      // login glpi
+      $auth = new \Auth();
+      $this->boolean((boolean)$auth->login('glpi', 'glpi', true))->isTrue();
```
orthagh: ``$this->login()``

inc/ticket.class.php
tomolimo: default value is here: https://github.com/tomolimo/glpi/blob/ccda0aef3271788b34ee382c57af7de8bb20f3d7/inc/ticket.class.php#L6879

inc/ticket.class.php
tomolimo: Needed to group users by assignment type: requester, observer, or technicians

tests/units/Ticket.php
tomolimo: DbTestCase::login doesn't permit login with 'glpi' user, but only with TU_USER

tests/units/Ticket.php
tomolimo: DbTestCase::login doesn't permit login with 'glpi' user, but only with TU_USER

tests/units/Ticket.php
tomolimo: DbTestCase::login doesn't permit login with 'glpi' user, but only with TU_USER

tests/units/Ticket.php
orthagh: that's exactly my point, i suggest a enhancement oh this function

tests/units/Ticket.php
orthagh: for this exact test , no need for glpi i think

tests/units/Ticket.php
orthagh: same as above

inc/ticket.class.php
orthagh: ok right

tests/units/Ticket.php
tomolimo: sorry I didn't understand, I agree, I'm going to extend it

inc/ticket.class.php
orthagh: right, i saw it

tests/units/Ticket.php
tomolimo: I wanted to be sure to get all items

trasher: The test user is superadmin; it should have all required rights ;)

tomolimo: then should I use TU_USER or glpi?

orthagh: only ``$this->login`` (So TU_USER)

PR #2607, opened 2017-08-17
Undefined index; fixes #2592

### LINKED ISSUE #2592: [9.2-dev] PHP notice after migration Undefined index: glpiactiveentities_string in C:\Projects\glpi9.2\inc\db.function.php at line 1740
Hello,

Doing a migration from 9.1.6 empty DB to 9.2-dev release, at the end I got the following PHP notices:

> PHP Notice: Undefined index: glpiactiveentities_string in C:\Projects\glpi9.2\inc\db.function.php at line 1740
> PHP Notice: Undefined index: glpiparententities in C:\Projects\glpi9.2\inc\db.function.php at line 1753
 
image

regards,
Tomolimo

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2592

### DIFF
--- a/inc/db.function.php
+++ b/inc/db.function.php
@@ -1736,6 +1736,10 @@ function getEntitiesRestrictRequest($separator = "AND", $table = "", $field = ""
    if (is_array($value)) {
       $query .= " IN ('" . implode("','", $value) . "') ";
    } else {
+      if (strlen($value) == 0 && !isset($_SESSION['glpiactiveentities_string'])) {
+         //set root entity if not set
+         $value = 0;
+      }
       if (strlen($value) == 0) {
          $query .= " IN (".$_SESSION['glpiactiveentities_string'].") ";
       } else {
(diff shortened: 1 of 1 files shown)


inc/db.function.php
```diff
@@ -1736,6 +1736,10 @@ function getEntitiesRestrictRequest($separator = "AND", $table = "", $field = ""
    if (is_array($value)) {
       $query .= " IN ('" . implode("','", $value) . "') ";
    } else {
+      if (strlen($value) == 0 && !isset($_SESSION['glpiactiveentities_string'])) {
+         //set root entity if not set
+         $value = 0;
+      }
       if (strlen($value) == 0) {
```
orthagh: regarding the original issue, if ``$_SESSION['glpiactiveentities_string']`` is not set, we still pass in this line and the notice will still trigger.

Maybe we could replace ``if`` by ``else if`` or move the ``!isset(...)`` part into the following block.

trasher: When `glpiactiveentities_string` does not exists; `$value` ist set to `0`; we therefore pass in the else part and work with root entity only.

orthagh: ok, you're right.

orthagh: strlen on an integer is not very kind to read

trasher: Use `trim($value) === ''` instead?

orthagh: not much better.
Keep as actually, i approve

Issue #2587, opened 2017-08-09
[9.2-dev] can't add document to followups in ticket timeline


tomolimo:
Hello,

When in a ticket, adding a new followup, if you attach a document to this followup, it will not be linked to the followup.

regards,
Tomolimo

PR #2610, opened 2017-08-17
Fix 2587

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2587


trasher:
I've unset the name parameter to fix ile attachment. I wonder if this is a good iedea to permit this to be changed anyways... File upload seems to always been waiting for a `filename` named field.

PR #2620, opened 2017-08-20
handle plugins CSS requirements

btry:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

If a plugin requires a library provided by GLPI, and this library requires an additional CSS, GLPI does not adds the CSS in the HEAD tag of the HTML.

Example:

The following loads chartist library, but the CSS included in Chartist is not added.

 ```php
// setup.php of a plugin
function plugin_init_flyvemdm() {
   // ...
   $CFG_GLPI['javascript']['plugins']['pluginflyvemdmmenu']['Menu'] = ['charts'];
   // ...
}
```

```php
// front/menu.php of the plugin

Html::header(PluginFlyvemdmMenu::getTypeName(Session::getPluralNumber()), '', 'admin', 'PluginFlyvemdmMenu', 'Menu');

```

### DIFF
--- a/inc/html.class.php
+++ b/inc/html.class.php
@@ -1152,7 +1152,11 @@ class Html {
                   $jslibs = $CFG_GLPI['javascript'][$sector][$item];
                }
             } else {
-               $jslibs = $CFG_GLPI['javascript'][$sector];
+               if (isset($CFG_GLPI['javascript']['plugins'][$item][$option])) {
+                  $jslibs = $CFG_GLPI['javascript']['plugins'][$item][$option];
+               } else {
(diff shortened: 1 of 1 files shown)


btry:
My bad, I confused the 2nd level of the array in $CFG_GLPI['javascript']. If I fix setup.php as the following, charts works again.

 ```php
// setup.php of a plugin
function plugin_init_flyvemdm() {
   // ...
   $CFG_GLPI['javascript']['admin']['pluginflyvemdmmenu']['Menu'] = ['charts'];
   // ...
}
```

Issue #2631, opened 2017-08-22
[9.2] Php-Errors on devices list on a computer form tab


tsmr:
When you are on the Devices tab in a computer, you have location and status dropdown menus and they do not display correctly and there are log errors

```
2017-08-22 12:08:30 [6@P365]
  *** PHP Warning(2): call_user_func_array() expects parameter 1 to be a valid callback, class 'Dropdown' does not have a method 'Number'
  Backtrace :
  inc\htmltableentity.class.php:183                  
  inc\htmltablecell.class.php:268                    HTMLTableEntity->displayContent()
  inc\htmltablerow.class.php:185                     HTMLTableCell->displayCell()
  inc\htmltablegroup.class.php:243                   HTMLTableRow->displayRow()
  inc\htmltablemain.class.php:261                    HTMLTableGroup->displayGroup()
  inc\item_devices.class.php:493                     HTMLTableMain->display()
  inc\item_devices.class.php:372                     Item_Devices::showForItem()
  inc\commonglpi.class.php:482                       Item_Devices::displayTabContentForItem()
  ajax\common.tabs.php:96                            CommonGLPI::displayStandardTab()
```

Issue #2629, opened 2017-08-22
[9.2] Devices Navigation


tsmr:
When you use arrow for navigate into components (devices) : 

![image](https://user-images.githubusercontent.com/13178158/29564447-25c6c744-8742-11e7-8c43-ea86ee56433f.png)

You have Fatal error 

Fatal error: Uncaught RuntimeException: Missing or incorrect device type called! in ..\front\device.form.php:36 Stack trace: #0 {main} thrown in ..\front\device.form.php on line 36

PR #2632, opened 2017-08-22
Fixes

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2631 #2629

### DIFF
--- a/front/device.form.php
+++ b/front/device.form.php
@@ -37,6 +37,8 @@ if (!isset($_GET['itemtype']) || !class_exists($_GET['itemtype'])) {
       'Missing or incorrect device type called!'
    );
 }
-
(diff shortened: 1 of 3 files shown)


front/device.form.php
```diff
@@ -37,6 +37,8 @@
       'Missing or incorrect device type called!'
    );
 }
-
+$options = [
+   'itemtype'  => $_GET['itemtype']
```
orthagh: double spaces before ``=>``

trasher: Amended

Issue #1843, opened 2017-02-20
Search in LDAP import form


yllen:
http://forum.glpi-project.org/viewtopic.php?id=156722

No LDAP to test and reproduce

PR #2636, opened 2017-08-23
Prevent existing users to be count in import display

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes


trasher:
Commit https://github.com/glpi-project/glpi/pull/2636/commits/1b5b10adb162c51200ad46de42c6d48618709472 must probably be backported to 9.1/bf branch. This is a side effect of the fix for #1843

PR #2640, opened 2017-08-24
Allowed to select multiple files in file dialog

edgardmessias:
Currently only allow send multiple files with Drag and Drop.

This change allow select multiple files in file dialog.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2640


trasher:
Thank you, but this part has been completely rewritten in 9.2; please open another PR against the 9.2 version to get that implemented.

PR #2671, opened 2017-08-30
Add is_deleted, is_dynamic and entities ID for OS; fixes #2630

### LINKED ISSUE #2630: [9.2] import of operating systems
I think that we need to add fields into glpi_items_operatingsystems table like other tables ?

`is_deleted` tinyint(1) NOT NULL DEFAULT '0',
  `is_dynamic` tinyint(1) NOT NULL DEFAULT '0',
  `entities_id` int(11) NOT NULL DEFAULT '0',

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2630

### DIFF
--- a/install/update_91_92.php
+++ b/install/update_91_92.php
@@ -1220,6 +1220,33 @@ Regards,',
       $migration->dropField('glpi_computers', 'os_kernel_version');
    }
 
+   if (!$DB->fieldExists('glpi_items_operatingsystems', 'is_deleted')) {
+      $migration->addField(
+         'glpi_items_operatingsystems',
+         'is_deleted',
+         "tinyint(1) NOT NULL DEFAULT '0'"
+      );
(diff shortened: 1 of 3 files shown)


install/update_91_92.php
```diff
@@ -1908,6 +1908,33 @@ function update91to92() {
    $migration->addKey('glpi_tasktemplates', 'users_id_tech');
    $migration->addKey('glpi_tasktemplates', 'groups_id_tech');
 
+   if (!$DB->fieldExists('glpi_items_operatingsystems', 'is_deleted')) {
```
orthagh: just one point, should this update part be placed after operatingsystems stuff, like ~ l1220

trasher: sure :)

PR #2672, opened 2017-08-30
Fixed missing scriptEnd at the end of Html::initEditorSystem function

tomolimo:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

### DIFF
--- a/inc/html.class.php
+++ b/inc/html.class.php
@@ -3880,7 +3880,6 @@ class Html {
       }
 
       // init tinymce
-      Html::scriptStart();
       $js = "$(function() {
          tinyMCE.init({
             language: '$language',


inc/html.class.php
```diff
@@ -3920,9 +3920,9 @@ static function initEditorSystem($name, $rand = '', $display = true, $readonly =
       });";
 
       if ($display) {
-         echo  Html::scriptBlock($js);
+         echo  Html::scriptEnd($js);
```
trasher: `scriptEnd` takes no arguments.

I'm not sure what is the best between your proposal, and keep `scriptBlock` and remove `scriptStart`. @orthagh ?

tomolimo: I agree, the question is real, but we should choose one or the other, but of course not a mix :)
It seems that I've been a bit too quick :)

trasher: Huh and by the way... With current proposal, `$js` is never sent to the buffer; it should not works

trasher: Looking at the code; `scriptBlock` is used about 69 times, where `scriptStart|End` is used only twice.

tomolimo:
you're welcome

PR #2683, opened 2017-09-01
api rest: add option lookup_recursive for getMyEntitiesMaster

garnier-quentin:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | not

With the webservice, we cannot get the complete list of entities. So an option, it's added to get the complete list (recursive entities).

### DIFF
--- a/inc/api.class.php
+++ b/inc/api.class.php
@@ -313,23 +313,23 @@ abstract class API extends CommonGLPI {
          $params['lookup_recursive'] = false;
       }
 
-      $myentities = array();
+      $myentities = [];
       foreach ($_SESSION['glpiactiveprofile']['entities'] as $entity) {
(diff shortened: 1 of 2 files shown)


trasher:
Documentation (apirest.md file) may be updated according to the change

garnier-quentin:
You re right. It's done.

inc/api.class.php
```diff
@@ -305,15 +305,29 @@ protected function changeActiveEntities($params = []) {
     *
     * @return array of entities (with id and name)
     */
-   protected function getMyEntities() {
+   protected function getMyEntities($params = []) {
```
orthagh: corresponding docblock missing

garnier-quentin:
Change done! Thanks :)

PR #2691, opened 2017-09-04
Feature/many entities

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

Some improvements on performances with many entities

### DIFF
--- a/ajax/getDropdownValue.php
+++ b/ajax/getDropdownValue.php
@@ -59,6 +59,9 @@ if (isset($_POST["entity_restrict"])
    }
    $_POST["entity_restrict"] = $entities;
 }
+if (isset($_POST['entity_restrict']) && 'default' === $_POST['entity_restrict']) {
+   $_POST['entity_restrict'] = $_SESSION['glpiactiveentities'];
+}
 
 // Security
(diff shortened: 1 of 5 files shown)


trasher:
This one should probably be backported to the 9.1 branch.

inc/session.class.php
```diff
@@ -520,7 +520,7 @@ static function loadGroups() {
                    LEFT JOIN `glpi_groups` ON (`glpi_groups_users`.`groups_id` = `glpi_groups`.`id`)
                    WHERE `glpi_groups_users`.`users_id`='" . self::getLoginUserID() . "' " .
                          getEntitiesRestrictRequest(" AND ", "glpi_groups", "entities_id",
-                                                    $_SESSION['glpiactiveentities'], true);
+                                                    $_SESSION['glpiactiveentities'], false);
```
orthagh: with 3520a42fa53318ee0c1afe6d747e81ee7ff06c0c commit, i think we shouldn't do this anymore, no ?

trasher: Indeed, this was not OK anyways :)

PR #2692, opened 2017-09-05
Don't use filename

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

The goal is not to use the real filename, this would prevent rejected file to be accessed while they're still in temporay directory.


trasher:
No cherry-pick for this one... Uploads have been rewrite for 9.2; need another PR

PR #2708, opened 2017-09-07
fix manual update

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

with the introduction of upgrade from stable to dev, manual upgrade (config/config.db.php absent) didn't work anymore.
I am forcing the missing parameter in hidden, because the code of detection of old and new version seems complex to factorize, your thoughts ?

### DIFF
--- a/inc/config.class.php
+++ b/inc/config.class.php
@@ -2589,4 +2589,26 @@ class Config extends CommonDBTM {
       return $values;
    }
 
+   /**
+    * Get message that informs the user he's using a development version
+    *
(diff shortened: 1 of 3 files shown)


trasher:
Doing the same as the standard update may not be possible; and would be hard.

The only thing that should cause a problem, is that the checkbox is never shown to the user in that case. Maybe should we just always display it if we're in a dev version?

PR #2710, opened 2017-09-07
Feature/entities apcu

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

Use APCu cache for entities sons and ancestors.

### DIFF
--- a/inc/db.function.php
+++ b/inc/db.function.php
@@ -342,7 +342,7 @@ function countElementsInTable($table, $condition = "") {
    $condition['COUNT'] = 'cpt';
 
    $row = $DB->request($table, $condition)->next();
-   return ($row ? $row['cpt'] : 0);
+   return ($row ? (int)$row['cpt'] : 0);
 }
 
 /**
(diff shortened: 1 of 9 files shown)


inc/db.function.php
```diff
@@ -1869,3 +1904,11 @@ function getEntitiesRestrictCriteria($table = '', $field = '', $value = '',
...
    return $crit;
 }
+
+function useCache() {
+   if (function_exists('apcu_exists') && (!defined('TU_USER') || defined('CACHED_TESTS'))) {
+      return true;
+   } else {
+      return false;
+   }
+}
```
orthagh: we could use directly

```php
return function_exists('apcu_exists') && (!defined('TU_USER') || defined('CACHED_TESTS'));
```

inc/db.function.php
```diff
@@ -786,65 +817,70 @@ function getSonsOf($table, $IDf) {
...
-             FROM `$table`
-             WHERE `$parentIDfield` = '$IDf'
-             ORDER BY `name`";
+   if (!count($sons)) {
+      // IDs to be present in the final array
+      $sons[$IDf] = "$IDf";
+      // current ID found to be added
+      $found = [];
+      // First request init the  varriables
+      $query = "SELECT `id`
```
orthagh: maybe replace this part by the iterator like below

PR #2713, opened 2017-09-08
Allowed to select multiple files in file dialog (See #2640)

edgardmessias:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2640

### DIFF
--- a/inc/document_item.class.php
+++ b/inc/document_item.class.php
@@ -624,7 +624,7 @@ class Document_Item extends CommonDBRelation{
       if ($item->getType() == 'Ticket') {
          echo "<input type='hidden' name='tickets_id' value='".$item->getID()."'>";
       }
-      Html::file();
+      Html::file(['multiple' => true]);
       echo "</td><td class='left'>(".Document::getMaxUploadSize().")&nbsp;</td>";
       echo "<td></td></tr>";
    }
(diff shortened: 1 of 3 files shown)


trasher:
Hello,

This seems OK globally, but I wonder if you did not miss one or 2 changes:
```diff
diff --git a/inc/document.class.php b/inc/document.class.php
index fc60923..795eb19 100644
--- a/inc/document.class.php
+++ b/inc/document.class.php
@@ -437,7 +437,7 @@ class Document extends CommonDBTM {
 
       echo "<td>".sprintf(__('%1$s (%2$s)'), __('File'), self::getMaxUploadSize())."</td>";
       echo "<td>";
-      Html::file();
+      Html::file(['multiple' => true]);
       echo "</td></tr>";
 
       $this->showFormButtons($options);
```

And another call to `Html::file()` in the Html class (~ line 5374).

edgardmessias:
In `document.class.php` is used in `front/document.form.php` for replace one file only.

And I will edit Html class (~ line 5374).

trasher:
Hum... So we cannot upload multiple files in once using the document form?

That may be OK, but this sounds a bit strange this is the only place we have to send document one by one. Any thoughts?

Nota: in edition mode, this should not be multiple anyways (so my patch is not complete).

PR #2714, opened 2017-09-08
collect mails with uid instead index

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | to fill (need check backlog)

Ok this is change which can finish all random issues with mail receiver.
Here is the scenario which needed this change.
- it a connection on gmail with imap
- on imap_delete, classic imap server (not gmail) will only mark the message for deletion, an imap_expunge will be necessary for real deletion. Gmail remove the message from INBOX (or any collected folder/label) directly on imap_delete and so on the next loop of message fetching, all emails have their index decreased).
- with this behavior, we have some warnings on imap_header and the current message is not parsed.
- additionally, on the next receiver fetch, a message already computed in ticket could be done again.

So i changed all imap_ functions that handles a message. They now use the message UID instead of the index in the folder.

### DIFF
--- a/inc/mailcollector.class.php
+++ b/inc/mailcollector.class.php
@@ -55,8 +55,8 @@ class MailCollector  extends CommonDBTM {
    public $charset         = "";
    /// IMAP / POP connection
    public $marubox         = '';
-   /// ID of the current message
-   public $mid             = -1;
+   /// UID of the current message
+   public $uid             = -1;
    /// structure used to store the mail structure
    public $structure       = false;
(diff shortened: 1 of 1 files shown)


orthagh:
i didn't find any corresponding issues in the current backlog (apart the internal teclib ticket)

trasher:
I do not remember any similar issues

PR #2718, opened 2017-09-11
optimise loading time in entity jstree

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

A wip for entity selector performance (currently testing on a flat tree of 37k nodes)

present optimisations atm:

- json send by ajax size divided by 3 (5Mo+ -> 1.7Mo)
- don't append additional html with ``redraw.jstree`` event
- some tests with [qload plugin](https://github.com/vakata/jstree/pull/1809)

for information, here is some 'benchmarks' with theses 37k entities on loading first sub node with chrome browser: 
- 9.2/bugfixes: 40s 
- base optimisations (without qload) -> 12s
- with qload -> 2s


inc/planning.class.php
```diff
@@ -660,6 +660,7 @@ static function showPlanning($fullview = true) {
                }
 
                // add classes to current event
+               added_classes = '';
```
orthagh: just an error detected when working on entity selector, it's non blocking

PR #2720, opened 2017-09-12
Migrate db.function into an object

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

Goals are:
- drop procedural remaining functions,
- get code coverage for the tests.

All current migrated classes are the ones that were tested directly. Some have been refactored a bit.


inc/db.function.php
```diff
@@ -1606,16 +1214,15 @@ function closeDBConnections() {
 /**
  * Format a web link adding http:// if missing
  *
- *@param $link link to format
+ * @param $link link to format
+ *
+ * @return formatted link.
  *
- *@return formatted link.
+ * @deprecated 9.2 Use DbUtils::isIndex()
```
orthagh: "use Toolbox::formatOutputWebLink"

inc/db.function.php
```diff
@@ -1832,77 +1376,29 @@ function getEntitiesRestrictRequest($separator = "AND", $table = "", $field = ""
...
 }
 
+/**
+ * Should APCu cache be used
+ *
+ * @return boolean
+ *
+ * @deprecated @see DbUtils::useCache()
+ */
 function useCache() {
```
orthagh: I'm not sure about this function should be here (or in DbUtils), maybe better in Toolbox or CommonGLPI

inc/dbutils.class.php
```diff
@@ -0,0 +1,903 @@
...
+    * Return the plural of a string
+    *
+    * @param string $string input string
+    *
+    * @return string plural of the parameter string
+    */
+   public function getPlural($string) {
+      $rules = [
+         //'singular'         => 'plural'
+         'criterias$'         => 'criterias',// Special case (criterias) when getPlural is called on already plural form
```
orthagh: sic, don't kown where it's used, but later (not now) we should investigate why and replace the couple by [criterion->criteria]

tests/units/DbUtils.php
```diff
@@ -345,6 +488,22 @@ function() {
 
 
    public function testIsIndex() {
+      $this
+         ->if($this->newTestedInstance)
+         ->then
+            ->boolean($this->testedInstance->isIndex('glpi_configs', 'fakeField'))->isFalse()
+            ->boolean($this->testedInstance->isIndex('glpi_configs', 'name'))->isFalse()
+            ->boolean($this->testedInstance->isIndex('glpi_users', 'locations_id'))->isTrue()
+            ->boolean($this->testedInstance->isIndex('glpi_users', 'unicity'))->isTrue()
```
orthagh: here; maybe add a real field but not indexed

inc/dbutils.class.php
trasher: :D

inc/db.function.php
trasher: Indeed; I'll move it to Toolbox (commondbtm seems too restrictive as well).
And since this is a very new function, I guess it should be dropped entirely from db.function.php.

inc/db.function.php
```diff
@@ -170,6 +117,8 @@ function getItemTypeForTable($table) {
  * @param string $fkname
  *
  * @return string ItemType name for the fkname parameter
+ *
+ * @deprecated 9.2 see DbUtils::getItemtypeForForeignKeyField()
  */
 function getItemtypeForForeignKeyField($fkname) {
    $table = getTableNameForForeignKeyField($fkname);
```
orthagh: miss the redirection i think

trasher: I forgot to add in DbUtils...

PR #2728, opened 2017-09-12
Restrict saved searches visibility; fixes #2725

### LINKED ISSUE #2725: [9.2] Saved Search (bookmarks) list: rights issue
Hello,

The list of saved searches must be filtered by user and the rights of the current profile.

ex this profile:

image

A user with such profile should be able to see only his saved searches and the public saved searches only, and not the one of the other users.

image

And here is the error that appears when such user clicks on a private saved search:

> Unhandled exception at line 179, column 1 in script block
> 
> 0x800a01b6 - JavaScript runtime error: Object doesn't support property or method 'select2'

and the screens are like this:

image

regards,
Tomolimo

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2725

### DIFF
--- a/inc/savedsearch.class.php
+++ b/inc/savedsearch.class.php
@@ -146,7 +146,7 @@ class SavedSearch extends CommonDBTM {
    function canCreateItem() {
 
       if ($this->fields['is_private'] == 1) {
-         return (Session::haveRight('config', CREATE)
+         return (Session::haveRight('config', UPDATE)
                  || $this->fields['users_id'] == Session::getLoginUserID());
       }
       return parent::canCreateItem();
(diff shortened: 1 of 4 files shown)


inc/savedsearch.class.php
```diff
@@ -1319,4 +1319,20 @@ public function createNotif() {
...
+    * Return visibility SQL restriction to add
+    *
+    * @return string restrict to add
+   **/
+   static function addVisibilityRestrict() {
+      $restrict = self::getTable() .'.is_private=1 AND ' . self::getTable() .
+         '.users_id='.Session::getLoginUserID();
+
+      if (Session::haveRight(self::$rightname, READ)) {
+         $restrict .= ' OR ' . self::getTypeName() . '.is_private=0';
```
tomolimo: getTypeName should be getTable

inc/savedsearch.class.php
```diff
@@ -345,14 +345,9 @@ function showForm($ID, $options = []) {
...
-      if ($ID > 0) {
-         $this->check($ID, UPDATE);
-      } else {
-         $this->check(-1, CREATE);
-      }
+      $this->initForm($ID, $options);
+      $this->showFormHeader($options);
 
-      echo "<form method='post' name='form_save_query' action='".$this->getFormURL()."'>";
       echo "<div class='center'>";
```
orthagh: this div will be inside of table

PR #2743, opened 2017-09-13
Fix lock objects

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

* 1st commit is just to have a space between "Unlock" and the itemtype
* 2nd one is a fix for regression added by https://github.com/glpi-project/glpi/commit/ea2d6c52998f250d21c8ef59cd8a82cb9661eb80#diff-0ea1de57520847d97d3ea2579cbdd928R690, param is_locked is not passed to the tab anymore due to the removal of $extraparamhtml. The regression results in a locked object, the lock message appears correctly, but fields are not set in readonly

### DIFF
--- a/inc/commonglpi.class.php
+++ b/inc/commonglpi.class.php
@@ -690,6 +690,10 @@ class CommonGLPI {
          if ($this->getType() == 'Ticket') {
             $this->input = $cleaned_options;
             $this->saveInput();
+            if (isset($cleaned_options['locked'])) {
+               $extraparamhtml = "&amp;".Toolbox::append_params(['locked' => $cleaned_options['locked']],
+                                                                '&amp;');
+            }
          } else {
             $extraparamhtml = "&amp;".Toolbox::append_params($cleaned_options, '&amp;');
          }
--- a/inc/objectlock.class.php
+++ b/inc/objectlock.class.php
@@ -173,8 +173,8 @@ class ObjectLock extends CommonDBTM {
 
       $msg = "<table><tr><td class=red>";
       $msg .= __("Locked by you!")."</td>";
-      $msg .= "<td><a class='vsubmit' onclick='javascript:unlockIt(this);'>".__('Unlock ').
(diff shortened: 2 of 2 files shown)


inc/objectlock.class.php
```diff
@@ -173,7 +173,7 @@ function callUnlock( ) {
 
       $msg = "<table><tr><td class=red>";
       $msg .= __("Locked by you!")."</td>";
-      $msg .= "<td><a class='vsubmit' onclick='javascript:unlockIt(this);'>".__('Unlock ').
+      $msg .= "<td><a class='vsubmit' onclick='javascript:unlockIt(this);'>".__('Unlock ')."&nbsp;".
```
tomolimo: 'Unlock ' should be 'Unlock' and ' #' should be "\&nbsp; #" to preserve only one space between words?

inc/commonglpi.class.php
```diff
@@ -690,6 +690,10 @@ function showTabsContent($options = []) {
          if ($this->getType() == 'Ticket') {
             $this->input = $cleaned_options;
             $this->saveInput();
+            if (isset($cleaned_options['is_locked'])) {
```
tomolimo: option is currently called 'locked' and not 'is_locked', see https://github.com/glpi-project/glpi/blob/8b53585c0930deb5bd8519a23c81af3746ed65b1/inc/objectlock.class.php#L454

inc/commonglpi.class.php
```diff
@@ -690,6 +690,10 @@ function showTabsContent($options = []) {
...
+               $extraparamhtml = "&amp;".Toolbox::append_params([$cleaned_options['is_locked']],
```
tomolimo: the option must be like ['locked' => 1] and not [$cleaned_options['is_locked']], otherwise the extraparam is '0=1' when it should be 'locked=1'

orthagh:
yes it's a good idea

PR #2746, opened 2017-09-14
Use ZendCache everywhere

### LINKED ISSUE #2712: Support other cache methods than APCu
Currently, locales are cached using APCu behind Zend Cache; but if APCu is working very well on linux boxes; it is still considered as beta on Windows (with IIS at least). And one may already have a caching system setted up he wants to use.

Zend Cache supports a lot of other adapters we may use:
https://docs.zendframework.com/zend-cache/storage/adapter/

We may add a parameter so we can use any extra supported adapter.

trasher:
Store overridable configuration; fixes #2712
Change cache prefix

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2712

Still use APCu per default, but along with Zend Cache.

### DIFF
--- a/inc/commontreedropdown.class.php
+++ b/inc/commontreedropdown.class.php
@@ -176,6 +176,7 @@ abstract class CommonTreeDropdown extends CommonDropdown {
 
 
    function prepareInputForUpdate($input) {
+      global $GLPI_CACHE;
 
       if (isset($input[$this->getForeignKeyField()])) {
          // Can't move a parent under a child
@@ -188,8 +189,8 @@ abstract class CommonTreeDropdown extends CommonDropdown {
             $input["ancestors_cache"] = '';
             if (Toolbox::useCache()) {
                $ckey = $this->getTable() . '_ancestors_cache_' . $this->getID();
-               if (apcu_exists($ckey)) {
-                  apcu_delete($ckey);
+               if ($GLPI_CACHE->hasItem($ckey)) {
+                  $GLPI_CACHE->removeItem($ckey);
                }
             }
             return $this->adaptTreeFieldsFromUpdateOrAdd($input);
@@ -210,13 +211,13 @@ abstract class CommonTreeDropdown extends CommonDropdown {
     * @param $changeParent
    **/
    function regenerateTreeUnderID($ID, $updateName, $changeParent) {
-      global $DB;
+      global $DB, $GLPI_CACHE;
 
       //drop from sons cache when needed
       if ($changeParent && Toolbox::useCache()) {
          $ckey = $this->getTable() . '_ancestors_cache_' . $ID;
-         if (apcu_exists($ckey)) {
-            apcu_delete($ckey);
+         if ($GLPI_CACHE->hasItem($ckey)) {
+            $GLPI_CACHE->removeItem($ckey);
          }
       }
 
@@ -276,13 +277,13 @@ abstract class CommonTreeDropdown extends CommonDropdown {
    /**
     * Clean sons of all parents from caches
     *
-    * @param null|integer $id   Parent id to clean. Default to current id
-    * @param boolean      $apcu Whether to clean APCu cache (defaults to true)
+    * @param null|integer $id    Parent id to clean. Default to current id
+    * @param boolean      $cache Whether to clean cache (defaults to true)
     *
     * @return void
     */
-   protected function cleanParentsSons($id = null, $apcu = true) {
-      global $DB;
+   protected function cleanParentsSons($id = null, $cache = true) {
+      global $DB, $GLPI_CACHE;
 
       if ($id === null) {
          $id = $this->getID();
@@ -302,14 +303,14 @@ abstract class CommonTreeDropdown extends CommonDropdown {
       $DB->query($query);
 
       //drop from sons cache when needed
-      if ($apcu && Toolbox::useCache()) {
+      if ($cache && Toolbox::useCache()) {
          foreach ($ancestors as $ancestor) {
             $ckey = $this->getTable() . '_sons_cache_' . $ancestor;
-            if (apcu_exists($ckey)) {
-               $sons = apcu_fetch($ckey);
+            if ($GLPI_CACHE->hasItem($ckey)) {
+               $sons = $GLPI_CACHE->getItem($ckey);
                if (isset($sons[$this->getID()])) {
                   unset($sons[$this->getID()]);
-                  apcu_store($ckey, $sons);
+                  $GLPI_CACHE->replaceItem($ckey, $sons);
                }
             }
          }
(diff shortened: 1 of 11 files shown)


inc/config.php
```diff
@@ -45,6 +45,25 @@
 Session::setPath();
 Session::start();
 
+//init cache
+try {
+   $cache_conf = GLPI_CACHE_CONFIG;
+   if (!isset($cache_conf['adapter']['options'])
+       || !isset($cache_conf['adapter']['options']['namespace'])) {
+      $cache_conf['options']['namespace'] = 'glpicache';
```
remicollet: is this a typo ? (missing [adapter] level)

inc/config.php
```diff
@@ -45,6 +45,25 @@
...
+   }
+   if (!defined('TU_USER') || defined('CACHED_TESTS')) {
+      $GLPI_CACHE = Zend\Cache\StorageFactory::factory($cache_conf);
+   } else {
+      $GLPI_CACHE = false;
+   }
+} catch (\Zend\ServiceManager\Exception\ServiceNotFoundException $e) {
```
remicollet: IIRC, exception name have changed in some version, probably better to catch all

inc/config.php
```diff
@@ -45,6 +45,25 @@
...
+try {
+   $cache_conf = GLPI_CACHE_CONFIG;
+   if (!isset($cache_conf['adapter']['options'])
+       || !isset($cache_conf['adapter']['options']['namespace'])) {
+      $cache_conf['options']['namespace'] = 'glpicache';
+   }
+   if (!defined('TU_USER') || defined('CACHED_TESTS')) {
+      $GLPI_CACHE = Zend\Cache\StorageFactory::factory($cache_conf);
+   } else {
+      $GLPI_CACHE = false;
```
remicollet: probably more robust to move default value above (before try), for now there's still run case without initialized value.

inc/commontreedropdown.class.php
```diff
@@ -188,8 +189,8 @@ function prepareInputForUpdate($input) {
             $input["ancestors_cache"] = '';
             if (Toolbox::useCache()) {
                $ckey = $this->getTable() . '_ancestors_cache_' . $this->getID();
-               if (apcu_exists($ckey)) {
-                  apcu_delete($ckey);
+               if ($GLPI_CACHE->hasItem($ckey)) {
+                  $GLPI_CACHE->removeItem($ckey);
                }
```
remicollet: but GPI_CACHE may be "false" ?

trasher: This is handle from Toolbox::useCache()

inc/config.php
trasher: No, the goal is to force namespace when not specified in GLPI_CACHE_CONFIG (see define.php)

inc/config.php
trasher: Oh... I was not aware of that. I'll change

inc/config.php
remicollet: but you test isset($cache_conf['adapter']['options']['namespace']) and then set $cache_conf['options']['namespace'] , this looks inconsistent

inc/config.php
remicollet: see https://github.com/glpi-project/glpi/commit/9474d574

inc/config.php
trasher: Indeed, you are right! Sorry :)

trasher: Fixed

inc/config.php
trasher: Done

inc/commontreedropdown.class.php
trasher: Te make things clear: the `Toolbox::useCache()` will return false if `$GLPI_CACHE` is false. The only exception is from tests, when the `CACHED_TESTS` constant is defined (that's why I had to declare `$GLPI_CACHE` in the tests).

remicollet: ok

inc/define.php
```diff
@@ -463,3 +463,7 @@
 //Maximum time, in miliseconds a saved search should not exeed
 //so we count it on display (using automatic mode).
 $CFG_GLPI['max_time_for_count'] = 200;
+
+if (!defined('GLPI_CACHE_CONFIG')) {
+   define('GLPI_CACHE_CONFIG', ['adapter' => 'apcu']);
```
remicollet: IIUC, the value can be defined in "local_define.php" ?

I haven't even notice this new file, but this seems messy.
Configuration is usually stored in the DB.
The only files in CONFIG_DIR are config_db (and slave) by design, as not storable in the DB

If sometime we add a UI to select the cache adapter (with connexion param for such case as redis, memcached...) how do the UI will save it ?

A GLPI_CONFIG_DIR/glpi_config_cache.php could work... but DB seems better (excepted if you can imagine some case where various PHP backends use the same DB but with different env, e.g. one server with APCu, one without)

inc/toolbox.class.php
```diff
@@ -2828,7 +2828,8 @@ public static function formatOutputWebLink($link) {
     */
    public static function useCache() {
       global $GLPI_CACHE;
-      return $GLPI_CACHE !== false && (!defined('TU_USER') || defined('CACHED_TESTS'));
+      return !($GLPI_CACHE instanceof Zend\Cache\Storage\Adapter\AbastractAdapter)
```
remicollet: bad negation ?

inc/define.php
trasher: This is a very advanced use case, and aslo more a system configuration than a user one.

Also, writing a configuration form for that would be a too huge job for the goal (I had to work on this, using some conf was just an extra that may helps).

If you think we should really provide UI to configure that (I do not tink), I'll open another ticket; and this will not be done for 9.2.

inc/toolbox.class.php
trasher: I don't think so (removing negation make tests to fail), but I'll check if the second part is OK :)

inc/config.php
```diff
@@ -45,6 +45,24 @@
 Session::setPath();
 Session::start();
 
+//init cache
+$GLPI_CACHE = false;
+try {
+   $cache_conf = GLPI_CACHE_CONFIG;
+   if (!isset($cache_conf['options'])
+       || !isset($cache_conf['options']['namespace'])) {
+      $cache_conf['options']['namespace'] = 'glpicache';
```
remicollet: Should probably use GLPI_VERSION to avoid bad answer in multi-glpi versions case or something related to the running instance.

inc/define.php
remicollet: All system configurations are stored in the DB and managed from the UI (DB master during install, DB replica, ldap, mail server, mail receiver, ...)

Of course the UI can be a feature of next major version

inc/config.php
trasher: Good idea, yes; I'll add it

inc/toolbox.class.php
trasher: I had a typo in the class name... ("Abastract")

tests/DbTestCase.php
```diff
@@ -51,10 +53,29 @@ public function setUp() {
...
+      }
+   }
 
    public function afterTestMethod($method) {
       global $DB;
 
+      if (in_array($method, $this->cached_methods)) {
+         global $GLPI_CACHE;
+         $GLPI_CACHE = false;
+      }
```
remicollet: Shouldn't you also call apcu_clear_cache, to ensure store values from one cache will not polute another TU ?

remicollet: rather $GLPI_CACHE::flush();

remicollet:
LGTM, btw I cannot test it

trasher:
Thanks for your help :)

Issue #2477, opened 2017-07-18
[9.2.1] Assign action on business rule for update


nqb:
Hello,

In GLPI 9.1.5, I notice that when you **assign** a technician group through business rules for tickets (update), current group on ticket is not replaced. Assign action seems to be identical to Add action.

Steps to reproduce : 
1. Create a business rule for ticket : 
  - Use rule for : Update
  - Criteria :  Category is "Bar" 
  - Actions : Technician group - Assign - "Bar" group
2. Create a ticket assign to "Foo" group
3. Select "Bar" category on ticket 
=> Groups "Foo" and "Bar" assign on ticket

Expected behavior : only "Foo" group assign on ticket. "Bar" group replace.

nqb:
No logs in `glpi/files/_log`, no logs in debug mode.

lcpj:
Same as #2294 and #2452

nqb:
Issue still presents in 9.2.1.

PR #2750, opened 2017-09-14
predefined fields could be not unique

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2477


install/update_91_92.php
```diff
@@ -1911,6 +1910,12 @@ function update91to92() {
       }
    }
 
+   // unique index added in a previous 9.2-dev
+   $migration->dropKey('glpi_tickettemplatepredefinedfields', 'unicity');
+   $migration->addKey('glpi_tickettemplatepredefinedfields',
```
trasher: Not sure it is really required

PR #2757, opened 2017-09-15
Check ticket status to permit followups and tasks; fixes #2656

### LINKED ISSUE #2656: [9.2] adding a followup on solved ticket is not harmonized
Hello,

Imagine a solved ticket (not yet closed) , and that you've kept the Followup tab along with the 'Processing Ticket' tab.
Then it's not possible to add a followup via the followup tab:
image

But via the 'Processing ticket' tab, it's possible:
image

Must be harmonized!

regards,
Tomolimo

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2656

### DIFF
--- a/inc/ticket.class.php
+++ b/inc/ticket.class.php
@@ -6944,10 +6944,11 @@ class Ticket extends CommonITILObject {
       $tmp = array('tickets_id' => $this->getID());
       $fup             = new TicketFollowup;
       $ttask           = new TicketTask;
-      $doc             = new Document;
 
-      $canadd_fup      = $fup->can(-1, CREATE, $tmp);
-      $canadd_task     = $ttask->can(-1, CREATE, $tmp);
+      $canadd_fup      = $fup->can(-1, CREATE, $tmp) && !in_array($this->fields["status"],
+                         array_merge($this->getSolvedStatusArray(), $this->getClosedStatusArray()));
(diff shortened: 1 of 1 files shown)


inc/ticket.class.php
```diff
@@ -6944,10 +6944,11 @@ function viewEditSubitem" . $this->fields['id'] . "$rand(e, itemtype, items_id,
       $tmp = array('tickets_id' => $this->getID());
       $fup             = new TicketFollowup;
       $ttask           = new TicketTask;
-      $doc             = new Document;
 
-      $canadd_fup      = $fup->can(-1, CREATE, $tmp);
-      $canadd_task     = $ttask->can(-1, CREATE, $tmp);
+      $canadd_fup      = $fup->can(-1, CREATE, $tmp) && !in_array($this->fields["status"],
```
orthagh: IMO, theses tests should be in canCreateItem methods of TicketFollowup & TicketTask

trasher: Seems this is already tested in both `canCreateItem` methods; but with other checks as well

trasher: Hum... I've read too quickly: only `getClosedStatusArray` is tested in those methods

PR #2759, opened 2017-09-15
Do not send notif if there were no changes; fixes #2378

### LINKED ISSUE #2378: Empty solution description
Dear All,
I recently upgraded to glpi 9.14. I have a question. If I click on save without any input in solution description field. It creates a new update instead of solving the ticket and sends notification regarding update of ticket. Has anyone come across such issue or am I the only one

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2378

### DIFF
--- a/inc/ticket.class.php
+++ b/inc/ticket.class.php
@@ -1349,8 +1349,6 @@ class Ticket extends CommonITILObject {
          }
       }
 
-      $this->updates[] = "actiontime";
-
       if (count($this->updates)) {
          // Update Ticket Tco
          if (in_array("actiontime", $this->updates)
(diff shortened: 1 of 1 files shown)


inc/ticket.class.php
```diff
@@ -1349,9 +1349,8 @@ function post_updateItem($history=1) {
          }
       }
 
-      $this->updates[] = "actiontime";
-
       if (count($this->updates)) {
+         $this->updates[] = "actiontime";
```
orthagh: the next part seems to be also strange as it test the presence actiontime in updates

PR #2762, opened 2017-09-15
Display ID in changes and problems like in tickets; fixes #2338

### LINKED ISSUE #2338: Difference in display of IDs between ticket and change
When editing a ticket, in the title we have the information of the id of the ticket.

ticket

However, for changes / problems, ..., the information is not present. 
change

We must activate the "Show GLPI ID" option.

Would it be possible to always have the id in the title for changes and problems even if the "Show GLPI ID" option is not enabled?

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2338

### DIFF
--- a/inc/change.class.php
+++ b/inc/change.class.php
@@ -636,8 +636,16 @@ class Change extends CommonITILObject {
          $showuserlink = 1;
       }
 
+      if (!$this->isNewItem()) {
+         $options['formtitle'] = sprintf(
+            __('%1$s - ID %2$d'),
+            $this->getTypeName(1),
+            $ID
+         );
+         //set ID as already defined
+         $options['noid'] = true;
+      }
       $this->showFormHeader($options);
-
 
       echo "<tr class='tab_bg_1'>";
       echo "<th class='left' width='$colsize1%'>".__('Opening date')."</th>";
--- a/inc/problem.class.php
+++ b/inc/problem.class.php
@@ -1055,6 +1055,15 @@ class Problem extends CommonITILObject {
          $showuserlink = 1;
       }
 
+      if (!$this->isNewItem()) {
+         $options['formtitle'] = sprintf(
+            __('%1$s - ID %2$d'),
+            $this->getTypeName(1),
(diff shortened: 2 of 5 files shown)


inc/change.class.php
```diff
@@ -636,9 +636,17 @@ function showForm($ID, $options=array()) {
          $showuserlink = 1;
       }
 
+      if (!$this->isNewItem()) {
+         $options['formtitle'] = sprintf(
+            __('%1$s - ID %2$d'),
+            $this->getTypeName(1),
```
yllen: It's not a correct gettext sentence.
You must use a sentence for ID %2$d for language readied from right to left

inc/problem.class.php
```diff
@@ -1055,6 +1055,15 @@ function showForm($ID, $options=array()) {
          $showuserlink = 1;
       }
 
+      if (!$this->isNewItem()) {
+         $options['formtitle'] = sprintf(
+            __('%1$s - ID %2$d'),
+            $this->getTypeName(1),
```
yllen: It's not a correct gettext sentence.

inc/ticket.class.php
```diff
@@ -3958,8 +3958,7 @@ function showForm($ID, $options=array()) {
       echo "<th colspan='4'>";
 
       if ($ID) {
-         $text = sprintf(__('%1$s - %2$s'), $this->getTypeName(1),
-                         sprintf(__('%1$s: %2$s'), __('ID'), $ID));
+         $text = sprintf(__('%1$s - ID %2$d'), $this->getTypeName(1), $ID);
          if ($ismultientities) {
```
yllen: It's not a correct gettext sentence.

inc/change.class.php
trasher: So this is alredy warong in some other places (commondbtm, commonglpi, ...)

PR #2776, opened 2017-09-18
[9.2] Error display planning when plugin not installed Fix #2775

### LINKED ISSUE #2775: [9.2] Error display planning when plugin not installed
image

amma35:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2775

### DIFF
--- a/inc/planning.class.php
+++ b/inc/planning.class.php
@@ -1247,8 +1247,9 @@ class Planning extends CommonGLPI {
          $group->getFromDB($actor[1]);
          $title = $group->getName();
       } else if ($filter_data['type'] == 'event_filter') {
-         if (!$filter_key::canView()
-               || !($item = getItemForItemtype($filter_key))) {
+         if (!($item = getItemForItemtype($filter_key))) {
+            return false;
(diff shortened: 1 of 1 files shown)


inc/planning.class.php
```diff
@@ -1247,8 +1247,9 @@ static function showSingleLinePlanningFilter($filter_key, $filter_data, $options
          $group->getFromDB($actor[1]);
          $title = $group->getName();
       } else if ($filter_data['type'] == 'event_filter') {
-         if (!$filter_key::canView()
-               || !($item = getItemForItemtype($filter_key))) {
+         if (!($item = getItemForItemtype($filter_key))) {
+            return false;
```
orthagh: if the ``!canView`` need to break first, just invert the two statements, no need for additional verbosity

tsmr: I have tested.No, same problem

orthagh: ok fine

PR #2795, opened 2017-09-20
Escape in titles

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

### DIFF
--- a/inc/levelagreement.class.php
+++ b/inc/levelagreement.class.php
@@ -298,7 +298,7 @@ abstract class LevelAgreement extends CommonDBChild {
                function delete_date$type$rand(e) {
                   e.preventDefault();
 
-                  if (nativeConfirm('".__('Also delete date ?')."')) {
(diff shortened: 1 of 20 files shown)


inc/levelagreement.class.php
```diff
@@ -298,7 +298,7 @@ function showForTicket(Ticket $ticket, $type, $tt, $canupdate) {
                function delete_date$type$rand(e) {
                   e.preventDefault();
 
-                  if (nativeConfirm('".__('Also delete date ?')."')) {
+                  if (window.confirm('".__s('Also delete date ?')."')) {
```
orthagh: please keep nativeConfirm

trasher: OK... But why?

PR #2804, opened 2017-09-21
Instanciate cache at the end, to prevent errors on some script calls

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

### DIFF
--- a/inc/config.php
+++ b/inc/config.php
@@ -47,26 +47,8 @@ Session::start();
 
 //init cache
 $GLPI_CACHE = false;
-try {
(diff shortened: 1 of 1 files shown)


inc/config.php
```diff
@@ -45,26 +45,6 @@
 Session::setPath();
 Session::start();
 
-//init cache
-$GLPI_CACHE = false;
```
remicollet: I think it is better to keep default value here

trasher: Changed ;)

PR #2815, opened 2017-09-22
remove slashes before single quote

orthagh:
@trasher you'll like this one :D

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes


orthagh:
maybe not, we'll see

PR #2862, opened 2017-09-28
Downstream

remicollet:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no (and yes in master)
| Tests pass?   | yes

With 9.2, config dir is empty, so make sense to also move this file.

1st commit keep BC and can be cherrypick in 9.2
2nd is for master only

### DIFF
--- a/inc/based_config.php
+++ b/inc/based_config.php
@@ -47,8 +47,10 @@ if (!empty($tz)) {
 }
 
 // If this file exists, it is load, allow to set configdir/dumpdir elsewhere
-if (file_exists(GLPI_ROOT ."/config/config_path.php")) {
-   include_once(GLPI_ROOT ."/config/config_path.php");
+if (file_exists(GLPI_ROOT . '/inc/downstream.php')) {
+   include_once (GLPI_ROOT . '/inc/downstream.php');
+} else if (file_exists(GLPI_ROOT . '/config/config_path.php')) { // For compatibility, deprecated
+   include_once (GLPI_ROOT . '/config/config_path.php');
 }
 
 // Default location for database configuration : config_db.php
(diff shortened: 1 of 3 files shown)


trasher:
While I agree; I use this to adapt paths locally sometimes. Will have to chek first if I can use `config/local_define.php` to declare my stuff.

remicollet:
Notice: current load order is:
* inc/config.php
* inc/based_config.php
* config/config_path.php
* inc/define.php
* config/user_local.php

So for now, user_local CANNOT override constant from config_path.

I think user_local could be moved to based_config, before config_path.php, but this means downstream will have to add "`if !(defined...)`" condition around some default values.

But, does it really make sense to override package layout ?

On IRC, you talk about XHPROF configuration, which is not defined anywhere, so which should work in user_local.

remicollet:
3rd commit (also need in 9.2/bf) ensure local_define.php is loaded before downstream.php

inc/based_config.php
```diff
@@ -46,9 +46,14 @@
    date_default_timezone_set(@date_default_timezone_get());
 }
 
+// If this file exists, it is load
+if (file_exists(GLPI_ROOT. '/config/local_define.php')) {
+   require_once GLPI_ROOT. '/config/local_define.php';
```
remicollet: cannot use GLPI_CONFIG_DIR which is not yet defined.
so probably need to be patched for downstream redirection

trasher:
Locally rebased, squashed and merged

trasher:
Also merged in 9.2/bf branch (compatibility layer kept).

PR #2868, opened 2017-09-29
fix #2867 Problem with id for load existing actors before playing rules

### LINKED ISSUE #2867: Problem with id for load existing actors before playing rules
GLPI version : 9.2

see #2047 Problem when executing the rules with ticket actors

With the commit : https://github.com/glpi-project/glpi/commit/39f2b51a3970ab951d216921e3dae310f64a4263

User IDs, groups and suppliers are not good.

I can give an example if I have to.

amma35:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2867

### DIFF
--- a/inc/ticket.class.php
+++ b/inc/ticket.class.php
@@ -1187,12 +1187,12 @@ class Ticket extends CommonITILObject {
          if (count($users)) {
             $field = 'users_id';
             foreach ($users as $user) {
-               if (!isset($input['_'.$field.'_'.$t]) || !in_array($user['id'], $input['_'.$field.'_'.$t])) {
+               if (!isset($input['_'.$field.'_'.$t]) || !in_array($user[$field], $input['_'.$field.'_'.$t])) {
                   if (!isset($input['_'.$field.'_'.$t])) {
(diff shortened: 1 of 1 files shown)


inc/ticket.class.php
```diff
@@ -1187,12 +1187,12 @@ function prepareInputForUpdate($input) {
          if (count($users)) {
             $field = 'users_id';
             foreach ($users as $user) {
-               if (!isset($input['_'.$field.'_'.$t]) || !in_array($user['id'], $input['_'.$field.'_'.$t])) {
+               if (!isset($input['_'.$field.'_'.$t]) || !in_array($user['users_id'], $input['_'.$field.'_'.$t])) {
```
trasher: Maybe use `$field` instead of raw value?

amma35: Yes of course I had not seen $field

trasher: It's easier to see it on the review ;)

Issue #2928, opened 2017-10-06
GLPI 9.2 - Multiple plugin requirements are displayed without newline


btry:
Hi

See this screenshot

![image](https://user-images.githubusercontent.com/14139801/31293560-d3fad8f6-aad7-11e7-805d-b8910ebbe3be.png)

PR #2939, opened 2017-10-10
Fix 2928

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2928

### DIFF
--- a/inc/plugin.class.php
+++ b/inc/plugin.class.php
@@ -541,7 +541,7 @@ class Plugin extends CommonDBTM {
                      $function   = 'plugin_' . $plug['directory'] . '_check_prerequisites';
                      $do_install = $this->checkVersions($plug['directory']);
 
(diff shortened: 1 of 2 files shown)


inc/plugin.class.php
```diff
@@ -541,7 +541,7 @@ function listPlugins() {
                      $function   = 'plugin_' . $plug['directory'] . '_check_prerequisites';
                      $do_install = $this->checkVersions($plug['directory']);
 
-                     if (function_exists($function)) {
+                     if ($do_install && function_exists($function)) {
```
btry: I guess this change is to fix something else ?

trasher: Yes, this is not directly related to the issue :)

PR #2944, opened 2017-10-11
Prevent kb items duplicates in list; fixes #2899

### LINKED ISSUE #2899: [9.2] Link knowbase Item to ticket
From /glpi/front/knowbaseitem.form.php?id=X

When you link a ticket to the knowbase article, the ticket appears four times like this : 

image

AND

The dates fields do not use Html::convDateTime

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2899

### DIFF
--- a/inc/dbmysqliterator.class.php
+++ b/inc/dbmysqliterator.class.php
@@ -94,13 +94,15 @@ class DBmysqlIterator implements Iterator, Countable {
                   case 'SELECT' :
                   case 'FIELDS' :
                      $field = $val;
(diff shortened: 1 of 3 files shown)


trasher:
You are right; but rather than play with GROUPBY issues; I've prefered to use DISTINCT.

DBmysqlIterator was not able to do that; so I've added the feature (+ tests)

Issue #2958, opened 2017-10-11
File missing or incorrect


yllen:
From an alert of search saved, if i click on List
_The requested URL /git92/front/savedsearch_alert.php was not found on this server._

yllen:
And when i change value of the alert, historical tab stay empty

Issue #2882, opened 2017-10-02
Entites informations not saved on saved searches


trasher:
When storing a saved search; entities informations (id, recursivity) are not saved in the database.

trasher:
Well... This is OK for public searches; but not on private ones (hardcoded values in the form)

trasher:
To precise: private bookmarks always have been attached to the root entity and set to not_recursive.

The solution would be to choose entitiy and recursivity adding private searches just as for the public ones.

trasher:
Closing, see #2978

PR #2961, opened 2017-10-12
Fix 2958

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2958

### DIFF
--- a/inc/savedsearch_alert.class.php
+++ b/inc/savedsearch_alert.class.php
@@ -42,7 +42,7 @@ class SavedSearch_Alert extends CommonDBChild {
    // From CommonDBChild
    static public $itemtype = 'SavedSearch';
    static public $items_id = 'savedsearches_id';
-   public $dohistory       = false;
+   public $dohistory       = true;
    protected $displaylist  = false;
 
    const OP_LESS     = 0;
@@ -382,5 +382,9 @@ class SavedSearch_Alert extends CommonDBChild {
             }
          }
       }
+   }
+
+   function getItemsForLog($itemtype, $items_id) {
+      return ['new' => $this];
    }
 }
--- a/inc/savedsearch_alert.class.php
+++ b/inc/savedsearch_alert.class.php
@@ -43,6 +43,7 @@ class SavedSearch_Alert extends CommonDBChild {
    static public $itemtype = 'SavedSearch';
    static public $items_id = 'savedsearches_id';
    public $dohistory       = false;
+   protected $displaylist  = false;
 
    const OP_LESS     = 0;
    const OP_LESSEQ   = 1;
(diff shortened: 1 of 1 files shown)


yllen:
OK for Historical tab.
But you delete link to go to saved search, you must stay it.
My problem was for List link not saved search name

trasher:
There is no alerts lists; this link was provided as a link to te saved search when you came from the saved search list. The "List..." link was a default I did not know we can remove initially; and the link to the saved search is already present in the form.

yllen:
I have the problem with this link List only after create a new alert.
Question: why in saved search i only have public search? The notification alert can be for private one.
And how can i modify an existing one? when i click on the name the bookmark is loaded

yllen:
Another point to alert.
If i have only 1 item i can't create 2 alerts (i have button but nothing done when i add, even message).
I you can create disable alert you can create several.

trasher:
> why in saved search i only have public search? 

In saved search management interface you mean? I've just tested with two saved searches created from computers, on private and one public;  can see both (with default glpi account).
Modification of private search is the same than public ones; from the management interface. The slide panel does not provide that.

> i have button but nothing done

You mean you even do not have the alert form? I've tested on previously created searches, I was able to create several alerts on each one...

Don't know what the problem is exactly right now; but I wonder if both problems does not have only one cause. Did you try relogin? Which profiles did you use?

yllen:
In Tools > saved search i only see public search
In star to load bookmarks i see public and private searches.
So i can't modify private because there are not in list

yllen:
>You mean you even do not have the alert form? I've tested on previously created searches, I was able to create several alerts on each one...

Example: i have only 1 phone in item Phone
I create a bookmark for this phone.
I create an alert for this bookmark.
I want to create a second one. I have the button Add a new alert. I click on it, have the form and complete it. On click on Save button, form is reload empty

yllen:
Test done this day and issue is the same. I used Super-Admin profil

trasher:
I still cannot reproduce :/ Tested on a phone asset as well; and I can see private searches in the list; and also create as many alerts as I want.

For the list problem, could you please send the query that is played in the management interface?

trasher:
OK, got it for private searches...
All private searches are in the root entity; and not recursive. In the panel, we take care of that, looking for entities restrictions only for public ones.

So the problem only appears when you use any entity but the root.

yllen:
I think i have understand the issue.
In search engine you only have searches for this entity or recursive one.
With bookmarks of my settings you have public bookmarks for this entity and private bookmark for all entities.
So if you have define a bookmark from root entity no recursive and private, you can load it from a sub-entity.

trasher:
I do not know why is_recursive is set to 0; it always been. On private searches; this does not really makes sense. I propose to change is_recusrive to 1 for all private searches.
Could you try if that fixes your issues?

yllen:
I can need to do a specific search for a sub-entity.
For me in bookmarks list of my settings i must have only private search for this entity

trasher:
So, this is related to https://github.com/glpi-project/glpi/issues/2882

Issue #2964, opened 2017-10-13
Added  label with ticket status in navigation header.


Drimmwald:
Hello!
I made some changes. Maybe it will be interesting for you.
Added  label with ticket status in navigation header.
So at the subconscious level it will be easy for users to perceive statuses by colors.

![status](https://user-images.githubusercontent.com/32536773/31536581-e5c7a386-b007-11e7-97bd-beacfb3d10e3.png)
commit d1a292415a61c64692580aa56b00bc3b011376d7

PR #2965, opened 2017-10-13
Add ticket status in navigation header

Drimmwald:
| Q             | A
| ------------- | ---
| Bug fix?    |  no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

Issues #2964


trasher:
Hello,

Thanks for the PR, I got some remarks. First, CS checks are failing (not very important, but this needs to be fixed).

Second, copying the entire `showNavigationHeader` is probably not the way to go; we need to find something more generic I guess.

trasher:
Maybe adding an option to existing method should does the trick (I've just took a very quick look)

trasher:
Great; that sounds better :)

Just one more remark: maybe this new code should be run only  if we're from a Ticket (or maybe from a CommonITILObject; I guess this should appply as well on Changes and Problems).

Drimmwald:
I think you are right. Now code working with Changes and Problems, but some styles of these statuses are absent. I will try fix in the near time.

orthagh:
Sorry to answer late on this submission, but this is not ok imo.
We have an existing iconography for ITIL status representation and the submitted colors didn't match at all.
image

This set may be outdated and subject to review but we need to do a spec on this (on spec wiki: https://github.com/glpi-project/spec/wiki).
And this change should aim next major version.

In parallel, we can try to transform theses images in full css.

trasher:
Reworked; see https://github.com/glpi-project/glpi/pull/3820

PR #2969, opened 2017-10-13
fix #2968 PHP error : on create ticket without groups_id

### LINKED ISSUE #2968: 9.2 PHP error : on create ticket without groups_id
GLPI 9.2

same error : see #1464

When I try to create a ticket from a plugin without groups_id.

2016-12-23 15:34:02
*** PHP Notice(8): Undefined index: groups_id
Backtrace :
inc\ticket.class.php:1486
inc\commondbtm.class.php:815 Ticket->prepareInputForAdd()

amma35:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2968

### DIFF
--- a/inc/ticket.class.php
+++ b/inc/ticket.class.php
@@ -1838,7 +1838,6 @@ class Ticket extends CommonITILObject {
                if ($item = getItemForItemtype($itemtype)) {
                   $item->getFromDB($items_id);
                   $input['items_locations'] = $item->fields['locations_id'];
-                  $input['items_groups'] = $item->fields['groups_id'];
                   if ($infocom->getFromDBforDevice($itemtype, $items_id)) {
                      $input['items_businesscriticities']
                         = Dropdown::getDropdownName('glpi_businesscriticities',


inc/ticket.class.php
```diff
@@ -1838,7 +1838,6 @@ function prepareInputForAdd($input) {
                if ($item = getItemForItemtype($itemtype)) {
                   $item->getFromDB($items_id);
                   $input['items_locations'] = $item->fields['locations_id'];
-                  $input['items_groups'] = $item->fields['groups_id'];
```
trasher: I do not understand why you remove it completely. Why not to solve as it has been in https://github.com/glpi-project/glpi/pull/1471/files?

amma35: The same is done after line 1847:
```
if (isset($item->fields['groups_id'])) {
    $input['items_groups'] = $item->fields['groups_id'];
 }
```

trasher: Indeed, you're right

PR #3001, opened 2017-10-19
Permit single quotes in user email; fixes #2993

### LINKED ISSUE #2993: Error while creating a new ticket and the requester has an e-mail address with a single quote in it
Hello,

I tried to create a ticket through the GLPI web interface for an user with single quote in his last name. The system rejected the request with the message that his e-mail address was not valid.

Name (like) : Jacques M'Boon
Address: Jacques.M'Boon@domain.com

In the e-mail field the address was cut like this: Jacques.M

If I put the e-mail address instead of the user name during ticket creation, everything is OK. So I think that the issue appears when the system retrieves the user e-mail address from the glpi user entry.

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2993

### DIFF
--- a/ajax/uemailUpdate.php
+++ b/ajax/uemailUpdate.php
@@ -115,7 +115,7 @@ if ((isset($_POST['field']) && ($_POST["value"] > 0))
                                                     'display' => false]);
    } else {
       $email_string = "<input type='text' size='25' name='".$_POST['field']."[alternative_email][]'
-                        value='".$default_email."'>";
+                        value='".htmlentities($default_email, ENT_QUOTES, 'utf-8')."'>";
    }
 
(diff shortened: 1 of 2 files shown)


trasher:
@kabassanov please test this fix proposal (including mail sending). Thank you.

kabassanov:
Yes, I confirm it works.
Just a small detail: in 9.1.6 the second file is (still) called inc/notificationmail.class.php

trasher:
This will be fixed for 9.2 only. Thanks for the feedback :)

PR #3005, opened 2017-10-19
enhance message to run composer

btry:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #none

Renders as is 

image

### DIFF
--- a/inc/autoload.function.php
+++ b/inc/autoload.function.php
@@ -380,7 +380,8 @@ if (!file_exists($autoload)) {
    }
 }
 if ($needrun) {
-   die('Run "composer install --no-dev" in the glpi tree');
+   die('Run "composer install --no-dev" in the glpi tree.<br>'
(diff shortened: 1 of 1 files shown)


inc/autoload.function.php
```diff
@@ -380,7 +380,8 @@ function glpi_autoload($classname) {
    }
 }
 if ($needrun) {
-   die('Run "composer install --no-dev" in the glpi tree');
+   die('Run "composer install --no-dev" in the glpi tree.<br>'
+       . 'To install composer please refer to <a href="https://getcomposer.org/">https://getcomposer.org/</a>');
 }
```
remicollet: HTML is not acceptable here, as will look badly for CLI commands

btry: Ah yes, you're right, so I'll add a check with isCommandLine()

PR #3013, opened 2017-10-20
check php version early

remicollet:
See http://forum.glpi-project.org/viewtopic.php?pid=339918#p339918

### DIFF
--- a/inc/define.php
+++ b/inc/define.php
@@ -47,7 +47,7 @@ if (substr(GLPI_VERSION, -4) === '-dev') {
    //for stable version
    define("GLPI_SCHEMA_VERSION", '9.2');
 }
-define('GLPI_MIN_PHP', '5.6.0');
+define('GLPI_MIN_PHP', '5.6.0'); // Must also be changed in top of index.php
 define('GLPI_YEAR', '2017');
 if (!defined('GLPI_DEMO_MODE')) {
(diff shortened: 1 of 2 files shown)


remicollet:
Of course, GLPI_MIN_PHP cannot be used as not yet defined.

trasher:
What about creating a new file with only the `defines` ad load it at the very begining?
Having to change the minimal PHP version at two different place may land to mistakes in the future :/

remicollet:
From phpcompatinfo static analyser report on index.php

`Requires PHP 5.3.0 (min), PHP 5.3.0 (all)`

remicollet:
> What about creating a new file with only the defines ad load it at the very begining?

To be honest, having 1 more file only for this seems messy
More, all the other checks have no value (never reached code), only the one here is useful.

trasher:
OK then. But there are chances this one will not be changed next time PHP version will be bumped :D

inc/define.php
```diff
@@ -47,7 +47,7 @@
    //for stable version
    define("GLPI_SCHEMA_VERSION", '9.2');
 }
-define('GLPI_MIN_PHP', '5.6.0');
+define('GLPI_MIN_PHP', '5.6.0'); // Must also be changed in top of index.php
```
trasher: 👍

Issue #3037, opened 2017-10-24
Possible migration issue with notifications


trasher:
Some users try to update their database for tests; and then restore a backup on this database without cleaning anything. 
New created tables stay in place; and parts of migration relying on table presence will not be executed. This has been spotted in #2871 and on the forums/mailing list for the notification issue.

We can change the way we some conditions are done to get that type of migration working ; but since user has made previous tests; his database will remain kind of crap; there is no solution.

Previously, newly created tables was removed (so current problem was solved); but this cause data loss when update is played several times.

tomolimo:
Hello,
Should we manage a status of migration, like: "migration started", "migration done".
Testing this status will inform migration procedure if it must alert for possible loss of data.
regards,
Tomolimo

trasher:
In that case; that would not work. Both migration can be successfull; if you restore a previous backup, GLPI cannot define what has been done.

We've been talking about that already; and I've not been able to find a real solution for all that :(

tomolimo:
And what about a glpi_configs_backup which will store the version of the backup table set?

trasher:
When user will restore his backup; this table will be overidded too... This would work only the first time

tomolimo:
If we have a way to store somewhere a kind of version for each table, will it be a solution?

trasher:
No, this would not be enough since data may be impacted as well... tickets migration issue affects table structure; but notification is only on data in the  table.

tomolimo:
The version can be either a schema evolution and/or a data evolution.
Or we may store not only the version of the table, but also a version of operations on the data.
using for example the table comment field like following:
##Schema: 1.0, Data: 1.0##
##Schema: 1.1, Data: 1.0##
##Schema: 1.2, Data: 1.0##
##Schema: 1.2, Data: 1.2##
or something like this, rules for version evolutions have to be defined, this is an idea only.

edgardmessias:
See this book: https://developers.redhat.com/promotions/migrating-to-microservice-databases/

trasher:
Notifications migratin sissues has been fixed as of #3041.

Issue #2871, opened 2017-09-29
Couldn't find created ticket after upgrading to GLPI 9.2


zawaye:
Dear GLPI Team,

I am using GLPI since 2 years ago. But earlier version of GLPI didn't found any error for upgrading from one version to another. Although I can upgrade from GLPI 9.1.6 to GLPI 9.2, I couldn't found any created ticket in my previous version even I search with ticket status = All. Therefore, I try to test with new ticket. But, the created new ticket also couldn't be found again. 
What happen in this complete version? 
Thanks,

trasher:
Check your logs, I guess something went wrong while upgrading your database.

AircraftRu:
I confirm it. I couldn't found all my tickets.

GLPI  9.1.4 -> GLPI 9.2

zawaye:
I haven't found any error message during my upgrade process. I just found just five warning as the following list:

1. ldap extension is not present
2. imap extension is not present
3. Zend OPcache extension is not present
4. APCu extension is not present
5. xmlrpc extension is not present

vinz38500:
Hey, I had exactly the same issue.. I can access a ticket with a direct link, but no chance to display the list of tickets.
I will redo the upgrade of the db as I may have forgotten an error message ..
Let's see

vinz38500:
Hello again
I did the upgrade from 9.1.4 tp 9.2 using cliupdate.php (to avoid timeout in the user interface)

I have the same behaviour, without any error in my log files

The only one I get is when I try to display tickets : 
`  Error: Unknown column 'glpi_tickets.time_to_resolve' in 'field list'
`
Result of my cliupdate 
`C:\LOCALDATA\website\glpi-9.2\tools>php cliupdate.php
Current GLPI version         : 9.1.4
New GLPI version             : 9.2
Current GLPI database version: 9.1.4
New GLPI database version    : 9.2
Default GLPI Language        : en_GB

========================================== Update to 9.2 ===========================================
** Api users tokens has been reset, if you use REST/XMLRPC api with personal token for authentication, please reset your user's token.
Configuration values added for login_remember_time, login_remember_default, use_notifications, notifications_mailing, notifications_ajax, notifications_ajax_check_interval, notifications_ajax_sound, notifications_ajax_icon_url, dbversion, sTask completed. (2 minutes 51 seconds)
Migration Done.`

trasher:
This is strange you get no errors during the upgrade :/
[...]

PR #3041, opened 2017-10-24
Migration issues

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #3037 #2871

### DIFF
--- a/install/update_91_92.php
+++ b/install/update_91_92.php
@@ -890,19 +890,19 @@ function update91to92() {
 
    if ($DB->tableExists('glpi_queuedmails')) {
       $migration->renameTable("glpi_queuedmails", "glpi_queuednotifications");
-      $migration->addPostQuery("UPDATE `glpi_crontasks`
-                                SET `itemtype` = 'QueuedNotification'
-                                WHERE `itemtype` = 'QueuedMail'");
-      $migration->addPostQuery("UPDATE `glpi_crontasks`
-                                SET `name` = 'queuednotification'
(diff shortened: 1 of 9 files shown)


install/update_91_92.php
```diff
@@ -678,160 +678,199 @@ function update91to92() {
...
+                  `olalevels_id` int(11) NOT NULL DEFAULT '0',
+                  `date` datetime DEFAULT NULL,
+                  PRIMARY KEY (`id`),
+                  KEY `tickets_id` (`tickets_id`),
+                  KEY `olalevels_id` (`olalevels_id`),
+                  KEY `unicity` (`tickets_id`,`olalevels_id`)
+               ) ENGINE=MyISAM DEFAULT CHARSET=utf8 COLLATE=utf8_unicode_ci;";
+      $DB->queryOrDie($query, "9.2 add table glpi_olalevels_tickets");
 
       $query = "INSERT INTO `glpi_crontasks`
```
orthagh: here, maybe ``REPLACE``

trasher: I do not think this is required since the add would only be run when table was not present; this should never be a duplicate.
Anyways; I can change that, this would have no impact

MatheusPRibeiro:
Problem solved we increased the "**Frequency of execution**" time of the action to 5 min.
Apparently the action, executed every 1 min was **fought**.

PR #3077, opened 2017-10-28
Feature/fix simcard ui

btry:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | 
| Fixed tickets | none

2 fixes
- when the browser is not wide enough, the eye button and the copy button go out of their expected place.
- when one click on the eye and goes out of it while pressing the mouse button, the field still shows its content.

## before
image

image

## after

image

image

### DIFF
--- a/css/styles.css
+++ b/css/styles.css
@@ -650,23 +650,16 @@ input.submit[type=reset] {
    padding-right: 50px;
 }
 
-/* #page .fa.disclose {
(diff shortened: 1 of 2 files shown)


trasher:
You got some CS issues ;)

Issue #3093, opened 2017-10-31
Bug for dropdown search with itemtype Software


yllen:
http://forum.glpi-project.org/viewtopic.php?id=159536

SELECT `glpi_softwares`.* 
                   FROM `glpi_softwares`
                   
                   WHERE 1  AND `glpi_softwares`.`is_deleted` = '0'  AND `glpi_softwares`.`is_template` = '0' AND ( `glpi_softwares`.`entities_id` = '1'  OR (`glpi_softwares`.`is_recursive`='1' AND `glpi_softwares`.`entities_id` IN ('0','303','68')) )  AND ( (`glpi_softwares`.`name`   LIKE '%7%'   OR `glpi_softwares`.`otherserial`   LIKE '%7%'  OR `glpi_softwares`.`serial`   LIKE '%7%' )) ORDER BY `glpi_softwares`.`entities_id`, `glpi_softwares`.`name`
                 LIMIT 0,5
  Error: Unknown column 'glpi_softwares.otherserial' in 'where clause'

Issue #2551, opened 2017-08-03
SQL error : Search in dropdown


amma35:
GLPI : 9.1.X

The search does not work for these types of objects (Cartridge model,  contract, phone)

![recherche](https://user-images.githubusercontent.com/15138371/28930209-888aa8ac-7872-11e7-985a-03a5cac37210.png)

```
  *** MySQL query error:
  SQL: SELECT `glpi_contracts`.* 
                   FROM `glpi_contracts`
                   
                   WHERE 1  AND `glpi_contracts`.`is_deleted` = \'0\'  
AND `glpi_contracts`.`is_template` = \'0\' 
AND ( `glpi_contracts`.`entities_id` IN (..)  )  
AND ( (`glpi_contracts`.`name`   LIKE \'%s%\'   OR `glpi_contracts`.`otherserial`   LIKE \'%s%\'  OR `glpi_contracts`.`serial`   LIKE \'%s%\' )) 
ORDER BY `glpi_contracts`.`entities_id`, `glpi_contracts`.`name`
                 LIMIT 0,100
  Error: Champ 'glpi_contracts.otherserial' inconnu dans where clause
  Backtrace :
  ajax\getDropdownValue.php:570    
```

PR #3094, opened 2017-10-31
fix dropdown search- see #3093

yllen:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #3093 #2551


trasher:
Finally, the fix is not correct. When linking a ticket, tickets ID used to be listed (and searchable). This is no longer the case with the fix.

PR #3104, opened 2017-11-03
Travis composer

remicollet:
Small test reorg to allow "composer test" to work in most cases:

- without LDAP
- without local server
- with plugins installed

BTW, the "api" directory could perhaps be renamed to "web"


trasher:
BTW; this one can target 9.2/bf branch; this will reduce differences between branches; and this does not impact code directly.

PR #3113, opened 2017-11-06
Tinymce/4.7.1

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

Master version of #3112 
Fix chrome issue (selecting image)

**Important**, with this upgrade, IE10 support on tinymce is dropped


tomolimo:
Here is a ticket in IE10:
image

:smile: 

regards,
Tomolimo

trasher:
Does not tinymce provide a way to know we're using a non supported browser?

I've used conditional comments for that in the past; but they've disappeared in ie10.

tomolimo:
> but they've disappeared in ie10

What do yo mean by 'disappeared'?

trasher:
There are no more conditional comments since ie10; need to use something else (js?)

tomolimo:
something like that:

```
if ($.browser.msie && $.browser.version === 10) {
  $("html").addClass("ie10");
}

```
?

trasher:
Sound correct; but i'd wrote `browser.version <= 10`. Could you confirm this works?

If so; we may add an advice on the page telling user he uses a not supported browser version.

trasher:
Finally, this solution in not relevant since jQuery current versions does not provide `$.browser`.

trasher:
I did not find any way to know if user has a supported browser or not from tinymce.

The only thing I've found is a bit of JS that *should* do the trick:
```
if (Function('/*@cc_on return document.documentMode===10@*/')()){
   //add a div at the top of the page for unsupported browser
}
```
Since I do not have IE 10, 11 nor edge; I cannot test on my side.

tomolimo:
Hello,
As a general comment on this evolution: should we migrate to this release?
Is it mandatory?
Does it solve something?
Thank you
Regards
Tomolimo

orthagh:
as mentioned on irc, we shouldn't keep old libraries and should stick to the latest version if possible.
General reasons:
 - security fixes
 - later migration could be painfull
 - And IE10 is not supported anymore by Microsoft itself so it's a good moment for migration i think.

trasher:
I totaly agree. My point was to inform the user his browser is outdated; to prevent "rich text is not working" tickets and to be fair. But this is not really mandatory.

Issue #2864, opened 2017-09-28
Can't update Operating system


yllen:
In operating system of a computer, if i change the field Name
"Cannot update item: not enough right's on the parent(s) item(s)"

In massive action, action indicated successfully but nothing done.
For change input "N/A (1516): Incompatible Items"

See also: http://forum.glpi-project.org/viewtopic.php?id=159124

yllen:
For me the correction is not correct. entities_id must stay for performance.
The correct correction is to add oeratingsytem in object managed with formard_entity_to of a computer

remicollet:
Seems OK forward_entity_to and optimization is for `CommonDBChild`, not needed for `CommonDBRelation`

My fault, sorry for the noise

Checker (may become a unit test)

```
<?php
require 'inc/includes.php';

echo "+ Login\n";
$auth = new Auth();
var_dump($auth->login('glpi', 'glpi', true));

var_dump(Session::changeActiveEntities(2, true));

echo "+ Disk\n";
$d = new ComputerDisk();
var_dump($d->can(1, READ));
var_dump($d->can(2, READ));

echo "+ OS\n";
$o = new Item_OperatingSystem();

var_dump($o->can(1, READ));
var_dump($o->can(2, READ));
```

All id=1 belongs to Entity 1, id=2 to Entity=2

Ouput
```
+ Login
  From Config::getCache() in /work/GLPI/92-bugfixes/inc/config.class.php line 2752
  Service with name "Zend\Cache\Storage\Adapter\Apcu" could not be created. Reason: ext/apcu is disabled - see 'apc.enabled' and 'apc.enable_cli' 
bool(true)
bool(true)
+ Disk
bool(false)
bool(true)
+ OS
bool(false)
bool(true)
```

So right check seems ok.

remicollet:
BTW, the `forward_entity_to` way seems a better fix

yllen:
$forward_entity_to is used in update of commonDBTM
and it also defined in calendar that iis a commonDropdown, not CommonDbChild or CommonDBRelation

trasher:
Should not have been closed... Was a revert :D

PR #3120, opened 2017-11-07
Fix 2864

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2864

### DIFF
--- a/inc/item_operatingsystem.class.php
+++ b/inc/item_operatingsystem.class.php
@@ -675,4 +675,12 @@ class Item_OperatingSystem extends CommonDBRelation {
       }
       parent::processMassiveActionsForOneItemtype($ma, $item, $ids);
    }
+
+   public function prepareInputForAdd($input) {
+      $item = getItemForItemtype($input['itemtype']);
+      $item->getFromDB($input['items_id']);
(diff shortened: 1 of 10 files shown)


inc/item_operatingsystem.class.php
```diff
@@ -675,4 +675,12 @@ static function processMassiveActionsForOneItemtype(MassiveAction $ma, CommonDBT
       }
       parent::processMassiveActionsForOneItemtype($ma, $item, $ids);
    }
+
+   public function prepareInputForAdd($input) {
+      $item = getItemForItemtype($input['itemtype']);
+      $item->getFromDB($input['items_id']);
+      $input['entities_id'] = $item->getField('entities_id');
```
orthagh: I don't like ``getField``for anything other than dislay, if it fails, you set NOT_AVAILABLE constant in the $input which contains "N/A" string.

trasher:
This PR is (one of?) the last blocker for 9.2.1 release. @yllen or @remicollet could you take a look and see if the fix is ok for you? Thanks!

PR #3122, opened 2017-11-07
Added unset for $_GET['id'] and $_POST['id']

### LINKED ISSUE #3119: [9.1 - 9.2] Problem when viewing existing ticket with malformed ticket id
Hello,

Try to view a ticket with a malformed ticket id like following:
http://glpi.acme.com/front/ticket.form.php?id=t251g952 
The ticket id must be a real one
then you get something strange, half 'New Ticket' and half ticket edition:
image
and as a side effect it will bypass any Lock mechanism. And it can also lead to screens without any save button (not reproduced, but viewed this morning by an end-user).

I believe the root cause is that in ticket.form.php line [189 for 9.1](https://github.com/glpi-project/glpi/blob/0254a61ec8b1cd35f435767b109c5373bdb3a052/front/ticket.form.php#L205), and [234 for 9.2](https://github.com/glpi-project/glpi/blob/c50b0e41fe06807fc2f2c39608bcb6121fe1044a/front/ticket.form.php#L234) we do an `unset($_REQUEST['id']);`, but in commonglpi.class.php line [1126 for 9.1](https://github.com/glpi-project/glpi/blob/0254a61ec8b1cd35f435767b109c5373bdb3a052/inc/commonglpi.class.php#L1126) and [970 for 9.2](https://github.com/glpi-project/glpi/blob/c50b0e41fe06807fc2f2c39608bcb6121fe1044a/inc/commonglpi.class.php#L970) we are testing `isset($_GET["id"])` and of course `id` it is still existing in `$_GET` as only `$_REQUEST['id']` has been unset.

thank you,
regards,
Tomolimo

tomolimo:
Fixes #3119

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #3119

### DIFF
--- a/front/ticket.form.php
+++ b/front/ticket.form.php
@@ -203,12 +203,9 @@ if (isset($_GET["id"]) && ($_GET["id"] > 0)) {
 } else {
    Html::header(__('New ticket'),'',"helpdesk","ticket");
    unset($_REQUEST['id']);
-   if (isset($_GET['id'])) {
-      unset($_GET['id']);
-   }
-   if (isset($_POST['id'])) {
-      unset($_POST['id']);
-   }
+   unset($_GET['id']);
+   unset($_POST['id']);
+   
    // alternative email must be empty for create ticket
    unset($_REQUEST['_users_id_requester_notif']['alternative_email']);
    unset($_REQUEST['_users_id_observer_notif']['alternative_email']);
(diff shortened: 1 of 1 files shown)


front/ticket.form.php
```diff
@@ -203,6 +203,12 @@
 } else {
    Html::header(__('New ticket'),'',"helpdesk","ticket");
    unset($_REQUEST['id']);
+   if (isset($_GET['id'])) {
```
orthagh: isset statements for GET and POST are superflu, unset do nothing (and did no warnings) if the var is not set

PR #3125, opened 2017-11-07
Fixed backslash in ticket title (Close #3123)

### LINKED ISSUE #3123: Backslash in ticket title
Hello,

When we use single quote in ticket tilte (') and change a parameter who relod form (for exemple category), GLPI add 2 backslashes on it each time for is realoded : 

1) Title set : 
image

2) Category changed & form reloaded :
image

3) Category changed angain :
image

...

Thank's

edgardmessias:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #3123 

### DIFF
--- a/inc/ticket.class.php
+++ b/inc/ticket.class.php
@@ -3708,11 +3708,10 @@ class Ticket extends CommonITILObject {
          }
       }
 
-      // Get default values from posted values on reload form
-      if (!$ticket_template) {
-         if (isset($_POST)) {
-            $options = $_POST;
-         }
+      if (isset($options['name'])) {
+         $order           = ["\\'", '\\"', "\\\\"];
+         $replace         = ["'", '"', "\\"];
+         $options['name'] = str_replace($order, $replace, $options['name']);
       }
 
       // Restore saved value or override with page parameter
@@ -4035,12 +4034,16 @@ class Ticket extends CommonITILObject {
          echo "<tr class='tab_bg_1'>";
          echo "<td>".sprintf(__('%1$s%2$s'), __('Title'), $tt->getMandatoryMark('name'))."<td>";
          if (!$tt->isHiddenField('name')) {
-            echo "<input type='text' maxlength='250' size='80' name='name'
-                       value=\"".$options['name']."\"";
+            $opt = [
+                'value'     => $options['name'],
+                'maxlength' => 250,
+                'size'      => 80,
+            ];
+
             if ($tt->isMandatoryField('name')) {
-               echo " required='required'";
+                $opt['required'] = 'required';
             }
-            echo ">";
+            echo Html::input('name', $opt);
          } else {
             echo $options['name'];
             echo "<input type='hidden' name='name' value=\"".$options['name']."\">";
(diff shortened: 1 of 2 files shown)


inc/ticket.class.php
```diff
@@ -3708,11 +3708,10 @@ function showFormHelpdesk($ID, $ticket_template = false) {
          }
       }
 
-      // Get default values from posted values on reload form
```
trasher: I guess some values will be lost on reload if you remove that

inc/ticket.class.php
```diff
@@ -4396,7 +4395,9 @@ function showForm($ID, $options = []) {
          $options['content'] = str_replace($order, $replace, $options['content']);
       }
       if (isset($options['name'])) {
-         $options['name'] = str_replace("\\\\", "\\", $options['name']);
+         $order           = ["\\\\\\'", '\\\\\\"', "\\\\\\\\"];
```
trasher: Huh... really? I know dealing with `\` stuff is not something easy on GLPI... But we cannot add backslashes again and again

inc/ticket.class.php
edgardmessias: This block is repeated
```php
      // Get default values from posted values on reload form
      if (!$ticket_template) {
         if (isset($_POST)) {
            $options = $_POST;
         }
      }
```

trasher: Oh, you're right, I did not pay attention

inc/ticket.class.php
edgardmessias: I'm seeing more details

trasher:
Thanks a lot for your investigations!

And indeed, 👍 to prevent double escaping rather than trying to fix it.

inc/ticket.class.php
```diff
@@ -4036,7 +4035,7 @@ function showFormHelpdesk($ID, $ticket_template = false) {
          echo "<td>".sprintf(__('%1$s%2$s'), __('Title'), $tt->getMandatoryMark('name'))."<td>";
          if (!$tt->isHiddenField('name')) {
             echo "<input type='text' maxlength='250' size='80' name='name'
```
orthagh: Here, to avoid code replication, maybe use 

```php
echo Html::input('name', [
   'value'     => $options['name'],
   'maxlenght' => 250,
   'size'      => 80
];
```

edgardmessias: I will change

edgardmessias: @orthagh, I noticed that there are more inputs like this, can I create a separate pull?

orthagh: for master branch, yes, for 9.2/bf, just here

edgardmessias: Because of `$tt->isMandatoryField('name')`, I can change Html class?

```php
// in ticket.class.php
echo Html::input('name', [
    'value'     => $options['name'],
    'maxlength' => 250,
    'size'      => 80,
    'required'  => $tt->isMandatoryField('name'),
]);


//in html.class.php
   static function formatAttribute($key, $value) {
      // begin new code
      if (is_bool($value)) {
         if ($value) {
            return sprintf('%1$s="%2$s"', $key, $key); //Ex. required="required"
         }
         return '';
      }
      // end new code
      if (is_array($value)) {
         $value = implode(' ', $value);
      }

      return sprintf('%1$s="%2$s"', $key, Html::cleanInputText($value));
   }
```

orthagh: for the modification, only if it's mandatory.
I suggest to prepare the ``$options`` array before and append required in an ``if`` statement

edgardmessias: Modifying keeps code cleaner. The Yii2 framework looks like this.

For master branch, I will make

Issue #3020, opened 2017-10-20
Update to PHPMailer 6


trasher:
PHPMailer has released a new major version; 5 series are only maintained for security fixes.

See https://github.com/PHPMailer/PHPMailer/blob/6.0/UPGRADING.md

trasher:
I've reported an issue regarding to address validation: https://github.com/PHPMailer/PHPMailer/issues/1229

trasher:
Done

PR #3126, opened 2017-11-07
Feature/phpmailer

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #3020

It seems we did not use any deprecated that has been dropped, nor any method that signature has changed.

The only thing I've seen is that `user@localhost` is no longer valid; and will not be. So, I've provided a workaround.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
         "ext-curl": "*",
         "jasig/phpcas": "~1.3.0",
         "iamcal/lib_autolink": "~1.7",
-        "phpmailer/phpmailer": "~5.2",
+        "phpmailer/phpmailer": "~6.0",
         "sabre/vobject": "~4.1",
         "simplepie/simplepie": "^1.5",
         "tecnickcom/tcpdf": "~6.2",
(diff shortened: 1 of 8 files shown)


remicollet:
I think we have an issue with the SetLanguage.

If I read correctly PHPMailer, default path is broken, so we have to give it.
For now getLibraryDir() will return "src" (according to test), but language are not under src, but in ../language.

And, for downstream packaging... default patch works.

So we probably need something like

```
     $lang_path = dirname(Config::getLibraryDir("PHPMailer"))  . '/language/';
     if (is_dir($lang_path)) {
         $this->SetLanguage("en", $lan_path);        
     } else {
         $this->SetLanguage("en"); // Probably system library, rely on default value
     }
```

Perhaps worth to add a unit test, close to https://git.remirepo.net/cgit/rpms/php/phpmailer/php-phpmailer6.git/plain/PHPMailerRpmTest.php in our test suite, to ensure everything is ok (both in upstream version, and in downstream distribution)

remicollet:
For memory https://bugzilla.redhat.com/1513669

PR #3130, opened 2017-11-07
Rotate e-mail queue for not sended e-mails

edgardmessias:
For environments with high volume of emails, unsent emails delay other sent.

Example: If I have 40 unsent emails (Example: Invalid address), the cron need executate 5 (minimal 5 minutes) times to send the others emails.

With the pull request, unsent emails try again later, allowing new emails to sent.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #3130

### DIFF
--- a/inc/notificationeventmailing.class.php
+++ b/inc/notificationeventmailing.class.php
@@ -285,8 +285,15 @@ class NotificationEventMailing extends NotificationEventAbstract implements Noti
             }
 
             $mmail->ClearAddresses();
-            $current->update(['id'        => $current->fields['id'],
-                                'sent_try' => $current->fields['sent_try']+1]);
+            $input = [
+                'id'        => $current->fields['id'],
+                'sent_try'  => $current->fields['sent_try'] + 1
+            ];
+
+            if ($CFG_GLPI["smtp_retry_time"] > 0) {
+               $input['send_time'] = date("Y-m-d H:i:s", strtotime('+' . $CFG_GLPI["smtp_retry_time"] . ' minutes')); //Delay X minutes to try again
+            }
+            $current->update($input);
          } else {
             //TRANS to be written in logs %1$s is the to email / %2$s is the subject of the mail
             Toolbox::logInFile("mail",
--- a/inc/define.php
+++ b/inc/define.php
@@ -35,7 +35,7 @@
 */
 
 // Current version of GLPI
-define('GLPI_VERSION', '9.2.1');
+define('GLPI_VERSION', '9.2.2');
 if (substr(GLPI_VERSION, -4) === '-dev') {
    //for dev version
    define('GLPI_PREVER', str_replace('-dev', '', GLPI_VERSION));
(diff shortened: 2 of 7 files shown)


trasher:
Whatever the point of vue is; many emails will not be sent because of not reachable emails.
Without a similar change, all emails would be blocked for a while until max retries is reached.

On many mails servers, a similar action is done:
- try to send the mail as it is received,
- on fail, postpone for some time (it is quite frequent the postpone time is increased).

I don't think we got any way to be more precise.

inc/notificationeventmailing.class.php
```diff
@@ -285,8 +285,11 @@ static public function send(array $data) {
             }
 
             $mmail->ClearAddresses();
-            $current->update(['id'        => $current->fields['id'],
-                                'sent_try' => $current->fields['sent_try']+1]);
+            $current->update([
+                'id'        => $current->fields['id'],
+                'send_time' => date("Y-m-d H:i:s", strtotime('+2 minutes')), //Delay 2 minutes to try again
```
trasher: Just a question... If your crontab is ran each 5 minutes; the mecanism will not be efficient; am I wrong?

I also guess this should made configurable (from an overridable constant).

edgardmessias: It will be efficient anyway, it will try to send the emails again, but in the next run, it will send the remaining emails instead of trying the wrong emails again

edgardmessias: Can I create the option to allow setting this

edgardmessias:
Added option for try to deliver again.

inc/notificationeventmailing.class.php
trasher: Each entry in the queue should be tried only once per run; if you have something different this is probably a  bug.
As of the option, I had in mind a constant in `inc/define.php`; not from the configuration, but this can be ok as well.

edgardmessias: But each list item is tested only once, will only be tested again the next run.
I will try to make an example

edgardmessias: Run step: 3

| 10:00 | 10:05 | 10:10 | 10:15 | 10:20 |
|---|---|---|---|---|
| 1W 10:00  | 4W 10:00 | 5W 10:04 | 1W 10:07 | 5W 10:12 |
| 2W 10:00  | 1W 10:02 | 6R 10:04 | 2W 10:07 | 1W 10:17 |
| 3R 10:00  | 2W 10:02 | 4W 10:07 | 4W 10:12 | 2W 10:17 |
| 4R 10:00  | 5W 10:04 | 1W 10:07 | 5W 10:12 | 4W 10:17 |
| 5W 10:04  | 6R 10:04 | 2W 10:07 |  |   |
| 6R 10:04 |   |   |   |   |

W - Wrong
R - Right

trasher:
Unit tests are failing; update is not OK. Dealing with versions with or without `-dev` requires some attention. As far as I know, we do n ot have any documentation on this (see https://github.com/glpi-project/docdev/issues/70).

edgardmessias:
Done.

PR #3146, opened 2017-11-10
Improved use of flush in output buffering

edgardmessias:
More details, see https://www.sitepoint.com/php-streaming-output-buffering-explained/ and https://www.sitepoint.com/faster-web-pages-php-buffer-flush/

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 

### DIFF
--- a/inc/html.class.php
+++ b/inc/html.class.php
@@ -971,7 +971,7 @@ class Html {
       }
 
       if (!$params['create']) {
-         Html::glpi_flush();
+         self::glpi_flush();
       }
    }
 
@@ -1272,6 +1272,7 @@ class Html {
 
       // End of Head
       echo "</head>\n";
+      self::glpi_flush();
    }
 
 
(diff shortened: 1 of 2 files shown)


inc/html.class.php
```diff
@@ -1272,6 +1272,7 @@ static function includeHeader($title = '', $sector = 'none', $item = 'none', $op
 
       // End of Head
       echo "</head>\n";
+      self::glpi_flush();
```
trasher: I'm not sure this one is really relevant.

trasher:
GLPI makes heavy use of javascript, which is only run once dom is loaded. What would be the benefit of playing with outbut buffering? Pages will not work anyways.

inc/html.class.php
edgardmessias: Once the browser has received the HTML head, it can begin to download all linked CSS files, favicons and other resources. These downloads can occur while the browser is idle and waiting for the main body content.

trasher: OK, good point :)

edgardmessias:
It will optimize search pages, reducing the wait time to start downloading the content.

trasher:
OK, as of https://github.com/glpi-project/glpi/pull/3146#discussion_r151645028 ; that is more clear to me,. Thanks

Issue #3144, opened 2017-11-10
Update select2


trasher:
Current select2 component is from... 2014. And the 3.x series is no longer maintained.

Also, current component has several limitations making hard to dynamically change some contents or parameters.

So we should upgrade to version 4.

trasher:
A kind of migration guide: https://select2.org/upgrading/migrating-from-35

trasher:
Done

PR #3173, opened 2017-11-17
Update select2

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | https://github.com/glpi-project/glpi/issues/3144


trasher:
There are 2 distinct commits. The first one is the entire migration to new select2 version.

The second one drops `_one_id` references in ajax calls. `_one_id` was part of select2 `initSelection` that has been removed in v4.
As far as I understand, the "one id" stuff was made to retrieve one key/value to instantiate probably on first load. Call was made only when select2 current id was different from element id; which does not happen on initial load.

Since slect2 now works differently, a `select` element is created with at least one option containing the current value. One id stuff does not seems to be usefull then.

trasher:
Gihub has bugged; this one has been merged :)

Issue #3189, opened 2017-11-20
Wrong results from changeActiveEntities end point


DIOHz0r:
The changeActiveEntities end point  of the API REST returns invalid status code and result if the param entities_id is missing and the is_recursive has an invalid value.

If you send the following body you will get an HTTP status 200 and a body with a false string.
```json
{"entities_id": 1, "is_recursive": "invalidValue"}
```

If you don't send the entities_id param and a wrong value for "is_recursive" in the body, you will get an HTTP status 200 and a true string.
```json
{"is_recursive": "invalidValue"}
```

The documentation https://dev.flyve.org/glpi/apirest.php/#change-active-entities says than it must returns a "400 (Bad Request) with a message indicating an error in input parameter."

PR #3194, opened 2017-11-21
Fix changeActiveEntities to return correct responses

DIOHz0r:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #3189

### DIFF
--- a/inc/api.class.php
+++ b/inc/api.class.php
@@ -283,7 +283,7 @@ abstract class API extends CommonGLPI {
     *   - 'entities_id': (default 'all') ID of the new active entity ("all" = load all possible entities). Optionnal
     *   - 'is_recursive': (default false) Also display sub entities of the active entity.  Optionnal
     *
-    * @return array
+    * @return array|bool
     */
    protected function changeActiveEntities($params = []) {
 
@@ -304,6 +304,8 @@ abstract class API extends CommonGLPI {
       if (!Session::changeActiveEntities($entities_id, $params['is_recursive'])) {
          return $this->returnError();
       }
+
+      return true;
    }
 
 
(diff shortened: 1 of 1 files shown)


inc/api.class.php
```diff
@@ -283,22 +283,27 @@ protected function retrieveSession() {
     *   - 'entities_id': (default 'all') ID of the new active entity ("all" = load all possible entities). Optionnal
     *   - 'is_recursive': (default false) Also display sub entities of the active entity.  Optionnal
     *
-    * @return boolean
+    * @return array
```
trasher: On success; this will return nothing.

inc/api.class.php
```diff
@@ -283,22 +283,27 @@ protected function retrieveSession() {
...
 
       if (!isset($params['is_recursive'])) {
          $params['is_recursive'] = false;
+      } else if (!is_bool($params['is_recursive'])) {
+         return $this->returnError();
       }
 
-      return Session::changeActiveEntities(intval($params['entities_id']),
-                                           $params['is_recursive']);
+      if (!Session::changeActiveEntities($entities_id, $params['is_recursive'])) {
```
orthagh: with this change, we have now no success response from the endpoint where we have ``true`` before.
Not sure if before answer was the best, but with this change, we have nothing.
Maybe, if the ``Session::changeActiveEntities(...)`` part return ok, we could call ``$this->getActiveEntities()``

DIOHz0r:
Hello, I've updated the function to return the true value, please review it again.

PR #3195, opened 2017-11-21
Don't read again DB for data already in memory

remicollet:
1/ don't read DB

2/ don't display relation type when not possible (ambiguous or deleted)


remicollet:
Notice: PR squashed, but keep 2 commits with really (but related) different changes.

PR #3243, opened 2017-12-04
Try to detect failing splits to avoid some issues on display

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

Example on tickets : if a ticket contents contains `$#$`; what follows this string is considered as the ID... That is wrong.


trasher:
Current PR fixes a specific issue; but there may be others cases that are still failing :/

Issue #3245, opened 2017-12-05
Sql error on 9.2.1


moobyfr:
After updating to 9.2.1, I got the sql-error.log flooded with SQL errors
```
  SQL: SELECT *
                      FROM `glpi_softwares`
                      WHERE `entities_id` = Entité racine &gt; IUT RS
                            AND `name` = 'FusionInventory Agent 2.3.19 (x64 edition)'
                            AND `manufacturers_id` = '2148'
  Error: You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'racine &gt; IUT RS
                            AND `name` = 'FusionInventory Age' at line 3
  Backtrace :
  inc/dbmysqliterator.class.php:274
  inc/dbmysql.class.php:542                          DBmysqlIterator->__construct()
  inc/transfer.class.php:1435                        DBmysql->request()
  inc/transfer.class.php:1473                        Transfer->copySingleSoftware()
  inc/transfer.class.php:1538                        Transfer->copySingleVersion()
  inc/transfer.class.php:1150                        Transfer->transferComputerSoftwares()
  inc/transfer.class.php:199                         Transfer->transferItem()
  ...ry/inc/inventorycomputerinventory.class.php:504 Transfer->moveItems()
  ...inventory/inc/inventoryruleimport.class.php:790 PluginFusioninventoryInventoryComputerInventory->rulepassed()
  inc/rule.class.php:1436                            PluginFusioninventoryInventoryRuleImport->executeActions()
  inc/rulecollection.class.php:1461                  Rule->process()
  ...ry/inc/inventorycomputerinventory.class.php:378 RuleCollection->processAllRules()
  ...ry/inc/inventorycomputerinventory.class.php:126 PluginFusioninventoryInventoryComputerInventory->sendCriteria()
  ...fusioninventory/inc/communication.class.php:235 PluginFusioninventoryInventoryComputerInventory->import()
  ...fusioninventory/inc/communication.class.php:463 PluginFusioninventoryCommunication->import()
  plugins/fusioninventory/front/communication.php:93 PluginFusioninventoryCommunication->handleOCSCommunication()
  plugins/fusioninventory/index.php:59               include_once()
```

This is about entities_id value not being quoted.

AnaelMobilia:
@trasher Will the new query builder take care of this kind of problems ? (just to know :))

moobyfr:
[...]

PR #3246, opened 2017-12-05
Update transfer.class.php for quote usage

moobyfr:
Fix the usage of entities_id if they contain non standard caracters
This is a fix for #3245

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #3245

### DIFF
--- a/inc/transfer.class.php
+++ b/inc/transfer.class.php
@@ -1428,7 +1428,7 @@ class Transfer extends CommonDBTM {
             }
             $query = "SELECT *
                       FROM `glpi_softwares`
-                      WHERE `entities_id` = ".$this->to."
+                      WHERE `entities_id` = '".$this->to."'
                             AND `name` = '".addslashes($soft->fields['name'])."'
                             $manufacturer";
 


trasher:
I guess the isse on #3245 is elsewhere. Even if we can properly quote the id; `entities_id` must be an integer, not a string. Seems like we get the entity name instead; and that is wrong.

moobyfr:
Perhaps it should be an id, but this fix is to be consistant on the usage of `entities_id` in this file , there are 9 search for `entities_id` and this one was the only one not quoted

trasher:
@moobyfr that's what I said. But there is another issue that should be fixed here

AnaelMobilia:
@trasher Will the new query builder take care of this kind of problems ? (just to know :))

trasher:
@AnaelMobilia you mean the DbIterator? Yes, it automatically enclose strings in single quotes

trasher:
BTW, this should reach 9.2/bf branch

PR #3250, opened 2017-12-05
Feature/insert update

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

TODOs:
- [x] Fix tests on DBIterator
- [x] Add tests on new methods
- [x] Add documentation

### DIFF
--- a/inc/certificate.class.php
+++ b/inc/certificate.class.php
@@ -689,9 +689,6 @@ class Certificate extends CommonDBTM {
       }
 
       $message      = [];
-      $items_notice = [];
-      $items_end    = [];
-
       foreach (Entity::getEntitiesToNotify('use_certificates_alert') as $entity => $value) {
          $before = Entity::getUsedConfig('send_certificates_alert_before_delay', $entity);
          // Check licenses
--- a/inc/log.class.php
+++ b/inc/log.class.php
@@ -228,15 +228,20 @@ class Log extends CommonDBTM {
          $new_value = Toolbox::substr($new_value, 0, 250);
       }
 
-      // Build query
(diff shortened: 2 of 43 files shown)


inc/certificate.class.php
```diff
@@ -689,9 +689,6 @@ static function cronCertificate($task = null) {
       }
 
       $message      = [];
-      $items_notice = [];
```
orthagh: seems strange, but i checked, and it's ok

inc/log.class.php
```diff
@@ -228,15 +228,20 @@ static function history ($items_id, $itemtype, $changes, $itemtype_link = '', $l
...
-                        '".addslashes($username)."', '$date_mod', '$id_search_option',
-                        '$old_value', '$new_value')";
-
-      if ($DB->query($query)) {
+      $params = [
+         'items_id'          => $items_id,
+         'itemtype'          => $itemtype,
+         'itemtype_link'     => $itemtype_link,
+         'linked_action'     => $linked_action,
+         'user_name'         => $username,
```
orthagh: addslashes dropped ? you kept one here: https://github.com/glpi-project/glpi/pull/3250/files#diff-11a09748e657af84b23fd2599f1f3101R1248

trasher: Indeed; this must stay

inc/certificate.class.php
trasher: Yep; they were not used

trasher:
Doc: https://github.com/glpi-project/docdev/pull/71

PR #3300, opened 2017-12-14
Feature/pdu

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes


inc/pdu.class.php
```diff
@@ -0,0 +1,254 @@
...
+
+      echo "<textarea cols='45' rows='3' id='comment' name='comment' >".
+           $this->fields["comment"];
+      echo "</textarea></td></tr>";
+
+      $this->showFormButtons($options);
+      return true;
+   }
+
+   /*function getSearchOptionsNew() {
```
orthagh: todo ?

inc/pdu.class.php
```diff
@@ -0,0 +1,254 @@
...
+            'rand'      => $rand
+         ]
+      );
+      echo "</td></tr>\n";
+
+      echo "<tr class='tab_bg_1'>";
+      echo "<td><label for='comment'>".__('Comments')."</label></td>";
+      echo "<td colspan='3' class='middle'>";
+
+      echo "<textarea cols='45' rows='3' id='comment' name='comment' >".
```
orthagh: not mandatory, but we have a function ``Html::textarea``

inc/pdu.class.php
trasher: Indeed, yes; I forget to do that part

inc/pdu.class.php
trasher: Just quickly tested... Seems that there are issues with labels (name is not part of the id, rand seems ignored, ...) :(

PR #3302, opened 2017-12-14
Fix sanitize the base64 img. closes #3301

### LINKED ISSUE #3301: sanitize img base64
With the use of htmlawed, the base64 img in knowbase, ticket solution... can be detected like a false positive and so denied the image to display

ddurieux:


ddurieux:
No because htmlawed test many combinaisons :(

ddurieux:
Yes I will add a test ;)

Issue #2926, opened 2017-10-06
Datacenter mode


orthagh:
Integration of racks plugins

magaiverpr:
It will be cool if the can add a way to control patch panels too.

eduardospinola:
And enclosures for Blade server ( or other hardwares ) too

trasher:
Current specs are available here: https://github.com/glpi-project/spec/wiki/GLPI-Datacenters-management

Please comment on https://github.com/glpi-project/spec/issues/1

orthagh:
@fcrispim77 the job is done and will be released in the next version of glpi.
Beta version will be announced soon.

PR #3316, opened 2017-12-19
add bijective base26 (excel like) x index in room

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2926 

image

### DIFF
--- a/inc/toolbox.class.php
+++ b/inc/toolbox.class.php
@@ -2859,4 +2859,20 @@ class Toolbox {
          && (!defined('TU_USER') || defined('CACHED_TESTS'));
    }
 
+   /**
+    * Convert a integer index into an excel like alpha index (A, B, ..., AA, AB, ...)
+    * @since 9.3
+    * @param  integer $index the numeric index
+    * @return string         excel like string index
+    */
+   public static function getBijectiveIndex($index = 0) {
+      $bij_str = "";
+      while ((int) $index > 0) {
+         $index--;
+         $bij_str = chr($index%26 + ord("A")) . $bij_str;
+         $index /= 26;
+      }
+      return $bij_str;
+   }
+
 }
--- a/js/common.js
+++ b/js/common.js
@@ -1044,6 +1044,22 @@ var replaceContent = function(editor, search, replace) {
 };
 
 /**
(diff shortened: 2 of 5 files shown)


inc/toolbox.class.php
```diff
@@ -2859,4 +2859,19 @@ public static function useCache() {
          && (!defined('TU_USER') || defined('CACHED_TESTS'));
    }
 
+   /**
+    * Convert a integer index into an excel like alpha index (A, B, ..., AA, AB, ...)
```
trasher: Missing @since

js/common.js
```diff
@@ -1043,6 +1043,21 @@ var replaceContent = function(editor, search, replace) {
    editor.setContent(body);
 };
 
+/**
+ * Convert a integer index into an excel like alpha index (A, B, ..., AA, AB, ...)
+ * @param  int index    the numeric index
```
trasher: Missing @since. s/int/integer/

tests/units/Toolbox.php
```diff
@@ -120,4 +120,18 @@ public function testUncleanHtmlCrossSideScriptingDeepImageBase64() {
...
+      foreach ([
+         1   => 'A',
+         2   => 'B',
+         27  => 'AA',
+         28  => 'AB',
+         53  => 'BA',
+         702 => 'ZZ',
+         703 => 'AAA',
+      ] as $number => $bij_string) {
+         $this->string($bij_string)->isIdenticalTo(\Toolbox::getBijectiveIndex($number));
```
trasher: That does not matter much, but string() is the string to test and isIdenticalTo() is the expected value :)

orthagh:
fixed

js/common.js
```diff
@@ -1043,6 +1043,22 @@ var replaceContent = function(editor, search, replace) {
    editor.setContent(body);
 };
 
+/**
+ * Convert a integer index into an excel like alpha index (A, B, ..., AA, AB, ...)
+ * @since  9.2
```
trasher: 9.3 :p

orthagh: my bad, fixed

Issue #3187, opened 2017-11-20
The simplified interface users can still change the ticket after a technician or group is assigned


arthurrschaefer:
After the simplified interface user opens a ticket. He is able to change some dropdowns like Type, Category, or even title and description. It would be ok if the ticket is not being processed by the tech team. Because the ticket is still in process of being analyzed and the tech team, probably, didn't get in touch with it.

But once the ticket is assigned to a technician or a group. It shouldn't be changed. At least the Type, category, title, and description should be fixed once the team started the attendance.

But users are yet in control of these fields, even after the assignments.

![image](https://user-images.githubusercontent.com/13874922/33023062-2e06ed6a-cdee-11e7-95b6-a28c353cdaf7.png)

GLPI 9.2.1
Installation mode: TARBALL

**Server**
 
Operating system: Linux ip-172-31-1-231 3.16.0-4-amd64 #1 SMP Debian 3.16.43-2 (2017-04-30) x86_64
PHP 5.6.30-0+deb8u1 apache2handler (Core, PDO, Phar, Reflection, SPL, SimpleXML, Zend OPcache, apache2handler, apc, apcu,
	bcmath, bz2, calendar, ctype, curl, date, dba, dom, ereg, exif, fileinfo, filter, ftp, gd, gettext, hash, iconv, imap, json,
	ldap, libxml, mbstring, mcrypt, mhash, mysql, mysqli, openssl, pcre, pdo_mysql, posix, readline, session, shmop, snmp, soap,
	sockets, standard, sysvmsg, sysvsem, sysvshm, tokenizer, wddx, xml, xmlreader, xmlrpc, xmlwriter, zip, zlib)
Setup: max_execution_time="30" memory_limit="128M" post_max_size="8M" safe_mode="" session.save_handler="files"
	upload_max_filesize="10M" 
Software: Apache/2.4.10 (Debian) (Apache/2.4.10 (Debian) Server at prototipo.bavierati.com.br Port 80)
	Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/62.0.3202.94 Safari/537.36
Server Software: (Debian)
	Server Version: 10.0.30-MariaDB-0+deb8u2
	Server SQL Mode: 
	Parameters: proto@localhost/prototipo
	Host info: Localhost via UNIX socket
	

- mysqli extension is installed
- ctype extension is installed
- fileinfo extension is installed
- json extension is installed
[...]

Issue #2504, opened 2017-07-20
[bug] Postonly can alter fields on form after created ticket (9.1.6)


ricardoasimoes:
When the postonly create a ticket, he can change almost every field on ticket.
All the field is avaliable to change but only works in the fields category and description.
In lasted releases (<=9.1.3) this thing doesn't working like that. All fields was not be able to be changed. This way is better

arthurrschaefer:
For now, I think I'm keeping 9.1.3 working here.

PierreCPF:
I have the same issue, but it's an issue or the new ticket access for post-only users ?

arthurrschaefer:
#3187

arthurrschaefer:
In the 9.2.2 post-only are just able to change Category, Urgency and Ticket Title.
Even though, I still don't believe it's a good thing.

I believe the user must be able to change things just until the ticket is not owned by some tech. If it's assigned to someone and work has begun, the ticket fields should be blocked. No changes during the service.

PR #3317, opened 2017-12-19
Feature/fix ticket self service

btry:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #2504 #3187

Somewhere between GLPI 9.1.2 and 9.1.6 post only users can edit not modifiable fields. These fields are not actually saved in DB, but displayed as HTML inputs.

## Comaprison before the regression
GLPI 9.1.2 (self service profile)
image

GLPI 9.2.1 (self service profile)
image

## with the fix

image


btry:
@trasher : more tests added. Fields impacted by the PR are checked.

PR #3327, opened 2017-12-20
add room from datacenter, and inherits locations from parents

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2926

### DIFF
--- a/front/dcroom.form.php
+++ b/front/dcroom.form.php
@@ -95,6 +95,16 @@ if (isset($_POST["add"])) {
 
 } else {
    Html::header(DCRoom::getTypeName(Session::getPluralNumber()), $_SERVER['PHP_SELF'], "management", "dcroom");
-   $room->display(['id' => $_GET["id"]]);
+   $options = [
+      'id' => $_GET["id"],
+   ];
+   if (isset($_REQUEST['_add_fromitem'])
+       && isset($_REQUEST['datacenters_id'])) {
+      $options['datacenters_id'] = $_REQUEST['datacenters_id'];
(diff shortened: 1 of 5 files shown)


front/dcroom.form.php
```diff
@@ -95,6 +95,16 @@
 
 } else {
    Html::header(DCRoom::getTypeName(Session::getPluralNumber()), $_SERVER['PHP_SELF'], "management", "dcroom");
-   $room->display(['id' => $_GET["id"]]);
+   $options = [
+      'id' => $_GET["id"],
+   ];
+   if (isset($_REQUEST['_add_fromitem'])
+       && isset($_REQUEST['datacenters_id'])) {
+      $options['datacenters_id'] = $_REQUEST['datacenters_id'];
```
trasher: I guess logic should be part of the class, not the front

front/rack.form.php
```diff
@@ -106,7 +106,11 @@
       $options['position'] = $_GET['position'];
    }
    if (isset($_GET['room'])) {
-      $options['room'] = $_GET['room'];
+      $room = new DCRoom;
```
trasher: Same here, logic should be in the class, not on the front.

front/dcroom.form.php
trasher: More globally, I think whe should set location from datacenter (when no location for the room itself has been set) every time (on not only when added from the datacenter button).

PR #3337, opened 2017-12-21
add reservation for items_racks

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2926 

image

### DIFF
--- a/inc/toolbox.class.php
+++ b/inc/toolbox.class.php
@@ -2672,20 +2672,21 @@ class Toolbox {
     * Because json can have been modified from addslashes_deep
     *
     * @param string $encoded Encoded JSON
+    * @param boolean $assoc  assoc parameter of json_encode native function
     *
     * @return mixed
     */
-   static public function jsonDecode($encoded) {
+   static public function jsonDecode($encoded, $assoc = false) {
       if (!is_string($encoded)) {
          Toolbox::logDebug('Only strings can be json to decode!');
          return $encoded;
       }
 
-      $json = json_decode($encoded);
+      $json = json_decode($encoded, $assoc);
 
       if (json_last_error() != JSON_ERROR_NONE) {
          //something went wrong... Try to stripslashes before decoding.
-         $json = json_decode(self::stripslashes_deep($encoded));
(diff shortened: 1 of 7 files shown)


inc/toolbox.class.php
```diff
@@ -2676,22 +2676,28 @@ static function cleanTagOrImage($content, array $tags) {
...
+    * @param boolean $assoc  assoc parameter of json_encode native function
+    * @param boolean $debug  generate a debug message in log
     *
     * @return mixed
     */
-   static public function jsonDecode($encoded) {
+   static public function jsonDecode($encoded, $assoc = false, $debug = true) {
       if (!is_string($encoded)) {
-         Toolbox::logDebug('Only strings can be json to decode!');
+         if ($debug) {
```
trasher: I'm not sure this message should be conditionned

inc/toolbox.class.php
```diff
@@ -2676,22 +2676,28 @@ static function cleanTagOrImage($content, array $tags) {
...
-      $json = json_decode($encoded);
+      $json = json_decode($encoded, $assoc);
 
       if (json_last_error() != JSON_ERROR_NONE) {
          //something went wrong... Try to stripslashes before decoding.
-         $json = json_decode(self::stripslashes_deep($encoded));
+         $json = json_decode(self::stripslashes_deep($encoded), $assoc);
          if (json_last_error() != JSON_ERROR_NONE) {
-            Toolbox::logDebug('Unable to decode JSON string! Is this really JSON?');
+            if ($debug) {
```
trasher: Same here. If we're expecting a json and there is something wrong; there is certainly a bug somewhere

tests/units/Toolbox.php
```diff
@@ -134,4 +134,18 @@ public function testgetBijectiveIndex() {
          $this->string(\Toolbox::getBijectiveIndex($number))->isIdenticalTo($bij_string);
       }
    }
+
+   public function testJsonDecode() {
```
trasher: It should be better to use a dataprovider. Also, maybe test a failing case

PR #3360, opened 2018-01-02
Fix SQL error in API (getItems) when class extend dbchild but not hav…

### LINKED ISSUE #3357: SQL error when use API getitems on TicketTemplateMandatoryField
The code `$item->isEntityAssign()` return yes because the `$itemtype` as an entities_id so it add in the `where` the field `entities_id` but this field not exist.

I think need fix the function `isEntityAssign` in `CommonDBConnexity` but not really sure

ddurieux:
…e field 'entities_id'. closes #3357

### DIFF
--- a/inc/api.class.php
+++ b/inc/api.class.php
@@ -1097,9 +1097,8 @@ abstract class API extends CommonGLPI {
 
       // filter with entity
       if ($item->isEntityAssign()
-          && !is_subclass_of($item, "CommonDBChild")
           // some CommonDBChild classes may not have entities_id fields and isEntityAssign still return true (like TicketTemplateMandatoryField)
-          || array_key_exists('entities_id', $item->fields)) {
+          && array_key_exists('entities_id', $item->fields)) {
          $where.= " AND (". getEntitiesRestrictRequest("",
                                              $itemtype::getTable(),
                                              '',
--- a/inc/api.class.php
+++ b/inc/api.class.php
@@ -1097,9 +1097,9 @@ abstract class API extends CommonGLPI {
 
       // filter with entity
       if ($item->isEntityAssign()
+          && !is_subclass_of($item, "CommonDBChild")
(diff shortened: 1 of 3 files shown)


inc/api.class.php
```diff
@@ -1083,14 +1083,14 @@ protected function getItems($itemtype, $params = [], &$totalcount = 0) {
          foreach ($params['searchText']  as $filter_field => $filter_value) {
             if (!empty($filter_value)) {
                $search = Search::makeTextSearch($filter_value);
-               $where.= " AND (`$table`.`$filter_field` $search
-                               OR `$table`.`id` $search)";
+               $where.= " AND (`$table`.`$filter_field` $search)";
```
trasher: I really do not know if this is ok :/

ddurieux: It's idiot, if we search `tickettemplates_id` = 1, it will get all right line + the line with `id` =1, so the list return is wrong because we don't want the `id=1`

ddurieux:
@orthagh so I update to manage the entities_id in the CommonDBChild::isEntityAssign ?

inc/api.class.php
```diff
@@ -1083,14 +1083,14 @@ protected function getItems($itemtype, $params = [], &$totalcount = 0) {
...
             }
          }
       }
 
       // filter with entity
-      if ($item->isEntityAssign()) {
+      if ($item->isEntityAssign
+              && array_key_exists('entities_id', $item->fields)) {
```
orthagh: last thing before merge for this part, i think it could be better to precise ``$item instanceof CommonDBChild`` to avoid removal in the future (seems duplicate with above line at first look) and maybe add a comment with an example (TicketTemplateMandatoryField doesn't have entities_id field)

orthagh:
branch now good to merge

ddurieux:
Thanks ;)

PR #3363, opened 2018-01-02
Apply TCPDF patch for PHP 7.2

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes


orthagh:
any issue upstream ?

trasher:
Zend patch is from upstream repo; but TCPDF is considered 'dead' (while the new one is still not stable... :/)

PR #3383, opened 2018-01-05
Add deleteOrDie method

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

### DIFF
--- a/inc/dbmysql.class.php
+++ b/inc/dbmysql.class.php
@@ -201,18 +201,19 @@ class DBmysql {
    }
 
    /**
-    * Execute a MySQL query
+    * Execute a MySQL query and die
+    * (optionnaly with a message) if it fails
     *
     * @since 0.84
     *
     * @param string $query   Query to execute
-    * @param string $message Explaination of query (default '')
+    * @param string $message Explanation of query (default '')
     *
     * @return mysqli_result Query result handler
     */
    function queryOrDie($query, $message = '') {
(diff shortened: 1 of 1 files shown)


inc/dbmysql.class.php
```diff
@@ -993,4 +993,25 @@ public function delete($table, $where) {
...
    }
+
+   /**
+    * Delete a row in the database
+    *
+    * @since 9.3
+    *
+    * @param string $table   Table name
+    * @param array  $where   WHERE clause
+    * @param string $message Explaination of query (default '')
```
orthagh: Explana**i**tion -> Explanation

inc/dbmysql.class.php
```diff
@@ -993,4 +993,25 @@ public function delete($table, $where) {
...
+    *
+    * @return mysqli_result|boolean Query result handler
+    */
+   function deleteOrDie($table, $where, $message = '') {
+      //TRANS: %1$s is the description, %2$s is the query, %3$s is the error message
```
orthagh: I think this comment should be moved before 1012

inc/dbmysql.class.php
```diff
@@ -993,4 +993,25 @@ public function delete($table, $where) {
       $result = $this->query($query);
       return $result;
    }
+
+   /**
+    * Delete a row in the database
```
orthagh: maybe add ``or die if it fails (with a message)``

PR #3384, opened 2018-01-05
Fix ticket self service for GLPI 9.3

btry:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #2504 #3187

This PR is a copy of #3317 and its purpose is to ehnance it to solve unit tests issues not encounteered with GLPI 9.2.

The difference between 9.2 and 9.3 is due to update of the select 2 component.

With GLPI 9.2, dropdowns are created from HTML input tags

with GLPI 9.3, dropdowns are created with HTML select tags.

Somewhere between GLPI 9.1.2 and 9.1.6 post only users can edit not modifiable fields. These fields are not actually saved in DB, but displayed as HTML inputs.

## Comparison before the regression
GLPI 9.1.2 (self service profile)
image

GLPI 9.2.1 (self service profile)
image

## with the fix

image

### DIFF
--- a/inc/levelagreement.class.php
+++ b/inc/levelagreement.class.php
@@ -244,8 +244,6 @@ abstract class LevelAgreement extends CommonDBChild {
     * @param  bool           $canupdate update right
     */
    function showForTicket(Ticket $ticket, $type, $tt, $canupdate) {
-      global $CFG_GLPI;
-
       list($dateField, $laField) = static::getFieldNames($type);
       $rand = mt_rand();
       $pre  = static::$prefix;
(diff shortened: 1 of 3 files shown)


AnaelMobilia:
@btry Seem you commit against 9.2/bugfixes and not master (if for 9.3... ;-))

btry:
@AnaelMobilia You're right. I changed the base of the PR, but the work is still unfinished.

PR #3386, opened 2018-01-06
Remove IE<= 10 CSS hacks

AnaelMobilia:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | none

Fix errors on CSS files (comments, bad values, unexisting files...)
Remove IE <= 10 specific CSS + JS


css/print.css
```diff
@@ -980,8 +980,8 @@ ul#onglet span.fake {
     /*css hack for IE*/
    position: absolute;
    bottom: auto;
-   top: expression((parseInt(document.documentElement.clientHeight)+parseInt(document.documentElement.scrollTop)-42)+'px');
-   width: expression(parseInt(document.documentElement.clientWidth)+'px');
+   /*top: expression((parseInt(document.documentElement.clientHeight)+parseInt(document.documentElement.scrollTop)-42)+'px');
```
trasher: This should be dropped, not commented ;)

css/styles.css
```diff
@@ -1152,12 +1152,12 @@ td.rotate {
   -ms-transform:rotate(-70deg); /* IE 9 */
   width: 10%;
   height: 80px;
-  <!--[if lte IE 7]>
-   filter:  progid:DXImageTransform.Microsoft.BasicImage(rotation=0.083);  /* IE6,IE7 */
-  <![endif]-->
-  <!--[if lte IE 8]>
-   -ms-filter: "progid:DXImageTransform.Microsoft.BasicImage(rotation=0.083)"; /* IE8 */
-  <![endif]-->
+  /*<!--[if lte IE 7]>*/
```
trasher: Same drop

AnaelMobilia:
Maybe we need to think to use a GUI framework ?
See #3392

css/jstree-glpi.css
```diff
@@ -329,8 +329,6 @@
   opacity: 0.8;
   filter: url("data:image/svg+xml;utf8,<svg xmlns=\'http://www.w3.org/2000/svg\'><filter id=\'jstree-grayscale\'><feColorMatrix type=\'matrix\' values=\'0.3333 0.3333 0.3333 0 0 0.3333 0.3333 0.3333 0 0 0.3333 0.3333 0.3333 0 0 0 0 0 1 0\'/></filter></svg>#jstree-grayscale");
   /* Firefox 10+ */
-  filter: gray;
-  /* IE6-9 */
   -webkit-filter: grayscale(100%);
```
trasher: According to MDN, `filter: grayscale(100%);` can be used for Firefox. `-webkit` must be kept for edge.

AnaelMobilia: Be warn that on this case, the developper write code and the line below the comments ;-)
Else IE6-9 have the webkit rendering engine 😨

trasher: Yes, I've understand that :D

So, `filter: url()` was for Firefox, but this does not seems needed right now.

orthagh:
@AnaelMobilia 
Do you mind if i push some additional commits to this pr ? i want to clean more the common.js part and fix some edge issues.

AnaelMobilia:
@orthagh No problems :-)
Thanks for asking !

PR #3388, opened 2018-01-06
Add used JS libraries + links to dev

AnaelMobilia:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | none

Add used JS libraries + links to dev
Will help to maintain up to date / get doc

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,3 +81,22 @@ We are working on a [markdown version](https://github.com/glpi-project/doc)
 * IRC : irc://irc.freenode.org/glpi
 * [Plugin directory](http://plugins.glpi-project.org)
 * [Plugin development documentation](https://github.com/pluginsGLPI/example)
+
+
+## Used JS libraries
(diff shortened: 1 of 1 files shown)


AnaelMobilia:
Tests are breaked due to a composer.org availibity issue... nothing to care with the PR :)

orthagh:
I'm not sure about this one.
The "best" way should be to use a dependency manager (npm/yarn/bower not sure about the correct one today).
But our last tests seems to add a lot of noise files in the archive

trasher:
This is indeed not the "best" way; dependency manager seems not to be the best way neither...

For me, this is OK; I did exactly the same for the Telemetry app, this would permit to know very quickly which versions are used and if upgrade are needed.

orthagh:
so ok for me

PR #3398, opened 2018-01-07
Protect /inc/ with .htaccess

AnaelMobilia:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | none

Prevent access on /inc/* with an .htaccess file (as for /config/).
With an .htaccess, all files are protected (some class files doesn't include check for GLPI_ROOT) and we delegate to the delivery software the job to protect files...

Regards


trasher:
I do not agree. htaccess fles targets apache HTTPD server only, and it can be disabled per configuration.

trasher:
I'm not against adding a .htaccess file; but php checks should not be removed.

If some are missing, this is probably a mistake, or maybe intended (in this case, a comment would be required).

As of checking if files are not readable from the web; the best option would be to prevent thos directories to be included in the webroot (which is a huge work).
Adding a check would be possible, but we will have to give support for that on every existing web server; so maybe just a warning.

AnaelMobilia:
Hi,

For each file in /inc/ missing the verification, no one appears to be intended, only developper mistakes (I checked every one) :-)

On one way, I understand your position.
On the other, I didn't understand why user's datas are less important than GLPI code.

If we have a non Apache server, any files on /files/_upload/ can be read. Eg licences serial document : can cause serious financial or business impact.

For me, these checks on /inc/ are false sense of security as we didn't handle the global access question.
I'm convinced that it's a reassuring code, but it's false security.

I think a more global way of handle this problem is an folder access check, on the same title than check for install script or usage of default users on the (after) login screen.

Moving folders out of web path could be an idea, imo, we would have many issues with IIS on this case. And a lot of work :-)

I'm convinced than the warning on (after) login screen will increase security as it will warn users after updating their versions with IIS or bad web server configuration.

I think getting feedback of others contributors could be interesting on this case.

Regards,
Anael

trasher:
>  On the other, I didn't understand why user's datas are less important than GLPI code.

This is not :)

>  I'm convinced that it's a reassuring code, but it's false security.

Yes... And no.

Once again, the point is not to prevent adding extra security; but to keep what does currently exists. Adding an extra check on all directories not being accessible from the web server is certainly a better way to go; but that was not the point of the current PR ;) 
Replacing the current system with a htaccess is not OK, just adding it would be.

What about:
1- add missing checks in php files when they're missing,
2- add the .htaccess file, apache users will have a better protection,
3- in another time (and another PR), add a feature to check if directories are not accessible from the web server (may target for 9.3).

btry:
Hi

I think this is not a good idea to remove the security check in inc/*.class.php. No matter .htaccess protects it.

trasher:
So; this one is not OK; closing

PR #3418, opened 2018-01-10
Prevent rules to be statically stored from unit tests

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes


trasher:
This one was to get fusion plugin tests fixed; but it seems no longer required

PR #3434, opened 2018-01-15
Network storage

trasher:
Add hard drive devices to network equipments (for NAS)
Migrate computerdisks table
Rename computerdisk to item_disk

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

### DIFF
--- a/front/item_disk.form.php
+++ b/front/item_disk.form.php
@@ -39,20 +39,23 @@ Session::checkCentralAccess();
 if (!isset($_GET["id"])) {
    $_GET["id"] = "";
 }
-if (!isset($_GET["computers_id"])) {
-   $_GET["computers_id"] = "";
+if (!isset($_GET["items_id"])) {
+   $_GET["items_id"] = "";
+}
+if (!isset($_GET["itemtype"])) {
+   $_GET['itemtype'] = '';
 }
 
-$disk = new ComputerDisk();
+$disk = new Item_Disk();
 if (isset($_POST["add"])) {
    $disk->check(-1, CREATE, $_POST);
 
    if ($newID = $disk->add($_POST)) {
-      Event::log($_POST['computers_id'], "computers", 4, "inventory",
+      Event::log($_POST['items_id'], $_POST['itemtype'], 4, "inventory",
                  //TRANS: %s is the user login
                  sprintf(__('%s adds a volume'), $_SESSION["glpiname"]));
       if ($_SESSION['glpibackcreated']) {
-         Html::redirect($disk->getFormURL());
+         Html::redirect($disk->getLinkURL());
       }
    }
    Html::back();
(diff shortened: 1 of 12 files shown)


inc/api.class.php
```diff
@@ -560,16 +560,17 @@ protected function getItem($itemtype, $id, $params = []) {
           && $itemtype == "Computer") {
```
orthagh: Here, we filter on only computers

inc/api.class.php
```diff
@@ -560,16 +560,17 @@ protected function getItem($itemtype, $id, $params = []) {
...
          $query = "SELECT `glpi_filesystems`.`name` AS fsname,
-                          `glpi_computerdisks`.*
-                   FROM `glpi_computerdisks`
+                          `glpi_items_disks`.*
+                   FROM `glpi_items_disks`
                    LEFT JOIN `glpi_filesystems`
-                             ON (`glpi_computerdisks`.`filesystems_id` = `glpi_filesystems`.`id`)
-                   WHERE `computers_id` = '$id'
+                             ON (`glpi_items_disks`.`filesystems_id` = `glpi_filesystems`.`id`)
+                   WHERE `itemss_id` = '$id'
```
orthagh: items**s**_id

inc/item_disk.class.php
```diff
@@ -204,6 +225,7 @@ function showForm($ID, $options = []) {
       Html::autocompletionTextField($this, "freesize");
       echo "&nbsp;".__('Mio')."</td></tr>";
 
+      //FIXME
```
orthagh: will it be fixed ?

inc/lock.class.php
```diff
@@ -112,7 +112,29 @@ static function showForItem(CommonDBTM $item) {
 
          }
 
-         $types = ['ComputerDisk', 'ComputerVirtualMachine'];
```
orthagh: why no just replace `ComputerDisk` by `itemDisk` ?

trasher: bcause `$params` are not the same

inc/define.php
```diff
@@ -311,7 +311,7 @@
 
 $CFG_GLPI["itemdevicenetworkcard_types"]  = ['Computer', 'NetworkEquipment', 'Peripheral', 'Phone', 'Printer'];
 
-$CFG_GLPI['itemdeviceharddrive_types']    = ['Computer', 'Peripheral'];
+$CFG_GLPI['itemdeviceharddrive_types']    = ['Computer', 'Peripheral', 'NetworkEquipment'];
```
wawax: Please add also Printer

front/item_disk.form.php
```diff
@@ -39,16 +39,19 @@
...
-$disk = new ComputerDisk();
+$disk = new Item_Disk();
 if (isset($_POST["add"])) {
    $disk->check(-1, CREATE, $_POST);
 
    if ($newID = $disk->add($_POST)) {
-      Event::log($_POST['computers_id'], "computers", 4, "inventory",
+      Event::log($_POST['items_id'], $_POST['itemtype'], 4, "inventory",
                  //TRANS: %s is the user login
                  sprintf(__('%s adds a volume'), $_SESSION["glpiname"]));
```
orthagh: actually not in the pr purpose, but a ``getLinkURL`` instead of ``getFormURL`` is the attended behavior

dsever:
Hi, 

Are we here only talk about NAS, or this class si much more generic, so it can be FC storage?

Dubravko

trasher:
Do not know what "FC storage" is, but basically a NAS device is just a NetworkEquipment asset (as it exists from a while) with harddrive(s) attached.

dsever:
Basicly it is the same, but instead of ethernet it uses Fibre Channel interfaces. As far I can see that is ok, because for Network devices there is an option to chose type of interface. But to cover storage devices as network device, there controllers are still missing, and other components like array type, luns and client mappings.

trasher:
If you need enhancements, you may use the [suggestion website](http://suggest.glpi-project.org). We did not have any negative feedback on that part, it won't change for 9.3 (since RC1 has been released, features are freezed).
Also, the whole DCIM is a first version, there are lots of things than can be "missing".

PR #3452, opened 2018-01-20
Fix #3358: the last global criterion can be deleted

### LINKED ISSUE #3358: Glpi 9.2 - global search criterion fails
It looks like #371 is back with most recent 9.2 version.

When adding a global search criterion (eg. a software characteristics) in the global computers search it cannot be removed anymore ...

[edit]: note that only the last criterion cannot be removed ... if more than one criterion is used, only the last one remains always present

mohierf:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #3358

### DIFF
--- a/inc/search.class.php
+++ b/inc/search.class.php
@@ -5122,7 +5122,8 @@ class Search {
       $default_values["criteria"]    = [0 => ['field' => $default_criteria,
                                                         'link'  => 'contains',
                                                         'value' => '']];
-      $default_values["metacriteria"]    = [];
+      // Do not set a default value for metacriteria else
+      // $default_values["metacriteria"]    = [];
 
       // Reorg search array
       // start


trasher:
Check unit tests, they're broken.

I'm a bit surprised the proposed patch fixes the issue... I wonder if this is the right way to get that fixed.

mohierf:
@trasher I am far from being an expert with the Glpi source code 😉  But I do confirm that the problem do not happen any more with this patch... 

The problem happens because the `$default_values["metacriteria"]` exists... I will check if there is another solution to cope with this ... any idea on your side ?

trasher:
I dot not have any idea right now; I'm not familiar enough with the search engine :(

orthagh:
better fix:

```diff
diff --git a/inc/search.class.php b/inc/search.class.php
index 991554ed89..991a8b5d99 100644
--- a/inc/search.class.php
+++ b/inc/search.class.php
@@ -5236,7 +5236,7 @@ class Search {
       foreach ($default_values as $key => $val) {
          if (!isset($params[$key])) {
             if ($usesession
-                && ($key == 'is_deleted' || !isset($saved_params[$key])) // retrieve session only if not a new request
+                && ($key == 'is_deleted' || !isset($saved_params['criteria'])) // retrieve session only if not a new request
                 && isset($_SESSION['glpisearch'][$itemtype][$key])) {
                $params[$key] = $_SESSION['glpisearch'][$itemtype][$key];
             } else {
```

mohierf:
Thanks @orthagh 👍

orthagh:
Replaced by #3472

PR #3488, opened 2018-01-31
add plugin rack migration script

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2926 

Here is the optional tasks:
- [ ] amps to power in models
- [ ] some missing fields (flow_rate, dissipation, amps)
- [x] other racked items

### DIFF
--- a/scripts/migrations/racks_plugin.php
+++ b/scripts/migrations/racks_plugin.php
@@ -1,0 +1,481 @@
+<?php
+/**
+ * ---------------------------------------------------------------------
+ * GLPI - Gestionnaire Libre de Parc Informatique
+ * Copyright (C) 2015-2018 Teclib' and contributors.
+ *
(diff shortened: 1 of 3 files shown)


scripts/migrations/racks_plugin.php
```diff
@@ -0,0 +1,461 @@
...
+      $exist = $DB->tableExists($table);
+      $error = checkResult($exist);
+   }
+
+   checkError($error);
+
+   return $error;
+}
+
+function TruncateCoreTables() {
```
trasher: Is this really required? I guess adding the possibility as an argument would be better.

orthagh:
import of other items done.
Ready for another review.

Issue #3522, opened 2018-02-08
Headers in lowercase


Gianfranco97:
* Steps to reproduce (which actions have you made) :
I have a problem trying to log in with a token. It is a bit strange, at first I thought the problem was mine but after brooding about it, I can not see what is wrong. I am trying to login with a code similar to the following:

```js
let myHeaders = new Headers()
myHeaders.append('Content-Type', 'application/json')
myHeaders.append('Authorization', 'user_token mG7y6hcMgwESTh2Z1KLRJL267bJxXsqoyXeWPbGT')

let options = {
method: 'GET',
headers: myHeaders
}
fetch ('https://dev.flyve.org/glpi/apirest.php/initSession', options)
.
.
.
```

**Request Headers:**

![captura de pantalla 2018-02-08 a la s 11 55 09 a m](https://user-images.githubusercontent.com/19965590/35982804-0c4f0154-0cc7-11e8-949e-31dfc8e2c003.png)

```
GET /glpi/apirest.php/initSession HTTP/1.1
Host: dev.flyve.org
Connection: keep-alive
Pragma: no-cache
Cache-Control: no-cache
authorization: user_token mG7y6hcMgwESTh2Z1KLRJL267bJxXsqoyXeWPbGT
Origin: http://localhost:3000
User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_12_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/63.0.3239.132 Safari/537.36
content-type: application/json
Accept: */*
Referer: http://localhost:3000/app
Accept-Encoding: gzip, deflate, br
Accept-Language: en,es-419;q=0.9,es;q=0.8
DNT: 1
```

**Response Headers:**

![captura de pantalla 2018-02-08 a la s 11 55 18 a m](https://user-images.githubusercontent.com/19965590/35982821-16cb7838-0cc7-11e8-90d6-0561eb60d3dc.png)

```
HTTP/1.1 400 Bad Request
Server: nginx
Date: Thu, 08 Feb 2018 15:53:15 GMT
Content-Type: application/json; charset=UTF-8
Content-Length: 201
Connection: keep-alive
Expires: Mon, 26 Jul 1997 05:00:00 GMT
Cache-Control: no-store, no-cache, must-revalidate
Pragma: no-cache
Access-Control-Allow-Origin: *
Access-Control-Expose-Headers: content-type, content-range, accept-range
```

* Expected result :

A valid session token

[...]

PR #3523, opened 2018-02-08
Handle case-insensitive headers fields 

DIOHz0r:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #3522

### DIFF
--- a/inc/apirest.class.php
+++ b/inc/apirest.class.php
@@ -446,6 +446,11 @@ class APIRest extends API {
       if (function_exists('getallheaders')) {
          //apache specific
          $headers = getallheaders();
+         if (false !== $headers) {
+            foreach ($headers as $key => $value) {
+               $headers[ucwords(strtolower($key), '-')] = $value;
+            }
+         }
       } else {
          // other servers
          foreach ($_SERVER as $server_key => $server_value) {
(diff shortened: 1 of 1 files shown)


inc/apirest.class.php
```diff
@@ -446,6 +446,11 @@ public function parseIncomingParams($is_inline_doc = false) {
       if (function_exists('getallheaders')) {
          //apache specific
          $headers = getallheaders();
+         if (false !== $headers) {
```
orthagh: `count($headers)` could be a better test

inc/apirest.class.php
```diff
@@ -446,6 +446,11 @@ public function parseIncomingParams($is_inline_doc = false) {
...
+            foreach ($headers as $key => $value) {
+               $headers[ucwords(strtolower($key), '-')] = $value;
```
orthagh: not mandatory, but before adding the new one, i think removing old key could avoid strange behavior, aka:
```unset($headers[$key]);```

It will avoid to have some "duplicates" keys like 
```
[
   'authorization': stuff,
   'Authorization': stuff
]
```

inc/apirest.class.php
DIOHz0r: I added both, the function returns false on some cases.

orthagh:
Ok for me

PR #3542, opened 2018-02-12
Automatic inventory infos

wawax:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #number

Move minimum inventory infos into the glpi to access these data even if inventory plugin is disabled. This will allow us to change an asset status based on it's last inventory date in the future.

inventory_infos


install/update_92_93.php
```diff
@@ -47,6 +47,37 @@ function update92to93() {
...
+   }
+
+   //Add automatic inventory infos for each itemtype
+   $fields = [
+      'date_last_inventory' => 'datetime', //Last inventory
+      'date_last_seen'      => 'datetime', //Last time the asset has been seen (either local or remote inventory)
+      'inventorytypes_id'   => 'integer', //Type of inventory (fusion, ocs, etc)
+      'inventory_source'    => 'string' //Which device has discovered the asset
+   ];
+   foreach ($CFG_GLPI["asset_types"] as $itemtype) {
```
trasher: Since asset_type may evolve in the future, it would be better to list current asset types ;)

install/update_92_93.php
```diff
@@ -294,51 +294,64 @@ function update92to93() {
...
-                  KEY `manufacturers_id` (`manufacturers_id`),
-                  KEY `racktypes_id` (`racktypes_id`),
-                  KEY `states_id` (`states_id`),
-                  KEY `users_id_tech` (`users_id_tech`),
-                  KEY `group_id_tech` (`groups_id_tech`),
-                  KEY `is_template` (`is_template`),
-                  KEY `is_deleted` (`is_deleted`),
-                  KEY `dcrooms_id` (`dcrooms_id`)
-                  ) ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_unicode_ci";
+                 `id` int(11) NOT NULL AUTO_INCREMENT,
```
trasher: For rack, enclosures and pdu tables, we have to take care of migration from an existing master

trasher:
@wawax this one seems OK for me in the core; but this will need FI and OCS plugins to be updated as far as I've understand, right?
I wonder if you've began to work ono FI port :)

trasher:
- [ ] Add new fields in forms

Issue #3549, opened 2018-02-14
[9.2.2] Crontask savedsearchesalerts doesn't send any alert to the user


Badoche1337:
Hi,

The crontask savedsearchesalerts doesn't send ajax notifications based on saved searches. After debugging it seems that no target are added to the notification.

Problem identified :

In class NotificationTargetSavedsearch_Alert, method getSpecificTargets seems to build an array with user information and pass it to addToRecipientsList method.

But : 
- the array built in  getSpecificTargets doesn't have "users_id" index.
- addToRecipientsList check if "users_id" is present and not null before adding a new recipient, so no target is added to the notification.

If I add a "users_id" index in getSpecificTargets with $user->getID() as value I can get the notification works

------------
* Steps to reproduce (which actions have you made) :

- Create a new saved search on ticket list (no criteria)
- Create a new alert on this saved search with comparaison operator ">=" and value = 0 (root entity and recursive)
- Enable notifications by mail, by ajax, etc
- Enable "Saved searches" notification and add a ajax template to it
- Create a ticket to get saved search count to 1
- Go to crontask list and execute it manually

* Expected result :

Receiving a browser notification which inform me that the saved search alert is triggered

* Actual result :

Nothing happened

* URL of the page :

N/A

* Screenshot of the problem (if pertinent) :

N/A

PR #3554, opened 2018-02-14
Savedsearch ajax notification doesn't work

Badoche1337:
Hello,

The crontask in charge of sending ajax notification to users if their saved searches trigger the goal is not working, a "users_id" index is missing so nobody is added as a target of the notification.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #3549


trasher:
Thank you :)

Issue #3254, opened 2017-12-06
Empty install stay 'work in progress"


trasher:
When upgrading to a new release without DB changes, setup stays on "Work in progress" state.

This is just a display issue; upgrade is OK, and "Use GLPI" button is displayed anyways.

PR #3555, opened 2018-02-14
fix messages div order in gui migration output

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #3254

### DIFF
--- a/inc/migration.class.php
+++ b/inc/migration.class.php
@@ -73,7 +73,7 @@ class Migration {
    function __construct($ver) {
 
       $this->deb = time();
-      $this->setVersion($ver);
+      $this->version = $ver;
    }
 
    /**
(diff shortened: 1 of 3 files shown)


inc/migration.class.php
```diff
@@ -73,7 +73,7 @@ class Migration {
    function __construct($ver) {
 
       $this->deb = time();
-      $this->setVersion($ver);
+      $this->version = $ver;
```
orthagh: For information, this change prevents the call of `addNewMessageArea` triggered in `setVersion` method (for construct case).
`setVersion` is called later in `update_xx_to_yy.php` files

Without that, we have this output:
```html
<div id="migration_message_9.2.2"><p class="center">
   Work in progress | Task Completed
</p></div>
<h3>Update to x.x.x</h3>
```

With the change, the `h3` and `div#migration` tags are in correct order.

tests/units/Migration.php
```diff
@@ -69,14 +69,6 @@ function () {
       }
    }
 
-   public function testConstructor() {
```
trasher: Maybe we can test output is empty; since this is what is expected? (same for the update test)

PR #3579, opened 2018-02-20
drop unneeded patch (merged upstream)

remicollet:
Cleanup.


trasher:
Should target 9.2; not master. Will rebase.

PR #3585, opened 2018-02-20
[wip] Manage accounts on sources, so can have same 'login' for different so…

ddurieux:
…urces (local, LDAPs, MAILs)
+ update travis to use the virtual machine instead the container


orthagh:
some thoughts on this pr after a speedy test:

- seems working :p (restapi also)
- as @trasher said, we could take the opportunity to get rid of images on login form
- on login form, we now have a dropdown for [local, ldap1, ldap2] where ldap name is set, i think it could be better if ldap servers have a field to set what we displays on this page (login_page_label).
Actually on my instance, it displays the fqdn of my ldap server  in the dropdown, not very user friendly and disclosure of internal information.

orthagh:
Oh, one more thing: don't display dropdown on login form if only local option is available

flegastelois:
- I already have username "toto" in DB, authenticated from "LDAP 1".
- Want to add username "toto" from "LDAP 2".

Administration > Users > [add user] ...  From an external source > Login=toto > Import from directories > Error "Unable to add. The user already exist."

image

- it would be necessary to add the LDAP directory dropdown on this form

PR #3587, opened 2018-02-20
Change travis method trying to kill timeouts

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: php
+sudo: required
 
 env:
   - DB=mysql


trasher:
As far as I know... There is no cons. We've switched to "no sudo" a while ago, because it should be faster than the other (but comparing build times with or without sudo; result is the same).

ddurieux:
It means we passe from container (4Go RAM and 2 CPU) to a virtual machine (7.5 Go RAM and 2 cpu bursted). 

Seems more stable...

PR #3590, opened 2018-02-21
league/climate

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

A test to implement league/climate for cli scripts.
Currently only rack migrations, but if you like, we could migrate other scripts


trasher:
I made a try with league on another cli command; it seems to be quite hard to get something generic and extensible for our needs. :-1: for me.

Issue #2995, opened 2017-10-18
Rights ticket to display button


yllen:
http://forum.glpi-project.org/viewtopic.php?id=159364

in 9.1.x display button of form was based on self::canxxxx
in 9.2 display is based on new parameter ($canupdate, $canpriority...).

Example:
in 9.1 you could update ticket with right, update, assign, steal own
in 9.2 you must have right to update or change priority to update the ticket

AircraftRu:
Hello,
If user has such rights:
![glpi - profiles](https://user-images.githubusercontent.com/17082164/31945011-9b8347c6-b8d6-11e7-8ca5-f6cfae50a407.png)
he can still update ticket.

User can't assigne specialists with that rights in GLPI 9.1.4.

AircraftRu:
Thank you! I'm looking forward to this solve!

AircraftRu:
Hello.
It's very big issue. Reise priority, please.

PR #3615, opened 2018-02-26
display save button in ticket form when right to assign/steal

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2995

### DIFF
--- a/tests/units/Ticket.php
+++ b/tests/units/Ticket.php
@@ -106,9 +106,8 @@ class Ticket extends DbTestCase {
    public function testCreateTicketWithActors($ticketActors) {
       $ticket = new \Ticket();
       $this->integer((int)$ticket->add([
-            'name'         => 'ticket title',
-            'description'  => 'a description',
-            'content'      => ''
+            'name'    => 'ticket title',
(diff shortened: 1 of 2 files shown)


tests/units/Ticket.php
```diff
@@ -990,21 +984,22 @@ public function testFormPostOnly() {
...
       $ticket = new \Ticket();
       $this->integer(
          (int)$ticket->add([
-            'name'         => '',
-            'description'  => 'A ticket to check displayed tech form',
-            'content'      => ''
+            'name'                => '',
+            'content'             => 'A ticket to check displayed tech form',
+            '_users_id_requester' => '3', // post-only
+            '_users_id_assign'    => '4', // tech
```
trasher: Better to get user id from name rather than use an hardcoded id (my change in the future)

trasher: Hum... Nevermind, ids are already used in other places in this test; let as it is OK.

PR #3622, opened 2018-02-26
fix auto-submitted rule (non empty string matched)

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #3597

With modified regex in #3597, rule matches on empty string (or non set `Auto-Reply` header).
So all collected mails would have been refused (without reason provided by ui furthermore)

Note: i moved the `i` modifier at the end, but the real change is `.*` to `.+`


orthagh:
yes you're right

PR #3624, opened 2018-02-27
Change variable declaration; fixes #3544

### LINKED ISSUE #3544: Bug on AuthLdap class
hello

i found a bug on the authldap class.
In getAllUsers function, a variable is not set correctly:

https://github.com/glpi-project/glpi/blob/9.2/bugfixes/inc/authldap.class.php#L1767-L1768

```
$userfound=false;
if (!empty($ldap_users[$user[$field_for_db]])
        || ($userfound = self::dnExistsInLdap($user_infos, $user['user_dn']))) {
   ....

```
the "userfound" variable can be not set if the first test is true. The second test is not checked, so the userfound variable stay at 'false'.

i suggest to set variable during declaration:

```

$userfound=self::dnExistsInLdap($user_infos, $user['user_dn']);
if (!empty($ldap_users[$user[$field_for_db]]) {
   ....

```

best regards
cristo

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #3544

### DIFF
--- a/inc/authldap.class.php
+++ b/inc/authldap.class.php
@@ -1763,9 +1763,8 @@ class AuthLDAP extends CommonDBTM {
          } else {
             //Ldap synchronisation : look if the user exists in the directory
             //and compares the modifications dates (ldap and glpi db)
-            $userfound = false;
-            if (!empty($ldap_users[$user[$field_for_db]])
-                || ($userfound = self::dnExistsInLdap($user_infos, $user['user_dn']))) {
+            $userfound = self::dnExistsInLdap($user_infos, $user['user_dn']);
+            if (!empty($ldap_users[$user[$field_for_db]]) || $userfound) {
                // userfound seems that user dn is present in GLPI DB but do not correspond to an GLPI user
                // -> renaming case
                if ($userfound) {


inc/authldap.class.php
```diff
@@ -1763,9 +1763,8 @@ static function getAllUsers(array $options, &$results, &$limitexceeded) {
          } else {
             //Ldap synchronisation : look if the user exists in the directory
             //and compares the modifications dates (ldap and glpi db)
-            $userfound = false;
-            if (!empty($ldap_users[$user[$field_for_db]])
-                || ($userfound = self::dnExistsInLdap($user_infos, $user['user_dn']))) {
+            $userfound = self::dnExistsInLdap($user_infos, $user['user_dn']);
+            if (!empty($ldap_users[$user[$field_for_db]])) {
```
orthagh: not approved finally, should we not have the part `|| $userfound` in the `if` block.
We have a `if` at 1770 for this var but also a `else` line 1784

trasher: Since `$userfound` is tested just below; it is not relevant in the first one.

trasher: Yes, you're right; fixed.

Issue #3507, opened 2018-02-05
[9.1.6 to 9.2.2] Regression : sorting preference lost in saved searches


nqb:
Hello,

After an upgrade from GLPI 9.1.6 to GLPI 9.2.2, I notice that bookmarks (now saved search) has lost there sorting preferences : 
- before upgrade : sorting on "Last update" column
- after upgrade : sorting on "Title" column

If I compare queries between 9.1.6 and 9.2.2, there are identical.

* Steps to reproduce (which actions have you made) :
  * Compare sorting preferences between bookmarks (9.1.6) and saved searches (9.2.2)

* Expected result : same sorting preference between bookmarks and saved searches

* Actual result : different sorting preference between bookmarks and saved searches

nqb:
@orthagh : patch applied.

I get back sorting preferences used in bookmarks on 9.1.6 in saved searches (9.2.2). You can merge.

PR #3635, opened 2018-02-28
default values for Search now merge Search-Itemtype-User defaults; se…

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #3507

### DIFF
--- a/inc/search.class.php
+++ b/inc/search.class.php
@@ -5172,17 +5172,28 @@ class Search {
       if (($itemtype != 'AllAssets')
           && class_exists($itemtype)) {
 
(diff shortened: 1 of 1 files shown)


orthagh:
i suggest for this pr a merge after 9.2.2.
reporter is aware of the fix and could apply itself.

But, my opinion the change seems to be a little too major to be merged just before release.
I'll prefer some days/weeks of tests before having public release of this change

Issue #3642, opened 2018-02-28
Search engine for knowbaseitem


yllen:
If i make a search in knowbaseitem with a string with empty character and the end the request is wrong
text to search: "yllen search "

 AND (`glpi_knowbaseitems`.`name` LIKE '%yllen search %' 
            OR `glpi_knowbaseitems`.`answer` LIKE '%yllen search %' )

So empty result because you can't have empty character front of % character.

Issue is still present in 9.2.x

PR #3643, opened 2018-02-28
fix search engine - #3642

yllen:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #3642

### DIFF
--- a/inc/search.class.php
+++ b/inc/search.class.php
@@ -6265,7 +6265,7 @@ class Search {
             $val = substr($val, $begin, $length-$end-$begin);
          }
 
-         $SEARCH = " $NOT LIKE '".(!$begin?"%":"").$val.(!$end?"%":"")."' ";
+         $SEARCH = " $NOT LIKE '".(!$begin?"%":"").trim($val).(!$end?"%":"")."' ";
       }
       return $SEARCH;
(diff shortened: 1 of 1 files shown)


yllen:
Test failed for PHP version, but 9.1 is still for PHP 5.6

trasher:
CircleCI s not setup for 9.1 branch. I'll try to disable it, you can just ignore it: only travis builds are required here (and they are OK ;).

inc/search.class.php
```diff
@@ -6265,7 +6265,7 @@ static function makeTextSearch($val, $not=false) {
             $val = substr($val, $begin, $length-$end-$begin);
          }
 
-         $SEARCH = " $NOT LIKE '".(!$begin?"%":"").$val.(!$end?"%":"")."' ";
+         $SEARCH = " $NOT LIKE '".(!$begin?"%":"").rtrim($val).(!$end?"%":"")."' ";
```
trasher: Why not just `trim()`? Problem would be the same with an extra space at the begginning of the searched string, isn't it?

PR #3646, opened 2018-02-28
Try to exclude 9.1/bf branch from circleci builds

trasher:
circleci is not setup on 9.1/branch; and there are still commits on this branch...


trasher:
Does not work. Will not spend more time for a dead branch

Issue #3641, opened 2018-02-28
[9.2.2] Bad french translations for "Public/Private Saved searches"


nqb:
Hello,

In 9.2.2, when you have several saved searches (private and public), french translation are : 
* Recherches sauvegardées publique
* Recherches sauvegardées privée

Should be (missing *s*) : 
* Recherches sauvegardées publique*s*
* Recherches sauvegardées privée*s*

trasher:
Translations are done on transifex ;)

nqb:
@trasher, 

there is already a translation for "Public Saved searches" : 

https://github.com/glpi-project/glpi/blob/39d4ad46a5e5f4ea326ea7b8bcaeccf59deb6b1f/locales/fr_FR.po#L15259

However, files mentioned seems no related to `front/savedsearch.php` or `ajax/savedsearch.php`

nqb:
Certainly : 
https://github.com/glpi-project/glpi/blob/39d4ad46a5e5f4ea326ea7b8bcaeccf59deb6b1f/inc/savedsearch.class.php#L764
https://github.com/glpi-project/glpi/blob/39d4ad46a5e5f4ea326ea7b8bcaeccf59deb6b1f/inc/savedsearch.class.php#L771

PR #3648, opened 2018-03-01
Plural for "private" and "public"; see #3641 comments

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

To be merge after 9.2.2 release; way too late to change translations (in particular adding plural forms).

### DIFF
--- a/inc/savedsearch.class.php
+++ b/inc/savedsearch.class.php
@@ -761,14 +761,20 @@ class SavedSearch extends CommonDBTM {
       $colspan = 2;
       echo "<table class='tab_cadre_fixehov'>";
       echo "<thead><tr><th colspan='$colspan' class='private_header'>" .
-                  sprintf(__('Private %1$s'), $this->getTypeName(count($searches['private']))) .
+                  sprintf(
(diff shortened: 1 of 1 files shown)


inc/savedsearch.class.php
```diff
@@ -761,14 +761,20 @@ function displayMine() {
       $colspan = 2;
       echo "<table class='tab_cadre_fixehov'>";
       echo "<thead><tr><th colspan='$colspan' class='private_header'>" .
-                  sprintf(__('Private %1$s'), $this->getTypeName(count($searches['private']))) .
+                  sprintf(
+                     _n('Private %1$s', 'Privates %1$s', count($searches['private'])),
```
remicollet: no 's' in english.... but _n can be used with twice the same string, as for `_n('Software', 'Software', 1);`

trasher: Indeed; thank you! :)

PR #3650, opened 2018-03-01
Improve checks on read files directory

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

Two things:
1- display a warning when checks cannot be done. User is aware of potential issue; it's up to him.
2- try to guess host (I almost never user localhost myself; and I guess most of the time called URI will not be localhost only).

### DIFF
--- a/inc/config.class.php
+++ b/inc/config.class.php
@@ -2512,13 +2512,19 @@ class Config extends CommonDBTM {
       ]);
 
       /* TODO: could be improved, only default vhost checked */
-      if ($fic = fopen('http://localhost'.$CFG_GLPI['root_doc'].'/index.php?skipCheckWriteAccessToDirs=1', 'r', false, $context)) {
+      $protocol = 'http';
+      if (isset($_SERVER['HTTPS'])) {
+         $protocol = 'https';
+      }
+      $uri = $protocol . '://' . $_SERVER['SERVER_NAME'] . $CFG_GLPI['root_doc'];
+
+      if ($fic = fopen($uri.'/index.php?skipCheckWriteAccessToDirs=1', 'r', false, $context)) {
          fclose($fic);
          if (!$fordebug) {
             echo "<tr class='tab_bg_1'><td class='b left'>".
                __('Web access to files directory is protected')."</td>";
          }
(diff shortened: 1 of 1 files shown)


inc/config.class.php
```diff
@@ -2512,13 +2512,19 @@ static function checkWriteAccessToDirs($fordebug = false) {
       ]);
 
       /* TODO: could be improved, only default vhost checked */
-      if ($fic = fopen('http://localhost'.$CFG_GLPI['root_doc'].'/index.php?skipCheckWriteAccessToDirs=1', 'r', false, $context)) {
+      $protocol = 'http';
+      if (isset($_SERVER['HTTPS'])) {
+         $protocol = 'https';
+      }
+      $uri = $protocol . '://' . $_SERVER['SERVER_NAME'] . $CFG_GLPI['root_doc'];
```
remicollet: indeed using server_name may be required for vhost/ssl, but mays be different from the client side (where is it retrieved) and from the server side.... btw no perfect solution... will need a JS check from the client

inc/config.class.php
```diff
@@ -2541,6 +2547,10 @@ static function checkWriteAccessToDirs($fordebug = false) {
                      __s('Web access to files directory is protected')."\"></td></tr>";
             }
          }
+      } else {
+            echo "<td><img src='".$CFG_GLPI['root_doc']."/pics/warning_min.png'>".
+                  "<p class='red'>".__('Web access to the files directory, should not be allowed')."<br/>".
+                  __('Automatic checks cannot be done; please review .htaccess file and the web server configuration.')."</p></td></tr>";
       }
```
remicollet: IIRC, there is tons of false positive... btw, ok to raise admin attention

inc/config.class.php
trasher: Yes, this is not perfect; but maybe a bit better than just test localhost without https ;)

inc/config.class.php
trasher: I totaly agree... But this will permit to get rid of "files directory is insecure". We all agree that the way to go is to store files outside the webroot ;)

PR #3667, opened 2018-03-05
WIP fix default values for ticket

btry:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | none

Issue revealed while working on Formcreator, and its unit tests.

The plugin tries to get the default values of a ticket from Ticket::getDefaultValues() and populate / replace data to do its job. It seems the default values of a ticket don't match the format expected in Ticket::prepareInputForAdd().


btry:
I may create a unit test which loads the defaults, sets title and description fields then call $ticket->add() and see if the ticket created successfully. But will the tests check php-errors.log ?

trasher:
Why should .log be tested?
Not handled errors, such as notices should break tests.

btry:
This PR fixes an inconsistency, and I did not found an actual bug related to this. This appears only with my current work on Formcreator, and it only generates PHP notices (and my unit tests checks for them)

tests/units/Ticket.php
```diff
@@ -0,0 +1,74 @@
...
+      $this->integer($input['_suppliers_id_assign'])->isEqualTo(0);
+      $this->array($input['_suppliers_id_assign_notif']['use_notification'])->contains('1');
+      $this->array($input['_suppliers_id_assign_notif']['alternative_email'])->contains('');
+
+      $this->string($input['name'])->isEqualTo('');
+      $this->string($input['content'])->isEqualTo('');
+      $this->integer((int) $input['itilcategories_id'])->isEqualTo(0);
+      $this->integer((int) $input['urgency'])->isEqualTo(3);
+      $this->integer((int) $input['impact'])->isEqualTo(3);
+      $this->integer((int) $input['priority'])->isEqualTo(3);
```
trasher: Something is wrong here

btry: What's wrong ?

Note that this is unfinished. I'l l let you know when the test are ready for review.

trasher: This is wrong:
```
+      $this->integer((int) $input['priority'])->isEqualTo(3);
+      $this->integer((int) $input['priority'])->isEqualTo(3);
+      $this->integer((int) $input['priority'])->isEqualTo(1); 
```
You did not say that was not finished.

Issue #3688, opened 2018-03-07
Plugin missing on session change


DIOHz0r:
* Steps to reproduce (which actions have you made) :

1. Install and activate any plugin
2. Make the API where a user has a different session
3. Check the plugins available on session 

* Expected result :

The plugins available on the first login must be available on session change

* Actual result :

The plugins are lost

* Screenshot of the problem (if pertinent) :

Before session check....
![imagen](https://user-images.githubusercontent.com/1426313/37119047-967569ac-222c-11e8-9acc-ce2b3d3aedd4.png)

After session change...
![imagen](https://user-images.githubusercontent.com/1426313/37119094-c4b6d044-222c-11e8-9d2e-1019310c2907.png)

* Your GLPI setup (you can find it in Setup > General menu, System tab) :
GLPi 9.2.2

fgendorf:
Same problem here

edgardmessias:
For REST: Add in `apirest.php`, before `include ('./inc/includes.php');`:

```php
if(isset($_GET['session_token'])){
    session_id($_GET['session_token']);
}
```

edgardmessias:
Solution for REST and XMLRPC: #3692

DIOHz0r:
Hello @edgardmessias ,

I did another PR (#3697) with a simple a better fix for this issue IMO with. Let's wait until @trasher review both PR and check which one is better.

edgardmessias:
My solution initialize only after detect the session id.
Your solution, the plugins are initialize twice times and on each request, a new session is created and destroyed.

DIOHz0r:
@edgardmessias, Why do you say that the plugins initialize twice and on each request?, If the plugin array exist on session they won't be initialized again, also I'm leaving all the logic about security like cross, sanitize parameters, check if the API is enable, etc. that you removed

edgardmessias:
@DIOHz0r, to you understand, the real problem is the double session.
On API, a temporary session is created, call all `plugin_init_<plugin>`, destroy and load the session from token.

In `plugin_init_<plugin>`, user is not logged, and if some plugin use `$_SESSION`, the user is none

edgardmessias:
@DIOHz0r, about security check, see https://github.com/edgardmessias/glpi/blob/api_session/inc/api.class.php#L111

Note: Current code return a HTML with message `api disable`, my solution return JSON for apirest and XML for apixmlrpc

edgardmessias:
The #3692 close this issue

PR #3692, opened 2018-03-08
Fixed session init for APIs

edgardmessias:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #3688

With this, the session is initialized only once
First check the session token, for later, initialize the session
For get the session, need parse the content, to run later.

### DIFF
--- a/inc/api.class.php
+++ b/inc/api.class.php
@@ -43,6 +43,7 @@ abstract class API extends CommonGLPI {
    protected $session_write = false;
 
    static $api_url = "";
+   static $content_type = "application/json";
    protected $format;
    protected $iptxt         = "";
    protected $ipnum         = "";
@@ -75,6 +76,12 @@ abstract class API extends CommonGLPI {
     */
    abstract protected function returnResponse($response, $code, $additionalheaders);
 
+   /**
+    * Upload and validate files from request and append to $this->parameters['input']
+    *
+    * @return void
+    */
+   abstract protected function manageUploadedFiles();
 
    /**
     * Constructor
@@ -84,8 +91,17 @@ abstract class API extends CommonGLPI {
     *
     * @return void
     */
-   public function __construct() {
+   public function initApi() {
       global $CFG_GLPI, $DB;
+
+      // Load GLPI configuration
+      include_once (GLPI_ROOT . '/inc/includes.php');
+      $variables = get_defined_vars();
+      foreach ($variables as $var => $value) {
+         if ($var === strtoupper($var)) {
+            $GLOBALS[$var] = $value;
+         }
+      }
 
       // construct api url
       self::$api_url = trim($CFG_GLPI['url_base_api'], "/");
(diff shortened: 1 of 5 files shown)


trasher:
API unit tests are failing

edgardmessias:
I will fix it

edgardmessias:
@trasher, I fixed, but not understand why not work on PHP 7.2

trasher:
Hum... There is a warning; maybe is this related. I'll take a look

edgardmessias:
The problem: `PHP Warning:  session_save_path(): Cannot change save path when session is active`

trasher:
Yes, there are lots of output in the local webserver; this has to be fixed... But on my computer, tests are OK even with those issues.

edgardmessias:
Fixed, wait result

inc/api.class.php
```diff
@@ -257,30 +257,11 @@ protected function retrieveSession() {
          $current = session_id();
          $session = trim($this->parameters['session_token']);
 
-         if (file_exists(GLPI_ROOT . '/inc/downstream.php')) {
```
trasher: You can't drop all of this. See http://glpi-install.readthedocs.io/en/latest/install/index.html#files-and-directories-locations

tests/DbTestCase.php
```diff
@@ -84,7 +84,7 @@ public function afterTestMethod($method) {
       // Cleanup log directory
       foreach (glob(GLPI_LOG_DIR . '/*.log') as $file) {
          if (file_exists($file)) {
-            unlink($file);
+            @unlink($file);
```
trasher: I do not agree; and this should really not be required.

inc/api.class.php
edgardmessias: I will revert

edgardmessias:
@trasher, fixed and tests passing

DIOHz0r:
Hi @trasher,

I also did another PR #3697 related with the same issue, Could you check both PR and decide which one will be merged?.

Thank tou.

inc/apirest.class.php
```diff
@@ -55,6 +55,20 @@ public static function getTypeName($nb = 0) {
       return __('Rest API');
    }
 
+   public function prepare() {
+      parent::prepare();
+
+      foreach ($_FILES as $filename => $files) {
```
orthagh: Is this move of fileupload var parsing is needed ?

edgardmessias: Yes, because the`GLPIUploadHandler::uploadFiles` call document validation, and only validade after load config

orthagh: ok, just a suggestion so, maybe we could move this code to a separate function `manageUploadedFiles` (for example) and call after the initialization.

inc/api.class.php
```diff
@@ -84,9 +85,18 @@
     *
     * @return void
     */
-   public function __construct() {
+   public function prepare() {
```
orthagh: pretty subjective, i dont like `prepare` term for the behavior.
I suggest `initApi`

edgardmessias: I will make this.

edgardmessias:
@orthagh, done.

inc/apirest.class.php
```diff
@@ -55,6 +55,23 @@ public static function getTypeName($nb = 0) {
       return __('Rest API');
    }
 
+   /**
+    * Upload and validate files from request and append to $this->parameters['input']
+    *
+    * @return void
+    */
+   public function manageUploadedFiles() {
```
orthagh: last thing (sorry)
make this function abstract in api.class, and do it also in xmlrpc (empty of course)

orthagh:
great, i'll do some additional tests but we'll could merge.

It could be done in another pr, but ideally, we could make a test for #3688 issue.
I can take care of that task.

edgardmessias:
@orthagh, done

PR #3694, opened 2018-03-09
Fixed e-mail images if contains "&tickets_id="

edgardmessias:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | no


trasher:
I guess we should add an unit test on that; your opinions?

edgardmessias:
I will work o unit test

Tachyonn:
Not work for me. 9.2.2. After making changes mailgate and queuednotification becomes "running" and never ends.

trasher:
There are parse errors; current PR cannot be tested right now.

trasher:
@edgardmessias I've fixed the parse error ;)

edgardmessias:
Thanks @trasher.

For unit tests, my idea is, create two static method, one for extract images docs from HTML, and another to replace URL in HTML.

Tachyonn:
With fixes made by @trasher now works... 9.2.2, php 5.6

trasher:
@edgardmessias yes for the tests; I was talking about something like that on another chan recently ;)

@Tachyonn great; thanks for the feedback!

trasher:
A  part of the split work has been done in #3698

PR #3695, opened 2018-03-09
Makes warning fatal on API tests

trasher:


trasher:
I do not know why this is failing on PHP 5.6... Anyways, I'm not going to spend much time on this.

PR #3697, opened 2018-03-09
Restart the plugins on API call

DIOHz0r:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #3688 

### DIFF
--- a/inc/api.class.php
+++ b/inc/api.class.php
@@ -271,6 +271,11 @@ abstract class API extends CommonGLPI {
             $_SESSION['glpi_use_mode'] = Session::NORMAL_MODE;
 
             Session::loadLanguage();
+
+            if (!isset($_SESSION["glpi_plugins"])) {
+               $plugin = new Plugin();
+               $plugin->init();
+            }
          }
       }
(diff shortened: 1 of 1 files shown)


inc/api.class.php
```diff
@@ -271,6 +271,11 @@ protected function retrieveSession() {
             $_SESSION['glpi_use_mode'] = Session::NORMAL_MODE;
 
             Session::loadLanguage();
```
edgardmessias: Change code to before `Session::loadLanguage();`, to work plugins locales.

DIOHz0r: I though the same but on the `inc/include.php` file the same `Session::loadLanguage();` is called before init the plugin, however your comment gave me the alert about there weren't tiggered the plugin's load & hook, so I'll fix that.

DIOHz0r:
As #3692 was merged and it has a different logic this PR has no effect so let's close it.

Issue #939, opened 2016-09-08
tools/generate_bigdump.php is outdated


remicollet:
It needed to be updated, as it doesn't work anymore.

trasher:
Fixed in 9.3/bugfixes

PR #3713, opened 2018-03-13
fix warnings in generate_bigdump

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #939

Just fixed the logged warnings, maybe we have some part non functional anymore (i think about computers operating systems for example)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -40,7 +40,8 @@
         "natxet/CssMin": "^3.0",
         "patchwork/jsqueeze": "^2.0",
         "atoum/atoum": "^3.3",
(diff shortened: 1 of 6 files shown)


orthagh:
oh also, maybe we can implements https://github.com/fzaninotto/Faker

trasher:
I had in mind to use Faker for unit tests bootstrap data as well. But I've experimented that in tests of another project, and I had issues :/ For bigdump; it should be ok anyways.

Issue #3709, opened 2018-03-13
Requester followup


allanlopesprado:
The requester no need received your own email, Is there a way to fix this?
![image](https://user-images.githubusercontent.com/17826867/37339432-6e47b716-2699-11e8-9f8a-583979f8471e.png)

edgardmessias:
Maybe this can souve your problem:

```diff
diff --git a/inc/notificationtargetcommonitilobject.class.php b/inc/notificationtargetcommonitilobject.class.php
index 3faa2b3cf..c5dfa414e 100644
--- a/inc/notificationtargetcommonitilobject.class.php
+++ b/inc/notificationtargetcommonitilobject.class.php
@@ -123,6 +123,12 @@ abstract class NotificationTargetCommonITILObject extends NotificationTarget {
             $author_lang  = $data["language"];
             $author_id    = $data['users_id'];

+            if (isset($this->options['followup_id'])
+                && isset($this->options['author_id'])
+                && $this->options['author_id'] == $author_id) {
+               continue;
+            }
+
             if (!empty($data['altemail'])
                 && ($data['altemail'] != $author_email)
                 && NotificationMailing::isUserAddressValid($data['altemail'])) {
diff --git a/inc/ticketfollowup.class.php b/inc/ticketfollowup.class.php
index fe7bf2553..e20bfa22d 100644
--- a/inc/ticketfollowup.class.php
+++ b/inc/ticketfollowup.class.php
@@ -248,6 +248,7 @@ class TicketFollowup  extends CommonDBTM {

       if ($donotif) {
          $options = ['followup_id' => $this->fields["id"],
+                          'author_id' => $this->fields['users_id'],
                            // Force is_private with data / not available
                           'is_private'  => $this->fields['is_private']];
          NotificationEvent::raiseEvent('delete_followup', $job, $options);
@@ -300,6 +301,7 @@ class TicketFollowup  extends CommonDBTM {
                //FIXME: _need_send_mail does not seems to be used

                $options = ['followup_id' => $this->fields["id"],
+                                'author_id'   => $this->fields['users_id'],
                                 'is_private'  => $this->fields['is_private']];

                NotificationEvent::raiseEvent("update_followup", $job, $options);
@@ -466,6 +468,7 @@ class TicketFollowup  extends CommonDBTM {

       if ($donotif) {
          $options = ['followup_id' => $this->fields["id"],
+                          'author_id'   => $this->fields['users_id'],
[...]

Issue #3662, opened 2018-03-02
Rework notifications


trasher:
Current notification system is buggy and very hard to maintain/understand.

What should be done:
- rework the way recipients are managed (especially remove the add/drop ugly stuff),
- send lighter notifications by default,
- use a templating system,
- ... ?

edgardmessias:
More control for the user by notification type (all, ignore my, when observe, ....)

allanlopesprado:
Notification of calls awaiting evaluation
Listbox - 1 week, 2 weeks, .... 1 month

Notification, something like

Hello, ##ticket.authors##
You have ##open.satisfaction## open satisfaction survey, to answer access ## ticket.urlsatisfaction ##
List: ##list.open.satisfaction## 
Thanks for answering

EC2311:
Hello
Depending of the status (of project, of ticket) differents users or groups defined in the ticket or project itself may received the notification : not all users or all groups may receive notifications.. It should be great :)

nbriche:
Hello,

Another suggestion: Overriding the reply-to address for a notification trigger.

For example, setting the ticket closing notification to a noreply would force the user to click a link to access the ticket, resulting in a lot less reopenings due to "thank you" replies.

Doing the same to solutions would also force the user to actually go and approve/reject the solution instead of just adding a followup.

trasher:
Ths ticket is about rewriting the complex notifications system; we plan to privde same features that existing, not to add anything new.
Please use [suggestion website](http://suggest.glpi-project.org) for RFE.

TarasSavchuk:
I believe the possible way to improve notification system is by integrating it in business rules for tickets.
Pros:
* we already have a lot of criteria here, i.e. we can match any ticket fields to decide if we need to send notifications (priority, etc) - very flexible notifications possible.
* this approach can be implemented without touching existing notification system.
We already have some notifications integrated there (sending approval requests). What would be great is to implement general action "send email with proper recipients (requestor, agent, agent group, etc)".

PR #3715, opened 2018-03-13
Requester no need received your own email

edgardmessias:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #3709


edgardmessias:
@orthagh, currently, there is an option to notify the follow-up author.

glpi

edgardmessias:
You should only send the notification only if you have configured it, not always

orthagh:
I close as requester agreed on existing feature covers it's need.
If you want, you may send a more global option (and permits to configure parts of it) in a new pr (need specs on this i think).

edgardmessias:
I will wait the #3662

PR #3717, opened 2018-03-14
Fixed next month for ticket recurrent for last day

edgardmessias:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | no

Old: `2018-01-31`, +1 MONTH, `2018-03-03`,  `2018-04-03`,  `2018-05-03`

New: `2018-01-31`, +1 MONTH, `2018-02-28`, `2018-03-31`, `2018-04-30`

### DIFF
--- a/inc/ticketrecurrent.class.php
+++ b/inc/ticketrecurrent.class.php
@@ -352,11 +352,16 @@ class TicketRecurrent extends CommonDropdown {
          $timestart  = strtotime($begin_date) - $create_before;
          $now        = time();
          if ($now > $timestart) {
-            $value = $periodicity;
-            $step  = "second";
+            $value  = $periodicity;
+            $step   = "second";
+            $prefix = "";
             if (preg_match('/([0-9]+)MONTH/', $periodicity, $matches)) {
                $value = $matches[1];
                $step  = 'MONTH';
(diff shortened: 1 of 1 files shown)


allanlopesprado:
I suggest the option is added (31 days)

image

cedric-anne:
This fix does not work if start date is `30th day of january` and interval is `1 month`. In this case computed date will be `1st day of march` or `2nd day of march` depending on day count in february.

It will also not work if start date is `29th day of february` and interval is `1 year`. In this case computed date will be `1st day of march`.

allanlopesprado:
I suggest again adding the 31st day in the calendar

edgardmessias:
@cedric-anne, the solution is only for last day of month

edgardmessias:
@allanlopesprado, open a issue with your request

trasher:
This fix is only for one particular case, while there are several other known issues. I've created a ticket (https://github.com/glpi-project/glpi/issues/4134), to get all fixed (and anyways, there is no guarantee this will not break something else).
I close this one.

PR #3724, opened 2018-03-15
WIP make getFromDBByCrit use one less SQL query

btry:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #3725 

CommonDBTM::getFromDbByQuery() deprecated without any alternative.

### DIFF
--- a/inc/commondbtm.class.php
+++ b/inc/commondbtm.class.php
@@ -272,27 +272,8 @@ class CommonDBTM extends CommonGLPI {
     * @return boolean|array
     */
    public function getFromDBByCrit(array $crit) {
-      global $DB;
-
-      $crit = ['SELECT' => 'id',
-               'FROM'   => $this->getTable(),
-               'WHERE'  => $crit];
-
-      $iter = $DB->request($crit);
-      if (count($iter) == 1) {
-         $row = $iter->next();
-         return $this->getFromDB($row['id']);
-      } else if (count($iter) > 1) {
-         Toolbox::logWarning(
-            sprintf(
-               'getFromDBByCrit expects to get one result, %1$s found!',
-               count($iter)
-            )
-         );
-      }
-      return false;
+      return $this->getFromDBByRequest(['WHERE' => $crit]);
    }
-
 
    /**
     * Retrieve an item from the database by request. The request is an array
--- a/inc/migration.class.php
+++ b/inc/migration.class.php
@@ -808,14 +808,14 @@ class Migration {
                                        AND `itemtype` = '$type'";
                      if ($result2 = $DB->query($query)) {
                         if ($DB->numrows($result2) == 0) {
-                              $DB->insert(
-                                 'glpi_displaypreferences', [
-                                    'itemtype'  => $type,
-                                    'num'       => $newval,
-                                    'rank'      => $rank++,
-                                    'users_id'  => $data['users_id']
-                                 ]
-                              );
+                           $DB->insert(
+                              'glpi_displaypreferences', [
+                                 'itemtype'  => $type,
+                                 'num'       => $newval,
+                                 'rank'      => $rank++,
+                                 'users_id'  => $data['users_id']
+                              ]
+                           );
                         }
                      }
                   }
@@ -824,14 +824,14 @@ class Migration {
             } else { // Add for default user
                $rank = 1;
                foreach ($tab as $newval) {
-                     $DB->insert(
-                        'glpi_displaypreferences', [
-                           'itemtype'  => $type,
-                           'num'       => $newval,
-                           'rank'      => $rank++,
-                           'users_id'  => 0
-                        ]
-                     );
+                  $DB->insert(
+                     'glpi_displaypreferences', [
+                        'itemtype'  => $type,
+                        'num'       => $newval,
+                        'rank'      => $rank++,
+                        'users_id'  => 0
+                     ]
+                  );
                }
             }
          }
--- a/tests/functionnal/CommonDBTM.php
+++ b/tests/functionnal/CommonDBTM.php
@@ -115,6 +115,31 @@ class CommonDBTM extends DbTestCase {
 
    }
 
+   public function testGetFromDBByCrit() {
+      $instance = new \Computer();
+      $instance->getFromDBByCrit(['name' => '_test_pc01']);
+      // the instance must be populated
+      $this->boolean($instance->isNewItem())->isFalse();
+
+      $instance = new \Computer();
+      $this->exception(
+         function() use ($instance) {
+            $instance->getFromDBByCrit(['entities_id' => '1']);
+         }
+      )->isInstanceOf(\RuntimeException::class)
+      ->message
+      ->contains('getFromDBByRequest expects to get one result, 3 found!');;
+      // the instance must not be populated
+      $this->boolean($instance->isNewItem())->isTrue();
+
+      $instance = new \Computer();
+      $success = $instance->getFromDBByCrit(['name' => 'non existent computer']);
+      // Must return false because no row returned by the SQL query
+      $this->boolean($success)->isFalse();
+      // the instance must not be populated
+      $this->string($instance->getField('name'))->isEqualTo(NOT_AVAILABLE);
+   }
+
    public function testGetFromDBByRequest() {
       $instance = new \Computer();
       $instance->getFromDbByRequest([
--- a/tests/units/Migration.php
+++ b/tests/units/Migration.php
@@ -124,9 +124,9 @@ class Migration extends \GLPITestCase {
 
       $this->array($this->queries)->isIdenticalTo([
          0 => 'SELECT * FROM `glpi_configs` WHERE `context` = \'core\' AND `name` IN (\'one\', \'two\')',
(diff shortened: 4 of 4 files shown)


inc/commondbtm.class.php
```diff
@@ -271,20 +271,43 @@ public static function getFromIter(DBmysqlIterator $iter) {
...
+    **/
+   public function getFromDBByRequest(array $request) {
       global $DB;
 
-      $crit = ['SELECT' => 'id',
-               'FROM'   => $this->getTable(),
-               'WHERE'  => $crit];
+      // Limit the request to the useful expressions
+      $request = array_intersect_key($request, [
+         'WHERE' => '',
```
trasher: I guess we should find a better way to get possible keys here; from iterator itself maybe; to avoid missing entries (order is already missing at least; but if we add new keys in the future, they'll be missing here :/ Any opinions?

tests/functionnal/CommonDBTM.php
```diff
@@ -114,4 +114,62 @@ public function testgetIndexNameOtherThanID() {
...
+      $instance->getFromDBByCrit(['name' => '_test_pc01']);
+      // the instance must be populated
+      $this->boolean($instance->isNewItem())->isFalse();
+
+      $instance = new \Computer();
+      $this->exception(
+         function() use ($instance) {
+            $instance->getFromDBByCrit(['entities_id' => '1']);
+         }
+      );
```
trasher: Maybe test exception class or message?

tests/functionnal/CommonDBTM.php
```diff
@@ -114,4 +114,62 @@ public function testgetIndexNameOtherThanID() {
...
+      $this->boolean($instance->isNewItem())->isFalse();
+
+      $instance = new \Computer();
+      $this->exception(
+         function() use ($instance) {
+            $instance->getFromDbByRequest([
+               'WHERE' => ['contact' => 'johndoe'],
+            ]);
+         }
+      );
```
trasher: same as the other exception

inc/commondbtm.class.php
```diff
@@ -271,20 +271,43 @@ public static function getFromIter(DBmysqlIterator $iter) {
...
+         'LEFT JOIN' => '',
+         'INNER_JOIN' => '',
+         'START' => '',
+         'LIMIT' => '',
+      ]);
+      $request['FROM'] = $this->getTable();
+      $request['SELECT'] = $this->getTable() . '.id';
```
trasher: This is maybe too restrictive (see my comment on `getFromDB` below.

Either drop the `SELECT` part  (or set it to `$this->getTable() . '*'`); or make it configurable as well?

inc/commondbtm.class.php
```diff
@@ -271,20 +271,43 @@ public static function getFromIter(DBmysqlIterator $iter) {
...
 
-      $iter = $DB->request($crit);
+      $iter = $DB->request($request);
       if (count($iter) == 1) {
          $row = $iter->next();
          return $this->getFromDB($row['id']);
```
trasher: Not directly related to the current pull request; but we should avoid calling `getFromDB` here, since it will make an extra call to the database. Using `getFromIter` sounds a better solution.

remicollet:
Don't even understand the need.... the $crit array may already contain 

      return $this->getFromDBByRequest(['WHERE' => $crit]); 

Above will break some usage of the functions (eaxctly the one you seems to want to handle)

OK, getFromDBByCrit name is probably confusing.

inc/commondbtm.class.php
btry: I considered that order would b e useless. However, reading your comment, this remindss me I did not wanted to allow START and LIMIT, and finally thought this may be useful. Then, ORDER may be useful too.

This array is actually a whitelist. If we do the reverse (a blacklist) we may solve your point.

tests/functionnal/CommonDBTM.php
btry: An exception is thrown only in the unit tests context, t hen UI considered useless to test its class or message.

I would more easily agree if getFromDBByCrit() or getFromDBByRequest() explicitly throw an exception. I did not t ook  the freedom to do it, and preferred to keep the original behaviour.

inc/commondbtm.class.php
btry: I get the point with the next comment.

As the method shall populate the instance ($instance->fields array) the select must be definitively ```table.*```

inc/commondbtm.class.php
btry: getFromIter itseld calls getFromDB, then this is not better than the current solution.

Also, getFromDB() calls post_getFromDB(), then I think we may hydrate the isntance with the full row, then call post_getFromDB().

btry:
@remicollet the need is to get an item with a SQL query more complex than a using only a WHERE clause.

I need to get an item with WHERE and JOINS. getFromDbByCrit() does not allows that as the argument **$crit** is intended to be only in the WHERE clause.

tests/functionnal/CommonDBTM.php
btry: The exception produces a message similar to this

```CommonDBTM::getFromDBByRequest() in /home/intra.dethegeek.eu.org/dethegeek/public_html/glpi-flyvemdm-93/inc/commondbtm.class.php line 311
getFromDBByRequest expects to get one result, 2 found! ```

The line of code is not predictible and may change often. I'll test  the type of exception only, as I don"'t see an obvious way to get the message and test a substring only.

Comment (trasher) [reply tests/functionnal/CommonDBTM.php]: See https://github.com/glpi-project/glpi/blob/9.2/bugfixes/tests/units/DBmysqlIterator.php#L52

File: tests/functionnal/CommonDBTM.php [hunk 6]
```diff
@@ -114,4 +114,66 @@ public function testgetIndexNameOtherThanID() {
       ]))->isFalse();
 
    }
+
+   /**
+    * @engine inline
```
trasher: Engine should not be required (I'd prefer avoid using this whan this is not required).

tests/units/Migration.php
```diff
@@ -97,6 +97,9 @@ function () {
       ]);
    }
 
+   /**
+    * @engine inline
```
trasher: Same here for engine annotation

tests/functionnal/CommonDBTM.php
btry: sorry, this was only for debug purposes. Removing

btry:
To be squashed before merge !

inc/migration.class.php
```diff
@@ -808,14 +808,14 @@ function updateDisplayPrefs($toadd = [], $todel = []) {
...
-                              );
+                           $DB->insert(
+                              'glpi_displaypreferences', [
+                                 'itemtype'  => $type,
+                                 'num'       => $newval,
+                                 'rank'      => $rank++,
+                                 'users_id'  => $data['users_id']
+                              ]
+                           );
                         }
```
btry: CS

inc/migration.class.php
```diff
@@ -824,14 +824,14 @@ function updateDisplayPrefs($toadd = [], $todel = []) {
...
-                        ]
-                     );
+                  $DB->insert(
+                     'glpi_displaypreferences', [
+                        'itemtype'  => $type,
+                        'num'       => $newval,
+                        'rank'      => $rank++,
+                        'users_id'  => 0
+                     ]
+                  );
```
btry: CS

tests/units/Migration.php
```diff
@@ -159,7 +162,24 @@ function () {
          'value'     => 'setted value'
       ]];
       $it = new \ArrayIterator($dbresult);
-      $this->calling($this->db)->request = $it;
+      $this->db = new \mock\DB();
```
trasher: I do not understand this change.

btry: I need to reinstanciate the mock to to mock the request() method. See a few lines below:
```php
$this->calling($this->db)->request[1] = $it;
$this->calling($this->db)->request[2] = [];
```
It is hard to know how namy calls to request() were done before reaching this scenario. The purpose is to reset the call counter of the method and be a bit more robust if later the count of calls changes as side effect of future development.

trasher: Stilll do not understand... `request` was called once; it is now called twice, and second is empty? 
Your changes should not affect Migration class; something sound not correct (or I just do not get the point) :/

trasher:
I wonder if this change is needed since #3738 has been merged

btry:
@trasher #3738 solves my issue with my current development. I renamed this PR to better match its current fucntion. 

Now this PR is an a ttempt to use less SQL queries to execute a getFromDBByCrit, like you sugggested when I submitted it first. But, it impacts the unit tests of Migration. I need to understand why the mock of $DB needs to be changed in this test case.

btry:
I did not identified exactly how the PR impacts the Migration unit tests. I need to allocate more time to finish this, but I'm too busy on other subjects for now.

trasher:
No hurry... You've targetted master; but it was still milestoned on 9.3; I saw that after my comments ;)

trasher:
This does not seems to be urgent, and it currently breaks tests... I close for the moment, feel free to reopen (or open another one) when you'll have time

PR #3739, opened 2018-03-20
Fix auth

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes


trasher:
Sounds like there is something wrong with circle on this one... Stucked on old buggy build.

Issue #3778, opened 2018-03-29
Assign group from user object


yllen:
Now groups are recursive.
So a user can only have groups from parents entities.

If a user have no group in current entity, you can't assign him a group from a parent entity.
Groups must be added from one entity present in completename of current entity.

PR #3779, opened 2018-03-29
fix assign group from a user - see #3778

yllen:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #3778

### DIFF
--- a/inc/group_user.class.php
+++ b/inc/group_user.class.php
@@ -148,30 +148,21 @@ class Group_User extends CommonDBRelation{
          echo "<tr class='tab_bg_2'><td class='center'>";
          echo "<input type='hidden' name='users_id' value='$ID'>";
 
-         // All entities "edited user" have access
-         $strict_entities = Profile_User::getUserEntities($ID, true);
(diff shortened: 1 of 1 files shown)


inc/group_user.class.php
```diff
@@ -148,30 +148,21 @@ static function showForUser(User $user) {
...
-
-            echo "</td><td class='tab_bg_2 center'>";
-            echo "<input type='submit' name='addgroup' value=\""._sx('button','Add')."\"
-                   class='submit'>";
-
-         } else {
-            _e('None');
-         }
+         $params = ['used'      => $used,
+                    'condition' => "is_usergroup
```
trasher: My first idea was to tell you to avoid using "hardcoded" SQL... But I took a look at the code, and this is currently not possible.
I took a look at the "condition" stuff from the dropdown, and I really don't like the way this is handled; I wonder if we can use a more reliable way.

Avoid using "condition" parameter would be the best option, or we'll probably need to rewrite that part :/

Issue #2374, opened 2017-06-28
Tell admin about new rights after update


trasher:
Defining rights after an update that provide new rights may not be possible; it is up to the admin to set correct right to profiles.

We should add a warning after upgrade as a reminder, telling user which are the new rights.

Armin83:
Great idea!

trasher:
Done

PR #3795, opened 2018-04-04
Add a method to add new rigths from update and display warning; close…

trasher:
…s #2374

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2374

### DIFF
--- a/inc/migration.class.php
+++ b/inc/migration.class.php
@@ -981,11 +981,12 @@ class Migration {
     * Add new right
     * Give full rights to profiles having config right
     *
-    * @param string $name Right name
+    * @param string  $name   Right name
+    * @param integer $rights Right to set (defaults to ALLSTANDARDRIGHT)
     *
     * @return void
     */
-   public function addRight($name) {
+   public function addRight($name, $rights = ALLSTANDARDRIGHT) {
       global $DB;
 
       $count = countElementsInTable(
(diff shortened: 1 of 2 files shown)


inc/migration.class.php
```diff
@@ -976,4 +976,53 @@ public function setContext($context) {
...
+      $count = countElementsInTable(
+         'glpi_profilerights',
+         ['name' => $name]
+      );
+      if ($count == 0) {
+         //new right for certificate
+         //give full rights to profiles having config right
+         foreach ($DB->request("glpi_profilerights", ['name' => 'config']) as $profrights) {
+            if ($profrights['rights'] && (READ + UPDATE)) {
+               $rightValue = CREATE | READ | UPDATE | DELETE  | PURGE | READNOTE | UPDATENOTE | UNLOCK;
```
orthagh: globally ok with this pr, but this part seems strange.
Here a list of issue i though about:
- DELETE is present only for itemtype with is_deleted  field
- READNOTE & UPDATE NOTEfor those which have notes tab

CREATE-UPDATE-PURGE-UNLOCK should be ok, 
But more generally some rights have only READ, ex reports, also child itemtypes have less rights, and some rights don't have any standards constants

trasher: I'm pretty sure we cannot find a really generic way... I've based the method on 2 existing cases from upgrade scripts. Maybe should we define a default standard "rightValue" (which rights for this one?) and permit to pass specific values calling the method?

trasher: I really do not know if we can have side effects... But if we cans be as generic as possible from the begining, it will be better. I've added a new commit to fix that.

PR #3797, opened 2018-04-04
Check for the right field

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

### DIFF
--- a/install/update_92_921.php
+++ b/install/update_92_921.php
@@ -63,7 +63,11 @@ function update92to921() {
    }
 
    if (!$DB->fieldExists('glpi_tickets', 'time_to_own')) {
-      $migration->addField("glpi_tickets", "time_to_own", "datetime", ['after' => 'due_date']);
+      $after = 'due_date';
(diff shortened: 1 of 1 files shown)


yllen:
**You can't change an already published file**. 
You must create a new one for next migration.

In your PR, you check is due_date is present but it was rename to time_to_resolve in update_91_92, line 773.
Done also in update_92_921, line 238.

So for me this field is not present.
And if it's always present, you must rename it to time_to_resolve

trasher:
This field should be present as of 9.1; but in some cases, there is a (fixed) bug, and field has not been renamed.
So we can have both cases, and this makes a fatal error. There is no way to get that fixed without changing the file.

yllen:
You must make a new file for new version.
- if field due_date present => rename it
- if field time_to_own not present => add it

trasher:
Upgrade will still fail on the original file.

PR #3886, opened 2018-04-19
use correct dates in stats > tickets (graph mode)

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

### DIFF
--- a/front/stat.graph.php
+++ b/front/stat.graph.php
@@ -81,7 +81,7 @@ switch ($_GET["type"]) {
    case "technicien" :
       $val1    = $_GET["id"];
       $val2    = "";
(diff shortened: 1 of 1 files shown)


front/stat.graph.php
```diff
@@ -81,38 +81,38 @@
    case "technicien" :
       $val1    = $_GET["id"];
       $val2    = "";
-      $values  = Stat::getItems($_GET["itemtype"], $_GET["date1"], $_GET["date2"], $_GET["type"]);
```
trasher: In `stat.class.php'; a link is defind with aprameters as GET; do not know when this is used.

trasher: nvermind, it' ok

PR #3897, opened 2018-04-22
Fixes #3896

### LINKED ISSUE #3896: [9.3 beta] new service level wrong url
9.3 beta fresh install  (or upgraded from 9.2.2)
setup>service level> add (+)

* Expected result :
go to glpi/front/slm.php

* Actual result :
go to glpi/**glpi**/front/slm.php

* URL of the page :
/glpifront/slm.php

9.3 beta ( tarball) fresh install ( but same behaviour with 9.3 upgraded from 9.2.2)

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #3896

Fixes the issue with the double root in URL when trying to add a new SLM item.

### DIFF
--- a/inc/auth.class.php
+++ b/inc/auth.class.php
@@ -105,12 +105,12 @@ class Auth extends CommonGLPI {
             $menu['options']['ldap']['title']           = AuthLDAP::getTypeName(Session::getPluralNumber());
             $menu['options']['ldap']['page']            = '/front/authldap.php';
             $menu['options']['ldap']['links']['search'] = '/front/authldap.php';
-            $menu['options']['ldap']['links']['add']    = AuthLDAP::getFormURL();
+            $menu['options']['ldap']['links']['add']    = AuthLDAP::getFormURL(false);
 
             $menu['options']['imap']['title']           = AuthMail::getTypeName(Session::getPluralNumber());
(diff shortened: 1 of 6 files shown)


trasher:
I wonder if the fix is correct... There are numerous calls of `getFormURL()`, and only a few `getFromURL(false)` in the current codebase. If that was the issue, I guess it should have been reported long time ago.
But I do not have any idea of what is wrong yet :/

cconard96:
I'm not sure. The fixes work for me, and originally I thought it was just 1 or 2 classes that had issues, but it seems to be more than that. I'm not that familiar with the source code, but I can check to see why it's working for some forms and not others later today.

The only relevant changes I saw was a commit to change some hard coded URLs to use method calls.

trasher:
As you've stated, issue seems global, we'll need to investigate a bit on that.

trasher:
Since I was not sure of all possible issues; I took the faultly commit, and I reviewed all the changes, changing one by one all errors introduced.
See https://github.com/glpi-project/glpi/pull/3906 (mainly the same as yours) ; thank you :)

PR #3942, opened 2018-04-30
Clean backup

AnaelMobilia:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | none

  - Update XML dump to have date and version on filename
  - Use the DB query builder

### DIFF
--- a/front/backup.php
+++ b/front/backup.php
@@ -40,9 +40,6 @@ if (isset($_POST['check_version'])) {
 
 Session::checkRight("backup", READ);
 
-// full path
-$path = realpath(GLPI_DUMP_DIR);
-
 Html::header(__('Maintenance'), $_SERVER['PHP_SELF'], "admin", "backup");
 
 $max_time = min(get_cfg_var("max_execution_time"), get_cfg_var("max_input_time"));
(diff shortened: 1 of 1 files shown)


front/backup.php
```diff
@@ -62,65 +62,50 @@
...
-
-      $query[$i] = "SELECT *
-                    FROM `$table`";
-      $i++;
+   global $DB, $path;
+
+   // Will contain the select * from $table
+   $query = [];
+   // Foreach DB tables...
+   foreach ($DB->listTables() as $id => $row) {
```
trasher: Do not call method in the loop, store in a variable before and use that on in the loop.

front/backup.php
```diff
@@ -62,65 +62,50 @@
...
+      // Save the query...
+      $query[$id] = "SELECT * FROM `" . $row['TABLE_NAME'] . "`";
    }
 
-   //le nom du fichier a generer...
-   //Si fichier existe deja il sera remplac�par le nouveau
-   $chemin = GLPI_DUMP_DIR."/backup.xml";
+   // Filename
+   $time_file = date("Y-m-d-H-i");
+   $filename = $path . "/glpi-" . GLPI_VERSION . "-$time_file.xml";
```
trasher: Maybe "glpi-backup-..."?

front/backup.php
```diff
@@ -62,65 +62,50 @@
...
    //File path
-   $A->FilePath = $chemin;
+   $A->FilePath = $filename;
 
-   // Type of layout : 1,2,3,4
-   // For details about Type see file genxml.php
-   if (empty($Type)) {
-      $A->Type = 4;
-   } else {
-      $A->Type = $Type;
```
trasher: Why this has been removed?

front/backup.php
```diff
@@ -143,45 +129,43 @@ function current_time() {
...
- * @param $table  table  name
- * @param $from   begin from
- * @param $limit  limit to
-**/
+/**
+ * Get data of a table
+ * @param DB $DB
+ * @param string $table table name
+ * @param int $from FROM xxx
+ * @param int $limit LIMIT xxx
```
trasher: Should be "integer", not "int" ;)

front/backup.php
trasher: I'd vote to keep the `GLPI_DUMP_DIR` constant rather than a global.

AnaelMobilia: @trasher : indeed, I re-used the [SQL dump code](https://github.com/glpi-project/glpi/blob/master/front/backup.php#L427) here.

+1 for using the global, I will update the SQL  code too !

AnaelMobilia:
Note : travis failure for the check... ;-)

PR #3952, opened 2018-05-02
Fix undefined method

gitdevmod:
Fix "PHP Fatal error:  Call to undefined method Html::entity_decode() in $DOC/glpi/ajax/getKbRevision.php on line 54"

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  |no
| BC breaks?    | 
| Deprecations? | 
| Tests pass?   | 
| Fixed tickets | 


gitdevmod:
After removing Html:: error is still present

PR #3988, opened 2018-05-10
Fix installation

### LINKED ISSUE #3997: GLPI 9.3 installing from tarball issue
Hi @trasher ,
while installing ( fresh ) or upgrading 9.3 from tarball the installation stops at this step testing DB engine. cant proceed after that , please refer the screen shots.
image

AnaelMobilia:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | none

  - Fix install which silently break
  - Fix bad CSS & JS URL

Replace PR #3940 
Fix #3997

### DIFF
--- a/css/style_install.css
+++ b/css/style_install.css
@@ -283,3 +283,8 @@ a.vsubmit:hover {
 .ui-widget {
    font-size: 12px;
 }
+
+/* Language dropdown */
+ul {
+    color: #000;
+}
--- a/inc/config.class.php
+++ b/inc/config.class.php
@@ -72,8 +72,6 @@ class Config extends CommonDBTM {
     *   @since 0.85
    **/
    static function getMenuContent() {
-      global $CFG_GLPI;
-
       $menu = [];
       if (static::canView()) {
          $menu['title']   = _x('setup', 'General');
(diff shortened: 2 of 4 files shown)


trasher:
> I think that it is responsability of the displayCheckDbEngine function to check if it is possible to do the check. Maybe it should display a warning like Unable to check DB engine since database connection is not established..

If you display a warning during the install process, we'll have reports and help requests about it; so I guess it is better to not display anything on the install at this time.

trasher:
> The step3 function of the install process should also include this check.

Agree with that. Fact is the check while installing must be done elsewhere, not dropped.

trasher:
I've added the check for installation

trasher:
@cedric-anne, @AnaelMobilia  are you ok with my change?

AnaelMobilia:
@trasher Seem good :-)
I made a minor commit to fix doc (int / integer)

Regards

trasher:
Great, thanks for the feedbacks. This one is OK for me.

PR #4003, opened 2018-05-14
Fix task template selection from select2; fixes #3895

### LINKED ISSUE #3895: [9.3 beta] ticket task template don't fill task category nor action time
create task template with description, category, status actiontime, actor, group.
use this template to create a ticket task task

* Expected result :
task created with filled fields

* Actual result :
only description and status are filled

* URL of the page :
http://localhost/glpi93/front/ticket.form.php?id=2236

template : 
image
created task : 
image

GLPI 9.3-dev (/glpi93 => /var/www/html/glpi93)
Installation mode: TARBALL

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #3895 

This PR fixes the select2 issue (`.select2('val', value)` is deprecated).
This works for dropdowns that already have the value loaded, but not ajax loaded ones (category, user, group, ...) :/

### DIFF
--- a/ajax/task.php
+++ b/ajax/task.php
@@ -46,6 +46,25 @@ if (isset($_POST['tasktemplates_id']) && ($_POST['tasktemplates_id'] > 0)) {
    $template = new TaskTemplate();
    $template->getFromDB($_POST['tasktemplates_id']);
 
-   $template->fields = array_map('html_entity_decode', $template->fields);
-   echo json_encode($template->fields);
+   $fields = $template->fields;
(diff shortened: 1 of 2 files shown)


trasher:
Found a "solution", but I do no like it. See 2a02c5e5fa31ebf5668294aff8227dc83d7159b5

trasher:
Since I'm pretty sure the provided solution is not correct; I'll not change for now.

PR #4011, opened 2018-05-15
Fix menu

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes


orthagh:
This menu could be strange to understand but "Field unicity" is a third level menu:
Setup > Checks > Fields unicity

Ok, Checks menu only contains one entry and it's indeed superflu to have an additional level.
But a long waited feature called "configuration" (like implemented in [Typology plugin](http://plugins.glpi-project.org/#/plugin/typology) ) will be set here in the future (when ? I can't say but we have a commercial lead about this).

So i never removed myself the "superflu" level and i don't know if i approve or refuse this pr.

ping @flegastelois, do "configuration management" feature is still expected ?

PR #4071, opened 2018-05-25
Feature/fix metacriteria search query

Badoche1337:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 

When you use meta criterias with OR standard criteria, metacriterias are only used in second part of where clause (after the OR keyword).

This PR handle metacriterias before standards ones and add these criterias in each part of the where clause.

### DIFF
--- a/inc/search.class.php
+++ b/inc/search.class.php
@@ -777,7 +777,7 @@ class Search {
                      if ($tmp_new_where !== false) {
                         $new_where .= $tmp_new_where;
                         if (!empty($METAWHERE)) {
-                          $new_where .= ' AND ' . $METAWHERE;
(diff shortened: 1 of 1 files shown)


trasher:
Could you please rebase on master? It will not be part of 9.3 since this is a hug change and we're already on a RC.

orthagh:
i close, i'll push in the month a better version for this issue.
Here you appy metacriteria after each normal-criteria.

In the planned feature, you will be able to set a meta criteria anywhere in the query and also do some group of criteria (critieria in criteria)

See this wip branch: https://github.com/orthagh/glpi/tree/feature/nested_search

PR #4103, opened 2018-05-30
Fixes #4099

### LINKED ISSUE #4099: User right bug read/write ticket author
From version 9.1.7 to version 9.2.2 the rights differ with the same configuration of rights 

### Example : 

Test GLPI 9.1.7

- Creation of a ticket by techA
    - Ticket assigned to techA
- techA has the right to modify its ticket but not techB
- techB has the right to change the status of the ticket

Test GLPI 9.2.2

 - Creation of a ticket by techA
    - Ticket assigned to techA
- techA and techB do not have the right to change the ticket

priorite

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #4099

### DIFF
--- a/tests/functionnal/Ticket.php
+++ b/tests/functionnal/Ticket.php
@@ -691,7 +691,7 @@ class Ticket extends DbTestCase {
       $this->boolean((boolean)$ticket->canMassiveAction('update', 'type', 'qwerty'))->isTrue();
       $this->boolean((boolean)$ticket->canMassiveAction('update', 'location', 'qwerty'))->isTrue();
       $this->boolean((boolean)$ticket->canCreateItem())->isTrue();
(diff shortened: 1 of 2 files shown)


tests/functionnal/Ticket.php
```diff
@@ -1020,33 +1020,33 @@ public function testFormTech() {
          $location = true
       );
 
-      //drop update ticket right from tech profile
+      //drop UPDATE ticket right from tech profile (still with OWN)
```
trasher: Maybe a new test with tech without update and without own should be a good idea.

Issue #4108, opened 2018-05-31
[9.2.X] Plugin Redirect


keha35:
* Steps to reproduce (which actions have you made) :
Use any plugin with redirection using $_GET['Redirect'] ('Plugin'+nameOfPlugin_Id) to simplified interface.

* Expected result :
Redirection to the plugin page

* Actual result :
Redirection to the home page

* URL of the page :
http://localhost/glpi92/index.php?redirect=PluginXX_1

* Your GLPI setup (you can find it in Setup > General menu, System tab) :
GLPI 9.2.3 With windows 10

PR #4109, opened 2018-05-31
Fixe #4108 Plugin Redirect

keha35:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #4108

### DIFF
--- a/inc/commondbtm.class.php
+++ b/inc/commondbtm.class.php
@@ -4807,7 +4807,7 @@ class CommonDBTM extends CommonGLPI {
          }
 
          //retrieve entity
-         $entities_id = $_SESSION['glpiactive_entity'];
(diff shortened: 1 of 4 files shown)


orthagh:
Hello @keha35 

ok for the issue for the redirect missing on helpdesk interface.

Some points:
- Since 0.90, this redirect feature is replaced by direct url calls, if login required, redirect will be automatically set after, we don't need anymore to "forge" redirect params.
- We can add your enhancement, but one change required: as the code you added is the same as the central interface, we need to factorise it in a function.
- Much better, with this function we could add an unit test to avoid regressions.

trasher:
I do not understand the factorization. It sems that some cases are now lost, I do not think this is expected.

trasher:
ping?

trasher:
No ping response; I'm closing this one.

PR #4129, opened 2018-06-04
Fix solutions migration

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes


trasher:
According to documentation, syntax is `IFNULL(expression, alt_value) `; so the first value should be the one in the table if this is present, and the one if the log if not.

Ping @tomolimo ?

trasher:
See http://forum.glpi-project.org/viewtopic.php?pid=348586#p348586

Issue #4267, opened 2018-06-29
Error Using Rack Template


Dackson:
* Steps to reproduce (which actions have you made) :
Add new rack model
-- Assets > Racks > Manage Templates > Add a template
Add new model based on model rack
-- Assets > Racks >  Add new rack > Choose a template
-- Save
-- Nothing happens

* Expected result :
A new Rack is added based on the template already created

* Actual result :
After adding a Rack template, and trying to use it nothing happens.
Note: The model is being saved in the glpi_rack table instead of the glpi_rackmodels table

* URL of the page :
Not

* Screenshot of the problem (if pertinent) :
![glpierro](https://user-images.githubusercontent.com/3226885/42108016-b29d87f4-7baf-11e8-9432-d1dc7c0924e0.png)

* Your GLPI setup (you can find it in Setup > General menu, System tab) :
Not

cconard96:
Rack templates are supposed to be stored in glpi_racks like other item types. the glpi_rackmodels table is used for models like "Dell Netshelter SX 24U". However, you are correct that adding racks from template fails add items to the DB. I should have a fix shortly.

vikipediaEx:
hello may I ask has this issue been fixed?

PR #4269, opened 2018-06-30
Fix creation from rack template

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #4267 

There was code missing from the rack class's prepareInputForAdd. I've copied the code from the monitor class, and that seems to restore expected functionality.

### DIFF
--- a/inc/rack.class.php
+++ b/inc/rack.class.php
@@ -895,9 +895,6 @@ JAVASCRIPT;
          if (isset($input["id"]) && ($input["id"] > 0)) {
             $input["_oldID"] = $input["id"];
          }
-         if (isset($input["size"]) && ($input["size"] == '')) {
-            unset($input["size"]);
-         }
          unset($input['id']);
          unset($input['withtemplate']);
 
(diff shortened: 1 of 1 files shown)


inc/rack.class.php
```diff
@@ -891,7 +891,19 @@ classes: 'qtip-shadow qtip-bootstrap rack_tipcontent'
    }
 
    function prepareInputForAdd($input) {
-      return $this->prepareInput($input);
+      if ($this->prepareInput($input)) {
+         if (isset($input["id"]) && ($input["id"] > 0)) {
+            $input["_oldID"] = $input["id"];
+         }
+         if (isset($input["size"]) && ($input["size"] == '')) {
```
trasher: I do not understand this one; there is no size field in racks

trasher: So I guess this should just be removed. Rest looks good (and fix the original issue)

Issue #3763, opened 2018-03-27
Do not use session as cache


trasher:
Some part of the code does use the SESSION as a kind of cache; this should be avoid.

trasher:
This concerns (among others):
* glpi_table_of
* glpi_foreign_key_field_of
* glpi_all_possible_rights
* glpi_item_device_affinities

trasher:
This may also concern some entries in `$CFG_GLPI`:
* languages
* glpitables (used?)
* *types
* itemdevices
* networkport_instantiations
* itemdevices_itemaffinity
* lock_lockable_objects
* inventory_lockable_objects
* priority_matrix

PR #4301, opened 2018-07-06
Feature/session as cache

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | yes
| Tests pass?   | yes
| Fixed tickets | #3763

### DIFF
--- a/inc/commondbtm.class.php
+++ b/inc/commondbtm.class.php
@@ -163,6 +163,20 @@ class CommonDBTM extends CommonGLPI {
 
 
    /**
+    * Get known tables
+    *
+    * @return array
+    */
+   public static function getTablesOf() {
+      global $GLPI_CACHE;
+      if ($GLPI_CACHE->has('table_of')) {
+         return $GLPI_CACHE->get('table_of');
+      }
+      return [];
+   }
+
+
+   /**
     * Return the table used to store this object
     *
     * @param string $classname Force class (to avoid late_binding on inheritance)
@@ -170,6 +184,7 @@ class CommonDBTM extends CommonGLPI {
     * @return string
    **/
    static function getTable($classname = null) {
+      global $GLPI_CACHE;
 
       if ($classname === null) {
          $classname = get_called_class();
@@ -179,11 +194,13 @@ class CommonDBTM extends CommonGLPI {
          return '';
       }
 
-      if (empty($_SESSION['glpi_table_of'][$classname])) {
-         $_SESSION['glpi_table_of'][$classname] = getTableForItemType($classname);
+      $glpi_tables = self::getTablesOf();
+      if (!isset($glpi_tables[$classname]) || empty($glpi_tables[$classname])) {
+         $glpi_tables[$classname] = getTableForItemType($classname);
+         $GLPI_CACHE->set('table_of', $glpi_tables);
       }
 
-      return $_SESSION['glpi_table_of'][$classname];
+      return $glpi_tables[$classname];
    }
 
 
@@ -195,7 +212,10 @@ class CommonDBTM extends CommonGLPI {
     * @return void
    **/
    static function forceTable($table) {
-      $_SESSION['glpi_table_of'][get_called_class()] = $table;
+      global $GLPI_CACHE;
+      $glpi_tables = self::getTablesOf();
+      $glpi_tables[get_called_class()] = $table;
+      $GLPI_CACHE->set('table_of', $glpi_tables);
    }
 
 
--- a/inc/commondbtm.class.php
+++ b/inc/commondbtm.class.php
@@ -169,7 +169,7 @@ class CommonDBTM extends CommonGLPI {
     */
    public static function getTablesOf() {
       global $GLPI_CACHE;
-      if ($GLPI_CACHE->has('table_of')) {
+      if ($GLPI_CACHE != null && $GLPI_CACHE->has('table_of')) {
          return $GLPI_CACHE->get('table_of');
       }
       return [];
@@ -197,7 +197,9 @@ class CommonDBTM extends CommonGLPI {
       $glpi_tables = self::getTablesOf();
       if (!isset($glpi_tables[$classname]) || empty($glpi_tables[$classname])) {
          $glpi_tables[$classname] = getTableForItemType($classname);
-         $GLPI_CACHE->set('table_of', $glpi_tables);
+         if ($GLPI_CACHE != null) {
+            $GLPI_CACHE->set('table_of', $glpi_tables);
+         }
       }
 
       return $glpi_tables[$classname];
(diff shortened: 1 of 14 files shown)


inc/includes.php
```diff
@@ -105,15 +98,14 @@
    $PLUGINS_INCLUDED = 1;
    $LOADED_PLUGINS   = [];
    $plugin           = new Plugin();
-   if (!isset($_SESSION["glpi_plugins"])) {
+   if ($plugin->getPlugins() === null) {
```
cedric-anne: `if (count($plugin->getPlugins()) === 0) {`

inc/migration.class.php
```diff
@@ -699,7 +699,7 @@ function executeMigration() {
       $this->storeConfig();
 
       // as some tables may have be renamed, unset session matching between tables and classes
-      unset($_SESSION['glpi_table_of']);
+      $GLPI_CACHE->removeItem('table_of');
```
cedric-anne: `global $GLPI_CACHE` missing.

inc/plugin.class.php
```diff
@@ -1876,4 +1868,83 @@ static function getState($state = 0) {
...
+   /**
+    * Set plugin unloaded from its name
+    *
+    * @since 9.4
+    *
+    * @param integer $name Plugin name
+    *
+    * @return void
+    */
+   public static function setUnloadedName($name) {
```
cedric-anne: I would have name it `setUnloadedByName`.

inc/plugin.class.php
```diff
@@ -347,14 +348,10 @@ function checkStates() {
    static function isAllPluginsCSRFCompliant() {
```
cedric-anne: This method is not used (only use of it is commented).

inc/plugin.class.php
```diff
@@ -1876,4 +1868,83 @@ static function getState($state = 0) {
...
+    *
+    * @param integer $id Plugin id
+    *
+    * @return void
+    */
+   public static function setUnloaded($id) {
+      global $GLPI_CACHE;
+      $plugins = $GLPI_CACHE->getItem('plugins');
+      unset($plugins[$id]);
+      $GLPI_CACHE->replcaeItem('plugins', $plugins);
```
cedric-anne: `replcaeItem` -> `replaceItem`

inc/commondbtm.class.php
```diff
@@ -126,18 +144,40 @@ static function getTable($classname = null) {
...
 
-      if (empty($_SESSION['glpi_foreign_key_field_of'][get_called_class()])) {
-         $_SESSION['glpi_foreign_key_field_of'][get_called_class()]
-            = getForeignKeyFieldForTable(static::getTable());
+   static function getForeignKeyField() {
+      $fkeys = self::getForeignKeyFielsdOf();
+      $classname = get_called_class();
+      if (!isset($fkeys[$classname]) || empty($fkeys[$classname])) {
+         $fkeys[$classname] = getForeignKeyFieldForTable(static::getTable());
+         $GLPI_CACHE->replaceItem('foreign_key_field_of', $fkeys);
```
cedric-anne: Missing `global $GLPI_CACHE;`.

inc/config.class.php
```diff
@@ -2922,12 +2924,41 @@ public static function getCache($optname, $context = 'core') {
...
+            $cache = Zend\Cache\StorageFactory::factory($opt);
+            $cache_class = get_class($cache);
+         }
          if (Session::DEBUG_MODE == $_SESSION['glpi_use_mode']) {
-            Toolbox::logError($e->getMessage());
+            Toolbox::logDebug($e->getMessage());
          }
       }
-      Toolbox::logDebug("CACHE $optname", $cache_class);
+      //Toolbox::logDebug("CACHE $optname", $cache_class);
```
cedric-anne: If you remove this line, `$cache_class` can be removed too.

scripts/cliupdate.php
```diff
@@ -57,6 +57,7 @@
 include_once (GLPI_ROOT . "/inc/db.function.php");
 include_once (GLPI_CONFIG_DIR . "/config_db.php");
 
+global $PHPLOGGER, $SQLLOGGER; //globals...
```
cedric-anne: I do not see why it is declared here.

inc/includes.php
trasher: Hum... We're originally trying to knw here if "something" has been loaded; we definitely cannot rely on the `getPlugins` method.

inc/plugin.class.php
trasher: Indeed, even if it's out of the current PR scope

inc/plugin.class.php
```diff
@@ -1876,4 +1868,83 @@ static function getState($state = 0) {
 
       return __('Error / to clean');
    }
+
+   /**
+    * Get plugins list
+    *
+    * @since 9.4
```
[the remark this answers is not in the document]
trasher: It is not important, and all other entries says "9.4" right now (even the one in the database). Out of scope.

scripts/cliupdate.php
trasher: They must exists before they can be used/assigned; declaring a global at the beginning rather than in any further line does not sounds illogical to me.

inc/plugin.class.php
trasher: https://github.com/glpi-project/glpi/pull/4460

inc/migration.class.php
trasher: fixed

inc/includes.php
```diff
@@ -40,25 +40,18 @@
...
-   }
-}
 
 /// TODO try to remove them if possible
 include_once (GLPI_ROOT . "/inc/db.function.php");
 
 // Standard includes
 include_once (GLPI_ROOT . "/inc/config.php");
 
+$GLPI_CACHE->removeItems(['table_of', 'foreign_key_field_of']);
```
cedric-anne: If items are erased on each process, maybe forcing the `Memory` adapter is a better solution.

trasher: Do not remember but that looks like a dev only line :D

cedric-anne: No, looks like it was already done before your modifications.

cedric-anne: `glpi_foreign_key_field_of` is computed using `CommonDBTM::getTable()` function, that can be overrided by items. Maybe it has been made to be sure to do not have cache problem when updating `CommonDBTM::getTable()` results.

cedric-anne: My bad, it was not cleaning but it was defining an empty array if variables was not set.

trasher:
>  Reading and writing it from too many places increase risk of having something that "breaks" stored data.

Cache entries are managed from a few specific methods; I do not understand "too many places".

orthagh:
One functional issue:
image

On opcache & apcu, it's already not very clear, but in case of file, it uselless i think

inc/commondbtm.class.php
```diff
@@ -179,11 +194,15 @@ static function getTable($classname = null) {
          return '';
       }
 
-      if (empty($_SESSION['glpi_table_of'][$classname])) {
-         $_SESSION['glpi_table_of'][$classname] = getTableForItemType($classname);
+      $glpi_tables = self::getTablesOf();
+      if (!isset($glpi_tables[$classname]) || empty($glpi_tables[$classname])) {
+         $glpi_tables[$classname] = getTableForItemType($classname);
+         if ($GLPI_CACHE != null) {
+            $GLPI_CACHE->set('table_of', $glpi_tables);
```
orthagh: One thing here for this entire block, as it is repeated in other place, could it be better to place the `getTableForItemType` and the cache set in the `getTablesOf` function, no ?

trasher:
I've reworked a bit. A low memory usage is a GLPI issue, not a user/config one; so I've dropped this part. Also, high was at 50%, I've set to 80.

For the UI part...
APCu:
screenshot_2018-10-23 glpi - setup 2
Filesystem:
screenshot_2018-10-23 glpi - setup 1
Memory:
screenshot_2018-10-23 glpi - setup

Cannot test other adapters.

inc/commondbtm.class.php
trasher: Where is this code repeated?

I do not understand what you have in mind, to me that's not really possible. Will talk about this wit you directly.

orthagh:
perfect for ui

PR #4316, opened 2018-07-09
Hide ticket filter on page load

InformatiqueOLLN:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | na

### DIFF
--- a/front/ticket.php
+++ b/front/ticket.php
@@ -44,6 +44,13 @@ echo Html::manageRefreshPage();
 
 Search::show('Ticket');
 
+echo'
+<script type="text/javascript">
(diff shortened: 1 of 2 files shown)


trasher:
I do not agrre with this change.
And this should probably not be done like this.

InformatiqueOLLN:
Hi,
It was simply a quick and dirty solution.
The best way to do this would be a parameter in user account.
Sorry for this patch that should not be applied.

trasher:
Ok, closing for now

Issue #4337, opened 2018-07-12
Massive action for a document target


pedroluisvsousa:
In massive action of document target in knowledge base, I can't delete permanently the profile item. But i have a message "No selected items".

I'm using GLPI 9.3.

![erro_base_conhecimento](https://user-images.githubusercontent.com/26510405/42650429-c1db971e-85e2-11e8-864c-10276dddf241.png)

trasher:
Cannot reproduce on current 9.3/bugifxes, nor on the demo. What is the browser?
Tested successfully with Firefox and Chromium.

pedroluisvsousa:
The problem related occurs only when the target is an **user profile**.
I have tested in Firefox and Chrome browsers.
Bug present in 9.2.x and 9.3.x.

Please, see the video below.

https://www.dropbox.com/s/g64yaaxt0490r1u/user_profile_target_erro.ogv?dl=0

trasher:
I still cannot reproduce.

cconard96:
I'm still able to reproduce on the demo site. Are you using the bugfixes branch? You already merged my fix PR a while ago.

pedroluisvsousa:
I tested using the brugfixes branch and the bug was solved.

PR #4361, opened 2018-07-18
Profile KnowbaseItem naming special case

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes 
| Fixed tickets | #4337 

Add special case for knowledgebaseitem due to class naming scheme oddity.
Actual: KnowbaseItem_Profile
Expected and the scheme for other itemtypes: Profile_KnowbaseItem

### DIFF
--- a/inc/commondbvisible.class.php
+++ b/inc/commondbvisible.class.php
@@ -276,7 +276,12 @@ abstract class CommonDBVisible extends CommonDBTM {
                echo "<tr class='tab_bg_1'>";
                if ($canedit) {
                   echo "<td>";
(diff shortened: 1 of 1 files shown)


cconard96:
It occurs to me now that an alternative solution would be to use class_alias.
class_alias("KnowbaseItem_Profile","Profile_KnowbaseItem);

At that point, should ItemType_User classes be aliased to User_ItemType as well to be consistent with Group_ItemType, Entity_ItemType, and other Profile_ItemType classes?

Or, rename the classes to be consistent and then use aliases to their current names to maintain compatibility (Remove alias in later version?).

Issue #1604, opened 2017-01-12
To be able to assign users to groups using LIKE % instead of complete name


tomolimo:
Hello
I propose to give more possibilities when searching LDAP for user members of groups.
Let's have an example:
![image](https://cloud.githubusercontent.com/assets/7569958/21894721/867c01b4-d8e0-11e6-839d-bd9e73583606.png)
Current test is to match exact string in the attribute value. This has the drawback if 'department' is not written the exact same way in LDAP, then it will not match

I propose to use LIKE instead of IN in the mySQL query, This will permit to have a more flexible use of assignment:
![image](https://cloud.githubusercontent.com/assets/7569958/21894808/cc0ee6ba-d8e0-11e6-8d1d-0372615f5e55.png)
Using LIKE operator will permit such criterion. In this case any user department matching the '163387 %' will be added to the group.

orthagh:
ping @wawax for this issue and the corresponding PR #1606 

Moved to 9.1.3 to have time for tests (9.1.2 freezed today for wanted date of release next monday)

PR #4365, opened 2018-07-19
fix call to undefined method DB::quoteField, fix #4364

### LINKED ISSUE #4364: Uncaught Error: Call to undefined method DB::quoteField()
From http://forum.glpi-project.org/viewtopic.php?id=164026

See: https://github.com/glpi-project/glpi/blob/155cc60fd56ea747854cd100cf726129f269704d/inc/user.class.php#L1424

Since 155cc60fd56

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #4364

Reading the previous code before https://github.com/glpi-project/glpi/commit/155cc60fd56ea747854cd100cf726129f269704d#diff-a9c4448deb6265f035355e2a9ca92bc0L1408, i think the LIKE part is completly useless.

By the way the new line in above commit was not equivalent to previous, it miss to include $lgroup in expression.

### DIFF
--- a/inc/dbmysqliterator.class.php
+++ b/inc/dbmysqliterator.class.php
@@ -403,8 +403,13 @@ class DBmysqlIterator implements Iterator, Countable {
             $ret .= " $bool ";
          }
          if (is_numeric($name)) {
-            // No Key case => recurse.
-            $ret .= "(" . $this->analyseCrit($value, $bool) . ")";
+            // no key and direct expression
+            if ($value instanceof QueryExpression) {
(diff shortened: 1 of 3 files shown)


orthagh:
Here is the "why" for the LIKE:
https://github.com/glpi-project/glpi/issues/1604

trasher:
LGTM but unt tests should be added on those ones I guess

PR #4402, opened 2018-07-30
fix(calendar): fix timezone from begin and end date

stonebuzz:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

When exporting ical, Outlook misconstrues the TZID directive, resulting a time shift.
The idea is to include in the start and end timestamp the UTC time zone


pvasener:
Tested and validated on my production platform.

PR #4420, opened 2018-08-02
Feature/deprecate rawsql

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

### DIFF
--- a/inc/ruledictionnarydropdowncollection.class.php
+++ b/inc/ruledictionnarydropdowncollection.class.php
@@ -230,10 +230,11 @@ class RuleDictionnaryDropdownCollection extends RuleCollection {
             // No item left : delete old item
             if ($result
                 && ($DB->result($result, 0, 0) == 0)) {
-               $Sql              = "DELETE
-                                    FROM `".$this->item_table."`
-                                    WHERE `id` = '$ID'";
-               $resdel           = $DB->query($Sql);
+               $DB->delete(
+                  $this->item_table, [
+                     'id'  => $ID
+                  ]
+               );
                $deletecartmodel  = true;
             }
 
@@ -252,10 +253,11 @@ class RuleDictionnaryDropdownCollection extends RuleCollection {
                      }
                      // Delete cartrodges_assoc
                      if ($deletecartmodel) {
-                        $sql = "DELETE
-                                FROM `glpi_cartridgeitems_printermodels`
-                                WHERE `printermodels_id` = 'id'";
-                        $DB->query($sql);
+                        $DB->delete(
+                           'glpi_cartridgeitems_printermodels', [
+                              'printermodels_id'   => $ID
+                           ]
+                        );
                      }
                      // Add new assoc
                      $ct = new CartridgeItem();
--- a/inc/transfer.class.php
+++ b/inc/transfer.class.php
@@ -2631,28 +2631,22 @@ class Transfer extends CommonDBTM {
          default :
             // Copy -> Copy datas
             if ($ID != $newID) {
-               $query = "SELECT *
-                         FROM `glpi_logs`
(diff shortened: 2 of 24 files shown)


inc/dbmysql.class.php
```diff
@@ -121,7 +121,7 @@ function connect($choice = null) {
          $this->dbh->set_charset(isset($this->dbenc) ? $this->dbenc : "utf8");
 
          if (GLPI_FORCE_EMPTY_SQL_MODE) {
-            $this->dbh->query("SET SESSION sql_mode = ''");
+            $this->dbh->query("SET SESSION sql_mode = ''", true);
```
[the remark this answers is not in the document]
cedric-anne: Maybe you made a mistalke, thinking this call was on `self::query()` function.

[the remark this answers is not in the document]
trasher: Yes, this one should not have been added :)

inc/dbmysql.class.php
```diff
@@ -154,10 +156,21 @@ function escape($string) {
     *
     * @throws GlpitestSQLError
     */
-   function query($query) {
+   function query($query, $new = false) {
       global $CFG_GLPI, $DEBUG_SQL, $SQL_TOTAL_REQUEST;
 
       $is_debug = isset($_SESSION['glpi_use_mode']) && ($_SESSION['glpi_use_mode'] == Session::DEBUG_MODE);
+
+      if ($new === false && ($is_debug || defined('TU_USER'))) {
```
cedric-anne: I do not like the idea of adding a parameter for this behaviour.

I think we should deprecate usage in this function and create a new **private** method, that consider the given query as secured, and call it from BDmysql methods that "compile" queries. Doing this will consider all raw `['INSERT', 'UPDATE', 'DELETE']` queries given from another class as deprecated, which seems to be what we want.

cedric-anne: Another solution is to use a code like this:
```
$caller = end(debug_backtrace(0, 2));
if ($caller && $callerClass = $caller['class'] && !in_array($callerClass, [DBmysql::class, DBmysqlIterator::class])) {
```
Problem is that DBmysqlIterator call this method, so we could not deprecate non provate usage.

trasher: Change reverted.

inc/ruledictionnarydropdowncollection.class.php
```diff
@@ -230,10 +230,11 @@ function replayRulesOnExistingDBForModel($offset = 0, $maxtime = 0) {
             // No item left : delete old item
             if ($result
                 && ($DB->result($result, 0, 0) == 0)) {
-               $Sql              = "DELETE
-                                    FROM `".$this->item_table."`
-                                    WHERE `id` = '$ID'";
-               $resdel           = $DB->query($Sql);
+               $DB->delete(
+                  $this-item_table, [
```
orthagh: missing `>`

inc/transfer.class.php
```diff
@@ -2627,28 +2627,21 @@ function transferHistory($itemtype, $ID, $newID) {
...
-                                          '".$data['old_value']."', '".$data['new_value']."')";
-                        $DB->query($query);
-                     }
-                  }
+               while ($data = $iterator->next()) {
+                  $data = Toolbox::addslashes_deep($data);
+                  $data = [
+                     'items_id'  => $newID,
+                     'itemtype'  => $itemtype
+                  ] + $data;
```
orthagh: this part seems ko:
We'll have an `id` key in this `$data` array we should remove, no ?

PR #4444, opened 2018-08-06
Fix responsiveness of Search Field

M-Falken:
screenshot_20180806_114846
Fix for responsiveness of the search field under 454px screen width, on some mobile (P20 pro for example).
The field was on top of the hamburger menu icon.
More width to the field added too, as it was possible.

| Bug fix?      | yes
| Tests pass?   | yes

Thanks

### DIFF
--- a/css/styles.css
+++ b/css/styles.css
@@ -5397,3 +5397,16 @@ input[name=as_map]:checked + label {
       margin-left: 200px;
    }
 }
+
+/** Responsive for search field */
+@media screen and (max-width: 454px) {
+   #c_recherche form #champRecherche input{
+      width:140px;
+   }
+
+   #c_recherche {
+      position:relative;
+      top:14px;
+      left:-47px;
+   }
+}


orthagh:
I'm not sure at all.

An example on one plus 3T with your changes:
image

By the way, p20 pro have a screen width of 1080, i don't know where you found 454px.

orthagh:
Expecting result with current glpi (it works for me):
image

orthagh:
Another thing, which browser are you using ?

M-Falken:
Ok, I've used FF and chrome, latest version for both on my PC.
The 454px is because when you test with the dev mode with adaptative mode the problem start at this width, and I mean on a vertical view.
My screenshot on first post is with my P20 pro on vertical view and with Chrome.
Will test few more just in case.

M-Falken:
I think I've found out what was this pseudo bug.
Unfortunately it seems that the ./css/style.min.css wasn't loaded correctly certainly because of cache issue.
So the rules were not the good ones.

After applied the 9.3.1-bugfixes and give it a try again with:

- Android 8.1.0 CLT-L09 + chrome 68.0.3440.85 (P20 Pro) both vertical and horizontal : test ok
- W10 + Firefox 61.0.1 : test ok
- W10 + Chrome 68.0.3440.84 : test ok
- W10 + Edge 42.17134.1.0 : test ok

Thanks again

Issue #3270, opened 2017-12-11
Font Awesome 5


trasher:
Font Awesome version 5 has just been released.

Need to checks size, compatibility, what their new deign would look, and the amount of required work to switch.

orthagh:
https://fontawesome.com/how-to-use/upgrading-from-4

cconard96:
@trasher 
The upgrade is mostly compatible and about the same size if we can remove unneeded fonts like the brands fonts (Extra 1 MB otherwise). Some icons like the trash icon used before is now pro and had to be changed. Others have different looks by default.

See https://github.com/cconard96/glpi/tree/feature/fa5 for POC. I can open a PR if you want, but more testing will need done to make sure I didn't miss any icons that are broken.

orthagh:
@cconard96 publish your poc in pr.
We will test it.

AnaelMobilia:
@trasher : I think this issue could now be closed (#4491 ) ;-)
Regards,
Anael

PR #4491, opened 2018-08-13
Upgrade Font Awesome to version 5.2.0

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #3270


orthagh:
first and rapid review:

most important part is not changed at the the moment:
image

cconard96:
Should be... test environment had it working as the code for those icons didn't need changed. I'll check I didn't have something cached on my browser when I get off work (+10 hours).

Its possible chrome handles them differently than Firefox as well and requires a higher font weight.

cconard96:
I haven't been able to re-create that issue on a fresh copy of the master branch and my commit. Tested in Linux with Chrome and Firefox, and Chrome from my phone.

cconard96:
This should be the final commit. I believe the missing icon issue has been corrected by setting font weights in 9c4717d.

orthagh:
Ok, it was my fault for missing icon, my working folder for master was broken.

With your last and a clean on my side, here is some additional notes:
- size for some icons changed: 

image

- Calendar pic is very ugly: 
image

Maybe we could use `calendar-alt`

- fa-global became also ugly, 
image

I suggest replace by `globe-americas`

orthagh:
One last thing also, instead adding font-weight, it could be better to add the corresponding classes provided by the lib (if not used in a before|after pseudo classes)

Issue #4530, opened 2018-08-23
[9.3] sla ola time to own - error in view with debug mode on rule auto assign


Armin83:
* Steps to reproduce (which actions have you made) :
assign ola time to own with ticket rules (Debug mode on)

* Expected result :
No errors in view.
(ola time to solve no error)
(sla time to own no error)
(sla time to solve no error)

* Actual result :
[2018-08-23 21:00:20] glpiphplog.ERROR: Toolbox::userErrorHandlerNormal() in /var/www/glpi/inc/toolbox.class.php line 657
  *** PHP Notice(8): Undefined index: date
  Backtrace :
  inc/levelagreement.class.php:272
  inc/ticket.class.php:4717                          LevelAgreement->showForTicket()
  inc/commonglpi.class.php:468                       Ticket->showForm()
  ajax/common.tabs.php:92                            CommonGLPI::displayStandardTab()
  {"user":"2@glpidev","mem_usage":"0.018\", 3.26Mio)"}
[2018-08-23 21:00:20] glpiphplog.ERROR: Toolbox::userErrorHandlerNormal() in /var/www/glpi/inc/toolbox.class.php line 657
  *** PHP Notice(8): Undefined index: olalevels_id
  Backtrace :
  inc/levelagreement.class.php:274
  inc/ticket.class.php:4717                          LevelAgreement->showForTicket()
  inc/commonglpi.class.php:468                       Ticket->showForm()
  ajax/common.tabs.php:92                            CommonGLPI::displayStandardTab()

* URL of the page :
http://192.168.1.108/front/ticket.form.php?id=8609

* Screenshot of the problem (if pertinent) :
![unbenannt](https://user-images.githubusercontent.com/19192087/44546527-83f93080-a718-11e8-8027-65fe01453391.PNG)

* Your GLPI setup (you can find it in Setup > General menu, System tab) :
```
Operating system: Linux glpidev 4.15.0-22-generic #24-Ubuntu SMP Wed May 16 12:15:17 UTC 2018 x86_64
PHP 7.2.7-0ubuntu0.18.04.2 apache2handler (Core, PDO, Phar, Reflection, SPL, SimpleXML, Zend OPcache, apache2handler, apc, apcu,
	calendar, ctype, curl, date, dom, exif, fileinfo, filter, ftp, gd, gettext, hash, iconv, imagick, imap, intl, json, ldap,
[...]

PR #4541, opened 2018-08-26
Don't use getFromIter for one result

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #4530 

Fix matches a previous patch to slalevel_ticket class

### DIFF
--- a/inc/olalevel_ticket.class.php
+++ b/inc/olalevel_ticket.class.php
@@ -82,8 +82,9 @@ class OlaLevel_Ticket extends CommonDBTM {
          ],
          'LIMIT'        => 1
       ]);
-      if (count($iterator)) {
-         return $this->getFromIter($iterator);
+      if (count($iterator) == 1) {
+         $row = $iterator->next();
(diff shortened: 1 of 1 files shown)


inc/olalevel_ticket.class.php
```diff
@@ -82,8 +82,9 @@ function getFromDBForTicket($ID, $olaType) {
          ],
          'LIMIT'        => 1
       ]);
-      if (count($iterator)) {
-         return $this->getFromIter($iterator);
+      if (count($iterator) == 1) {
+         $row = $iterator->next();
+         return $this->getFromDB($row['id']);
```
trasher: Not OK since it will make one extra query on the database. I ad in mind getFromIter return one result only

trasher: OK, I was wrong; change is OK, thank you!

PR #4554, opened 2018-08-30
Collector : can now specify where the requester's email is found

AdrienClairembault:
'from' : current behavior
'reply-to' : use the email found in the 'reply-to' header

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

### DIFF
--- a/inc/mailcollector.class.php
+++ b/inc/mailcollector.class.php
@@ -82,6 +82,9 @@ class MailCollector  extends CommonDBTM {
    const REFUSED_FOLDER  = 'refused';
    const ACCEPTED_FOLDER = 'accepted';
 
+   // Values for requester_field
+   const REQUESTER_FIELD_FROM = 0;
+   const REQUESTER_FIELD_REPLY_TO = 1;
 
    static function getTypeName($nb = 0) {
       return _n('Receiver', 'Receivers', $nb);
@@ -297,12 +300,12 @@ class MailCollector  extends CommonDBTM {
       Dropdown::showYesNo("use_mail_date", $this->fields["use_mail_date"]);
       echo "</td></tr>\n";
 
-      echo "<tr class='tab_bg_1'><td>" . __('Caller\'s email location') . "</td>";
+      echo "<tr class='tab_bg_1'><td>" . __('Field for requester') . "</td>";
       echo "<td>";
-      Dropdown::showFromArray("caller_location", [
-         "from" => __('From'),
-         "reply-to" => __('Reply-To')
-      ], ["value" => $this->fields['caller_location']]);
+      Dropdown::showFromArray("requester_field", [
+         self::REQUESTER_FIELD_FROM => __('From'),
+         self::REQUESTER_FIELD_REPLY_TO => __('Reply-To')
+      ], ["value" => $this->fields['requester_field']]);
       echo "</td></tr>\n";
 
       echo "<tr class='tab_bg_1'><td>".__('Comments')."</td>";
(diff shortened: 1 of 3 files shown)


install/mysql/glpi-empty.sql
```diff
@@ -4292,6 +4292,7 @@ CREATE TABLE `glpi_mailcollectors` (
   `errors` int(11) NOT NULL DEFAULT '0',
   `use_mail_date` tinyint(1) NOT NULL DEFAULT '0',
   `date_creation` datetime DEFAULT NULL,
+  `caller_location` enum('from', 'reply-to') NOT NULL DEFAULT 'from',
```
orthagh: Don't use enum, two possibility:
- integer with const in MailCollector class for value (prefered method)
- direct varchar

orthagh:
BTW, functional tests : OK

install/update_93_94.php
```diff
@@ -61,6 +61,12 @@ function update93to94() {
    }
    /** /Add default group for a user */
 
+   /** Add caller location on glpi_mailcollectors */
+   if (!$DB->fieldExists('glpi_mailcollectors', 'requester_field')) {
+      $migration->addField("glpi_mailcollectors", "requester_field", "int(11) NOT NULL DEFAULT '0'");
```
trasher: You should use "integer" automatic type instead of full  declaration

inc/mailcollector.class.php
```diff
@@ -1286,6 +1298,7 @@ function getHeaders($uid) {
 
       $sender       = $mail_header->from[0];
       $to           = $mail_header->to[0];
+      $reply_to     = $mail_header->reply_to[0];
```
trasher: Hum... What happens here if there is no reply to in headers?

tomolimo: I totally agree with @trasher 
I'm not sure that email clients provide this field in a easy to use way.
And more than that, my email client gives the possibility to insert several Reply-to persons....See below:

```
Received: from XXXXXXXX ([10.10.0.189]) by
 YYYYYYYYY ([10.10.0.245]) with mapi id 14.03.0382.000; Thu, 6
 Sep 2018 08:45:45 +0200
Content-Type: application/ms-tnef; name="winmail.dat"
Content-Transfer-Encoding: binary
From: "trucmuche" <trucmuc@machinchose>
To: helpdesk <glpi@machinchose>
Subject: FW: [glpi-project/glpi] Collector : can now specify where the
 requester's email is found (#4554)
Thread-Topic: [glpi-project/glpi] Collector : can now specify where the
 requester's email is found (#4554)
Thread-Index: AQHURappg+gmjzKyfkOia/og9Uucl6TiziHA
Disposition-Notification-To: "trucmuche" <trucmuche@machinchose>
Date: Thu, 6 Sep 2018 08:45:45 +0200
Message-ID: <080DF555E8A78147A053578EC592E83901258D1FAD@arexdb711.ar.ray.group>
References: <glpi-project/glpi/pull/4554@github.com>
 <glpi-project/glpi/pull/4554/review/152786541@github.com>
In-Reply-To: <glpi-project/glpi/pull/4554/review/152786541@github.com>
Reply-To: "trucmuche" <trucmuche@machinchose>, "bidule"
	<bidule@machinchose>
Accept-Language: en-US
Content-Language: en-US
X-MS-Has-Attach:
X-MS-Exchange-Organization-SCL: -1
X-MS-TNEF-Correlator: <080DF555E8A78147A053578EC592E83901258D1FAD@XXXXXXXXXXX>
MIME-Version: 1.0
X-MS-Exchange-Organization-AuthSource: YYYYYYYYYYYY
X-MS-Exchange-Organization-AuthAs: Internal
X-MS-Exchange-Organization-AuthMechanism: 03
X-Originating-IP: [10.19.69.87]
X-EsetResult: clean, is OK
X-EsetId: 16EFCE3994A44B3842A890
```

AdrienClairembault: After some testing it appears that the 'reply-to' value is always set in $mail_header even if the original mail doesn't have a 'reply-to' header (in this case it will be a copy of the 'from' value). 

From a comment on the function doc http://php.net/manual/en/function.imap-rfc822-parse-headers.php#89758 : 

> Please be aware that the result will always contain fields that have not been set in the message. For example reply_to and sender will always be set even if there is no Reply-To and Sender header fields are defined in the message.

trasher: That's fine for me

PR #4570, opened 2018-09-04
fix negative install date

mzulfahmy:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | no
| Fixed tickets | #4568


tsmr:
For me it's not the good solution. The problem came from OCS. It's resolved in 1.5.3 release of the plugin.

https://github.com/pluginsGLPI/ocsinventoryng/blob/master/inc/software.class.php#L149

PR #4586, opened 2018-09-06
fix location of local_define.php

jnsc:
According to the documentation on https://glpi-install.readthedocs.io/fr/latest/install/ a local_define.php file can be placed in the GLPI_CONFIG_DIR. This fixes the issue.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #number

### DIFF
--- a/inc/based_config.php
+++ b/inc/based_config.php
@@ -53,7 +53,7 @@ if (!defined("GLPI_CONFIG_DIR")) {
 }
 
 // If this file exists, it is load
-if (file_exists(GLPI_CONFIG_DIR. '/local_define.php')) {
+if (file_exists(GLPI_CONFIG_DIR. '/local_define.php') && !defined('TU_USER')) {
    require_once GLPI_CONFIG_DIR. '/local_define.php';
 }
 
--- a/inc/based_config.php
+++ b/inc/based_config.php
@@ -42,11 +42,6 @@ if (!empty($tz)) {
    date_default_timezone_set(@date_default_timezone_get());
 }
 
-// If this file exists, it is load
-if (file_exists(GLPI_ROOT. '/config/local_define.php') && !defined('TU_USER')) {
-   require_once GLPI_ROOT. '/config/local_define.php';
-}
-
 // If this file exists, it is load, allow to set configdir/dumpdir elsewhere
 if (file_exists(GLPI_ROOT . '/inc/downstream.php')) {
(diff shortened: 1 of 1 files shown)


inc/based_config.php
```diff
@@ -57,6 +52,11 @@
    define("GLPI_CONFIG_DIR", GLPI_ROOT . "/config");
 }
 
+// If this file exists, it is load
+if (file_exists(GLPI_CONFIG_DIR. '/local_define.php')) {
```
trasher: Miss second part of original condition

jnsc: You are right, I missed it as it's not included in the release I installed 9.3, but a git blame showed me that you added this condition, so if it resolve the problem, thats fine for me.

remicollet:
-1

local_define must be load first, before downstream which only defines undefined value.
Once downstream is loaded, local_defined can't change anything.

Ex: 
* /etc/glpi/local_define.php

`define('GLPI_DOC_DIR',        '/var/lib/glpi/docs');`

* /usr/share/glpi/inc/downstream.php

`defined('GLPI_VAR_DIR')        or define('GLPI_VAR_DIR',        '/var/lib/glpi/files');`

* /usr/share/glpi/inc/based_config.php

```
if (!defined("GLPI_DOC_DIR")) {
   define("GLPI_DOC_DIR", GLPI_VAR_DIR);
}
```

remicollet:
More information:

* **local_define** can define everything excepted **config** directory (obviously)
* if **downstream** redirect **config** directory, it also have to take care of loading local_define from the new location

trasher:
thanks @remicollet that's why I've requested your review. But one issue remains: install docs says exactly the inverse, cannot remember who's right here (it has changed several times).
If this PR is rejected, docs must be fixed. Any thoughs?

trasher:
> it also have to take care of loading local_define from the new location

Hum.. Maybe should it be automated if CONF_DIR differs from default? Anyways, I'll update docs to reflect that

remicollet:
Indeed, example of downstream.php is not correct (while above example is)

local_define loaded in base_config is only for standard layout, without any redirection (.../glpi/config/...)

Common user should only play with .../glpi/config/local_define.php
IMHO, Only packager should play with inc/downstream.php

trasher:
See https://github.com/glpi-project/doc-install/pull/28

PR #4601, opened 2018-09-10
Drop useless menu entries

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

Cannot find where those are used.


trasher:
Used in breadcrumbs.

PR #4621, opened 2018-09-12
Replace get type items

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes


inc/document_item.class.php
```diff
@@ -941,6 +941,109 @@ static function getRelationMassiveActionsSpecificities() {
...
+    * Get items for an itemtype
+    *
+    * @since 9.3.1
+    *
+    * @param integer $items_id Object id to restrict on
+    * @param string  $itemtype Type for items to retrieve
+    *
+    * @return DBMysqlIterator
+    */
+   public static function getTypeItems($items_id, $itemtype) {
```
cedric-anne: Could you explain why it was not possible to override the `getListForItem` and/or 'getListForItemParams' function  instead of redefining about 100 lines of code ?

inc/commondbrelation.class.php
```diff
@@ -1569,8 +1572,25 @@ protected static function getListForItemParams(CommonDBTM $item, $noent = false)
       $link = new $link_type;
```
cedric-anne: Still there is still cases where `$link_type` is not the name of a CommonDBTM class, we should throw an exception if `!is_a($link_type, CommonDBTM::class, true)`.

cedric-anne:
Some global remarks :
 - No need to describe default value of method parameters, it will be automatically generated using the one existing in method declaration.
 - You should add `array` prototype before new `$where` parameters declaration -> `array $where = []`.
 - Added methods should be tagged as `@since 9.3.2`.

inc/document_item.class.php
```diff
@@ -941,6 +941,109 @@ static function getRelationMassiveActionsSpecificities() {
...
+    *
+    * @param integer $items_id Object id to restrict on
+    * @param string  $itemtype Type for items to retrieve
+    * @param boolean $noent    Flag to not compute entity informations (see Document_Item::getTypeItemsQueryParams)
+    * @param array   $where    Inital WHERE clause. Defaults to []
+    *
+    * @return array
+    */
+   protected static function getMainTypeItemsQueryParams($items_id, $itemtype, $noent = false, $where = []) {
+      global $DB;
```
cedric-anne: Most of this part could be removed, as result is identical to `self::getListForItemParams()` result. The only different part is that entities table is  left joined here and inner joined in `self::getListForItemParams()`.

```diff
diff --git a/inc/document_item.class.php b/inc/document_item.class.php
index eab6c856e..e0179d4ad 100644
--- a/inc/document_item.class.php
+++ b/inc/document_item.class.php
@@ -954,62 +954,13 @@ class Document_Item extends CommonDBRelation{
     * @return array
     */
    protected static function getMainTypeItemsQueryParams($items_id, $itemtype, $noent = false, $where = []) {
-      global $DB;
-
-      $item = getItemForItemtype($itemtype);
-      $order_col = $item->getNameField();
-
-      if ($item instanceof CommonDevice) {
-         $order_col = "designation";
-      } else if ($item instanceof Item_Devices) {
-         $order_col = "itemtype";
-      } else if ($itemtype == 'Ticket') {
-         $order_col = 'id';
-      }
 
-      if (!count($where)) {
-         $where = [static::getTable() . '.' . static::$items_id_1  => $items_id];
-      }
+      $item = new Document();
+      $item->getFromDB($items_id);
 
-      $link_table = $item->getTable();
-      $link_id    = 'items_id';
-
-      $params = [
-         'SELECT'    => [
-            static::getTable() . '.id AS linkid',
-            $link_table . '.*'
-         ],
-         'FROM'      => static::getTable(),
-         'LEFT JOIN' => [
-            $link_table => [
-               'FKEY' => [
-                  static::getTable()   => $link_id,
-                  $link_table          => 'id'
-               ]
-            ]
-         ],
-         'WHERE'     => $where,
-         'ORDER'     => $link_table . '.' . $order_col
-      ];
-
-      $rel_class = static::class;
-      $rel = new $rel_class();
-      if ($rel->maybeDynamic()) {
-         $params['SELECT'][] = static::getTable() . '.is_dynamic';
-      }
-
-      if ($rel->maybeRecursive()) {
-         $params['SELECT'][] = static::getTable() . '.is_recursive';
-      }
-
-      if ($DB->fieldExists(static::getTable(), 'itemtype')) {
-         $params['WHERE'][static::getTable() . '.itemtype'] = $itemtype;
-      }
-
-      if ($item->maybeTemplate()) {
-         $params['WHERE'][$link_table . '.is_template'] = 0;
-      }
+      $params = self::getListForItemParams($item, true, $itemtype, $where);
 
+      $link_table = getTableForItemtype($itemtype);
       if ($noent === false && $item->isEntityAssign() && $itemtype != Entity::getType()) {
          $params['SELECT'][] = 'glpi_entities.id AS entity';
          $params['LEFT JOIN']['glpi_entities'] = [
```

inc/commondbrelation.class.php
```diff
@@ -1538,29 +1538,32 @@ static function processMassiveActionsForOneItemtype(MassiveAction $ma, CommonDBT
...
       $inverse = $item->getType() == static::$itemtype_1;
 
       $link_type  = static::$itemtype_1;
       $link_id    = static::$items_id_1;
       $where_id   = static::$items_id_2;
+      $item_type  = $item->getType();
 
       if ($inverse === true) {
          $link_type  = static::$itemtype_2;
          if ($link_type == 'itemtype') {
```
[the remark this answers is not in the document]
trasher: Yes there is something wrong... I see that exact case recently, but I'm still not sure what to do when `itemtype_1` is `itemtype` (this appears only twice in the core :( )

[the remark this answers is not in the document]
trasher: Note that this *may* also be true in current 9.3 codebase, I'll have to check.

trasher:
Closing for now, seems too much work for the moment, that may not be really useful in the future

Issue #1962, opened 2017-03-21
Like computer, add device for network and printer in search engine


yllen:
http://forum.glpi-project.org/viewtopic.php?id=157179

yllen:
http://forum.glpi-project.org/viewtopic.php?id=157180

PR #4656, opened 2018-09-18
Add component search options to printers and network devices

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1962


trasher:
Tests are broken. I guess you took search options (and theirs ID) from computer class?
I was about to say those ones should be factorized in a `getSearchOptionsToAdd` method; but we'll have to take care of the duplicate id before :(

trasher:
Hi,

I've opened another pull request with a more reliable way to go (factorizing search options among child classes). 

I've based my work/reflexion on what you've proposed; thank you!

Issue #4635, opened 2018-09-13
Small bugs in /ajax/comments.php and /inc/dropdown.class.php


loicbenoit:
* Steps to reproduce (which actions have you made) :
PHP errors that happen when using custom dropdowns in custom objects created with plugin GenericObject.

* Expected result :
No PHP error for "class not found because plugin autoloader was not loaded" or "trying to access an array item on non-array".

* Actual result :
PHP errors

* See proposed fixes in pull-request.

trasher:
Hello,

You should open your pull request here and not on your own fork :) Also, since this is a bug, you must target 9.3/bugfixes branch.

Thank you

loicbenoit:
Thanks, I will move my PR to 9.3/bugfixes branch.

About using my own fork of GLPI, I'm asking for some leniency from the community, as I'm required to do so for my time spent on this...

trasher:
I do not understand. 

You can use your fork to do the work (in facts, we encourage you to do so).

But when you want your fix to reach GLPI; you have to open a pull request from your fork to GLPI repository, and you opened a PR from your fork to your fork.

loicbenoit:
These 2 level of forking is the quirk I was refering to.
The fork from Ville de Montreal will make the official pull request to GLPI.

trasher:
OK... I'll wait the PR to be opened here.

PR #4670, opened 2018-09-19
9.3/bugfixes pr2

marckhouzam:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #4635

Handle case when the dropdown class is not found and $tmpname is not an array. This makes GLPI more robust even though the situation arises because of a plugin.

Use case: Creating custom dropdowns with plugin GenericObject. Autoloaders from plugins are not available when running some GLPI ajax scripts, such as /ajax/comments.php.


marckhouzam:
The developper that developed this fix is @loicbenoit.  He will address any needed modifications.

trasher:
We had a problem with CS checks, so I've rebased your PR and fix CS issues.

marckhouzam:
Thanks!

Issue #4685, opened 2018-09-21
[9.3.1] Used/Managed items takes into account only the last group


lufik:
* Root cause of the problem in php code:
```
      $iterator = $DB->request([
         'SELECT'    => [
            'glpi_groups_users.groups_id',
            'glpi_groups.name'
         ],
         'FROM'      => 'glpi_groups_users',
         'LEFT JOIN' => [
            'glpi_groups' => [
               'FKEY' => [
                  'glpi_groups_users'  => 'groups_id',
                  'glpi_groups'        => 'id'
               ]
            ]
         ],
         'WHERE'     => ['glpi_groups_users.users_id' => $ID]
      ]);
      $number = count($iterator);

      $group_where = [];
      while ($data = $iterator->next()) {
         $group_where[$field_group] = $data['groups_id'];
         $groups[$data["groups_id"]] = $data["name"];
      }
```
As you can see the $group_where[$field_group] is rewritten in every cycle so only the last value is used later in queries.
I'm going to find how to fix it following new code (there were string concat of OR every group in 9.2).

* Steps to reproduce (which actions have you made) :
Assign user to groups and assign different asset(s) to different groups.
Go to: Administration -> Users -> choose the user -> go to the tab Used items.

* Expected result :
See all the assets (from all user's groups).

* Actual result :
See the assets only from the one (last) group.

* URL of the page :
https://glpi.devel.invasys.org/front/user.form.php?id=7

* Your GLPI setup (you can find it in Setup > General menu, System tab) :
See #4682

lufik:
The fix is really easy. Someone just forget to write [] so the name => value is added instead of replaced. I created another pull request.

PR #4688, opened 2018-09-21
Bug fix for multiple groups for used/managed items

lufik:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #4685 

Bug fix for Issue #4685 to 9.3/bugfix branch.

### DIFF
--- a/inc/user.class.php
+++ b/inc/user.class.php
@@ -3914,7 +3914,7 @@ class User extends CommonDBTM {
 
       $group_where = [];
       while ($data = $iterator->next()) {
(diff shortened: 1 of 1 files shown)


lufik:
BTW even with 1 value it uses IN (e.g. "field in ('1')" ) but IMHO this can be improved in the dbmysqliterator class.

trasher:
Don't think this is an iterator issue; but anyways I'm not sure that change anything

PR #4709, opened 2018-09-26
Bug fix for Computer -> Connections ignore is_deleted flag

lufik:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #4705 

Bug fix for Issue #4705 to 9.3/bugfix branch.

### DIFF
--- a/inc/computer_item.class.php
+++ b/inc/computer_item.class.php
@@ -750,7 +750,7 @@ class Computer_Item extends CommonDBRelation{
                    || Peripheral::canView()
                    || Monitor::canView()) {
                   if ($_SESSION['glpishow_count_on_tabs']) {
-                     $nb = self::countForMainItem($item);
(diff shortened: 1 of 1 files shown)


trasher:
I do not agree with the proposal. Deleted items are here displayed with a different BG color, which is intended.
That was maybe not the case in older releases, this was very inconsistent (sometimes, deleted were displayed, sometimes not...).

lufik:
Ok, I'm going to change the pull request so Monitor -> Connections will use the same logic as Computer -> Connections.

lufik:
I wait for the decision how to fix the problem (in issue #4705).

trasher:
i'm closing this one since this is not OK for the moment; you can reopen once this will be ok. Thank you

Issue #4717, opened 2018-09-27
fatal error 500 with API + LDAP + user token


btry:
Hi

Here is the backtrace shared by a community user.

```
[Thu Sep 27 14:54:59.078312 2018] [:error] [pid 10799] [client 10.251.14.140:53404] 
PHP Fatal error:  Uncaught exception 'RuntimeException' with message 'Something went wrong searching in LDAP directory' in /var/www/html/glpi3/inc/authldap.class.php:2839\n
Stack trace:\n
#0 /var/www/html/glpi3/inc/auth.class.php(670): AuthLDAP::searchUserDn(Resource id #7, Array)\n
#1 /var/www/html/glpi3/inc/api.class.php(229): Auth->login('', '', false, false, '')\n
#2 /var/www/html/glpi3/inc/apirest.class.php(135): API->initSession(Array)\n
#3 /var/www/html/glpi3/apirest.php(45): APIRest->call()\n
#4 {main}\n  thrown in /var/www/html/glpi3/inc/authldap.class.php on line 2839
```

There is an uncaught exception.

PR #4718, opened 2018-09-27
Add missing try / catch block

btry:
When trying to login from the A1PI with a user token, when GLPI has LDAP authentication sources,
AuthLdap::searchUserDn() fails to find the user by login or by DN and raises an exception

Signed-off-by: Thierry Bugier <tbugier@teclib.com>

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #4717 

### DIFF
--- a/inc/auth.class.php
+++ b/inc/auth.class.php
@@ -657,16 +657,21 @@ class Auth extends CommonGLPI {
                            AuthLdap::IDENTIFIER_LOGIN => $ldap_method["login_field"],
                         ],
                      ];
-                     $user_dn
-                        = AuthLdap::searchUserDn($ds,
-                                                 ['basedn'      => $ldap_method["basedn"],
-                                                       'login_field' => $ldap_method['login_field'],
-                                                       'search_parameters'
-                                                                     => $params,
-                                                       'user_params'
-                                                         => ['method' => AuthLDAP::IDENTIFIER_LOGIN,
-                                                                  'value'  => $login_name],
-                                                         'condition'   => $ldap_method["condition"]]);
+                     try {
+                        $user_dn
+                           = AuthLdap::searchUserDn($ds,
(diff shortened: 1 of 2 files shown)


inc/authldap.class.php
```diff
@@ -2757,6 +2757,7 @@ static function tryLdapAuth($auth, $login, $password, $auths_id = 0, $user_dn =
     *          - condition : ldap condition used
     *
     * @return array|boolean dn of the user, else false
+    * @throws \RuntimeException
```
btry: i would suggest a specific class instead of \RuntimeException ...

inc/auth.class.php
```diff
@@ -657,16 +657,20 @@ function login($login_name, $login_password, $noauto = false, $remember_me = fal
...
+                                                   ['basedn'      => $ldap_method["basedn"],
+                                                         'login_field' => $ldap_method['login_field'],
+                                                         'search_parameters'
+                                                                        => $params,
+                                                         'user_params'
+                                                            => ['method' => AuthLDAP::IDENTIFIER_LOGIN,
+                                                                     'value'  => $login_name],
+                                                            'condition'   => $ldap_method["condition"]]);
+                     } catch (\RuntimeException $e) {
+                        $user_dn = false;
```
trasher: I think nothing gets logged as is.

btry: We had this case with Flyve MDM. 

I need to do tests to find if we need to update our  code to provide an extra argument to our call ro tje api endpoint initSession.

But it is clear that something must catch the possible exception thrown by this call

trasher: I do not understand. I'm just saying you should call a logError or something similar (see similar try/catch in authldap class).

btry: Done

PR #4728, opened 2018-10-02
use selinux extension when available, avoid terrible exec calls

remicollet:


remicollet:
> Should target 9.3/bf

master is fine for me, as you prefer.

PR #4734, opened 2018-10-03
Search: nested criteria

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | maybe 

Purpose:
- metacriteria can now be set as normal criteria (with a meta flag). This permits to have meta in any place of the where (not only at the end)
- criteria can now contains criteria, and sql query includes this sub criteria with parenthesis
- 1st criteria can now be set as NOT

TODO:

- [x] Dockblocks
- [x] Tests on criteria with meta flag
- [x] Test on nested criteria
- ~~[ ] I would like to remove static keywords for sql construction (WHERE, HAVING, JOIN etc as class property) to simplify methods declaration~~ will do later
- [x] Check bookmarks (~~it should be ok~~ it's not :D )
- [x] display folding (currently removed)
- [x] align first dropdown of meta with normal criteria
- [x] nested criteria add is_deleted,  as_map & itemtype parameters in form, we need to remove them
- [x] Customer order: add new action "notcontains"
- [x] Customer order: add metacriteria to changes
- [x] Bug: metacriteria button still available when it should not


ajax/search.php
```diff
@@ -0,0 +1,76 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with GLPI. If not, see <http://www.gnu.org/licenses/>.
+ * ---------------------------------------------------------------------
+ */
+
+/**
+ * @since 0.85
```
trasher: Pretty sure not :D

inc/savedsearch.class.php
```diff
@@ -923,7 +925,7 @@ private function displaySavedSearchType($searches) {
                   Toolbox::logError($e);
                   $data = false;
                }
-               if ($data) {
+               if (isset($data['data']['totalcount'])) {
```
trasher: Won't this fail if `$data === false`?

inc/search.class.php
```diff
@@ -2113,104 +2241,752 @@ static function showGenericSearch($itemtype, array $params) {
...
+   /**
+    * construct the default criteria for an itemtype
+    *
+    * @since 9.4
+    *
+    * @param  string $itemtype
+    *
+    * @return array  criteria
+    */
+   static function getDefaultCriteria($itemtype = '') {
```
trasher: Not sure it can be empty

trasher:
Just tested; this seems to works well (included new and existing bookmarks --well quite simple in my tests) :)

A small issue when you try to search on a form that has not been filled (just add rules and/or global rules and/or groups and send).

inc/savedsearch.class.php
trasher: `$data && isset($data['data']) && isset($data['data']['totalcount'])` :/

inc/search.class.php
```diff
@@ -6216,8 +6962,8 @@ static function showFooter($type, $title = "", $count = null) {
                $font       = $_SESSION['glpipdffont'];
                //$subsetting = false;
             }
-            $pdf->setHeaderFont([$font, 'B', 8]);
-            $pdf->setFooterFont([$font, 'B', 8]);
+            $pdf->setHeaderFont([$font, 'B', $fonsize]);
```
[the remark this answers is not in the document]
trasher: Yes I know ;)

PR #4749, opened 2018-10-08
display softwares in computer only for the current entity

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

Match the count of the tab.
internal ref: 15015


orthagh:
need to to complete  pr with software side.

Issue #3609, opened 2018-02-23
Emotes in emails


Hevelious:
When an user sends ticket by email, the description is blanked out after an emote is used.
------------
* Steps to reproduce (which actions have you made) :
Send email with emote f.e. :)

* Expected result :
Emote is still there, description intact.

* Actual result :
Text after emote is removed.

* URL of the page :
Any ticket.

* Screenshot of the problem (if pertinent) :
Can't show  because of ticket content. User :) this text disappears.

* Your GLPI setup (you can find it in Setup > General menu, System tab) :
GLPI 9.1.6
php 7.0.25
Ubuntu 16.04.1
Apache 2.4.18
MySQL 5.7.21

orthagh:
Hello @Hevelious 

I just tested with 9.2/bugfixes, here the results:
![image](https://user-images.githubusercontent.com/418844/36681796-e4e94e92-1b19-11e8-970f-5405fa225f83.png)

So it seems the code works.

By the way, you are using a non-supported version (9.1.6) and the bug may be already fixed, please check with last version 9.2.1 (or 9.2.2, it will be released this week)

Hevelious:
We will upgrade to 9.2.2 when released, will give info after that.

Hevelious:
We just upgraded our production, the bug still exists.

![1](https://user-images.githubusercontent.com/25429596/37787579-9e6c2cde-2dff-11e8-8f3b-f2027c0187b4.png)
![2](https://user-images.githubusercontent.com/25429596/37787580-9e885e54-2dff-11e8-9e65-8abedee3d160.png)
![3](https://user-images.githubusercontent.com/25429596/37787581-9ea82b6c-2dff-11e8-8f22-9f2c19ac9a52.png)

Hevelious:
The screen capture shows email notification.
The result is the same in GLPI.

When I create a ticket by clicking it trough GUI everything works correctly.
The issue manifests itself when I send the same text with emotes via email.

Fchallier:
I have the same plrobleme here... 
Even if i activate the riche texte in the setup. 

It's impossible tu put an emotes  in the description of the ticket.

trasher:
Just tested mail creation from a collector; and I cannot reprduce any issue:
- txt mail without rich text:
![screenshot-2018-5-3 glpi - tickets - 1101](https://user-images.githubusercontent.com/224733/39562270-ff48409e-4eaa-11e8-9394-6f42b69109a9.png)
[...]

PR #4765, opened 2018-10-11
Fix truncated data with emojis on mail

stonebuzz:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #3609

When an email is collected by GLPI
 if it contains an emoji, the description of the ticket / followup is truncated from the emojis.

This PR encode emojis into a shortcode to be properly saved in database
```php
echo LitEmoji::encodeShortcode('Very hard ticket 🔥');

// 'Very hard ticket :fire:'
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,7 +32,8 @@
         "paragonie/random_compat": "^2.0",
         "monolog/monolog": "^1.23",
         "sebastian/diff": "^1.4 || ^2.0 || ^3.0",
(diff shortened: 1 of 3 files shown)


hisakiyo:
It would be nice to be able to see emojis from GLPI interface instead of seeing text like ```:blush: ```

PR #4784, opened 2018-10-14
Allow locking of personalizations

### LINKED ISSUE #1062: Permit to lock palettes for user preferences
The user can change the palette which adminstrator has defined. 

It's a good idea to lock this feature for the anothers users

see http://glpi.userecho.com/topics/74-permit-to-lock-palettes-for-user-preferences/#comment-631

cconard96:
closes #1062

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1062

Allows locking of user personalizations to the defaults set by GLPI administrator.

### DIFF
--- a/inc/profile.class.php
+++ b/inc/profile.class.php
@@ -48,7 +48,7 @@ class Profile extends CommonDBTM {
                                           'reservation', 'rssfeed_public',
                                           'show_group_hardware', 'task', 'ticket',
                                           'tickettemplates_id', 'ticket_cost',
-                                          'ticketvalidation', 'ticket_status'];
+                                          'ticketvalidation', 'ticket_status','personalization'];
 
 
    /// Common fields used for all profiles type
@@ -93,6 +93,7 @@ class Profile extends CommonDBTM {
                   $ong[3] = __('Assistance'); // Helpdesk
                   $ong[4] = __('Life cycles');
                   $ong[6] = __('Tools');
+                  $ong[8] = __('Setup');
                } else {
                   $ong[2] = __('Assets');
                   $ong[3] = __('Assistance');
@@ -152,7 +153,11 @@ class Profile extends CommonDBTM {
                break;
 
             case 8 :
-               $item->showFormSetup();
+               if ($item->fields['interface'] == 'helpdesk') {
+                  $item->showFormSetupHelpdesk();
+               } else {
+                  $item->showFormSetup();
+               }
                break;
          }
       }
@@ -1405,6 +1410,11 @@ class Profile extends CommonDBTM {
       $rights = [['itemtype'  => 'Config',
                             'label'     => __('General setup'),
                             'field'     => 'config'],
+                      ['rights'  => [
+                         READ    => __('Read'),
+                         UPDATE  => __('Update')],
+                        'label'  => __('Personalization'),
+                        'field'  => 'personalization'],
                       ['itemtype'  => 'DisplayPreference',
                             'label'     => __('Search result display'),
                             'field'     => 'search_config'],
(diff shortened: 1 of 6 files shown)


cconard96:
Makes sense. I'll modify this PR with the requested change.

install/mysql/glpi-empty.sql
```diff
@@ -2665,6 +2666,11 @@ INSERT INTO `glpi_displaypreferences` VALUES ('261','SavedSearch','9','1','0');
 INSERT INTO `glpi_displaypreferences` VALUES ('262','SavedSearch','3','1','0');
 INSERT INTO `glpi_displaypreferences` VALUES ('263','SavedSearch','10','1','0');
 INSERT INTO `glpi_displaypreferences` VALUES ('264','SavedSearch','11','1','0');
+INSERT INTO `glpi_displaypreferences` VALUES ('265','Plugin','4','1','0');
```
trasher: Pretty sure this should not be present ;)

install/mysql/glpi-empty.sql
```diff
@@ -1373,6 +1373,7 @@ INSERT INTO `glpi_configs` VALUES ('206','core','purge_all','0');
 INSERT INTO `glpi_configs` VALUES ('207','core','purge_user_auth_changes','0');
 INSERT INTO `glpi_configs` VALUES ('208','core','purge_plugins','0');
 INSERT INTO `glpi_configs` VALUES ('209','core','display_login_source','1');
+INSERT INTO `glpi_configs` VALUES ('210','core','lock_palette','0');
```
trasher: You can drop this one

trasher:
Than you.

Some remarks:
- you have a little CS issue,
- updating a master database does not add new right on existing profiles, it has created 8 new profiles named "personalization" :D
- user have been able to update the personalization tab with READ and no UPDATE.

Another thing, maybe the lang dropdown can be moved to the first tab as well; I'm not sure it is a good idea to get that locked ;-)

cconard96:
Language has been present on the both tabs previously. You can update the language on the first tab even without personalization rights.

trasher:
Update is still not ok (tests are all failing because of that)

trasher:
You're right about the language... It's present in the 2nd tab only for the global configuration I guess.

cconard96:
Not sure what is broken with update. I can't test locally right now because I cannot get the test DB installed.

I check again tonight

install/update_93_94.php
```diff
@@ -290,6 +290,19 @@ function update93to94() {
    $migration->addField("glpi_changevalidations", "timeline_position", "tinyint(1) NOT NULL DEFAULT '0'");
    $migration->addField("glpi_problemtasks", "timeline_position", "tinyint(1) NOT NULL DEFAULT '0'");
 
+   /** Give all existing profiles access to personalizations for legacy functionality */
```
orthagh: The whole piece of code can be replaced by
```suggestion
   /** Give all existing profiles access to personalizations for legacy functionality */
   $migration->addRight('personalization', READ | UPDATE);
```

inc/profile.class.php
```diff
@@ -1479,6 +1489,48 @@ function showFormSetup($openform = true, $closeform = true) {
    }
 
 
+   /**
+    * Print the helpdesk form for a profile
+    *
+    * @param $openform     boolean  open the form (true by default)
```
trasher: You've missed the `@since 9.4.0` tag, and the `@return`. For other params, syntax is `@param type $name A bit of documentation` ;)

PR #4785, opened 2018-10-15
Fix SQL dump backup

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | --

Seems to have broke in #3942 when switching to iterator. Applicable to master only, not 9.3/bugfixes.

Error:
screenshot from 2018-10-14 23-06-04

### DIFF
--- a/front/backup.php
+++ b/front/backup.php
@@ -142,19 +142,18 @@ function get_content($DB, $table, $from, $limit) {
 
    $content = "";
 
-   $result = $DB->request($table, ['START' => $from, 'LIMIT' => $limit]);
+   $iterator = $DB->request($table, ['START' => $from, 'LIMIT' => $limit]);
 
-   if ($result) {
-      $num_fields = $DB->num_fields($result);
+   if ($iterator->count()) {
 
-      while ($row = $DB->fetch_row($result)) {
+      while ($row = $iterator->next()) {
          $insert = "INSERT INTO `$table` VALUES (";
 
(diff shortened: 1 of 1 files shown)


front/backup.php
```diff
@@ -586,7 +585,7 @@ function backupMySql($DB, $dumpFile, $duree, $rowlimit) {
    echo "<br/>" . __('You should rather use a dedicated tool on your server.');
    echo "</li></ul></div>";
    echo "</td></tr><tr><td>";
-   echo "<img src='".$CFG_GLPI["root_doc"]."/pics/sauvegardes.png' alt=\"".__s('Deleted')."\">".
```
trasher: Is this image still used? It can be removed if not.

cconard96: I had no idea why it is there. It appears to be decorative, but has a tooltip of 'Deleted'. It's also used in the Rules menu but with a tooltip 'Users'. I could remove or replace with Font Awesome save icon.

trasher: Just remove the `alt` here, there is no alt on `i` tags; and it should have been empty anyways (decorative images must have an empty `alt`)... Do not change anything else , it's out of the scope of the current pull request ;)

trasher: OK, I'll do the simage change on my side for 9.3/bf and will rebase your PR once done. you do not need to worry about that, thanks

Issue #4799, opened 2018-10-17
add device firmware in search engine for a computer


yllen:
https://forum.glpi-project.org/viewtopic.php?id=163430

buggzbeny:
Hello,
We need this research too, bios and bios date. Can you add them ?
Regards.

Issue #4689, opened 2018-09-21
Remove hard coded duplicate type names


trasher:
Back to #2681 ; we've added type name as a prefix on search options for sub items.

While this fixes the orginal issue; this produces a weird display; and this cause item type to be duplicated in the search results list (append from search engine).

We have to remove hard coded type names; and find another solution to solve the original issue.
Maybe this may be of interest: https://select2.org/searching#matching-grouped-options

PR #4800, opened 2018-10-17
add device firmware in search engine of a computer - see #4799

yllen:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #4799

### DIFF
--- a/inc/computer.class.php
+++ b/inc/computer.class.php
@@ -1011,6 +1011,20 @@ class Computer extends CommonDBTM {
          ]
       ];
 
+      $tab[] = [
+      'id'                 => '96',
+      'table'              => 'glpi_devicefirmwares',
+      'field'              => 'designation',
+      'name'               => __('Firmware'),
+      'forcegroupby'       => true,
+      'usehaving'          => true,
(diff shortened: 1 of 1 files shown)


trasher:
This is not really ok since firmwares can be attached to other assets types than computers.

Please also note you should not add prefix on search options (see https://github.com/glpi-project/glpi/issues/4689).

Also, this is not ok for master: changes has been done, and components search options must now be handled in component class (see https://github.com/glpi-project/glpi/commit/cba3cf582bb8cceea5409bd17510812416cdd61d). They'll be automatically added to all assets linked from configuration.

yllen:
For NetworkEquipment firmware version was defined in rawSearchOptions
For Peripheral firmware version is not defined in searchOptions, for printer too (i didn't check all)
So i think you define firmware version only when asked otherwise it should have been defined everywhere and not just in network equipment.

For prefix, OK i dit not see this commit.

For master i saw it was different so i don't added need_cherrypick label.

PR #4802, opened 2018-10-18
Feature/search plugins

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | kind of
| Tests pass?   | yes

Several extra changes has been done as well to get that working:
- add a `noremove` option to search options, making impossible to remove required ones;
- `getSpecificValueToDisplay` method will now receive in its `$options` a `raw_data` entry containing current row full data (previously, only *current column* was sent).

In order to make this new `raw_data` entry; I had to change the way rows columns are indexed in the search (as this was not predictable; data was impossible to use outside of search loops themselves).
This is the most huge change here.

An alternative to all those changes would have been to maintain a mapping array; but that was not more easy to do and would be more complex to maintain in the future.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,11 @@ The present file will list all changes made to the project; according to the
 #### Changes
 - Plugins are now loaded in ajax files.
 - `TicketFollowup` has been replaced by `ITILFollowup`
+- `$num` parameter has been removed from several `Search` class methods:
+   - `addSelect()`,
+   - `addOrderBy()`,
+   - `addHaving()`,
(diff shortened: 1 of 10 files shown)


trasher:
Extra output has been fixed (a call I did not see first). I've changed icons (after FA has been updated on master) but for the colors... Well green and red (like enable button) is too much; and I do not really have proposals on my side.
Separate actions and messages is not really possible... Messages are echoed during checks that are made to display one action or another. Trying to get that separated would imply to run same checks twice, at least.

PR #4812, opened 2018-10-20
Subqueries in JOIN

btry:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | none

This PR is a proposal to allows to create subqueries in JOIN. Example in the unit tests.

Example of the need in Flyve MDM: https://github.com/flyve-mdm/glpi-plugin/pull/785

```php
      $result = $DB->request("SELECT `l1`.*
         FROM `$table`  AS `l1` INNER JOIN
         (SELECT  MAX(`date`) AS `maxdate`, `topic`
            FROM `$table`
            WHERE `itemtype` = '$itemtype' AND `items_id` = '$itemId'
            GROUP BY `topic`
         ) AS `l2`
         ON `l1`.`topic` = `l2`.`topic` AND `l1`.`date` = `l2`.`maxdate`
         WHERE `itemtype`='$itemtype' AND `items_id` = '$itemId'
         GROUP BY `topic`");
```

In this example I need to get a single row for the column topic, with the latest associated value. Without a subquery, mysql messes the rows and may give a value of the column message for an date which is not the highest. Using a subquery like this one is requires to have accurate results.

### DIFF
--- a/inc/dbmysqliterator.class.php
+++ b/inc/dbmysqliterator.class.php
@@ -211,7 +211,14 @@ class DBmysqlIterator implements Iterator, Countable {
                               break;
                         }
                         foreach ($val as $jointable => $joincrit) {
-                           $join .= " $jointype JOIN " .  DBmysql::quoteName($jointable) . " ON (" . $this->analyseCrit($joincrit) . ")";
+                           if (isset($joincrit['SUBQUERY'])) {
+                              $subquery = new QuerySubquery($joincrit['SUBQUERY']);
+                              unset($joincrit['SUBQUERY']);
+                              $alias = DBMysql::quoteName($jointable);
+                              $join .= " $jointype JOIN (" . $subquery->getSubQuery() . ") AS $alias ON (" . $this->analyseCrit($joincrit) . ")";
+                           } else {
+                              $join .= " $jointype JOIN " .  DBmysql::quoteName($jointable) . " ON (" . $this->analyseCrit($joincrit) . ")";
+                           }
                         }
                      } else {
                         trigger_error("BAD JOIN, value must be [ table => criteria ]", E_USER_ERROR);
--- a/inc/dbmysqliterator.class.php
+++ b/inc/dbmysqliterator.class.php
@@ -211,10 +211,12 @@ class DBmysqlIterator implements Iterator, Countable {
                               break;
                         }
                         foreach ($val as $jointable => $joincrit) {
-                           if (isset($joincrit['SUBQUERY'])) {
-                              $subquery = new QuerySubquery($joincrit['SUBQUERY']);
-                              unset($joincrit['SUBQUERY']);
+                           $joinExpressionKeys = ['FKEY' => null, 'ON' => null];
+                           $subquery = array_diff_key($joincrit, $joinExpressionKeys);
+                           if (count($subquery)) {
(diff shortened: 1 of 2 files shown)


inc/dbmysqliterator.class.php
```diff
@@ -194,12 +194,10 @@ function buildQuery ($table, $crit = "", $log = false) {
                      break;
 
                   case 'JOIN' : // deprecated
+                     Toolbox::deprecated('"JOIN" is deprecated, please use "LEFT JOIN" instead.');
```
trasher: I do not like this way; and as far as I remember this already cause difficult to understand issues.

trasher: I do not like this way; and as far as I remember this already cause difficult to understand issues.

trasher: I do not like this way; and as far as I remember this already cause difficult to understand issues.

trasher:
Not able to add a comment directly in the code... For the JOIN depreacated, I do not like this way; and as far as I remember this already cause difficult to understand issues. Please leave as is in 9.3.

inc/dbmysqliterator.class.php
trasher: I do not like this way; and as far as I remember this already cause difficult to understand issues. Please leave as is in 9.3.

btry:
I cleaned up duplicates of your comment.

btry:
I cleaned up duplicates of your comment; but there is something weird today.... the last one gone as well :/

inc/dbmysqliterator.class.php
```diff
@@ -215,7 +215,16 @@ function buildQuery ($table, $crit = "", $log = false) {
                               break;
                         }
                         foreach ($val as $jointable => $joincrit) {
-                           $join .= " $jointype JOIN " .  DBmysql::quoteName($jointable) . " ON (" . $this->analyseCrit($joincrit) . ")";
+                           $joinExpressionKeys = ['FKEY' => null, 'ON' => null];
```
btry: Might be an opportuniy to have  this array in a private property and mutualize it with this liine 
https://github.com/glpi-project/glpi/blob/9.3/bugfixes/inc/dbmysqliterator.class.php#L490

btry:
I updated the PR. Removed the refactor, and added an extra commit to not have the subquery and the FKEY in the same array. It looks smarter to me.

@trasher 

As dusussed, this is compatible with Postgres as it allows a derived table to be used for JOIN, at least since 7.2

inc/dbmysqliterator.class.php
trasher: This should be factorized, yes.

I'm do not know yet if I prefer your prévious proposal or this one :D But I'm globally OK with the change, thank you.

btry: the new one avoids the ugly key "SUBQUERY". This is smoother to write, and seems to me more natural and easier to learn.

trasher: Building queries from an array is not natural :D

trasher:
Could you please rebase your branch on master? Thank you :)

btry:
@trasher : done

Issue #4824, opened 2018-10-25
Bug in select profil in user tab


yllen:
https://forum.glpi-project.org/viewtopic.php?id=165879

yllen:
@orthagh  Look at my notes on your PR

PR #4826, opened 2018-10-25
fix retrieving of profiles under current one

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #4824

Since eef9748aff5


orthagh:
@yllen Could you check again with last commit ?

PR #4909, opened 2018-11-13
Modem component

btry:
Hi

This PR implements modem components in GLPI (it seems that something related to modems existed between GLPI 0.6 and GLPI 0.78)

The PR is still WIP : I need to cllect more precise data about modems in mobile devices. Those devices may have several simcards (up to 4), and those having several sim cards may have several modems.

Currently I implemetned the component with the ability to use 4 simcards with 4 IMEI. The component also stores the size of the simcard for each slot (some devices have 2 sim slots with 2 different sizes).

The component also contains a field for the version of its firmware.

### DIFF
--- a/inc/define.php
+++ b/inc/define.php
@@ -235,7 +235,7 @@ $CFG_GLPI["itemdevices"]                  = ['Item_DevicePowerSupply', 'Item_Dev
                                                   'Item_DeviceMotherBoard', 'Item_DeviceNetworkCard',
                                                   'Item_DeviceSoundCard', 'Item_DeviceControl',
                                                   'Item_DeviceHardDrive', 'Item_DeviceDrive', 'Item_DeviceMemory',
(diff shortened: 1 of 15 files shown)


inc/define.php
```diff
@@ -339,7 +341,7 @@
 
 $CFG_GLPI['itemdevicefirmware_types']     = ['Computer', 'Peripheral', 'Phone', 'NetworkEquipment', 'Printer'];
 
-$CFG_GLPI['itemdevicesimcard_types']      = ['Computer', 'Peripheral', 'Phone', 'NetworkEquipment', 'Printer'];
+$CFG_GLPI['itemdevicesimcard_types']      = $CFG_GLPI['itemdevicemodem_types'];
```
btry: simcards must be associated to the assets which can have a modem

trasher:
As far as I know, you do not have all required informations right now; PR canbe reopened once OK.

PR #4930, opened 2018-11-18
Multiple Search Folding Fixes

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

Use unique ids for more-criteria list items (Required for other fixes listed)
Add new criteria before the more-criteria list item (Fixes issues with deleting any criteria that wasn't last)
When removing criteria, make sure the first list item is still the headerRow and visible (Fix issue with deleting first criteria while folded)
Removed duplicate folding JavaScript (Fix issue trying to fold after adding a new group but before committing the search)

### DIFF
--- a/inc/search.class.php
+++ b/inc/search.class.php
@@ -2312,7 +2312,6 @@ JAVASCRIPT;
 
          $(document).on("click", ".remove-search-criteria", function() {
             var rowID = $(this).data('rowid');
-            console.log('remove-search-criteria', rowID);
             $('#' + rowID).remove();
             $('#searchcriteria ul li:first-child').addClass('headerRow').show();
          });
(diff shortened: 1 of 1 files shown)


inc/search.class.php
```diff
@@ -2406,7 +2409,8 @@ static function displayCriteria($request = []) {
             ]);
          }
          echo "<i class='far fa-minus-square' alt='-' title=\"".
-                  __s('Delete a rule')."\" onclick=\"$('#$rowid').remove();\"></i>&nbsp;";
+                  __s('Delete a rule')."\" onclick=\"$('#$rowid').remove();
+                  $('#searchcriteria ul li:first-child').addClass('headerRow').show();\"></i>&nbsp;";
```
orthagh: PR seems good for.
One point here and for the two following, i think we need to remove `onclick` attributes (i miss to do it) and write in a (existing?) jquery block with click events

PR #4938, opened 2018-11-20
Allow specify of creation and modification dates during add;closes #1731

### LINKED ISSUE #1731: Bypass creation and modification dates
In `CommonDBTM::add()`, it is not possible to change creation nor modification date. They're both automatically retrived from `$_SESSION['glpi_currenttime']`.

In `CommonDBTM::update()`, we can change both dates.

This should be made consistent, I propose to allow to specify dates at creation as well

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1731

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ The present file will list all changes made to the project; according to the
 - Add followups to Changes and Problems
 - Add timeline to Changes and Problems
 - Search on devices from Printers and Network equipments
+- Ability to specify creation and modification dates during CommonDBTM object add method
 
 ### Changed
 - `license_id` field in `glpi_items_operatingsystems` table has been renamed to `licenseid`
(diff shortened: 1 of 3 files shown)


trasher:
I'm unsure about this proposal.

Dates if they're passed in the `$input` array are silently ignored; I wonder if that may cause unexpected issues :/ @orthagh what is your opinion?

trasher:
And oh, wait for orthagh feedback first; but this will certainly need unit tests ;)

trasher:
I'm doing a few tests on the core; if all goes well, it's ok for me.

Maybe a line explaining that change may be added in the changelog?

trasher:
Oh, also, can you add a quick unit test on that? Thanks!

trasher:
Tested, see https://github.com/glpi-project/glpi/pull/4988

cconard96:
I may be wrong but $input[] is assigned to $this->input early in the add method. I am not sure how anything would be ignored...

I did test briefly on the split/merge PR work and it seemed to take dates in the array OK.

I will work on tests tonight.

PR #4957, opened 2018-11-23
send queue notification also when no mailing enabled

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes


trasher:
OK for me, but after the 9.3.3 release, since this can have impact (not all notification workflow is tested).

PR #4966, opened 2018-11-24
WIP : experiment - give input to canUpdateItem

btry:
Hi

In Flyve MDM there is a growing need to allow or forbid update of an itemtype depending on the data to update, the user ID compared against users_id foreign key and the current profile of the user.

the method canCreateItem() can access the payload, see CommonDBTM::can()
https://github.com/glpi-project/glpi/blob/9.3/bugfixes/inc/commondbtm.class.php#L2755

If item does not exists yet, the payload is added in $item->input, then canCreatep() and canCreateItem() are called. They may access $item->input and do some specific checks.

However this is not possible when we update an item. The idea is to ad  the same mechanism to allow itemtypes to do checks in the data from the DB (already accessible via $item->fields) and also check if some fields may be updated depending on current user context.

This allows to keep rights checks in can*() and can*Item() methods in one hand, and payload preparation in prepareInputForUpdate() in the other hand. 

First commits of the PR implement the need, and one of them changes many fields in front/ folder to pass the payload. This commit is useful to ensure that calls from the UI and from the API are done the same way. Please not that some checks on the UPDATE right are not updated because they are specific cases and need deeper analysis.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |

### DIFF
--- a/front/apiclient.form.php
+++ b/front/apiclient.form.php
@@ -43,7 +43,7 @@ if (isset($_POST["add"])) {
    Html::back();
 
 } else if (isset($_POST["update"])) {
-   $client->check($_POST["id"], UPDATE);
+   $client->check($_POST["id"], UPDATE, $_POST);
    $client->update($_POST);
    Html::back();
 
--- a/front/budget.form.php
+++ b/front/budget.form.php
@@ -88,7 +88,7 @@ if (isset($_POST["add"])) {
    $budget->redirectToList();
 
 } else if (isset($_POST["update"])) {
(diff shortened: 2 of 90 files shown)


trasher:
If you want to introduce changes, they must be discussed before.
We want to avoid stalled pull requests; and this one would stay stalled from a while since this has never been discussed.

Anyways, this is certainly too early for that kind of chnage; this is to be done once front files has been factorized.

PR #4978, opened 2018-11-27
add new rule criteria to check creation date is a working hour in a c…

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

### DIFF
--- a/inc/ticket.class.php
+++ b/inc/ticket.class.php
@@ -1822,11 +1822,12 @@ class Ticket extends CommonITILObject {
 
       // add calendars matching date creation (for business rules)
       $calendars = [];
-      foreach ($DB->request([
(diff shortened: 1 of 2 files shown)


inc/ticket.class.php
```diff
@@ -1820,6 +1820,23 @@ function prepareInputForAdd($input) {
          $input['_users_id_assign'] = Session::getLoginUserID();
       }
 
+      // add calendars matching date creation (for business rules)
+      $calendars = [];
+      foreach ($DB->request([
```
trasher: Maybe better not to call request in a loop

PR #5019, opened 2018-12-04
paste image at current cursor position in tinymce

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

Before this fix image was always pasted at the end of the editor due to a racing condition when storing pictures with placing cursor at end.
I also cleaned some superflu code

internal ref 15413

### DIFF
--- a/js/common.js
+++ b/js/common.js
@@ -730,24 +730,6 @@ var getSize = function (size) {
 };
 
 /**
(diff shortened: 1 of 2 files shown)


orthagh:
done

PR #5038, opened 2018-12-07
[WIP] Enhance map dipslay and link to item list, specially when have more t…

ddurieux:
…han 1 location with same GPS coordinates

### DIFF
--- a/ajax/map.php
+++ b/ajax/map.php
@@ -67,23 +67,24 @@ if (!isset($_POST['itemtype']) || !isset($_POST['params'])) {
 
    $rows = $data['data']['rows'];
    $points = [];
+   $int_points = [];
    foreach ($rows as $row) {
       $idx = $row['raw']["ITEM_$lat_field"] . ',' . $row['raw']["ITEM_$lng_field"];
       $level = count(explode(' > ', $row['raw']["ITEM_$name_field"]));
       if (isset($points[$idx])) {
          $points[$idx]['count'] += 1;
-         if ($level < $points[$idx]['level']) {
+         if ($level < $int_points[$idx]['level']) {
             // It's the higher level
-            $points[$idx]['level'] = $level;
+            $int_points[$idx]['level'] = $level;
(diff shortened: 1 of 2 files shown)


ddurieux:
And some screenshots
screenshot_2018-12-07 glpi - tickets 1
screenshot_2018-12-07 glpi - tickets
screenshot_2018-12-07 glpi - tickets 2

ajax/map.php
```diff
@@ -69,15 +69,34 @@
...
    foreach ($rows as $row) {
       $idx = $row['raw']["ITEM_$lat_field"] . ',' . $row['raw']["ITEM_$lng_field"];
+      $level = count(explode(' > ', $row['raw']["ITEM_$name_field"]));
       if (isset($points[$idx])) {
          $points[$idx]['count'] += 1;
+         if ($level < $points[$idx]['level']) {
+            // It's the higher level
+            $points[$idx]['level'] = $level;
+            $points[$idx]['title'] = $row['raw']["ITEM_$name_field"];
+            $points[$idx]['loc_ids'] = [$row['raw']['loc_id']];
```
trasher: `loc_ids` does not seems used

ddurieux: It's used to define if the location name if yet added in title.
The behavior is:
* if the location level is higher, it replace the name and the id in _loc_ids_ list
* if the location level is same, it check if the id in _loc_ids_ list, if not it add the name to the previous title (so will have many locations names in same location level)

trasher: Ok, so this is used  locally only, but sent along with the ajax result. I think this is useless to return it since this is not used ;)

ddurieux: You're right, I will modify it

ddurieux: I add code to fix that in a new commit, not squashed to see the diff

ddurieux:
J'ai mis en WIP, j'ai détecté un bug

trasher:
@ddurieux english only on the tracker please.

trasher:
Are you planning to end this or should we close?

trasher:

PR #5042, opened 2018-12-07
Incorrect CFG_GLPI name - see #5041

yllen:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #5041


trasher:
`$CFG_GLPI['itemdevice_types']` is defined and used only in that place (do not even understand why this has been used). Its name did not change since 2013, I wonder what is wrong here.

trasher:
Also, please use "close" keyword instead of see in your commits.

PR #5165, opened 2018-12-21
Fix QueuedNotification clean cron

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

Fixes:
SQL: DELETE FROM `glpi_queuednotifications` WHERE `is_deleted` = '1' AND UNIX_TIMESTAMP(`send_time`) < '1542826920')
  Error: You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near ')' at line 1


cconard96:
Fixed in 2429ed9

PR #5172, opened 2018-12-26
fix #5171 Duplicated action between RuleImportEntity & Ruleasset

### LINKED ISSUE #5171: Duplicated action between RuleImportEntity & Ruleasset
RuleImportEntity: 
2 actions are duplicated with the new AuleAsset rule engine: Location and Group in charge of the hardware

amma35:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #5171

### DIFF
--- a/inc/ruleimportentity.class.php
+++ b/inc/ruleimportentity.class.php
@@ -52,7 +52,7 @@ class RuleImportEntity extends Rule {
    **/
    function maxActionsCount() {
       // Unlimited
-      return 4;
+      return 2;
    }
 
    function executeActions($output, $params, array $input = []) {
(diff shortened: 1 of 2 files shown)


orthagh:
I'm not sure at all about this PR.
You're removing existing configurations for admin without warning them at all (nor aksing them to choose).

But I'm not sure also on criteria/actions removal as the old rules cannot be correctly displayed if we don't provide any migration to the other engine.

The only thing that can be removed surely is any pre-configuration in empty sql file,  if any exists.

a possible scenario:
- warn user on some entity rules are udapted  (if engine enabled as is it currently only used by ocs plugin)
- disable old rules
- don't permits to use old criteria / action, keep them deprecated.

trasher:

PR #5205, opened 2019-01-07
permits to sort change's items searchoption

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes


yllen:
@orthagh we never allow sort for item of an object because it never sort items but object.
Example: if you select only one object with many items and want to sort them, action on column title will do nothing.

Issue #5098, opened 2018-12-12
[9.3.3][BETA 9.4] Add an authorization to a user tab & profiles selection


seb1435:
**Describe the bug**

https://forum.glpi-project.org/viewtopic.php?pid=356072

Only users with Super-admin profile can select all the profiles available in the Authorization tab when editing an user.

Other profiles like Admin can only select Admin or Self-service

**To reproduce**

1. Add a user with Super-admin and Admin profiles
2. Log-in using this user and select Super-admin profile
3. Edit an user
4. Go on the Authorizations tab
5. Open the Profile dropdown: all available profile are listed
6. Select Admin profile
7. Open the Profile dropdown: only Admin and Self-service are available for selection

**Expected behavior**

As Admin is allowed to read profiles in Administration > Profiles > Administration, all profiles with lower rights should be available for selection in the dropdow list (as Technician or Supervisor).

**Your GLPI setup (you can find it in Setup > General menu, System tab)**

```
[code]
 
GLPI 9.3.3 ( => /var/www/glpi)
Installation mode: TARBALL
```

**Additional context**

Behavior also presents in beta 9.4

OscarBeiro:
This issue is inherited.
Already present on 9.2.x

cconard96:
I think the dropdown is working properly, but default profile rights are not correct (tested from 9.4).
Example, a Technician has read access to public RSS and public reminders, but an Admin does not. Another example is a Technician and Supervisor has access to read entities but an Admin does not.

Therefore, the Admin profile is less privileged than other profiles and cannot assign them to users. Other issues exist between other profiles.

seb1435:
> Therefore, the Admin profile is less privileged than other profiles and cannot assign them to users. Other issues exist between other profiles.

So maybe the problem is the way GLPI calculates the rank of priviledge of a profile? For exemple, in my setup, Admin profile has more rights on configuration and  administration tabs than Supervisor but less on Assistance tab.

cconard96:
The way I understand it, the admin (or which ever profile you use) has to have at the least the same level of access for every right. I assume this is to prevent an Admin from creating a more privileged profile and assigning it to themselves to gain more rights.

seb1435:
Indeed, but GLPI should weight the differents rights, at least for administration and configuration rights don't you think ? A supervisor should be able to create another supervisor, a technician, etc.

ConteZero:
@trasher @cconard96 current merged commit doesn't completely solve the bug.
[...]

Issue #5214, opened 2019-01-08
[9.4] An user with Admin profile is not able to change his password


ConteZero:
In GLPI version 9.4 RC1 an user with Admin profile is not able to change his password from /front/user.form.php page.
This is caused by 2 missing inserts in glpi_profilerights table from glpi-empty.sql, global_validation and rule_asset are not present into the table and this is the cause of the problem.
Manually adding them to glpi_profilerights fixes the bug.

I've created a new profile with the same permissions of Admin default profile and looking at the result of

SELECT * FROM `glpi_profilerights` WHERE `profiles_id`= 3;

and comparing with the same query repeated for the newly created profile there are many differences into the values.

For example for Admin you have

datacenter 	255

instead for the other profile you have

datacenter 	31

this means that the Admin has permissions on 'Read notes' 'Update notes' 'Select/unselect all' that are not enabled on the other profile.
'Read notes' and 'Update notes' are not selectable from the interface and 'Select/unselect all' should only be a javascript function to enable the full row.

From my point of view, having these differences into the database could potentially impact functions that compare profiles permissions and could cause unexpected results.
The same inconsistencies in default permissions values are present also for other profile created by glpi-empty.sql file.
I think that this problem is also related with the issue https://github.com/glpi-project/glpi/issues/5098

cconard96:
Related to my comment in #5098 . I am trying to look into the rights discrepancies for the default profiles.

cconard96:
Actually, not sure the issues are related. Seems like a UI inconsistency between /front/user.form.php and /front/preference.php when editing your own user.

ConteZero:
function currentUserHaveMoreRightThan in /inc/profile.class uses getUnderActiveProfileRestrictCriteria that gives not correct results with current default profiles created by installer.
The discrepancies for the default profiles have effect also on new user creation, in my config I have the default Admin profile and a newly create profile (AdminTest) with the same permissions:
- Admin is able to create a new user with Admin or AdminTest profiles
- AdminTest is able to create a new user with AdminTest profile but not with Admin profile

PR #5215, opened 2019-01-09
Fix default profile rights

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #5098 

Some rights had extra bits assigned than when assigning rights in UI (not including object locks). Some profile rights were not set in more privileged profiles.

Tried to keep rights in line with:
https://glpi-user-documentation.readthedocs.io/fr/latest/modules/administration/07_Profils/01_Profils.html


ConteZero:
I think that rule_asset and global_validation should be added to all the default profiles, or you'll get wrong results from functions that compare profile permissions
https://github.com/glpi-project/glpi/issues/5214#issuecomment-452602444

When a new profile is created from the interface, 79 different permissions are written into glpi_profilerights table, every profile created by installer should have all these permissions or you get unexpected results when comparing profile permissions.

ConteZero:
In case of upgrade from old GLPI versions the problem with the default profiles will remain.
It could be handled with a db migration, but I think it's tricky to do it without the risk of break something.

orthagh:
> but I think it's tricky to do it without the risk of break something.

exactly, and we can't know when an administrator has altered on purpose its profiles rights.
So currently, we can only fix new installation.

ConteZero:
Perhaps add some info into the release notes, so the administrators could do the fixes by themselves if it's needed.

PR #5226, opened 2019-01-10
Fix linking when merging tickets

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

Merge target should become parent of merged ticket instead of son.
If there is a an opposite link between tickets already, merge would of failed while adding the new link. So, delete the old link first.

### DIFF
--- a/inc/ticket.class.php
+++ b/inc/ticket.class.php
@@ -2789,10 +2789,18 @@ class Ticket extends CommonITILObject {
                      //Add relation (this is parent of merge target)
                      $tt = new Ticket_Ticket();
                      $linkparams = [
-                        'link'         => Ticket_Ticket::PARENT_OF,
+                        'link'         => Ticket_Ticket::SON_OF,
                         'tickets_id_1' => $id,
                         'tickets_id_2' => $input['_mergeticket']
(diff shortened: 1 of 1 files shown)


inc/ticket.class.php
```diff
@@ -2789,10 +2789,18 @@ static function processMassiveActionsForOneItemtype(MassiveAction $ma, CommonDBT
...
                      $tt = new Ticket_Ticket();
                      $linkparams = [
-                        'link'         => Ticket_Ticket::PARENT_OF,
+                        'link'         => Ticket_Ticket::SON_OF,
                         'tickets_id_1' => $id,
                         'tickets_id_2' => $input['_mergeticket']
                      ];
+
+                     $existinglinks = Ticket_Ticket::getLinkedTicketsTo($id);
+                     foreach ($existinglinks as $ttkey => $link) {
```
orthagh: i think this whole part could be replaced by 
```php
$tt->deleteByCriteria([
...
]);
```

PR #5240, opened 2019-01-11
fix nested criteria when searchoption id is not already displayed

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | internal ref 15769

2 issues fixed:
- nested criteria didn't push their id into `$data['toview']` cause the whole part was still for old flat mecanism, so i add a clojure for parsing recursively criteria. i didn't saw it because i think all my tests was on searchoptions already present in search columns

- if a nested **criterion** is the first criterion of a query, a superflu `AND` was added in front of the sub sql.

### DIFF
--- a/tests/functionnal/Search.php
+++ b/tests/functionnal/Search.php
@@ -339,9 +339,9 @@ class Search extends DbTestCase {
                      'value'      => 'test',
                   ], [
                      'link'       => 'OR',
-                     'field'      => 'view',
+                     'field'      => 2,
                      'searchtype' => 'contains',
-                     'value'      => 'toto',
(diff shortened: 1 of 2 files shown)


tests/functionnal/Search.php
```diff
@@ -376,18 +375,73 @@ public function testNestedAndMetaComputer() {
...
+      $this->array($data)
```
trasher: A very big :+1:  for that change!

PR #5375, opened 2019-02-13
bug using rest api if auth Cas is configured

salleman33:
Avoid redirect to cas authentification to use rest api


trasher:
Sounds good to me because tests are still ok.  You have a small CS issue to fix (see circleci logs)

PR #5380, opened 2019-02-13
Fix CSRF and DB function in install.php

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

Close method on PDO does not seem to exist and breaks form CSRF (closeForm wouldn't get called).
Script would stop when calling function quoteName on class AbstractDatabase. This in turn called an abstract function  getQuoteNameChar which stopped the script.

### DIFF
--- a/install/install.php
+++ b/install/install.php
@@ -320,7 +320,6 @@ function step3($host, $user, $password, $update) {
          echo "<input type='hidden' name='install' value='Etape_3'>";
          echo "<p class='submit'><input type='submit' name='submit' class='submit' value='".
                __('Continue')."'></p>";
-         $link->close();
          Html::closeForm();
 
       } else if ($update == "yes") {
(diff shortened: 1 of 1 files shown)


install/install.php
```diff
@@ -434,7 +432,7 @@ function next_form() {
       }
    } else if (!empty($newdatabasename)) { // create new db
       // try to create the DB
-      if ($link->query("CREATE DATABASE IF NOT EXISTS ".AbstractDatabase::quoteName($newdatabasename))) {
+      if ($link->query("CREATE DATABASE IF NOT EXISTS ".DBmysql::quoteName($newdatabasename))) {
```
trasher: `DBmysql` should not be used directly. Most of the time we can rely on `$DB` but here it is not yet instantiated.

trasher:
pin @tsmr ; this should solve the issue you were talking about on IRC channel ;)

tomolimo:
> ping @tsmr ; this should solve the issue you were talking about on IRC channel ;)

which one?

PR #5523, opened 2019-03-05
All tables must use InnoDB

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

No longer allow tables using myisam.


trasher:
As far as I can see; checks are done from the Update class, so this should be ok from command line as well.

PR #5577, opened 2019-03-12
Password token date was not removed

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #number

### DIFF
--- a/inc/user.class.php
+++ b/inc/user.class.php
@@ -4515,7 +4515,7 @@ class User extends CommonDBTM {
                }
                $input2 = [
                   'password_forget_token'      => '',
-                  'password_forget_token_date' => null,
+                  'password_forget_token_date' => 'NULL',
                   'id'                         => $this->fields['id']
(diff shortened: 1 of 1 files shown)


remicollet:
IMHO the issue is NOT fixed, as date are not clean in the 9.4.1 upgrade script, this will only avoid no more bad situation

IMHO, another fix is requires

```
-            if (($input['password_forget_token'] == $this->fields['password_forget_token'])
+            if ($input['password_forget_token'] && ($input['password_forget_token'] == $this->fields['password_forget_token'])
```

trasher:
If the stored date is older than 24h, it was not possible to use the exploit; that's why dates has not been reset (even if they should have, it's a bit late and problem will be solved until we release a new version).

PR #5742, opened 2019-04-04
Feature/95 imap tests

trasher:
PHPdoc, rename numberOf* methods
Factorize prepareInput, rename some methods
Ignore mocks when getting table name

Same as https://github.com/glpi-project/glpi/pull/5422 but target 9.5.


trasher:
I do not see why collector would have been overrided; and I think this would be a bad idea... Tha'ts why I've decided no to handle complex depreciations

PR #5772, opened 2019-04-10
fix user pictures

remicollet:
GLPI_PICTURES_DIR is NOT a sub-directory of GLPI_DOC_DIR


remicollet:
Please also apply in 9.3 bugfixes

PR #5778, opened 2019-04-11
Fix usage of INFORMATION_SCHEMA db on MySQL 8

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #number

### DIFF
--- a/inc/console/migration/timestampscommand.class.php
+++ b/inc/console/migration/timestampscommand.class.php
@@ -48,7 +48,7 @@ class TimestampsCommand extends AbstractCommand {
       parent::configure();
 
       $this->setName('glpi:migration:timestamps');
-      $this->setDescription(__('Convert DATETIME to timestamps to use timezones.'));
+      $this->setDescription(__('Convert "datetime" fields to "timestamp" to use timezones.'));
    }
 
    protected function execute(InputInterface $input, OutputInterface $output) {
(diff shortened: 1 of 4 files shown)


cedric-anne:
Tests are OK now.

From [official MySQL8 changelog](https://dev.mysql.com/doc/relnotes/mysql/8.0/en/news-8-0-0.html#mysqld-8-0-0-feature)
image

If I understand right the [lower_case_table_names configuration](https://dev.mysql.com/doc/refman/8.0/en/server-system-variables.html#sysvar_lower_case_table_names), it should work for any value.

inc/console/migration/timestampscommand.class.php
```diff
@@ -48,39 +48,39 @@ protected function configure() {
...
+                     'AND' => ['information_schema.TABLES.TABLE_TYPE' => 'BASE TABLE']
                   ]
                ]
             ]
          ],
          'WHERE'       => [
-            'INFORMATION_SCHEMA.COLUMNS.TABLE_SCHEMA' => $this->db->dbdefault,
-            'INFORMATION_SCHEMA.COLUMNS.COLUMN_TYPE'  => 'DATETIME'
+            'information_schema.COLUMNS.TABLE_SCHEMA' => $this->db->dbdefault,
+            'information_schema.COLUMNS.DATA_TYPE'    => 'datetime'
```
cedric-anne: I used the `DATA_TYPE` here as `The DATA_TYPE value is the type name only with no other information. The COLUMN_TYPE value contains the type name and possibly other information such as the precision or length.`

See https://dev.mysql.com/doc/refman/8.0/en/columns-table.html

trasher: MySQL does not care about case; postgresql do; and all tables/columns names are lowercase. I prefer to keep all lowercase to prevent confusions.

cedric-anne: OK, I will change it

PR #5819, opened 2019-04-20
enable DISTINCT request

salleman33:
As it is specified in https://glpi-developer-documentation.readthedocs.io/en/master/devapi/database/dbiterator.html?highlight=distinct#fields-selection for GLPI 10.0.0 (?!)


trasher:
This is already implemented on master (and maybe 9.5) branch. No new feature or core change in 9.4.

PR #5834, opened 2019-04-26
Wanted menu AllAssets to appear at all time

FVallieres:
current logic suggests that if none the menus ['Computer', 'Monitor', 'Peripheral', 'NetworkEquipment', 'Phone',
'Printer'] is activated then [allAssets] menu does not show up.

We want to see that menu at all time, especially when we dont use any of the objects above.

topic open at :
https://glpi.userecho.com/communities/1/topics/1027-wanted-to-see-allassets-menuat-all-times

Q	A
Bug fix?	yes
New feature?	yes
BC breaks?	no
Deprecations?	no
Tests pass?	yes
Fixed tickets


trasher:
There are many things not related to the inital "issue"; and I'm not OK with the fix anyways.

loicbenoit:
Hi Trasher,

The objective is to make it possible to display menu "Assets/Global", even when using only GenericObjects.

We welcome any proposal you might have on how to best achieve this objective.

Thanks.

PR #5891, opened 2019-05-10
Using a cache to know whether a table exits or not

FVallieres:
Using a cache to know whether a table exits or not, instead of requesting DB each time.

It partially fixes the bug of having the same queries sent too many times to database in the same HTTP request, which slow down considerably the application.
The correction consists of saving the DB return values of some queries into a static (class) variable (cache).

A part of the issue 5273 was solved otherwise by community on subsequent versions, so this PR contains only the part that was left unsolved.

Q | A
-- | --
Bug fix? | no
New feature? | yes
BC breaks? | no
Deprecations? | no
Tests pass? | yes
Fixed tickets | glpi-project#5273

### DIFF
--- a/inc/dbmysql.class.php
+++ b/inc/dbmysql.class.php
@@ -67,6 +67,9 @@ class DBmysql {
 
    private $cache_disabled = false;
 
+   // Ticket GitHub #5273, to avoid requesting DB each time.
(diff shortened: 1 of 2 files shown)


trasher:
See #5904

PR #5928, opened 2019-05-20
Fix 'is_requester' condition

AdrienClairembault:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

### DIFF
--- a/inc/ruleticket.class.php
+++ b/inc/ruleticket.class.php
@@ -416,7 +416,7 @@ class RuleTicket extends Rule {
       $criterias['_groups_id_assign']['name']               = __('Technician group');
       $criterias['_groups_id_assign']['linkfield']          = '_groups_id_assign';
       $criterias['_groups_id_assign']['type']               = 'dropdown';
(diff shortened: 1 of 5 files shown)


inc/ruleticket.class.php
```diff
@@ -546,7 +546,7 @@ function getActions() {
       $actions['_groups_id_requester']['name']              = __('Requester group');
       $actions['_groups_id_requester']['type']              = 'dropdown';
       $actions['_groups_id_requester']['table']             = 'glpi_groups';
-      $actions['_groups_id_requester']['condition']         = '`is_requester`';
+      $actions['_groups_id_requester']['condition']         = ['is_requester' => 1];
```
trasher: Not sure array is supported here in 9.5/bf, need to check

trasher: Talking about that directly, seems that is is :)

PR #5947, opened 2019-05-23
Put data insert out of SQL script

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

This makes data insertion database independent, and maybe will help to change the behavior when an error occurs in the install SQL script (currently, the whole script is displayed, not only the error)


install/empty_data.php
```diff
@@ -0,0 +1,7994 @@
...
+      'event'        => 'new',
+      'is_recursive' => 1,
+      'is_active'    => 1
+    ], [
+      'id'           => 3,
+      'name'         => 'Update Ticket',
+      'itemtype'     => 'Ticket',
+      'event'        => 'update',
+      'is_recursive' => 1,
+      'is_active'
```
orthagh: One idea as we are now in a php script.
Could we set some constants to identify 
- profiles id (ex PROF_SUPERADMIN_ID, PROF_OBSERVER_ID, etc)
- users id (USER_GLPI_ID, USER_TECH_ID, etc)
- Rights values (READ, UPDATE, etc)

trasher: I agree; but this is out of the scope of the current PR, I'd prefer to work on that on a separate one

PR #5954, opened 2019-05-24
Impact analysis 1/2

AdrienClairembault:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | -

First PR for impact analysis.


inc/impact.class.php
```diff
@@ -0,0 +1,993 @@
...
+            // Ignore default value (Dropdown::EMPTY_VALUE)
+            if (value == "default") {
+               return;
+            }
+
+            values = value.split("::");
+
+            $.ajax({
+               type: "POST",
+               url: "../ajax/impact.php",
```
cedric-anne:
```suggestion
               url: CFG_GLPI.root_doc + "/ajax/impact.php",
```

inc/impact.class.php
```diff
@@ -0,0 +1,993 @@
...
+      $ticket = new Ticket();
+      $problem = new Problem();
+      $change = new Change();
+      $imageName = strtolower(get_class($item));
+
+      $newNode = [
+         'id'     => $key,
+         'label'  => $item->fields['name'],
+         'shape'  => "image",
+         'image'  => "../pics/impact/$imageName.png",
```
cedric-anne: Path should not be relative.

inc/problem.class.php
```diff
@@ -1573,4 +1573,78 @@ function getRights($interface = 'central') {
...
+    * get active problems for an item
+    *
+    * @since 9.5
+    *
+    * @param string $itemtype     Item type
+    * @param integer $items_id    ID of the Item
+    *
+    * @return integer
+    */
+   public function GetActiveProblemsForItem($itemtype, $items_id) {
```
cedric-anne:
```suggestion
   public function getActiveProblemsForItem($itemtype, $items_id) {
```

install/mysql/glpi-empty.sql
```diff
@@ -1380,6 +1380,26 @@ INSERT INTO `glpi_configs` VALUES ('206','core','purge_all','0');
 INSERT INTO `glpi_configs` VALUES ('207','core','purge_user_auth_changes','0');
 INSERT INTO `glpi_configs` VALUES ('208','core','purge_plugins','0');
 INSERT INTO `glpi_configs` VALUES ('209','core','display_login_source','1');
+INSERT INTO `glpi_configs` VALUES ('210','core','impact_assets_list','');
+
+### Dump table glpi_impacts
+
+CREATE TABLE `glpi_impacts` (
+	`id` INT(11) NOT NULL AUTO_INCREMENT,
+	`source_asset_type` VARCHAR(255) NOT NULL DEFAULT '' COLLATE 'utf8_unicode_ci',
```
cedric-anne: Some automatic relation handling is made based on field names. So it is important to fits existing naming rules.
For example: https://github.com/glpi-project/glpi/blob/9.5/bugfixes/inc/commondbrelation.class.php#L109
```suggestion
	`itemtype_source` VARCHAR(255) NOT NULL DEFAULT '' COLLATE 'utf8_unicode_ci',
```

install/mysql/glpi-empty.sql
```diff
@@ -1380,6 +1380,26 @@ INSERT INTO `glpi_configs` VALUES ('206','core','purge_all','0');
...
+	`source_asset_id` INT(11) NOT NULL DEFAULT '0',
```
cedric-anne:
```suggestion
	`items_id_source` INT(11) NOT NULL DEFAULT '0',
```

install/mysql/glpi-empty.sql
```diff
@@ -1380,6 +1380,26 @@ INSERT INTO `glpi_configs` VALUES ('206','core','purge_all','0');
...
+	`impacted_asset_type` VARCHAR(255) NOT NULL DEFAULT '' COLLATE 'utf8_unicode_ci',
```
cedric-anne:
```suggestion
	`itemtype_impacted` VARCHAR(255) NOT NULL DEFAULT '' COLLATE 'utf8_unicode_ci',
```

install/mysql/glpi-empty.sql
```diff
@@ -1380,6 +1380,26 @@ INSERT INTO `glpi_configs` VALUES ('206','core','purge_all','0');
...
+	`impacted_asset_id` INT(11) NOT NULL DEFAULT '0',
```
cedric-anne:
```suggestion
	`items_id_impacted` INT(11) NOT NULL DEFAULT '0',
```

install/update_94_95.php
```diff
@@ -282,6 +282,51 @@ function update94to95() {
       ) ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_unicode_ci";
       $DB->queryOrDie($query, "add table glpi_itilfollowuptemplates");
    }
+   /** Impact analysis */
+
+   // Impact config
+   $impactConfig = $DB->request([
```
cedric-anne: You can use `$migration->addConfig(['impact_assets_list' => ''])`.

js/impact_network.js
```diff
@@ -0,0 +1,560 @@
...
+   window.colors[BACKWARD]  = DEPENDS_COLOR;
+   window.colors[BOTH]      = IMPACT_AND_DEPENDS_COLOR;
+
+   // Get start node type and id
+   var startNodeDetails = window.startNode.split('::');
+
+   // Load the graph
+   $.ajax({
+      type: "POST",
+      url: "../ajax/impact.php",
```
cedric-anne:
```suggestion
      url: CFG_GLPI.root_doc + "/ajax/impact.php",
```

js/impact_network.js
```diff
@@ -0,0 +1,560 @@
...
+                  alert(getLocale("duplicateAsset"));
+                  return;
+               }
+
+               /* We may have new nodes and eges linked to the node that we are
+                 inserting into the graph, let's check by building a new graph
+                 from this node */
+               $.ajax({
+                  type: "POST",
+                  url: "../ajax/impact.php",
```
cedric-anne:
```suggestion
                  url: CFG_GLPI.root_doc + "/ajax/impact.php",
```

inc/impact.class.php
```diff
@@ -0,0 +1,993 @@
+<?php
+
+if (!defined('GLPI_ROOT')) {
+   die("Sorry. You can't access this file directly");
+}
+
+class Impact extends CommonDBRelation {
```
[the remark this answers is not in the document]
cedric-anne: I added relation properties and it seems to work.

inc/impactitem.class.php
```diff
@@ -0,0 +1,81 @@
...
+/**
+ * @since 9.5.0.
+ */
+class ImpactItem extends CommonDBTM {
+
+   function prepareInputForUpdate($input) {
+      return $input;
+   }
+
+   // public function update(array $input, $options = [], $history = true) {
```
cedric-anne: You should remove dead code.

inc/html.class.php
```diff
@@ -1286,6 +1286,11 @@ static function includeHeader($title = '', $sector = 'none', $item = 'none', $op
             echo Html::css('public/lib/codemirror.css');
             Html::requireJs('codemirror');
          }
+
+         if (in_array('vis', $jslibs)) {
```
cedric-anne: `vis` lib is not used anymote, you should remove these lines.

install/update_94_95.php
```diff
@@ -452,6 +452,77 @@ function update94to95() {
    $migration->addKey('glpi_tickettemplatepredefinedfields', 'tickettemplates_id');
    /** /ITiL templates */
 
+   /** Impact analysis */
+
+   /** Impact analysis */
```
```suggestion
```

js/impact.js
```diff
@@ -0,0 +1,2332 @@
...
+// Constants for delta action
+var DELTA_ACTION_ADD    = 1;
+var DELTA_ACTION_UPDATE = 2;
+var DELTA_ACTION_DELETE = 3;
+
+// Load cytoscape
+var cytoscape = window.cytoscape;
+
+// Needed for JS lint validation
+var _ = window._;
```
cedric-anne: For linting validation, you can add `/* global _ */` instead of creating a new var.

js/impact.js
```diff
@@ -0,0 +1,2332 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with GLPI. If not, see <http://www.gnu.org/licenses/>.
+ * ---------------------------------------------------------------------
+ */
+
+// Constants to represent nodes and edges
+var NODE = 1;
```
cedric-anne: As JS code is still loaded directly in window context (still not wrapped into modules), I would prefer these constants to be declared inside `impact` namespace.

js/impact.js
```diff
@@ -0,0 +1,2332 @@
...
+var DELTA_ACTION_DELETE = 3;
+
+// Load cytoscape
+var cytoscape = window.cytoscape;
+
+// Needed for JS lint validation
+var _ = window._;
+var showMenu = window.showMenu;
+
+var impact = {
```
cedric-anne: 1. Namespaces should avec a name starting with an uppercase char (like a class name).
2. To prevent conflict with third party components, prefixing namespace with `GLPI` could be a good idea (as already done for `GLPIPlanning`.

```suggestion
var GLPIImpact = {
```

js/impact.js
```diff
@@ -0,0 +1,2332 @@
...
+               parent_id               : node.parent,
+               position_x              : node.position.x,
+               position_y              : node.position.y,
+               zoom                    : impact.cy.zoom(),
+               pan_x                   : impact.cy.pan().x,
+               pan_y                   : impact.cy.pan().y,
+               impact_color            : impact.edgeColors[FORWARD],
+               depends_color           : impact.edgeColors[BACKWARD],
+               impact_and_depends_color: impact.edgeColors[BOTH],
+               nodes_positions         : nodesPositions,
```
cedric-anne: I cannot save impact graph unless I made this change:
```suggestion
               nodes_positions         : JSON.stringify(nodesPositions),
```

js/impact.js
```diff
@@ -0,0 +1,2332 @@
...
+      );
+   },
+
+   /**
+    * Handler for "removeFromCompound" action
+    *
+    * @param {JQuery.Event} event
+    */
+   menuOnRemoveFromCompound: function(event) {
+      var parent = impac
```
cedric-anne: I have a JS error when I click on the export link:
image

ajax/dropdownTrackingDeviceType.php
```diff
@@ -36,9 +36,19 @@
 
 Session::checkLoginUser();
 
+// Read parameters
+$context  = $_POST['context'] ?? '';
+$itemtype = $_POST["itemtype"] ?? '';
```
trasher: `itemtype` is mandatory (see L. 51), so this is wrong to set it to `''` here. `$isValidItemtype` should be conditionned on `$_POST['itemtyp']`presence.
Also, if it is missing, a warning should be logged IMHO

ajax/impact.php
```diff
@@ -0,0 +1,170 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with GLPI. If not, see <http://www.gnu.org/licenses/>.
+ * ---------------------------------------------------------------------
+ */
+
+const DELTA_ACTION_ADD    = 1;
```
trasher: Constants here? Better in a class, no?

ajax/impact.php
```diff
@@ -0,0 +1,170 @@
...
+
+// Send UTF8 Headers
+header("Content-Type: application/json; charset=UTF-8");
+Html::header_nocache();
+
+Session::checkLoginUser();
+
+switch ($_SERVER['REQUEST_METHOD']) {
+   // GET request: build the impact graph for a given asset
+   case 'GET':
```
trasher: Different HTTP method will do things very different? I'm really not fan

ajax/impact.php
```diff
@@ -0,0 +1,170 @@
...
+      $itemtype =  $_GET["itemtype"]  ?? "";
+      $items_id   =  $_GET["items_id"]    ?? "";
+
+      // Check required params
+      if (empty($itemtype) || empty($items_id)) {
+         http_response_code(400);
+         die;
```
trasher: do not just die, but die with a return code > 0. Also, you should add a log entry explaining what went wrong

css/impact.css
```diff
@@ -0,0 +1,378 @@
+#network_container {
```
trasher: Should be a `.scss` file I guess

inc/commondbtm.class.php
```diff
@@ -5143,4 +5143,55 @@ private function assetBusinessRules($condition) {
          }
       }
    }
+
+   /**
+    * Get incidents, request, changes and problem linked to this object
+    *
+    * @return array
+    */
+   public function getITILTickets(bool $count = false) {
```
trasher: Not sure this is the right place. Seems too large for an ITIL specific method

inc/define.php
```diff
@@ -510,3 +511,16 @@
 //Maximum time, in miliseconds a saved search should not exeed
 //so we count it on display (using automatic mode).
 $CFG_GLPI['max_time_for_count'] = 200;
+
+$CFG_GLPI["impact_asset_types"]  = [
+   'Computer'         => "pics/impact/computer.png",
```
trasher: Is this really required to list all images? Using lowered type name seems enough

inc/html.class.php
```diff
@@ -5551,6 +5551,7 @@ static function redefineAlert() {
                $(this).parent().prev('.ui-widget-overlay').addClass('glpi_modal');
             },
             close: function(){
+               $(this).parent().prev('.ui-widget-overlay').removeClass('glpi_modal');
```
trasher: hum... this looks like something added from jquery-ui, I do not understand why you have to remove it explicitely

inc/impact.class.php
```diff
@@ -0,0 +1,1091 @@
...
+      // Only enabled for CommonDBTM
+      if (!is_a($item, "CommonDBTM")) {
+         throw new InvalidArgumentException(
+            "Argument \$item ($class) must be a CommonDBTM)."
+         );
+      }
+
+      // If on an asset, get the number its direct dependencies
+      if (isset($CFG_GLPI['impact_asset_types'][$class])) {
+         $total = count($DB->request([
```
trasher: As far as I know, counting should only be done if the configuration is enabled

inc/impact.class.php
```diff
@@ -0,0 +1,1091 @@
...
+      $class = get_class($item);
+
+      // Only enabled for CommonDBTM
+      if (!is_a($item, "CommonDBTM")) {
+         throw new InvalidArgumentException(
+            "Argument \$item ($class) must be a CommonDBTM)."
+         );
+      }
+
+      $ID = $item->getID();
```
trasher: Be very careful with `getID` usage (even if it looks like a good idea to use it, in facts this is not on many cases, but required in some others :/)

inc/impact.class.php
```diff
@@ -0,0 +1,1091 @@
...
+
+         // Search for a valid linked item of this ITILObject
+         $found = false;
+         foreach ($linked_items as $linked_item) {
+            $class = $linked_item['itemtype'];
+            if (isset($CFG_GLPI['impact_asset_types'][$class])) {
+               self::printAssetSelectionForm($linked_items);
+               $found = true;
+               $item = new $class;
+               $item->getFromDB($linked_item['items_id']);
```
trasher: `$found = true` should be set only if `getFromDB` is true aswell

inc/impact.class.php
```diff
@@ -0,0 +1,1091 @@
...
+               break;
+            }
+         }
+
+         // No valid linked item were found, tab shouldn't be visible
+         if (!$found) {
+            return true;
```
trasher: Do you mean `false`?

inc/impact.class.php
```diff
@@ -0,0 +1,1091 @@
...
+      return true;
+   }
+
+   /**
+    * Load the cytoscape and spectrum-colorpicker librairies
+    *
+    * @since 9.5
+    */
+   public static function loadLibs() {
+      echo Html::css('public/lib/spectrum-colorpicker.css');
```
trasher: There is another mechanism to load CSS and JS

inc/impact.class.php
```diff
@@ -0,0 +1,1091 @@
...
+      global $CFG_GLPI;
+
+      // Dropdown values
+      $values = [];
+
+      // Add a value in the dropdown for each items, grouped by type
+      foreach ($items as $item) {
+         if (isset($CFG_GLPI['impact_asset_types'][$item['itemtype']])) {
+            // Add itemtype group if it doesn't exist in the dropdown yet
+            $itemtype_label = __($item['itemtype']);
```
trasher: I'm not sure all itemtypes do have a translation. Better to use `getTypeName()` I guess

inc/impact.class.php
```diff
@@ -0,0 +1,1091 @@
...
+               $values[$itemtype_label] = [];
+            }
+
+            $key = $item['itemtype'] . "::" . $item['items_id'];
+            $values[$itemtype_label][$key] = $item['name'];
+         }
+      }
+
+      Dropdown::showFromArray("impact_assets_selection_dropdown", $values);
+      echo "<br><br>";
```
trasher: Really required?

inc/impactcompound.class.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+if (!defined('GLPI_ROOT')) {
+   die("Sorry. You can't access this file directly");
+}
+
+/**
+ * @since 9.5.0
+ */
+class ImpactCompound extends CommonDBTM {
```
trasher: Why an emtpy class?

AdrienClairembault:
@cedric-anne 
>     * Already used items should not be proposed on "Add new asset" function in impact graph.
This is planned for the second release as it is more difficult than expected.

All your others points should be resolved, I'm working on the ones reported by @trasher now.

ajax/impact.php
AdrienClairembault: These constants are only relevant for updating the impact data through this ajax file.
They are only meant to indicate the action for each data sent to this endpoint (if we already have const to differentiate add/update/delete then I could use them instead but I didn't found any).

ajax/impact.php
AdrienClairembault: This seems fine to me, `GET` allow to access impact data and `POST` to send a new impact payload (update).

Without this we would need another parameter like "action" to differentiate our two possible actions (or split them in two ajax files).
Using the HTTP method seems a simple and clean way to do it.

@cedric-anne @orthagh, thoughts ?

inc/commondbtm.class.php
AdrienClairembault: Where should we define this ? @orthagh @cedric-anne

inc/define.php
AdrienClairembault: At first we used lowecase itemtype name but we switched to this as it makes it easier for plugins to add their own types and images (or to modify the existing ones).

trasher: or to modify the existing ones => I'm pretty sure this is a bad idea

inc/impact.class.php
AdrienClairembault: I'm not sure to understand, isn't it already the case as we are checking `$CFG_GLPI['impact_asset_types'][$class]`  ?

trasher: I'm talking about `$_SESSION['glpishow_count_on_tabs']`

inc/impact.class.php
AdrienClairembault: Why ? What else can it be replaced with ?

trasher: `$object->fields['id']`

inc/impact.class.php
AdrienClairembault: I think we did it that way because the files will be needed in a lot a pages (each asset type) and some plugin might add new assets type that will require them aswell. Can you confirm @cedric-anne ?

inc/impactcompound.class.php
AdrienClairembault: Compounds are used internally to define and store parent relations in the impact graphs but they can't be accessed "directly" in GLPI therefore I don't need to override anything from CommonDBTM.

ajax/dropdownTrackingDeviceType.php
AdrienClairembault: Fixed in latest commit.

ajax/impact.php
AdrienClairembault: Fixed in latest commit.

inc/html.class.php
AdrienClairembault: Indeed, I don't remember changing this, it must be a mistake.
Fixed in latest commit.

inc/impact.class.php
AdrienClairembault: Fixed in latest commit.

inc/impact.class.php
AdrienClairembault: Fixed in latest commit.

inc/impact.class.php
AdrienClairembault: Fixed in latest commit.

inc/impact.class.php
AdrienClairembault: Fixed in latest commit.

inc/impact.class.php
AdrienClairembault: Fixed in latest commit.

AdrienClairembault:
I'm working on the units tests.

ajax/impact.php
cedric-anne: I usually do it the same way, but it is not how it is done in GLPI right now. So I do not really have an opinion about this.

inc/commondbtm.class.php
cedric-anne: It could be a trait but it would require to add this trait to lots of elements. So maybe in CommonDBTM it is a good place.

tests/units/Impact.class.php
```diff
@@ -0,0 +1,287 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with GLPI. If not, see <http://www.gnu.org/licenses/>.
+ * ---------------------------------------------------------------------
+*/
+
+namespace tests\units;
+
+class Impact extends \GLPITestCase {
+
+   private function clearImpactItems() {
```
cedric-anne: Instead of doing this, it is preferable to extends `DbTestCase` which is using transation to rollback every request done during a test on `afterTestMethod`.

inc/impact.class.php
```diff
@@ -0,0 +1,1099 @@
...
+               ]
+            ]
+         ]));
+      } else if ($isITILObject) {
+         // Tab name for an ITIL object : always 0
+         $total = 0;
+      }
+
+      return self::createTabEntry(
+         _n('Impact ', 'Impacts ', Session::getPluralNumber()),
```
cedric-anne: Why there is a whitespace on tabname ?

tests/units/Impact.class.php
```diff
@@ -0,0 +1,287 @@
...
+
+   public function testGetTabNameForItem_tabCountDisabled() {
+      $this->clearImpactItems();
+      $oldSession = $_SESSION['glpishow_count_on_tabs'];
+      $_SESSION['glpishow_count_on_tabs'] = false;
+
+      $impact = new \Impact();
+      $computer = new \Computer();
+
+      $this->string($impact->getTabNameForItem($computer))
```
cedric-anne: To be sure to put back `$_SESSION` to original state if test is failing, you should restore its value before doing asserts.
```
$tabname = $impact->getTabNameForItem($computer);
$_SESSION['glpishow_count_on_tabs'] = $oldSession;
$this->string($tabname)->isEqualTo("Impacts ");
```

This remark is valid for all tests.

tests/units/Impact.class.php
```diff
@@ -0,0 +1,287 @@
...
+      $impact = new \Impact();
+      $impactRelationManager = new \ImpactRelation();
+
+      // Get computer
+      $computer = getItemByTypeName('Computer', '_test_pc02');
+
+      // Create an impact graph
+      $impactRelationManager->add([
+         'itemtype_source'   => "Computer",
+         'items_id_source'   => "1",
```
cedric-anne: You should use real id value. It may happen (on developper machine) that db contains altered bootstrap data. In this case, your test may fail.

To get id of a computer, you can use `getItemByTypeName('Computer', '_test_pc01', true);` (third param is used to retrieve only id).

cedric-anne:
When displaying impact view for an element for the first time, I have a blank graph. If I click on `Add a new asset to the impact network`, I see the base element at the top left corner.

AdrienClairembault:
> When displaying impact view for an element for the first time, I have a blank graph. If I click on `Add a new asset to the impact network`, I see the base element at the top left corner.

This should be fixed now, I think we are ready to merge if there are no others issues concerning the first iteration.

cedric-anne:
> > When displaying impact view for an element for the first time, I have a blank graph. If I click on `Add a new asset to the impact network`, I see the base element at the top left corner.
> 
> This should be fixed now, I think we are ready to merge if there are no others issues concerning the first iteration.

It works now.

PR #5956, opened 2019-05-26
Master

perseushub:
new feature for Chinese local character more localization


trasher:
Translations are handled on https://www.transifex.com/glpi/GLPI/dashboard/

PR #6003, opened 2019-06-07
Improved display consistency for deleted Computer_Item objects

AdrienClairembault:
Fix regression, Computer_Item used to be filtered on is_deleted but this was lost during the raw queries migration.

Related : #4705 

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | pending


AdrienClairembault:
I've updated the description as we found the root cause of this issue (regression during queries migration).

The fix is now much smaller :)

PR #6029, opened 2019-06-14
Fixes issue on notifications where 'Ticket' was hardcoded

tomolimo:
In notifications for Changes, the text behind the ##lang.change.description## was 'Ticket: Description' instead of 'Change: Description'.
Same issue for Problem notifications.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #number

### DIFF
--- a/inc/notificationtargetcommonitilobject.class.php
+++ b/inc/notificationtargetcommonitilobject.class.php
@@ -1204,7 +1204,7 @@ abstract class NotificationTargetCommonITILObject extends NotificationTarget {
                     $objettype.'.url'                   => __('URL'),
                     $objettype.'.category'              => __('Category'),
                     $objettype.'.content'               => __('Description'),
-                    $objettype.'.description'           => sprintf(__('%1$s: %2$s'), __('Ticket'),
+                    $objettype.'.description'           => sprintf(__('%1$s: %2$s'), $this->obj->getTypeName(1),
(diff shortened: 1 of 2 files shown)


trasher:
Commit on Search class does not seem related

tomolimo:
Hello @trasher 
This has been a mishandling from my side. and I was not able to correct it.
I wanted to do 2 distinct PR, and unfortunately I pushed from the same branche.
Sorry for that,
Thank you,
Regards,
Tomolimo

trasher:
OK, I'll cherry pick the first commit only. By the way, I'll do it on 9.4/bugfixes branch 9.3 is no longer maintained.

tomolimo:
Thank you,
And for the other commit, can I do another PR on 9.4/BF then?
By the way, could it be possible to "close" (or prevent PR to them) the branches that are no longer maintained?

trasher:
I do not think we can close branches or prevent PR on them; we can just remove them as far as I know :/

For the another PR, if this is a bugfix, yes, it should target 9.4; but it seems this is more a feature, so it should reach 9.5/bugfixes.

trasher:
First commit merged into 9.4/bf, thank you for the fix

PR #6031, opened 2019-06-14
Allow to disable stats computation after a followup is added

AdrienClairembault:
This change was needed because the formcreator plugin need to add a followup to a ticket without triggering the take into account time computation.

Internal ref : 17135
Linked PR on formcreator repository: [#1425](https://github.com/pluginsGLPI/formcreator/pull/1425)

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | pending

### DIFF
--- a/tests/functionnal/ITILFollowup.php
+++ b/tests/functionnal/ITILFollowup.php
@@ -225,35 +225,44 @@ class ITILFollowup extends DbTestCase {
 
       // Normal behaviior, no flag specified
       $ticketID = $this->getNewITILObject('Ticket');
+      $this->integer($ticketID);
 
       $ITILFollowUp = new \ITILFollowup();
-      $ITILFollowUp->add([
+      $this->integer($ITILFollowUp->add([
          'date'                            => $_SESSION['glpi_currenttime'],
          'users_id'                        => \Session::getLoginUserID(),
          'content'                         => "Functionnal test",
          'items_id'                        => $ticketID,
          'itemtype'                        => \Ticket::class,
-      ]);
+      ]));
 
-      $ticket->getFromDB($ticketID);
+      $this->boolean($ticket->getFromDB($ticketID))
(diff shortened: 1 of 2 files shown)


trasher:
Also, this is a new feature, and therefore should reach 9.5/bugfixes branch; please rebase ;)

tests/functionnal/ITILFollowup.php
```diff
@@ -207,4 +207,53 @@ public function testUpdateAndDelete() {
...
+      ];
+
+      $_SESSION['glpiset_default_tech'] = 0;
+      $_SESSION['glpiset_default_requester'] = 0;
+
+      // Normal behaviior, no flag specified
+      $ticketID = $this->getNewITILObject('Ticket');
+
+      $ITILFollowUp = new \ITILFollowup();
+      $ITILFollowUp->add([
```
trasher: You should test result of all add, update, getFromDB etc; to make sure each step is working as expected

AdrienClairembault:
> Also, this is a new feature, and therefore should reach 9.5/bugfixes branch; please rebase ;)

While "technically" speaking it is indeed a new feature, this was specifically added to fix a bug of formcreator (see linked formcreator PR in description).

On it's own this change add no new feature to the user, it just fix something that we didn't know was not working as intended.

For this reasons, shouldn't this belong to the bugfixes branch instead ?

AdrienClairembault:
> Also, this is a new feature, and therefore should reach 9.5/bugfixes branch; please rebase ;)

While "technically" speaking it is indeed a new feature, this was specifically added to fix a bug of formcreator (see linked formcreator PR in description).

On it's own this change add no new feature to the user, it just fix something that we didn't know was not working as intended.

For this reasons, shouldn't this belong to the bugfixes branch instead ?

PR #6039, opened 2019-06-17
Saves and restores cron context during call of SavedSearch_Alert::cronSavedSearchesAlerts

tomolimo:
To be sure that logs will be in GLPI with default datetime and language
and that notifications are sent even if $_SESSION['glpinotification_to_myself'] is false
and to restore default cron $_SESSION and $CFG_GLPI global variables for next cron task

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #6037

### DIFF
--- a/inc/savedsearch_alert.class.php
+++ b/inc/savedsearch_alert.class.php
@@ -305,9 +305,9 @@ class SavedSearch_Alert extends CommonDBChild {
 
    /**
     * Summary of saveContext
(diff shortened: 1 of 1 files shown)


tomolimo:
me neither!

PR #6107, opened 2019-07-08
Use DB transactions in transfer process

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

If an exception occurs at any point during a transfer, the whole process should be rolled back to avoid data loss.


trasher:
As discussed off-gh; all queries that will be don on the after a transaction has started will be part of it. So all added/changed data are still available; and all queries will be commited/rolled-back at once :)

PR #6116, opened 2019-07-09
Add canViewItem check in API

AdrienClairembault:
Internal ref : 16937

- Add canViewItem check to the API 
- Fix the canViewItem of ITILFollowup which had an unused condition (comparing an object to the string "Ticket").

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | pending

### DIFF
--- a/tests/web/APIRest.php
+++ b/tests/web/APIRest.php
@@ -347,7 +347,7 @@ class APIRest extends APIBaseClass {
       $ticketTMF = new \TicketTemplateMandatoryField();
 
       $tt_id = $ticketTemplate->add([
-         'entities_id' => 0,
+         'entities_id' => 1,
          'name'        => 'test'
       ]);
       $this->boolean((bool)$tt_id)->isTrue();
--- a/tests/web/APIRest.php
+++ b/tests/web/APIRest.php
@@ -347,7 +347,7 @@ class APIRest extends APIBaseClass {
       $ticketTMF = new \TicketTemplateMandatoryField();
 
       $tt_id = $ticketTemplate->add([
-         'entities_id' => 1,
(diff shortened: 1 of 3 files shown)


AdrienClairembault:
About performance : for ITILFollowups we lose about 3 ms per canViewItem check.
Some of this time is recovered when the data is not allowed (skipping code thanks to the continue statement).

However one issue is that the range parameter is a lot less relevant : it is used to build the SQL query (LIMIT) and we only check if each item is valid AFTER the query is executed.
So for a range of 20 element we might get 15 allowed and 5 forbidden items, resulting in only 5 results instead of the 20 results required.

With the current framework of the API I am not sure this could be fixed, we would need to build queries specifically for each item types.

tests/web/APIRest.php
```diff
@@ -347,7 +347,7 @@ public function testGetItemsCommonDBChild() {
       $ticketTMF = new \TicketTemplateMandatoryField();
 
       $tt_id = $ticketTemplate->add([
-         'entities_id' => 0,
+         'entities_id' => 1,
```
trasher: You should not rely on a hardcoded value, but rather sue something like:
```php
$ent1 = getItemByTypeName('Entity', '_test_child_1', true);
```

AdrienClairembault: Done.

AdrienClairembault:
@orthagh @cedric-anne 

Following our discussion I found that the API was already using Search::addDefaultWhere.
This has no effect for followups because there was no "ITILFollowup" case defined, I think we just need to add it.

Moving to a new PR: #6150

PR #6127, opened 2019-07-12
To avoid PHP notice

yllen:
image


trasher:
See https://github.com/glpi-project/glpi/pull/6129

Issue #2226, opened 2017-05-22
Project: automatic %


yllen:
http://forum.glpi-project.org/viewtopic.php?id=157580

yllen:
I agree with this ask.

PR #6128, opened 2019-07-14
Automatically update project percent done

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2226

Add option to automatically calculate a project's percent done based on child projects and tasks.

I've also added a check to prevent circular parent-child relations for projects and project tasks. This would have caused an infinite loop when updating percent done, and already causes a loop when trying to view the GANTT chart.


inc/dbutils.class.php
```diff
@@ -1820,4 +1820,42 @@ function getItemtypeForForeignKeyField($fkname) {
       $table = $this->getTableNameForForeignKeyField($fkname);
       return $this->getItemTypeForTable($table);
    }
+
```
trasher: maybe better in CommonDBConnexity? ping @orthagh

orthagh: +1

trasher: I've just realized  that call is done from Project, so maybe it can be moved into CommonDBTM, or it can stay here.

cconard96: I moved it to CommonDBTM and squashed my commits.

inc/project.class.php
```diff
@@ -1409,11 +1430,27 @@ function showForm($ID, $options = []) {
...
+      ];
+      if ($this->fields['auto_percent_done']) {
+         $percent_done_params['specific_tags'] = ['disabled' => 'disabled'];
+      }
+      Dropdown::showNumber("percent_done", $percent_done_params);
+      $auto_percent_done_params = [
+         'type'      => 'checkbox',
+         'name'      => 'auto_percent_done',
+         'title'     => __('Automatically calculate'),
+         'onclick'   => "$(\"select[name='percent_done']\")[0].disabled = !$(\"input[name='auto_percent_done']\")[0].checked;"
```
cedric-anne:
```suggestion
         'onclick'   => "$(\"select[name='percent_done']\").prop('disabled') = !$(\"input[name='auto_percent_done']\").prop('checked');"
```

inc/projecttask.class.php
```diff
@@ -212,6 +212,28 @@ function post_getEmpty() {
...
 
+      if (isset($this->input['auto_percent_done']) && $this->input['auto_percent_done']) {
+         // Force recalculation of this and parents
+         $this->recalculatePercentDone();
+      }
+      // Update parent percent_done
+      if ($this->fields['projecttasks_id'] > 0) {
+         $task = new self();
+         $task->getFromDB($this->fields['projecttasks_id']);
+         error_log($this->fields['projecttasks_id']);
```
cedric-anne:
```suggestion
```

inc/projecttask.class.php
```diff
@@ -212,6 +212,28 @@ function post_getEmpty() {
...
+         error_log($task->fields['auto_percent_done']);
```
cedric-anne:
```suggestion
```

inc/projecttask.class.php
```diff
@@ -541,12 +575,28 @@ function projecttasktemplate_update(value) {
...
+      ];
+      if ($this->fields['auto_percent_done']) {
+         $percent_done_params['specific_tags'] = ['disabled' => 'disabled'];
+      }
+      Dropdown::showNumber("percent_done", $percent_done_params);
+      $auto_percent_done_params = [
+         'type'      => 'checkbox',
+         'name'      => 'auto_percent_done',
+         'title'     => __('Automatically calculate'),
+         'onclick'   => "$(\"select[name='percent_done']\")[0].disabled = !$(\"input[name='auto_percent_done']\")[0].checked;"
```
cedric-anne:
```suggestion
         'onclick'   => "$(\"select[name='percent_done']\").prop('disabled') = !$(\"input[name='auto_percent_done']\").prop('checked');"
```

inc/project.class.php
```diff
@@ -237,6 +237,25 @@ static function getAdditionalMenuLinks() {
...
+         // Update parent percent_done
+         $project = new Project();
+         $project->getFromDB($this->fields['projects_id']);
+         if ($project->fields['auto_percent_done']) {
+            $project->recalculatePercentDone();
+         }
+      }
+
+      if (isset($this->input['projects_id']) && $this->fields['projects_id'] > 0) {
+         // Recalculate previous parent percent done
```
cedric-anne:
```suggestion
         // Recalculate previous parent percent done
         $project = new Project();
```

inc/project.class.php
```diff
@@ -237,6 +237,25 @@ static function getAdditionalMenuLinks() {
...
+      } else {
+         // Update parent percent_done
+         $project = new Project();
+         $project->getFromDB($this->fields['projects_id']);
+         if ($project->fields['auto_percent_done']) {
+            $project->recalculatePercentDone();
+         }
+      }
+
+      if (isset($this->input['projects_id']) && $this->fields['projects_id'] > 0) {
```
cedric-anne:
```suggestion
      if ($this->oldvalues['projects_id'] > 0 && $this->oldvalues['projects_id'] != $this->fields['projects_id']) {
```

inc/project.class.php
```diff
@@ -237,6 +237,25 @@ static function getAdditionalMenuLinks() {
...
+         // Recalculate previous parent percent done
+         if ($project->getFromDB($this->fields['projects_id']) && $project->fields['auto_percent_done']) {
```
cedric-anne:
```suggestion
         if ($project->getFromDB($this->oldvalues['projects_id']) && $project->fields['auto_percent_done']) {
```

inc/project.class.php
```diff
@@ -237,6 +237,25 @@ static function getAdditionalMenuLinks() {
    function post_updateItem($history = 1) {
       global $CFG_GLPI;
 
+      if (isset($this->input['auto_percent_done']) && $this->input['auto_percent_done']) {
```
cedric-anne:
```suggestion
      if ($this->fields['auto_percent_done']) {
```

inc/projecttask.class.php
```diff
@@ -212,6 +212,40 @@ function post_getEmpty() {
    function post_updateItem($history = 1) {
       global $CFG_GLPI;
 
+      if (isset($this->input['auto_percent_done']) && $this->input['auto_percent_done']) {
```
cedric-anne:
```suggestion
      if ($this->fields['auto_percent_done']) {
```

inc/projecttask.class.php
```diff
@@ -212,6 +212,40 @@ function post_getEmpty() {
...
+         if ($this->fields['projects_id'] > 0) {
+            $project = new Project();
+            $project->getFromDB($this->fields['projects_id']);
+            if ($project->fields['auto_percent_done']) {
+               $project->recalculatePercentDone();
+            }
+         }
+      }
+
+      if (isset($this->input['projects_id']) && $this->fields['projects_id'] > 0) {
```
cedric-anne:
```suggestion
      if ($this->oldvalues['projects_id'] > 0 && $this->oldvalues['projects_id'] != $this->fields['projects_id']) {
```

inc/projecttask.class.php
```diff
@@ -212,6 +212,40 @@ function post_getEmpty() {
...
+         // Recalculate previous parent percent done
+         if ($project->getFromDB($this->fields['projects_id']) && $project->fields['auto_percent_done']) {
```
cedric-anne:
```suggestion
         if ($project->getFromDB($this->oldvalues['projects_id']) && $project->fields['auto_percent_done']) {
```

inc/projecttask.class.php
```diff
@@ -212,6 +212,40 @@ function post_getEmpty() {
...
+            $project->recalculatePercentDone();
+         }
+      }
+      if (isset($this->input['projecttasks_id']) && $this->fields['projecttasks_id'] > 0) {
```
cedric-anne:
```suggestion
      if ($this->oldvalues['projecttasks_id'] > 0 && $this->oldvalues['projecttasks_id'] != $this->fields['projecttasks_id']) {
```

inc/projecttask.class.php
```diff
@@ -212,6 +212,40 @@ function post_getEmpty() {
...
+         // Recalculate previous parent percent done
+         if ($task->getFromDB($this->fields['projecttasks_id']) && $task->fields['auto_percent_done']) {
```
cedric-anne:
```suggestion
         if ($task->getFromDB($this->oldvalues['projecttasks_id']) && $task->fields['auto_percent_done']) {
```

inc/project.class.php
```diff
@@ -1883,4 +1920,52 @@ static function showGantt($ID) {
...
+         ],
+         'FROM'   => ProjectTask::getTable(),
+         'WHERE'  => [
+            $this->getForeignKeyField()   => $this->getID()
+         ]
+      ]);
+      $union = new QueryUnion([$query1, $query2], false, 'all_items');
+      $iterator = $DB->request([
+         'SELECT' => [
+            new QueryExpression('CAST(AVG('.$DB->quoteName('percent_done').') AS INT) AS percent_done')
```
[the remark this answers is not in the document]
cconard96: This was the simplest idea I had. The second method was to weight each sub-project/task based on the planned duration relative to the sum of all child durations. The second option would require more computation (and maybe another DB field) and I don't know how to handle items with no planned start/end dates.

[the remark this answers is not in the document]
cedric-anne: As we cannot know if data is correctly filled by users, starting with a simple computation as you did may be a good option.
Maybe adding a tooltip next to the `auto_percent_done` field could be helpful to explain how it works (i.e. `Html::showToolTip(__('Percentage is computed based on ....'));`

inc/project.class.php
```diff
@@ -1409,12 +1438,33 @@ function showForm($ID, $options = []) {
...
+         'type'      => 'checkbox',
+         'name'      => 'auto_percent_done',
+         'title'     => __('Automatically calculate'),
+         'onclick'   => "$(\"select[name='percent_done']\").prop('disabled') = !$(\"input[name='auto_percent_done']\").prop('checked');"
+      ];
+      if ($this->fields['auto_percent_done']) {
+         $auto_percent_done_params['checked'] = 'checked';
+      }
+      $checkbox_id = "check_".mt_rand();
+      Html::showCheckbox($auto_percent_done_params, $checkbox_id);
```
cedric-anne:
```suggestion
      $auto_percent_done_params['id'] = $checkbox_id;
      Html::showCheckbox($auto_percent_done_params);
```

inc/project.class.php
```diff
@@ -1409,12 +1438,33 @@ function showForm($ID, $options = []) {
...
+      echo "<span class='very_small_space'>";
+      Html::showToolTip(__('Percentage is computed based on the average of all child project and task percent done'), [
+         'applyto'   => $checkbox_id
```
cedric-anne: This does not work, as the "real" checkbox is not displayed. I guess displaying the `i` icon is the best option here.

inc/project.class.php
```diff
@@ -237,6 +237,25 @@ static function getAdditionalMenuLinks() {
    function post_updateItem($history = 1) {
       global $CFG_GLPI;
 
+      $project = new Project();
+      if ($this->fields['auto_percent_done']) {
+         // Force recalculation of this and parents
+         $this->recalculatePercentDone();
```
cedric-anne: I have only one project in my test instance, and it loops here. Indeed, `recalculatePercentDone` calls `update` which calls `post_updateItem` and so on.

To prevent this behaviour, I propose to :
1. Have a `recalculatePercentDone` method that only return the new value (without trigerring itself the update);
2. Update the field in `preprareInputForUpdate` (`$input['percent_done'] = $this->recalculatePercentDone();`), but only when `auto_percent_done` field value changed;
3. Update the parent with new computed value in `post_updateItem` (`$parent->update(['id' => $parent->getID(), 'percent_done' => $parent->recalculatePercentDone()])`), but only if percent_done value changed for self;

I am not sure that will work, need to be tested.

inc/project.class.php
```diff
@@ -1409,12 +1438,33 @@ function showForm($ID, $options = []) {
...
+      ];
+      if ($this->fields['auto_percent_done']) {
+         $percent_done_params['specific_tags'] = ['disabled' => 'disabled'];
+      }
+      Dropdown::showNumber("percent_done", $percent_done_params);
+      $auto_percent_done_params = [
+         'type'      => 'checkbox',
+         'name'      => 'auto_percent_done',
+         'title'     => __('Automatically calculate'),
+         'onclick'   => "$(\"select[name='percent_done']\").prop('disabled') = !$(\"input[name='auto_percent_done']\").prop('checked');"
```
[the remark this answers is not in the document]
cconard96: This change doesnt make sense to me. Initial JS was OK, but the last suggestion caused the error.

inc/project.class.php
cconard96: I will need to revert some of the last round of changes, test on a clean install, and create unit tests. Recursion on the calculation should not be possible. Function triggers the parent to update only (after updating its percent done) which recalculates for its parent and so on.

inc/project.class.php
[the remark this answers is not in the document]
cconard96: Sorry, I didn't realize "prop" was a getter and setter. It makes sense now.

inc/project.class.php
cconard96: I made a mistake in one of the later commits (used fields instead of input). The loop should be fixed now.

cconard96:
I had to add a reference to the old projects and project tasks instead of using oldvalues. The oldvalues array gets cleared during the DB update, so it cannot be used in post_updateItem().

inc/project.class.php
```diff
@@ -1256,6 +1295,19 @@ function prepareInputForAdd($input) {
...
+         unset($input['percent_done']);
+      }
+      if (isset($input['projects_id']) && $input['projects_id'] > 0) {
+         if (self::checkCircularRelation($input['id'], $input['projects_id'])) {
+            Session::addMessageAfterRedirect(__('Circular relation found. Parent not updated.'), false,
+                                          ERROR);
+            unset($input['projects_id']);
+         }
+      }
+      if (isset($input['projects_id']) && ($input['projects_id'] != $this->fields['projects_id'])) {
```
cedric-anne: A PHP notice is triggered when defining a parent on a project that had no parent.
```suggestion
      if ($this->fields['projects_id'] > 0 && isset($input['projects_id'])
          && ($input['projects_id'] != $this->fields['projects_id'])) {
```
```
  *** PHP Notice(8): Undefined index: auto_percent_done
  Backtrace :
  inc/project.class.php:1968                         
  inc/project.class.php:252                          Project::recalculatePercentDone()
  inc/commondbtm.class.php:1433                      Project->post_updateItem()
  front/project.form.php:94                          CommonDBTM->update()
```

trasher:
@cconard96 could you please rebase to fix conflicts? If you could also squash all commits, it would be perfect ;)

PR #6183, opened 2019-07-30
Add kanban view to projects

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

Add Kanban view for projects.

Related:
[Note on GitHub roadmap](https://github.com/glpi-project/glpi/projects/2#card-16300055)
[UserEcho Request](https://glpi.userecho.com/communities/1/topics/538-kanban-view)

- [x] Display card counts per board (Needs to dynamically update when dragging or adding items)
- [x] Add items from Kanban
- [x] Show user/team badges on cards
- [x] Add type, milestone, and progress to card content
- [x] Global Kanban view
- [x] Collapsible boards
- [x] Plugin hooks for adding content to cards before and after the main content (Specifically thinking about the Tags plugin).

Screenshot from 2019-09-17 04-32-09
Screenshot from 2019-09-17 04-32-35


orthagh:
Some UI / UX remarks:

MUST:
- cards in columns should fit their parent's width.
- add some border radius to cards and columns.
- (not for darker theme) cards should have white bg with a soft border.
- count cards in columns.
- add a button to create new cards directly from columns.
- add padding to cards.
- Cards titles should be black (with underline on hover).
- I dont think we need task's content displayed (title  assignment should be enough, we can consider step, type and milestone field )

GOOD TO HAVE
- collapsible and movable columns
- display user mini picture when assigned to a task (we should have also an alternative mini badge with user's initials when no picture available)
- add progress bar to display the "percent done" field


For some examples, check
- [github project view](https://github.com/glpi-project/glpi/projects/2)
- [gitlab board](https://gitlab.com/gitlab-org/gitlab-ce/-/boards?milestone_title=10.0&)

---
By the way, thank you for starting this work.

Technically, nothing much to say for the moment, some additional remarks:
- It could great to have the possibility to re-use the kanban view in some other place, (like ticket child-parent view for example). So if we have a generic kanban component distinct from project implementation, we could follow by developing other view once this pr merged.
- More i think for projects, we will need a global (inter-project) view like we have for gantt view (pretty much dated also...)
image
- Maybe separate the css for kanban in its own file (need to have @canne pov for this point)
- for your js file, try to have an object including your current function

cconard96:
@orthagh Ready for UI and technical review. Global view works, but is not linked anywhere. I wasn't sure where it should go since there is already an item for 'Summary'.

orthagh:
Hello.
I'm returning from vacation, so here is a fast review for today:

First, good job, it's a big enhancement comparing to previous iteration.

**UI:**
- add `pointer` class to folding and add icons
- headers should take the state color, see an example from gitlab:
image
- soften separators between headers and content 
- add a (very) soft border to cards
- align cards content to left (also their titles) and add more padding to cards
- user initials should be uppercase
- adding new card should have the same ui (white bg, rounded corner, etc)
- i'm not satisified with the curent progress bar design, but i don't have suggestions atm.

**Backend:**

I have this error when trying to add a new task:
```
[2019-09-02 14:42:42] glpisqllog.ERROR: DBmysql::query() in /var/www/html/glpi/9.5-git/inc/dbmysql.class.php line 296
  *** MySQL query error:
  SQL: SELECT * FROM `glpi_items_projects` WHERE `projecttasks_id` = '10'
  Error: Unknown column 'projecttasks_id' in 'where clause'
  Backtrace :
  inc/dbmysqliterator.class.php:95                   
  inc/dbmysql.class.php:850                          DBmysqlIterator->execute()
  inc/dbutils.class.php:441                          DBmysql->request()
  inc/db.function.php:258                            DbUtils->getAllDataFromTable()
  inc/notificationtargetprojecttask.class.php:447    getAllDataFromTable()
  inc/notificationtarget.class.php:1214              NotificationTargetProjectTask->addDataForTemplate()
  inc/notificationtemplate.class.php:253             NotificationTarget->getForTemplate()
  inc/notificationeventabstract.class.php:108        NotificationTemplate->getTemplateByLanguage()
  inc/notificationevent.class.php:163                NotificationEventAbstract::raise()
  inc/projecttask.class.php:236                      NotificationEvent::raiseEvent()
  inc/commondbtm.class.php:1100                      ProjectTask->post_addItem()
  ajax/kanban.php:83                                 CommonDBTM->add()
```
- 400

orthagh:
for summary add new icon temporary, we will decide later what to do with the old one

orthagh:
a new suggestion also:

- if possible, add a filtering input in top of the page, to filter displayed cards.
- And on the right of this input add a button to add new list (task state)

Ex on gitlab:
image

They also have a dropdown to switch boards, maybe on global view we can have the same to filter projects

orthagh:
Review for today iteration, we're very close on design i think.

**UI:**

- align column title (state name) on the left and align number of cards on the right
- add a soft border to columns
- hovering cards should change the pointer to `grab`
- review the filter input css, something like that:
```css
    font-size: 14px;
    margin: 10px 0 15px;
    border-radius: unset !important;
    padding: 5px 10px;
```
- css for inputs in adding form:
```css
    border-radius: unset;
    margin: 5px 0;
    padding: 5px;
```
- in the adding form add a close control in top right
- `.kanban-item-header` (this only one in the class list) have too much vertical padding, i suggest: `8px 10px 0px`

**Good to have (not mandatory):**

- dragging card should change the cursor to `grabbing`
- Make columns draggable 
- I think we should change default colors for project states (in empty.sql), here is a suggestion:
  * New: #a8d695
  * Processing: #f0ad4e
  * Closed : #958282

cconard96:
For dragging columns, would the order be stored client-side or in the DB (in project table)?

orthagh:
i think DB is better in this case but cookie is possible.
If you attend to do this part, store also folding state.

Data must be stored by user

orthagh:
Hello @cconard96 

I don't know if the current branch is still in wip since the previous review.
I encounter some issues :)

- I can't move anymore task between lists
- folded lists overflows outside the containers and the toggle button moved
image


For new features:

- I can add multiple time the same status (and the second column fail so to display associated tasks). I think the Dropdown::show... should have the `used` key to remove used entries.
- We miss this one, i think, we can't remove a column, ex for gitlab:
image
- button for adding list, should have a real text  (named Add list)
- the input for filtering should have a placeholder "Search of filter results"
- the ux to add a new status is not clear  (and fails with a js error). Here is an gif example how gitlab achieves this need:
gitlab_create_new_list

cconard96:
>     * folded lists overflows outside the containers and the toggle button moved

It seems it breaks on smaller screens or for columns with longer names. I do not have a solution for this at the moment as no fix seems to work right for it.

Nothing fails with an error here (or without), but I will re-test on a clean 9.5 base later.
For text, I am working on getting localized text over Ajax.

orthagh:
Ok, not so important atm, when things will be stabilized, i'll take this issue in charge myself.

cconard96:
I dropped support for tasks without a parent project (No adding them on global view) due to issues with right checks. It didn't seem logical to have a project task without a project as well...

I redid the add/create column UI to look more like GitLab's and added the other suggestions.

I tested on a clean install of this branch and it appeared to work properly for me (except the CSS issue with collapsed columns).

orthagh:
I showed to my colleague @flegastelois , and it's great for us, very good job !!!

We didn't found the way to display task in the global view, any hint ?

I'll do some additional css and it will be ready for technical review (ping @trasher and @cedric-anne )

ajax/getUserPicture.php
```diff
@@ -0,0 +1,89 @@
...
+include ('../inc/includes.php');
+
+header("Content-Type: application/json; charset=UTF-8");
+Html::header_nocache();
+
+Session::checkLoginUser();
+
+if (!isset($_GET['users_id'])) {
+   // Bad request
+   http_response_code(400);
```
trasher: Maybe add an entry in logs saying that's missing?

trasher: Same remark for all other missing/bad data

inc/projecttask.class.php
```diff
@@ -350,6 +350,27 @@ static function getAllForProject($ID) {
...
+    * @since 9.5.0
+    * @param $ID        integer  Id of the project task
+    *
+    * @return array of tasks ordered by dates
+   **/
+   static function getAllForProjectTask($ID) {
+      global $DB;
+
+      $tasks = [];
+      foreach ($DB->request('glpi_projecttasks',
```
trasher: Avoird calling method in loop, better to store it in a variable before

trasher:
Quickly tested, this seem to work :)

orthagh:
Did a commit for UI polishing, rotations now correctly displayed. For your information, the rotation need to be done on text+icon, not on the container (it's not impossible to do it on container, but much easier on sub element).

I took the opportunity to do some additional minor fixes (cursor, spacing on forms, review headers background, review filter bar).

orthagh:
screenshot after 
projects_kanban

orthagh:
By the way, i didn't have any issues like while the previous review, seems good for me so (functional -> OK)

prybak:
Awesome :).
Any chances that it will be propagated to Changes/Problems and in future as an alternative to ticket list?

cconard96:
I made it as generic as possible, so it could be done. @orthagh what is your opinion for this?

orthagh:
Main goal, we mainly need this feature in projects (i agree for itil objects btw, still with less priority) and we should merge before to keep the pr and the review easier.

After that, we need to finish / review the opened pr for 9.5 as we don't want to repel the release date forever. 

So depends on complexity to add this view to itil objects: 

- easy -> go for it in 9.5.
- hard -> wait after release (rather, we will merge after release).

cconard96:
After a review of ITIL Object code, I believe it best to wait for the next version. Adding the Kanban view itself is fairly easy, but there was no clear way of restricting the visibility of tickets, changes or problems like there was with Projects (no getVisibilityCriteria).

orthagh:
One more review from @cedric-anne (he needed to move out so i transfer his words)

- Cards order in columns should be saved (important in scrum when the product owner do triage)
- Adding a new card should be at the place where the form was initially (currently the card is moved to the bottom), maybe the form should be added at the bottom and launch a scrollto event
- We need to display tasks (not only projects) in global view. It's an interesting view to have all data at the same time.
- Do saves of kanban views are shared between users ? (maybe my fault if not) We didn't test for this review, but we we think a board should be the same for all (product owner and technicians).

cconard96:
Kanban views are per user as originally requested. Do we link the view to the Project itself and only let the managing group/user change it? This would include all state data (folded columns, column order, etc).

I will look at the rest tonight. I think for tasks without a project, some changes need made to canUpdateItem right checks because they were failing with these tasks. After that, I can re-add them as a supported itemtype on the global view.

cconard96:
I think I initially misunderstood the requirement of having tasks in the global view. Please let me know if the sample below is not what you wanted.
Screenshot from 2019-09-19 20-08-15
Note that subtasks and subprojects of a project that is finished are hidden to avoid a lot of clutter.

Right now the card order is being saved, but the restore isn't implemented yet. Also, views are still per-user as I didn't have time to switch them over either.

orthagh:
Hello @cconard96 

Thank you for your patience.

> Note that subtasks and subprojects of a project that is finished are hidden to avoid a lot of clutter.

Can you explain how you define a finished card ? For global view, i think we don't want any filter. 

We discussed again with @cedric-anne this morning about visibility of projects and global board.

* For projects -> save must be done on projects_id, views must be shared between users of this project (visibility restriction is already done by entities or membership of the project).
* For global view, it's harder. For scrum method, it doesn't add any value as you manage the boards project by project (and you can have a parent project to have a "meta" view). We think it's much a personal view where you can find all the tasks and project you have access.  We think this view cannot be shared between users, because we'll have some issues with not displayable projects or tasks and also entities mismatch. Maybe it's better for this global view to keep the save on user's id.

> only let the managing group/user change it?

Yes that's the plan.
Manager could be considered as "Product owner".
He have the right to change the view, ordering vertically the cards.

Other users (in the project's team) should only have the right to move cards between columns (and place it at the bottom, if it's possible with sortable lib).
We can also say a "simple" user has the right to add cards (only task i think) only in "No status" column and the manager will move after the cards (the manager has the right to create anywhere).

orthagh:
added a little commit to do a proper placeholder when moving columns:
image

orthagh:
3 more commits:

- expand grab cursor to the whole column header
- review add colum form (on this one, a second click on main button to display the form move the form to the right, i didn't find why)
- moved the new card to 5px bottom (precedence of css rules was broken) 

By the way, when trying to find why the add column form move to the right, i found you have
```$(self.add_column_form).on('click', '.kanban-create-column', function() {``` event register.
I didn't find where this code apply, did i miss something or is it dead code ?

**Comment by cconard96** [comment]
> Can you explain how you define a finished card ?
Any task or project whose parent has a state where "is_finished" is true. Parent projects are always visible regardless of finished state. I don't know about others, but we have a lot of closed projects with a lot of tasks. We don't need to see tasks from a 4 year old project when looking at a global view.
Perhaps this is just an issue with how our IT department uses Projects. Some projects we have are generalized like "Summer 2019" with a bunch of more specific Projects under it. Since we are a school district with limited number of technicians, we have to be able to efficiently coordinate all work that needs done while students are away. Most of our large migrations and changes happen during this 3 month period.

I will remove the limitation if you want.

**Comment by orthagh** [comment]
> Any task or project whose parent has a state where "is_finished" is true

I wasn't aware of this field, it's ok for me. You can keep this hidden filter.

**Comment by cconard96** [comment]
> By the way, when trying to find why the add column form move to the right, i found you have
> `$(self.add_column_form).on('click', '.kanban-create-column', function() {` event register.
> I didn't find where this code apply, did i miss something or is it dead code ?

Applies to this button:
Screenshot from 2019-09-27 15-21-56

**Comment by orthagh** [comment]
Hello @cconard96 

do you have an eta for this pr ?

Some issues on current work : 

- could you rebase and fix conflicts ?
- i can't get the global view working. Hints, my projects don't have any status set at the moment but i think it's also related to columns edition issue (see below)
- in the switcher, i see a project where i'm not actor nor manager and when i switch on it, i have a forbidden
- it seems columns are not editable anymore:
  - adding do nothing
  - deletion are done on dom but not saved
- i saw you add manager/actor right filtering, we talked with @cedric-anne about let only the manager move cards vertically, do you think it's possible (not mandatory btw)

If you need more from me, don't hesitate.

**Comment by cconard96** [comment]
@orthagh I made multiple fixes and rebased. If there aren't any further issues, I believe this is done now.

**Comment by orthagh** [comment]
i did a bunch of additional tests, and found some minors issues, most relative to the last points i mentioned : 

- forbidden projects in top left dropdown (issue triggered without the Project READALL right)
- global view issues :
   - realname/name difference between contact and user
   - empty in (if projects has no states)
- sql error on ambiguous projects_id column name (i think it triggered on subproject cards in any views)
- clean unused vars

**Comment by orthagh** [comment]
For the last one:

The cards order was correctly saved in `glpi_items_kanbans` but not restored in the saved order.

the `state` var [here](https://github.com/glpi-project/glpi/pull/6183/files#diff-afb73a35def00b53eb2fdda38c2cec2cR1422) contains the correct order but later when doing a [loadColumn](https://github.com/glpi-project/glpi/pull/6183/files#diff-afb73a35def00b53eb2fdda38c2cec2cR1434) the `self.user_state.state` was not set yet so it loads with default order.

Now it's ok for me, i didn't find anymore issues.
@cconard96 could you please validate my last commits and we will be done and ready for merge.

Btw, thank you very much for this work.

**Comment by cconard96** [comment]
Works for me

**Comment by trasher** [comment]
I'm testing on my side, in touch with @orthagh ; I'll merge soon ;)

**Comment by cconard96** [comment]
Before merging, do you think the background refresh should be on by default? It shouldn't be disruptive (doesnt refresh the full page and wont refresh if the user is sorting) but I left it disabled by default. It is a user preference.

**Comment by orthagh** [comment]
> do you think the background refresh should be on by default

We already have an option for auto-refreshing list of tickets:
image

Maybe we can merge the new option with the old one and re-label it, something like:

> Automatically refresh data (tickets list, project kanban) in minutes.

What do you think.

One last point also before merge, we encounter some strange issues with @trasher:

- on his side, he didn't have any save done on actions he did on kanban (could you confirm @trasher)
- on my side, delay of saving is very long (like 5s before trigger).

Do you have any hints for that @cconard96 ?

**Comment by cconard96** [comment]
>     * on my side, delay of saving is very long (like 5s before trigger).

Background save interval is set to 10 seconds. It used to send the full state immediately after each change, but I was concerned about network load especially for larger projects or global view. I changed it to check every 10 seconds if the state was changed, and if it was, send the state to the server. Also, all unsent changes should be immediately sent when navigating away.
This is configured in the "background_save_interval" variable in kanban.js.
We can probably reduce it to 3 seconds to make it feel more responsive when collaborating.

**Comment by orthagh** [comment]
> but I was concerned about network load especially for larger projects or global view

Make sense. I agree for reducing it to 3s

**Comment by trasher** [comment]
Indeed, I have issues on this feature :/

On the global view, changing cards (from processing to closed for example) works as expected.
Tasks order seems to work (on the screen this is correct), but this is lost wen reloading the page (event after I've seen a POST call to ajax/kanban.php in my browser) :/

on the Kanban tab, I can drag elements, but they're never dropped and goes back to their original emplacement.

I'm using firefox.

**Comment by cconard96** [comment]
> on the Kanban tab, I can drag elements, but they're never dropped and goes back to their original emplacement.

Only project managers can move or order cards when not in global view. Otherwise I don't know what this issue is.

File: ajax/kanban.php [hunk 3]
```diff
@@ -0,0 +1,215 @@
...
+
+if (isset($_REQUEST['action']) && $_REQUEST['action'] === 'get_translated_strings') {
+   header("Content-Type: application/json; charset=UTF-8", true);
+   echo json_encode((Kanban::getLocalizedKanbanStrings()), JSON_FORCE_OBJECT);
+   return;
+}
+
+$nonkanban_actions = ['update', 'add_item'];
+if (!isset($_REQUEST['action']) || !isset($_REQUEST['itemtype']) ||
+   (!in_array($_REQUEST['action'], $nonkanban_actions) && !class_uses($_REQUEST['itemtype'], 'Kanban'))) {
```
cedric-anne: `class_uses` is not used correclty here.
```suggestion
   (!in_array($_REQUEST['action'], $nonkanban_actions) && !in_array('Kanban', class_uses($_REQUEST['itemtype'])))) {
```

ajax/kanban.php
```diff
@@ -0,0 +1,215 @@
...
+      return;
+   }
+   if ($_REQUEST['items_id'] == -1) {
+      echo $itemtype::getFormURL(true).'?showglobalkanban=1';
+      return;
+   }
+   $item->getFromDB($_REQUEST['items_id']);
+   $tabs = $item->defineTabs();
+   $tab_id = array_search(__('Kanban'), $tabs);
+   if (is_null($tab_id)) {
```
cedric-anne:
```suggestion
   if (is_null($tab_id) || false === $tab_id) {
```

cedric-anne:
I do some functionnal tests and I had some bugs and unexpected behaviours:

1) On both project and global Kanban, when I refresh the page, badges colors are changing. It may be confusing.

2) On both project and global Kanban, when I move some tasks inside a column, order seems to be saved correctly, but on rendering, cards order is incorrect.

3) On global Kanban only, when I hide a column that contains cards, I am not able to  display it anymore.

4) On both project and global Kanban, I am able to move the "no status" column. Should we allow this ?

5) On project Kanban, if I am not the manager of the project, I cannot change task status. I agree with the fact that only managers should be able to change order inside a column (as order correspond to a priority defined by project manager), but I should be able to change the status of a task that is assigned to me.
Also, the drop placeholder should not be shown if dropping the card is not possible.

cconard96:
1. Options for consistent badge colors:
- Randomize on creation of users, suppliers, etc and store in DB. Allow users to set these colors?
- Create the badge client-side and cache using browser local storage. Could take up a lot of space client-side.
- Compute the color code based on fields in DB. Not sure how we could do this without having a lot of badges with nearly the same color.
Option 1 is probably the best.

2. I need to take another look at how the user state is saved and loaded. The current implementation seems unstable.

3. I think I just fixed this in my last commit.

4. I'll add an option to block sorting columns with the protected flag (same flag that prevents deletion).

5. I agree and this would make more sense for users. They could update the status from the item form but not the Kanban?

cconard96:
Re-wrote how the user state is handled and I am more confident in its reliability than before. Needs a bit more testing, but seems functional.

Badge colors are now stored in session storage.

Made it easier to identify invalid locations for cards. If a user moves a card without the ability to order cards, it automatically gets moved to the bottom of the column.

Please let me know if any of the previous issues you had still occur.

orthagh:
On first kanban display :

```log
Fatal error: Uncaught Error: Call to undefined method Project::canMoveKanbanCard() in /var/www/html/glpi/9.5-git/inc/project.class.php:2442 Stack trace: 
#0 /var/www/html/glpi/9.5-git/inc/project.class.php(163): Project::showKanban('2') 
#1 /var/www/html/glpi/9.5-git/inc/commonglpi.class.php(587): Project::displayTabContentForItem(Object(Project), '3', '') 
#2 /var/www/html/glpi/9.5-git/ajax/common.tabs.php(92): CommonGLPI::displayStandardTab(Object(Project), 'Project$3', '', Array) 
#3 {main} thrown in /var/www/html/glpi/9.5-git/inc/project.class.php on line 2442
```

Did you miss to push something ?

orthagh:
i tried to reset my previous view (by dropping the `glpi_items_kanbans` table)
When i go on a existing project with tasks and statuses,  all columns are displayed.

We should have only "no status" state displayed in this case, no ?

Also all my actions doesn't seems to store data in the previous table (and so nothing is saved).
Maybe you miss to address the post update/install state of glpi and you are currently working on an existing state when developping. 

Additional logs i found also:

**after display**
```
[2019-10-18 10:58:09] glpisqllog.ERROR: DBmysql::query() in /var/www/html/glpi/9.5-git/inc/dbmysql.class.php line 297
  *** MySQL query error:
  SQL: SELECT `date_mod`, `state` FROM `glpi_items_kanbans` WHERE `users_id` = '0' AND `itemtype` = 'Project' AND `items_id` = '2'
  Error: Unknown column 'date_mod' in 'field list'
  Backtrace :
  inc/dbmysqliterator.class.php:95                   
  inc/dbmysql.class.php:851                          DBmysqlIterator->execute()
  inc/item_kanban.class.php:114                      DBmysql->request()
  ajax/kanban.php:203                                Item_Kanban::loadStateForItem()
  {"user":"2@LU002"} 
[2019-10-18 10:58:11] glpisqllog.ERROR: DBmysql::query() in /var/www/html/glpi/9.5-git/inc/dbmysql.class.php line 297
  *** MySQL query error:
  SQL: INSERT INTO `glpi_items_kanbans` (`date_creation`, `itemtype`, `items_id`, `users_id`, `state`, `date_mod`) VALUES ('2019-10-18 10:58:11', 'Project', '2', '0', '{"0":{"column":"0","folded":"false","cards":{"0":"Project-5","1":"ProjectTask-9","2":"ProjectTask-12","3":"ProjectTask-13","4":"ProjectTask-14","5":"ProjectTask-15","6":"ProjectTask-16","7":"ProjectTask-17","8":"ProjectTask-18","9":"ProjectTask-19"}},"1":{"column":"1","folded":"false","cards":{"0":"ProjectTask-7","1":"ProjectTask-10"}},"2":{"column":"2","folded":"false","cards":{"0":"ProjectTask-2"}},"3":{"column":"3","folded":"false","cards":{"0":"ProjectTask-8","1":"ProjectTask-11"}},"4":{"column":"10","folded":"false","cards":{"0":"ProjectTask-20"}},"5":{"column":"11","folded":"false","cards":{"0":"ProjectTask-21"}}}', '2019-10-18 10:58:11')
  Error: Unknown column 'date_creation' in 'field list'
  Backtrace :
  inc/dbmysql.class.php:1120                         
  inc/item_kanban.class.php:85                       DBmysql->insert()
  ajax/kanban.php:198                                Item_Kanban::saveStateForItem()
  {"user":"2@LU002","mem_usage":"0.000\", 8.13Mio)"} 
```

cconard96:
The default view with no state is supposed to show all columns with a Project/Task. For me, I see this behavior. States in the DB may need dropped and re-created. The old states may not work 100% with some of the changes I made.

Yes, I missed a rename and didn't push the update changes. Everything else is up to date here.

orthagh:
indeed, global behavior is better with this part https://github.com/glpi-project/glpi/pull/6183/commits/11c624c42bdb621f0ee29659f613e151966a61d6#diff-cbe7eb957448e96608cadb0781f4a472R830.

cedric-anne:
Hi @cconard96 ,

I added a commit to improve sorting behaviour.
1. When trying to reorder cards into a column, without having right for it, card was not put back at initial place. I fixed it.
2. When someone move a card from a column to another, if he does not have to order cards, placeholder is now always displayed at the bottom of the column.
3. When trying to reorder cards into a column, without having right for it, placeholder is not displayed with "invalid-position" if position is same as initial.

If my commit is OK for you, then we should be ready to merge.

Regards

cconard96:
Looks good to me. Thank you.

PR #6204, opened 2019-08-07
Fix HAVING in search

AdrienClairembault:
Search that use an aggregate function (e.g. "Computer - number of CPU") lead to incorrect results as the condition is split between the `WHERE` and `HAVING` clause.

For example, if we search for "A = count(x) OR b = 1", the generated request will be like this:
```sql
SELECT ...
FROM ...
WHERE b = 1
GROUP BY ...
HAVING a = count(x)
```
To be correct, the request need to be like this instead:
```sql
SELECT ...
FROM ...
GROUP BY ...
HAVING b = 1 OR a = count(x)
```

The proposed changes do 3 actions to get the correct request:
- When building the request, add the `HAVING` conditions directly in the `WHERE` clause
- Once the `WHERE` is fully built, copy it into into the `HAVING` clause then remove it
- Check that every fields specified into the `HAVING` clause exists in the `SELECT` clause (and add them if not)

The last action in needed because: 
>The SQL standard requires that `HAVING` must reference only columns in the `GROUP BY` clause or columns used in aggregate functions. However, MySQL supports an extension to this behavior, and permits `HAVING` to refer to columns in the `SELECT` list and columns in outer subqueries as well.

**This also mean that this will only work with MySQL.**

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

### DIFF
--- a/inc/search.class.php
+++ b/inc/search.class.php
@@ -532,6 +532,7 @@ class Search {
 
          if (isset($criteria['field'])
             && isset($searchopt[$criteria['field']]["usehaving"])
+            || (isset($criteria['meta']) && $criteria['meta'] && $criteria['link'] == "AND NOT")
          ) {
             return true;
          }
@@ -1045,7 +1046,6 @@ class Search {
                      $criterion['searchtype'],
                      $criterion['value']
                   );
-
                   $sql .= $new_where;
                   continue;
                }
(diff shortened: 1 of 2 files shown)


trasher:
Fixes https://github.com/glpi-project/glpi/issues/6225

PR #6203, opened 2019-08-06
Add supervisor search option for User

btry:
Users have a new supervisor (or Responsible) property; but the search option is missing

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #number

PR #6208, opened 2019-08-08
Fix self join with linkfield

AdrienClairembault:
A very rare type of searchOption fail with the currently fail because of different alias in the `LEFT JOIN` and `WHERE` clauses (see #6203).

These changed introduce a new flag `is_self_join` that resolve this problem and make sure the correct alias is used in the `WHERE` clause.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

### DIFF
--- a/inc/search.class.php
+++ b/inc/search.class.php
@@ -3918,7 +3918,7 @@ JAVASCRIPT;
       $inittable = $table;
       $addtable  = '';
       if (($table != 'asset_types')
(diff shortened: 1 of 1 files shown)


AdrienClairembault:
As we discussed, this should be fixed directly in #6203 as it highly specific to this new searchOption.
There is no need to keep this one open.

PR #6209, opened 2019-08-08
Always use left join for meta criteria

AdrienClairembault:
The current behavior for meta criteria is that they are added with a "LEFT JOIN" only if the link method is NOT or AND NOT (else it is an INNER JOIN).

Why is this needed ? Shouldn't all JOINs be LEFT JOIN ?

Example of the current behavior and why is it not working as expected: 
image

The generated request will have `INNER JOIN glpi_monitors` which mean that the "OR" works like an "AND" instead (if the row doesn't match the monitor criteria then it is excluded from the mysql result set which is not what we with an "OR" ...).

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

### DIFF
--- a/inc/search.class.php
+++ b/inc/search.class.php
@@ -1140,7 +1140,7 @@ class Search {
          if (!in_array(getTableForItemType($m_itemtype), $already_link_tables)) {
             $FROM .= self::addMetaLeftJoin($data['itemtype'], $m_itemtype,
                                            $already_link_tables,
-                                           $criterion['value'] == "NULL" || strstr($criterion['link'], "NOT"),
+                                           true,
                                            $sopt["joinparams"]);
          }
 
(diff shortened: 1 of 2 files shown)


inc/search.class.php
```diff
@@ -1140,7 +1140,7 @@ static function constructAdditionalSqlForMetacriteria($criteria = [],
          if (!in_array(getTableForItemType($m_itemtype), $already_link_tables)) {
             $FROM .= self::addMetaLeftJoin($data['itemtype'], $m_itemtype,
                                            $already_link_tables,
-                                           $criterion['value'] == "NULL" || strstr($criterion['link'], "NOT"),
+                                           true,
```
trasher: The `addMetaLeftJoin` method seems to be used from `Search` only, I guess cwe can entirely remove the parameter that is now hardcoded, and clean the `addMetaLeftJoin` method a bit.

orthagh: +1 for removal

PR #6214, opened 2019-08-09
allow joins between a table and itself

btry:
Tyhis is an attemptp to make useable the responsible of an user with the search engine.

It includes the missing search option of his PR: https://github.com/glpi-project/glpi/pull/6203 

I successfully access the responsible user in a list and also succeeded to do a simple search with this data.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #number

### DIFF
--- a/inc/search.class.php
+++ b/inc/search.class.php
@@ -3395,6 +3395,12 @@ JAVASCRIPT;
          $addtable .= "_".$searchopt[$ID]["linkfield"];
       }
 
+      if ($table == getTableForItemType($itemtype) && isset($searchopt[$ID]['joinparams']['jointype'])
+          && $searchopt[$ID]['joinparams']['jointype'] == 'reflexive') {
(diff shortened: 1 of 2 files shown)


inc/search.class.php
```diff
@@ -4027,6 +4033,11 @@ static function addWhere($link, $nott, $itemtype, $ID, $searchtype, $val, $meta
 
          if (!empty($complexjoin)) {
             $table .= "_".$complexjoin;
+         } else {
```
btry: I guess the else is not necessary, but i preferred to be safe

trasher: you should use a `elseif` and avoid unnecessary level

orthagh:
replaced by #6260

PR #6215, opened 2019-08-09
Add defaultWhere for TicketTask

AdrienClairembault:
Internal ref : 16937.
Following #6150, this does the same but for TicketTask

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

### DIFF
--- a/inc/itilfollowup.class.php
+++ b/inc/itilfollowup.class.php
@@ -1088,9 +1088,9 @@ class ITILFollowup  extends CommonDBChild {
     */
    public static function buildParentCondition(
       $itemtype,
-      $target,
-      $user_table,
-      $group_table
+      $target = "",
+      $user_table = "",
+      $group_table = ""
    ) {
       // An ITILFollowup parent can only by a CommonItilObject
       if (!is_a($itemtype, "CommonITILObject", true)) {
@@ -1103,10 +1103,6 @@ class ITILFollowup  extends CommonDBChild {
       if (Session::haveRight($itemtype, $itemtype::READALL)) {
          return "(`itemtype` = '$itemtype') ";
       }
-
-      $requester = CommonITILActor::REQUESTER;
-      $assign    = CommonITILActor::ASSIGN;
-      $obs       = CommonITILActor::OBSERVER;
 
       $user   = Session::getLoginUserID();
       $groups = "'" . implode("','", $_SESSION['glpigroups']) . "'";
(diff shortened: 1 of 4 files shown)


inc/tickettask.class.php
```diff
@@ -83,7 +83,7 @@ function canViewItem() {
          return false;
       }
 
-      if (Session::haveRightsOr(self::$rightname, [parent::SEEPRIVATE, parent::SEEPUBLIC])) {
+      if (Session::haveRight(self::$rightname, parent::SEEPRIVATE)) {
```
AdrienClairembault: I've also added this change as the current check doesn't seem correct (unless I am missing something ?).

inc/itilfollowup.class.php
```diff
@@ -1123,73 +1119,7 @@ public static function buildParentCondition(
...
-            // Add tickets where the users is the validator
-            // Subquery for validator
-            $validation_query = "SELECT `$target`
-               FROM `glpi_ticketvalidations`
-               WHERE `users_id_validate` = '$user'";
-            $condition .= "OR `items_id` IN ($validation_query) ";
-         }
-
-         return $condition .= ")) ";
+         return $condition . Ticket::buildCanViewCondition("items_id") . ")) ";
```
trasher: Does not seems correct to use `Ticket` here

inc/tickettask.class.php
trasher: Dunno if the original code is working or not; but it sounds correct. I'm not really ok with this change

AdrienClairembault: The original code is checking for either public or private right to allow acces to ALL tasks.

The suggested changes make it check only for private rights (this is the way it's done for followups in ITILFollowup::CanViewItem).

inc/itilfollowup.class.php
AdrienClairembault: What would you use instead ?
This is building the condition for ITILFollowups whose parents are Tickets.
The new code for TicketTask needed to use it too (every TicketTask have a parent Ticket) so it seemed right to export in into Ticket.

inc/tickettask.class.php
trasher: I'm not sure followups and tasks can be handled in the same logic. Will need @orthagh point of view.

inc/itilfollowup.class.php
trasher: "What would you use instead ?" => `$itemtype`;)

AdrienClairembault: This line is inside this condition: 
```php
if ($itemtype == "Ticket")
```
So $itemtype can only be "Ticket".

I think it's better to use the string rather than $itemtype as it makes the code easier to read and may enable some features of your IDE like "Find references" or "Go to definition" if supported.

inc/tickettask.class.php
orthagh: Regarding the code of ItilFollowup, the changes seems ok.

inc/itilfollowup.class.php
orthagh: if it's only for TicketFollowup, inherits buildParentCondition in this class,  call parent for need code and apply specific behavior.
We should not have the kind of condition mentioned above

orthagh: nevermind, `TicketFollowup` is deprecated.

PR #6226, opened 2019-08-14
Bug in search engine for Having - fixed #6225

### LINKED ISSUE #6225: Bug in search engine for HAVING
You can't have WHERE and HAVING in a same request.
So if you need to use HAVING you must change WHERE to only use HAVING.

yllen:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #6225 

Tested for many items in computer with OR as criteria (memory size or name).
Not tested for all items

### DIFF
--- a/inc/search.class.php
+++ b/inc/search.class.php
@@ -660,14 +660,20 @@ class Search {
 
       // Add search conditions
       // If there is search items
+      $useglobalhaving = false;
       if (count($data['search']['criteria'])) {
          $WHERE  = self::constructCriteriaSQL($data['search']['criteria'], $data, $searchopt);
          $HAVING = self::constructCriteriaSQL($data['search']['criteria'], $data, $searchopt, true);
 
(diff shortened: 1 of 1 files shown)


flegastelois:
Seems to be a duplication of this PR: https://github.com/glpi-project/glpi/pull/6204 which is already validated.

yllen:
Before purpose a PR i always done a git pull from git branch 9.4/bugfixes and this issue was still present.

trasher:
This duplicates https://github.com/glpi-project/glpi/pull/6204. Since test are KO and the other add a specific new test, we do not keep this one.

Please test and give feedback on https://github.com/glpi-project/glpi/pull/6204; thanks :)

trasher:
Also, please avoid creating branches on main project, prefer using your fork.

Issue #6254, opened 2019-08-30
User can't schedule task for another


stonebuzz:
**Describe the bug**

If user try to schedule (with recall) a task for another tech, planning recall is not saved in BDD.

**Page(s) URL**

ticket tasks, change tasks, problem tasks

**To reproduce**

1. With Tech1 create task without schedul
2. With tech2 update task and add schedule with recall
4. Tech1 and Tech2 can't see schedule and recall
5. Infor are not inserted into DB

**Expected behavior**

Anohter Tech (like supervisor) should be able to add schedule on task

**Logs**

No logs

**Screenshots**

**Your GLPI setup (you can find it in Setup > General menu, System tab)**

```

[code]   GLPI 9.4.3-git-bugfix-recall-update-9cc2c66d1 (/GLPI/9.4-bugfixes => /home/stanislas/Teclib/dev/GLPI/9.4-bugfixes) Installation mode: GIT
--

[...]

PR #6255, opened 2019-08-30
fix(task): add right to another user to add schedule on task

stonebuzz:
This PR allow to another user to add schedule and recall on task where he's not the author

| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #6254

### DIFF
--- a/inc/planningrecall.class.php
+++ b/inc/planningrecall.class.php
@@ -53,7 +53,7 @@ class PlanningRecall extends CommonDBChild {
 
 
    function canCreateItem() {
-      return $this->fields['users_id'] == Session::getLoginUserID();
+      return true;
    }
 
 


inc/planningrecall.class.php
```diff
@@ -53,7 +53,7 @@ static function canCreate() {
 
 
    function canCreateItem() {
-      return $this->fields['users_id'] == Session::getLoginUserID();
+      return true;
```
trasher: Hum... Not sure for this one. Should we not look for some rights here, ping @orthagh

orthagh: regarding #6234, this make sense, but i think the both pr are wrong.

In the relative notification, only the task creator or the assignee can be recipient.
So only them should be allowed to add a recall.

For other users, planning recalls dropdown should not be displayed (or maybe in read-only, but i'm not sure it's needed).
Other scenario like a supervisor adding recall for users but not creator nor assignee requires a new right to do it, so an evolution not a bug.

orthagh: i revert my comment, 

Recipient Request retrieved from the planning recall not from the task.
So the current scenari is anybody who see the task can add a recall on it to be notified himself.

So the previous code changed by this pr is the good one and #6234 is wrong also

PR #6303, opened 2019-09-13
check same rights for followups in all ITIL objects

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?  | hope yes

Internal ref : 17062

For changes and problems, you need to have UPDATE right on the main object to add followups.
I move rights from ticket to CommonItilObject to have the same behavior in all ITIL objects

image

Following this fix, we may rename the constants to be named relative to itil objects not ticket (Ex ADDMYTICKET -> ADDMY) but it's out of the need for this fix.

### DIFF
--- a/inc/commonitilobject.class.php
+++ b/inc/commonitilobject.class.php
@@ -170,8 +170,18 @@ abstract class CommonITILObject extends CommonDBTM {
     *
     * @return boolean
     */
-   function canAddFollowups() {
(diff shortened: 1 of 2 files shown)


orthagh:
For constants renaming (and also corresponding gettext sentences), i think we should target 9.5

orthagh:
One strange thing, the previous check was also based on CREATE:followup, this right doesn't exists at all

PR #6318, opened 2019-09-18
Fix cyrillic characters in slugify; closes #2946

### LINKED ISSUE #2946: Slugify cyrillic characters fails
On some languages; slugify fails and can result in an empty string... 

For exemple:
```php
Toolbox::slugify('Έρευνα ικανοποίησης - Αιτήματα'); //==> ""
```

A solution would be to use:
```php
transliterator_transliterate("Any-Latin; NFD; [:Nonspacing Mark:] Remove; NFC; [:Punctuation:] Remove; Lower();", 'Έρευνα ικανοποίησης - Αιτήματα'); //==> "ereuna ikanopoieses  aitemata"
```

That sound better; but this requires additional php-intl extension.

I also see an existing library - https://github.com/cocur/slugify - but I'm affraid they support less languages than GLPI.

Slugs are used to make anchors on KB contents, and in charts (so charts are currently failing in some languages).

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2946

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,6 +19,7 @@
         "ext-gd": "*",
         "ext-zlib": "*",
         "ext-curl": "*",
+        "ext-intl": "*",
         "iamcal/lib_autolink": "~1.7",
         "phpmailer/phpmailer": "~6.0",
         "sabre/vobject": "~4.1",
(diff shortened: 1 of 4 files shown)


trasher:
Backport from master.

PR #6322, opened 2019-09-18
Added feature: to be able to view lock messages as a div in the DOM

tomolimo:
Hello,

Added feature: to be able to view lock messages as a div in the DOM instead of a floating dialog box.
By default, it will by the former behavior: lock message will be floating. 
It can be changed in the GLPI general settings > default values, and then it can be set per user.

Thank you,
Regards,

Tomolimo

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |


orthagh:
We just tested the feature.

- What the need having an option ? The new non floating design is better than the previous. We already have hundreds of options and we need to keep the configuration readable. So do one design, it's sufficient.

- Design is a bit strange. Alert box is not aligned with anything. Mostly in lefttab layout.

image

Imo, you have 2 options:

- align with full page under the header.
- align above the item form (+tabs) 

image

- do a soft red background with bold red text. This should draw attention.

- Is javascript is still mandatory to display this alert box ? We should have this into the html generation.

- Not relative to the content of this pr; When you send a proposal modification, please attach a screenshot of the change. It's easier to figure the change on our side.

Regards

tomolimo:
Hello @orthagh 

> - What the need having an option ? The new non floating design is better than the previous. We already have hundreds of options and we need to keep the configuration readable. So do one design, it's sufficient.

The need of the option is that I would like to preserve the current behavior for people having the use of it, and if these people like the "floating messages". That's the reason for the option -> to preserve existing behavior. 

> - Design is a bit strange. Alert box is not aligned with anything. Mostly in lefttab layout.

That's due to the fact that I tried to be independent of the layout and to find a criterion available in the three different layouts (lefttabs, classic, and split).

> Imo, you have 2 options:
> 
> - align with full page under the header.
> - align above the item form (+tabs)

During the design phase, I tried both, and for the lefttabs layout the best is to be in the tabs div, but this is not ok when in the other layouts. To be under the header and above the navigationheader is a bit strange, as the lock message is related with the item, and in the navigation header is related to navigation to other items.

so what I propose is to:
- move it into the tabs div when in lefttabs layout, and to place it under the navigationheader when in split layout, and to move it into the item div when in classic layout. Is this ok for you?

> - do a soft red background with bold red text. This should draw attention.

I'm going to look at what is feasible

> - Is javascript is still mandatory to display this alert box ? We should have this into the html generation.

I think yes, but I'm open to any proposals

> - Not relative to the content of this pr; When you send a proposal modification, please attach a screenshot of the change. It's easier to figure the change on our side.

Yes you're right, sorry for that, I didn't think about that, but I will :smiley: 

I'm going to propose another design in a new commit. Thank you for your feedback.

Regards,
Tomolimo

trasher:
There is no need for a configuration option for this change, please remove that. It is not possible to keep all changes that are done the old way (or it lends to dead code); and the code is more complelx to read and understand.

tomolimo:
> There is no need for a configuration option for this change, please remove that. It is not possible to keep all changes that are done the old way (or it lends to dead code); and the code is more complelx to read and understand.

Apart of the code consideration for which I totally agree, we must also deal with the use of the users and we must maximize user acceptance of the new versions/features...
This is called user experience...
Each time we do a change in the use of the users, we must do an estimation of the user acceptance (gap analysis). 
to give an example: in GLPI 9.2 the menus are verticals, and in GLPI 9.3 they are no longer vertical: this a problem for the acceptance of the new version...

orthagh:
> During the design phase, I tried both, and for the lefttabs layout the best is to be in the tabs div, but this is not ok when in the other layouts. To be under the header and above the navigationheader is a bit strange, as the lock message is related with the item, and in the navigation header is related to navigation to other items.

For information, in pr #5815 , we changed the item header. (about that, shame on me to not provides screenshot ).
Here is a capture of a new design:
image

It's integrate much better with all layout and ease your own integration.

The issue for this pr was the single action part was not convincing people in fine.
I should have separate header redesign from the the single action feature.


And So if we have the new header merged, i think the option for user is not needed at all as we can insert your message between the item header and the item form.

tomolimo:
> > During the design phase, I tried both, and for the lefttabs layout the best is to be in the tabs div, but this is not ok when in the other layouts. To be under the header and above the navigationheader is a bit strange, as the lock message is related with the item, and in the navigation header is related to navigation to other items.
> 
> For information, in pr #5815 , we changed the item header. (about that, shame on me to not provides screenshot ).
> Here is a capture of a new design:
> image
> 
> It integrates much better with all layouts and ease your own integration.
> 
> The issue for this pr was the single action part was not convincing people in fine.
> I should have separate header redesign from the the single action feature.
> 
> And So if we have the new header merged, i think the option for user is not needed at all as we can insert your message between the item header and the item form.

Ok for me, this new header seems to be very nice. With that we'll be able to align the Lock messages. :+1: 

Keep me informed, and I'll change the my commits. :smiley:

orthagh:
#5815 merged

tomolimo:
Hello @orthagh , @trasher 
What do you think about something like that?
image

orthagh:
Yes this is exactly what i had in mind :+1:

trasher:
ok for me aswell :)

tomolimo:
Hello
I've to check for the other display modes, and I'll keep it like that.
Thank you

tomolimo:
Hello, 
My last commit has some problems, don't use it...

trasher:
ping?

trasher:
Please rebase on 9.5/bugfixes

tomolimo:
This is what I did :(

tomolimo:
I really don't understand how these arrived, and I don't either know how to get ride of these....

trasher:
OK, I'll take a look, but I'm busy right now

tomolimo:
An idea, would be to delete this branch and to re-create a new one.
Do you want me to do it like this?

trasher:
I've force pushed your branch. Please keep a backup of your previous branch (rename it) before pulling the new one, and check if I did nt made mistakes solving conflicts.

BTW, it seems that you did not rebase but merge, but don't kown what you've done wrong to.

trasher:
Or just try on another clean branch, yes (did no tsee your comment before)

install/mysql/glpi-empty.sql
```diff
@@ -1237,6 +1237,7 @@ CREATE TABLE `glpi_impactitems` (
 	KEY `source` (`itemtype`, `items_id`),
 	KEY `parent_id` (`parent_id`)
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_unicode_ci;
+INSERT INTO `glpi_configs` VALUES ('211','core','lock_floating_message','1');
```
trasher: This has been moved in `install/empty_data.php` for glpi 9.5

tomolimo: oups!

trasher: I'm against adding a new preference.

This is a display detail, not something like a feature that has drastically changed.

inc/objectlock.class.php
```diff
@@ -279,11 +315,11 @@ function requestLock() {
                }
          ");
 
-      $msg = "<table><tr><td class=red nowrap>";
-      $msg .= __('Warning: read-only!')."</td>";
-      $msg .= "<td nowrap><a class='vsubmit' onclick='javascript:requestLock();'>".
+      $msg = "<span class=red style='padding-left:5px;'>";
```
trasher: There are many inliine stles. Better to put them in css files

PR #6353, opened 2019-09-25
Move & Improve isIndex to support pirmary and compound keys

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | yes
| Tests pass?   | yes


trasher:
I have doubts about all replacements I made in upgrade files; I'll have to double check

PR #6365, opened 2019-09-27
fix bad sql query

btry:
regression in https://github.com/glpi-project/glpi/pull/6260

Let's be careful with this PR. It solves the issue in Formcreator : https://github.com/pluginsGLPI/formcreator/issues/1506

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #number

### DIFF
--- a/inc/search.class.php
+++ b/inc/search.class.php
@@ -4060,7 +4060,7 @@ JAVASCRIPT;
       }
 
       if (($table != 'asset_types')
-         && !empty($complexjoin)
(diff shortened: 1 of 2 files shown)


btry:
The PR #6260 requires that we enter when the complex join is not empty. Before the change, we could enter when the table and tie itemtype's table are not the same. This PR limits the use of Complex joins to case where the tables are identical and allows again to enter when  the tables are different.

image

PR #6394, opened 2019-10-03
set execution time only if request succeeds

btry:
If we call the Search endpoint of the API and the search fails due to a SQL error, the max execution time is set. This prevents detection of the failure in the API and returns inconsistent JSON result.

see: https://github.com/glpi-project/glpi/blob/9.4/bugfixes/inc/api.class.php#L1609

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #number

### DIFF
--- a/inc/search.class.php
+++ b/inc/search.class.php
@@ -1264,15 +1264,15 @@ class Search {
          }
       }
 
-      $data['data']['execution_time'] = $DBread->execution_time;
-      if (isset($data['search']['savedsearches_id'])) {
-         SavedSearch::updateExecutionTime(
-            (int)$data['search']['savedsearches_id'],
-            $DBread->execution_time
-         );
-      }
-
       if ($result) {
+         $data['data']['execution_time'] = $DBread->execution_time;
+         if (isset($data['search']['savedsearches_id'])) {
+            SavedSearch::updateExecutionTime(
+               (int)$data['search']['savedsearches_id'],
+               $DBread->execution_time
+            );
+         }
+
          $data['data']['totalcount'] = 0;
          // if real search or complete export : get numrows from request
          if (!$data['search']['no_search']


inc/search.class.php
```diff
@@ -1264,7 +1264,6 @@ static function constructData(array &$data, $onlycount = false) {
          }
       }
 
-      $data['data']['execution_time'] = $DBread->execution_time;
       if (isset($data['search']['savedsearches_id'])) {
          SavedSearch::updateExecutionTime(
```
trasher: Maybe this one sould be updated as well only on success?

inc/search.class.php
```diff
@@ -1272,238 +1277,235 @@ static function constructData(array &$data, $onlycount = false) {
          );
       }
 
-      if ($result) {
```
btry: From this line to the end of the method : indentation level decreased. (diff unreadable)

inc/search.class.php
```diff
@@ -1272,238 +1277,235 @@ static function constructData(array &$data, $onlycount = false) {
...
 
-         $data['data']['count'] = count($data['data']['rows']);
-      } else {
-         echo $DBread->error();
+         $data['data']['rows'][$i] = $newrow;
+         $i++;
       }
+
+      $data['data']['count'] = count($data['data']['rows']);
+      return true;
```
btry: return a boolean instead of nothing.

inc/search.class.php
```diff
@@ -1264,6 +1264,11 @@ static function constructData(array &$data, $onlycount = false) {
          }
       }
 
+      if (!$result) {
+         echo $DBread->error();
+         return false;
+      }
```
btry: reverse the if ($result) {...} else {...} found below, for readability, and return a boolean instead of nothing

inc/search.class.php
btry: OK for me, commit amended.

An extra commit added for a refactor I would like to do. If you think it is too risky, then I  may remove it from the PR.

Calls to Search::constructDatas() never get the returned value

image

Issue #6390, opened 2019-10-01
Error  unknown column 'glpi_itilfollowups.date' in 'field list'


cp61:
**Describe the bug**

The Ticket tab for users in Post Only no longer displays open tickets

The sql-errors.log log returns the Error: Unknown column 'glpi_itilfollowups.date' in 'field list' (see log below)

**Page(s) URL**

glpi/front/ticket.php

**To reproduce**

1. Go to /glpi/front/ticket.php
2. no open ticket appears

**Expected behavior**

the tickets opened by the user should appear

**Logs**

Log de sql-errors.log

[2019-10-01 17:22:49] glpisqllog.ERROR: DBmysql::query() in C:\xampp\htdocs\glpi\inc\dbmysql.class.php line 188
*** MySQL query error:
SQL: SELECT DISTINCT glpi_tickets.id AS id, 'ocsinventoryng' AS currentuser,
glpi_tickets.id AS ITEM_Ticket_2, glpi_tickets.name AS ITEM_Ticket_1,
glpi_tickets.id AS ITEM_Ticket_1_id,
glpi_tickets.id AS ITEM_Ticket_1_id, glpi_tickets.content AS ITEM_Ticket_1_content, glpi_tickets.status AS ITEM_Ticket_1_status, glpi_tickets.status AS ITEM_Ticket_12, glpi_tickets.date AS ITEM_Ticket_15, glpi_tickets.date_mod AS ITEM_Ticket_19, GROUP_CONCAT(DISTINCT glpi_users_a5e2026cbcb683e774f93a59e24a62a4.id SEPARATOR '$$##$$')
AS ITEM_Ticket_4,
GROUP_CONCAT(DISTINCT CONCAT(glpi_tickets_users_d1524bb2ea1d461ab90aed3b5f0e7e60.users_id, ' ',
glpi_tickets_users_d1524bb2ea1d461ab90aed3b5f0e7e60.alternative_email)
SEPARATOR '$$##$$') AS ITEM_Ticket_4_2,
GROUP_CONCAT(DISTINCT glpi_users_f201be21cd638ee780d08cba4ceff0d4.id SEPARATOR '$$##$$')
AS ITEM_Ticket_5,
GROUP_CONCAT(DISTINCT CONCAT(glpi_tickets_users_819efb92c8b927b345e489211ec8e43b.users_id, ' ',
glpi_tickets_users_819efb92c8b927b345e489211ec8e43b.alternative_email)
[...]

Issue #6288, opened 2019-09-11
see private tasks option unchecked, profile can't see tickets list on the tickets page


M-Falken:
**Describe the bug**

Hi,
With this commit b052d61 applied on _search.class.php_ part of a 9.4.3
If a profile don't have the "see private tasks option checked" the user that use this profile can't see tickets list on the tickets page.
Just no results at all on the page.
Without this commit, all is fine again.

**Page(s) URL**

_/front/ticket.php_

**To reproduce**
![Screenshot_2019-09-11 GLPI - Profils - 9(1)](https://user-images.githubusercontent.com/1380639/64677929-d10d8900-d478-11e9-9fe4-7e9eee4904bc.png)

1. Go to a profile set it like the screen above and confirm that the "see private tasks option" is unchecked.
2. Assign this profile to your own account, and use it.
3. Go to _/front/ticket.php_
4. You should see the ticket list.
5. Now apply the b052d61
6. Refresh or change your profile once to any other then back to the test one.
7. Go to _/front/ticket.php_
8. Ticket list is empty.

**Expected behavior**

We should see the ticket list whatever the "see private tasks option" checked or not.

**Logs**

```
no error log for this and nothing with debug
```

**Your GLPI setup (you can find it in Setup > General menu, System tab)**

```
GLPI 9.4.3 PHP 7.3.9 + Opcache + APCu 5.1.17 under Litespeed | 10.2.26-MariaDB
```

trasher:
I do not reproduce with an up to date 9.4/bugfixes branch, please try with up to date current bugfixes branch (or [latest nightly](https://forge.glpi-project.org/tarballs/))

trasher:
I still do not reproduce.

M-Falken:
Ok, I give up on this one for now.
I suspect a specific modified plugin to interfere.
So I will wait the next official release then I'll do a new bunch of tests at this time.
Thanks for the time.

PR #6399, opened 2019-10-03
Fix followup and task search on limited profiles

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #6390

Only seems to affect lower-privileged profiles. The wrong table name was used in order.
May be related to #6288 as well


cconard96:
@M-Falken Could you check to see if this change fixes the issue you reported as well?

chauchse:
Unfortunately, i did not have time to try. I need to create a test environnement to try it...

M-Falken:
@cconard96 Sorry for very very late reply.
I can confirm this is ok now

PR #6430, opened 2019-10-08
Added a filter in canRequesterUpdateItem to be sure that ticket is not read only

### LINKED ISSUE #6402: Object is not lock when autolock = false
Hello,

we have a bug on gpli 9.4.3, PHP 7.3 without plugin. 

When autolock mode is set to false, if the user is requester or writer, 
the save button is displayed and active and we can update all the informations in the ticket, except the actors. 
there is No error in debug mode.

regards

tomolimo:
Added isReadOnly in ObjectLock class
Fixes #6402

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #6402

### DIFF
--- a/inc/objectlock.class.php
+++ b/inc/objectlock.class.php
@@ -460,6 +460,26 @@ class ObjectLock extends CommonDBTM {
 
 
    /**
+    * Summary of isReadOnly
+    * @param mixed $itemtype
+    * @param mixed $id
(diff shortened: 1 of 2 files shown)


trasher:
see #6556

Issue #5689, opened 2019-03-26
Bugs with closed ticket


yllen:
In a closed ticket, i can link a problem.
But in this problem i can't link this closed ticket

In a closed problem, i can link a ticket.
But in this ticket i can't link this closed problem

same comportment for change_ticket

In a ticket i can't select deleted item.
But in dustbin i can create a ticket from deleted item

In a closed ticket i can add item, cost, link with other items (knowbase, project, problem, change...) ?
For me in a closed ticket you only can reply to satisfaction survey or reopen the ticket

yllen:
i need theses fixes for my customers, so i'm goieng to do PR for 9.1, 9.2 and 9.3/bugfixes
If you can validate them

mab18:
Hi all,

With /inc/knowbaseitem_item.class.php of fix #5689 (#6437), I have an error :
Fatal error: Uncaught Error: Call to undefined method KnowbaseItem::getClosedStatusArray() in /var/www/html/glpi/inc/knowbaseitem_item.class.php:133 Stack trace: #0 /var/www/html/glpi/inc/knowbaseitem_item.class.php(94): KnowbaseItem_Item::showForItem(Object(KnowbaseItem), '') #1 /var/www/html/glpi/inc/commonglpi.class.php(485): KnowbaseItem_Item::displayTabContentForItem(Object(KnowbaseItem), '1', '') #2 /var/www/html/glpi/ajax/common.tabs.php(92): CommonGLPI::displayStandardTab(Object(KnowbaseItem), 'KnowbaseItem_It...', '', Array) #3 {main} thrown in /var/www/html/glpi/inc/knowbaseitem_item.class.php on line 133
When accessing http://my-IP/glpi/front/knowbaseitem.form.php?id=5  (any number)
tab is "élement associé" / "associated element"

no error with previous version. What's wrong ?

PR #6437, opened 2019-10-09
fix adding problem - see #5689

yllen:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #5689


yllen:
I made one commit by part.
Now you have the same comportment on the both sides

Issue #2182, opened 2017-05-10
[API REST 9.1.2] Return incorrect information


jp-morvan:
Hi,

I use API for checking some informations and the return is not the value in database.

**Example :** 
I want to get a phone item with "name" field equals to "000121212". So, I use this curl query 

> curl -s -X GET 'https://myglpi/apirest.php/search/Phone?criteria\[0\]\[field\]=1&criteria\[0\]\[searchtype\]=contains&criteria\[0\]\[value\]=000121212' -H 'Content-Type: application/json' -H 'Session-Token: mytoken'

here is the return :
```
{
   "totalcount":1,
   "count":1,
   "sort":1,
   "order":"ASC",
   "data":[
      {
         "1":121212,
         "7":"",
         "70":null,
         "10127":null,
         "23":null,
         "40":null,
         "4":null,
         "31":null
      }
   ],
   "content-range":"0-0/1"
}
```

The value returned is "121212" and not "000121212" like in database or in GLPI UI. 
The problem exists on every item type.

Thank you

ladenree76000:
i think my bug is the same : 
whatever  criteria[...][searchtype]  (0,1,2 ie : contains equals notequal) it behaves like "contains"

golgoth31:
the  api use some json_encode options when returning resultes:
if ($response !== null) {
         $json = json_encode($response, JSON_UNESCAPED_UNICODE
                                      | JSON_UNESCAPED_SLASHES
                                      | JSON_NUMERIC_CHECK
                                      | ($this->debug
                                          ? JSON_PRETTY_PRINT
                                          : 0));

Remove "JSON_NUMERIC_CHECK" and the returning json will be only with strings. the drawback is taht some values will be unusable as is. example:
memory size before: 16000
memory size after: 16000.00000

trasher:
I've not been able to find any correct solution to get that fixed right now :(

We'll certainly have to drop `JSON_NUMERIC_CHECK` anyways; but we'll have to type properly each value... Database results are _always_ text, we do not have any way to know/typehint values that are returned.

Yoda-BZH:
A possibility: Add a method to each class that can return data through the API.

This new method would transform (if necessary) the returned fields to the target/desired format.

This way, JSON_NUMERIC_CHECK would be unnecessary.

trasher:
> This new method would transform (if necessary) the returned fields to the target/desired format.

The `if necessary` part is the problem. We cannot know.

[...]

PR #6495, opened 2019-10-21
disable JSON_NUMERIC_CHECK by forcing mysql to not cast int/float types to strings

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | this is the point we need to check
| Fixed tickets | #2182


inc/dbutils.class.php
```diff
@@ -704,7 +704,7 @@ public function getSonsOf($table, $IDf) {
 
       if (!is_array($sons)) {
          // IDs to be present in the final array
-         $sons[$IDf] = "$IDf";
```
orthagh: i need another pov for this one

orthagh:
For information, after tests fixing, i think it's preferable to target 9.5 version to have a good beta/RC external revies

orthagh:
For information, in case of PDO, we can have the same behavior with 
```
pdo->setAttribute(PDO::ATTR_STRINGIFY_FETCHES, false);
pdo->setAttribute(PDO::ATTR_EMULATE_PREPARES, false);
```

inc/dbutils.class.php
trasher: This does not change anything, in the end, we just use this array to the `exportArrayToDB` method, which I guess does not care about type hinting.

orthagh: is cache reset on update ?

trasher: Cache is updated at the end of the method

Issue #6546, opened 2019-11-06
[9.5 bugfixes] - debug mode fills logs with php warnings


btry:
With debug mode enabled I have garbage in php-errors.log

```
[2019-11-06 09:58:19] glpiphplog.ERROR: Toolbox::userErrorHandlerNormal() in /home/tbugier/public_html/glpi-95/inc/toolbox.class.php line 663
  *** PHP Warning(2): print_r(): Couldn't fetch mysqli
  Backtrace :
  :                                                  
  inc/html.class.php:3928                            print_r()
  inc/html.class.php:3917                            Html::printCleanArray()
  inc/html.class.php:3917                            Html::printCleanArray()
  inc/html.class.php:828                             Html::printCleanArray()
  inc/html.class.php:1723                            Html::displayDebugInfos()
  plugins/formcreator/front/targetticket.form.php:92 Html::footer()
  {"user":"2@LU006","mem_usage":"0.000\", 10.64Mio)"} 
```

$_SESSION['glpicondition'] contains now objects which are displayed via print_r(). 

![image](https://user-images.githubusercontent.com/14139801/68285026-cfc69a00-007e-11ea-9c3d-440311ed44e7.png)

https://stackoverflow.com/questions/2010427/php-php-incomplete-class-object-with-my-session-data

It is possible that the session is deserialized before the required classes are loaded.

btry:
I'm not sure about the best solution yet. It would be useful to know if there are other cases where $_SESSION may contain an object.

This case might be source of bugs as the code reusing a condition from session expects a string, not an object.

https://github.com/glpi-project/glpi/blob/9.4/bugfixes/inc/dropdown.class.php#L2318

PR #6548, opened 2019-11-06
fix warning with objects in session

btry:
A proposal to fix warnings in debug mode when QuerySubQuery objects are added in $_SESSION['glpicondition']

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #6546

### DIFF
--- a/inc/html.class.php
+++ b/inc/html.class.php
@@ -3925,7 +3925,7 @@ JAVASCRIPT;
                   }
                } else {
                   if (is_object($val)) {
-                     print_r($val);
+                     echo (string) $val; // It is expected that the object implements __toString()
                   } else {
                      echo htmlentities($val);
                   }
--- a/inc/html.class.php
+++ b/inc/html.class.php
@@ -3925,7 +3925,11 @@ JAVASCRIPT;
                   }
                } else {
                   if (is_object($val)) {
-                     echo (string) $val; // It is expected that the object implements __toString()
+                     if (method_exists($val, '__toString')) {
+                        echo (string) $val;
+                     } else {
+                        echo "(object) " . get_class($val);
+                     }
                   } else {
                      echo htmlentities($val);
                   }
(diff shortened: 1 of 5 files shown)


inc/querysubquery.class.php
```diff
@@ -74,4 +74,8 @@ public function getQuery() {
       }
       return $sql;
    }
+
+   public function __toString() {
```
trasher: I guess it should be implemented in:
- AbstractQuery
- QueryParam
- QueryExpression

inc/html.class.php
```diff
@@ -3925,7 +3925,7 @@ static function printCleanArray($tab, $pad = 0, $jsexpand = false) {
                   }
                } else {
                   if (is_object($val)) {
-                     print_r($val);
+                     echo (string) $val; // It is expected that the object implements __toString()
```
trasher: Thinking about it, maybe could we add an interface for Queries objects, and add a test on the interface implementation  before the `is_object` (and keep `print_r` for this case).

btry: I agree

As the involved classes are used as substring of a final SQL query string, I suggest one of the following names
- QueryFragmentInterface
- QueryStringInterface

trasher: Or maybe just test for the `__toString` method? 4ve never tried, dunno if that works :)

btry: if will work
```php
Interactive mode enabled

php > class foo { }
php > class bar { public function __toString() { return ''; }}
php > var_dump(method_exists('foo', '__toString'));
php shell code:1:
bool(false)
php > var_dump(method_exists('bar', '__toString'));
php shell code:1:
bool(true)
```

btry: this is simpler, and is future proof IMHO.

trasher: :+1:

PR #6576, opened 2019-11-13
Fix getLoginUserID strict comparision

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #6575

Some existing strict comparisons will fail now that DB data is not always a string. The noted issue was when comparing an ID from a form input (string) to the current user's id in the session (integer now).

### DIFF
--- a/front/reservation.form.php
+++ b/front/reservation.form.php
@@ -48,7 +48,7 @@ if (isset($_POST["update"])) {
    list($begin_year,$begin_month) = explode("-", $_POST['resa']["begin"]);
    Toolbox::manageBeginAndEndPlanDates($_POST['resa']);
    if (Session::haveRight("reservation", UPDATE)
-       || (Session::getLoginUserID() === $_POST["users_id"])) {
+       || (Session::getLoginUserID() == $_POST["users_id"])) {
       $_POST['_target'] = $_SERVER['PHP_SELF'];
       $_POST['_item']   = key($_POST["items"]);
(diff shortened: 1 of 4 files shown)


inc/knowbaseitem.class.php
```diff
@@ -106,7 +106,7 @@ static function canView() {
       global $CFG_GLPI;
 
       return (Session::haveRightsOr(self::$rightname, [READ, self::READFAQ])
-              || ((Session::getLoginUserID() === false) && $CFG_GLPI["use_public_faq"]));
```
trasher: This one sould stay as is, because the method `return false;` on some cases

trasher:
I'd advice you to be very careful trying to do this kind of changes... Apart the comment I've done, I did not see anything really wrong reviewing your current changes; but test fails on tickets ACLs tests.
I suspect some difficult to see/reproduce cases, and many of them that could not be unit tested :/

front/reservation.form.php
```diff
@@ -113,7 +113,7 @@
             $input['users_id'] = $_POST['users_id'];
```
trasher: maybe it would be better `$input['users_id'] = $_POST['users_id'];` and keep the `===`?

trasher: maybe the same for user.class

PR #6590, opened 2019-11-14
Check canViewItem before showing KBI in new task form

AdrienClairembault:
Internal ref: 18414.

KBI that are not visible by the user are still shown and can be selected in the following dropdown :
image

This changes removes theses KBI from the dropdown if the user is not allowed to see them.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

### DIFF
--- a/inc/knowbaseitem.class.php
+++ b/inc/knowbaseitem.class.php
@@ -1928,13 +1928,18 @@ class KnowbaseItem extends CommonDBVisible {
    /**
     * Get ids of KBI in given category
     *
-    * @param $category_id
-    * @return array Array of ids
+    * @param int           $category_id   id of the parent category
+    * @param KnowbaseItem  $kbi           used only for unit tests
+    *
+    * @return array        Array of ids
     */
-   public static function getForCategory($category_id) {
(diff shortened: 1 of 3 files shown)


inc/knowbaseitem.class.php
```diff
@@ -1924,4 +1924,39 @@ public function revertTo($revid) {
          return false;
       }
    }
+
+   /**
+    * Get ids of KBI in given category
+    *
+    * @param $category_id
```
trasher: `@parma type $var blah`

inc/knowbaseitem.class.php
```diff
@@ -1924,4 +1924,44 @@ public function revertTo($revid) {
...
+
+      // Filter on canViewItem
+      $ids = array_filter($ids, function($id) use ($kbi) {
+         $kbi->getFromDB($id);
+         return $kbi->canViewItem();
+      });
+
+      // Avoid empty IN
+      if (count($ids) === 0) {
+         $ids[] = -1;
```
trasher: This case i not tested as far as I can see

PR #6598, opened 2019-11-15
Always create a doc_item, use -1 value in timeline to not display images

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

When using the new clean orphan docs crontask, all documents without a doc_item entry are clened.

### DIFF
--- a/inc/document.class.php
+++ b/inc/document.class.php
@@ -675,9 +675,6 @@ class Document extends CommonDBTM {
          [
             'COUNT'     => 'cpt',
             'FROM'      => 'glpi_reminders',
-            'WHERE'     => [
-               'text' => ['REGEXP', $this->getSelfUrlRegexPattern()]
-            ]
          ],
          Reminder::getVisibilityCriteria()
       );
@@ -739,35 +736,6 @@ class Document extends CommonDBTM {
       if ($result['cpt'] > 0) {
          return true;
       }
-
-      // Inlined images do not have entries in glpi_documents_items table.
-      // Check in KnowbaseItem content
-      $request = [
-         'FROM'      => 'glpi_knowbaseitems',
-         'COUNT'     => 'cpt',
-         'LEFT JOIN' => [
-            'glpi_entities_knowbaseitems' => [
-               'FKEY' => [
-                  'glpi_knowbaseitems'          => 'id',
-                  'glpi_entities_knowbaseitems' => 'knowbaseitems_id'
-               ]
-            ]
-         ],
-         'WHERE'     => [
-            'glpi_knowbaseitems.answer' => ['REGEXP', $this->getSelfUrlRegexPattern()],
-         ]
-      ];
-
-      if (array_key_exists('LEFT JOIN', $visibilityCriteria) && count($visibilityCriteria['LEFT JOIN']) > 0) {
-         $request['LEFT JOIN'] += $visibilityCriteria['LEFT JOIN'];
-      }
-      if (array_key_exists('WHERE', $visibilityCriteria) && count($visibilityCriteria['WHERE']) > 0) {
-         $request['WHERE'] += $visibilityCriteria['WHERE'];
-      }
-
-      $result = $DB->request($request)->next();
-
-      return $result['cpt'] > 0;
    }
 
    /**
@@ -808,8 +776,6 @@ class Document extends CommonDBTM {
       }
 
       // Check ticket and child items (followups, tasks, solutions) contents
-      $regexPattern = $this->getSelfUrlRegexPattern();
-
       $itil_table = $itil->getTable();
       $itil_key   = $itil->getForeignKeyField();
       $task_table = getTableForItemType($itil->getType() . 'Task');
@@ -841,26 +807,10 @@ class Document extends CommonDBTM {
          ],
          'WHERE'     => [
             $itil_table . '.id' => $items_id,
-            'OR' => [
-               $itil_table . '.content'     => ['REGEXP', $regexPattern],
-               'glpi_itilfollowups.content' => ['REGEXP', $regexPattern],
-               $task_table . '.content'     => ['REGEXP', $regexPattern],
-               'glpi_itilsolutions.content' => ['REGEXP', $regexPattern]
-            ]
          ]
       ])->next();
 
       return $result['cpt'] > 0;
-   }
-
-   /**
-    * Gives URL regex pattern for current document.
-    * This pattern can be use to find link to document into rich text contents.
-    *
-    * @return string
-    */
-   private function getSelfUrlRegexPattern() {
-      return 'document\\\.send\\\.php\\\?docid=' . $this->fields['id'] . '[^0-9]+';
    }
 
    static function rawSearchOptionsToAdd($itemtype = null) {
--- a/inc/document.class.php
+++ b/inc/document.class.php
@@ -665,21 +665,6 @@ class Document extends CommonDBTM {
       );
 
       $result = $DB->request($criteria)->next();
-      if ($result['cpt'] > 0) {
-         return true;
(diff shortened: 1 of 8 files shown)


inc/document.class.php
```diff
@@ -739,35 +736,6 @@ private function canViewFileFromKnowbaseItem() {
       if ($result['cpt'] > 0) {
          return true;
       }
```
cedric-anne: `return $result['cpt'] > 0;`

install/update_94_95.php
```diff
@@ -927,6 +927,50 @@ function update94to95() {
    }
    /** /Fix mixed classes case in DB */
 
+   /** A doc_item to rule them all! */
+   $itemtypes = [
+      'ITILFollowup' => 'content',
+      'ITILSOlution' => 'content',
```
```suggestion
      'ITILSolution' => 'content',
```

install/update_94_95.php
```diff
@@ -927,6 +927,50 @@ function update94to95() {
...
+      $result = $DB->request([
+         'SELECT' => ['id', $field],
+         'FROM'   => $itemtype::getTable(),
+         'WHERE'  => [
+            $field => ['REGEXP', $regexPattern]
+         ]
+      ]);
+
+      while ($data = $result->next()) {
+         //TODO create Document_Item entry, with a timeline_position set to Document_Item::NO_TL
```
cedric-anne: There is a TODO here :)

install/update_94_95.php
```diff
@@ -927,6 +927,50 @@ function update94to95() {
...
+   ];
+   foreach (['Change', 'Problem', 'Ticket'] as $itiltype) {
+      $itemtypes[$itiltype] = 'content';
+      $itemtypes[$itiltype . 'Task'] = 'content';
+   }
+
+   $docs_input =[];
+   foreach ($itemtypes as $itemtype => $field) {
+      // Check ticket and child items (followups, tasks, solutions) contents
+      $regexPattern = 'document\\\.send\\\.php\\\?docid=[0-9]+';
```
```suggestion
      $regexPattern = $DB->escape('document\.send\.php\?docid=[0-9]+');
```

install/update_94_95.php
```diff
@@ -927,6 +927,50 @@ function update94to95() {
...
+         'SELECT' => ['id', $field],
+         'FROM'   => $itemtype::getTable(),
+         'WHERE'  => [
+            $field => ['REGEXP', $regexPattern]
+         ]
+      ]);
+
+      while ($data = $result->next()) {
+         //TODO create Document_Item entry, with a timeline_position set to Document_Item::NO_TL
+         preg_match('/document\\.send\\.php\\?docid=([0-9]+)/', $data[$field], $matches);
```
```suggestion
         preg_match('/document\.send\.php\?docid=([0-9]+)/', $data[$field], $matches);
```

install/update_94_95.php
```diff
@@ -927,6 +927,50 @@ function update94to95() {
...
+         $docs_input[] = [
+            'documents_id' => $matches[1],
+            'itemtype'     => $itemtype,
```
cedric-anne: For ITILFollowup and ITILSolution, I think `'itemtype'` should be the itemtype of the parent ITIL item.

cedric-anne:
> If config `attach_ticket_documents_to_mail` is set to true, inlined images will be attached twice. One having with `Content-Disposition: inline;` and the second second having `Content-Disposition: attachment;`.

I quickly debugged it and it can be fixed by doing this: 
```diff
diff --git a/inc/notificationeventmailing.class.php b/inc/notificationeventmailing.class.php
index 125ecf8c91..24ecb4b474 100644
--- a/inc/notificationeventmailing.class.php
+++ b/inc/notificationeventmailing.class.php
@@ -172,7 +172,7 @@ class NotificationEventMailing extends NotificationEventAbstract implements Noti
             foreach ($documents_ids as $document_id) {
                $doc->getFromDB($document_id);
                // Add embeded image if tag present in ticket content
-               if (preg_match_all('/'.Document::getImageTag($doc->fields['tag']).'/',
+               if (preg_match_all('/'.preg_quote($doc->fields['tag']).'/',
                                   $current->fields['body_html'], $matches, PREG_PATTERN_ORDER)) {
                   $image_path = Document::getImage(
                      GLPI_DOC_DIR."/".$doc->fields['filepath'],
```
Not sure about side effects.

install/update_94_95.php
trasher: Dunno... ping @orthagh

trasher: I do not agree.
Doing this, this would be impossible to link image from its original real place (say a Fup); while it is possible to link it currently to the parent itemtype throw the Fup.

cedric-anne: This is not consistent with what will be done when a document will be added from the timeline form (see https://github.com/glpi-project/glpi/pull/6598/files#diff-7f569484c496f62a6ccf54647fed2b1bR5070).

Another problem if we do not use the parent ITIL item is that `ITILFollowup` and `ITILSolution` are not defined in `$CFG_GLPI['document_types']`.

Consequences of this are that `Document::getItemtypesThatCanHave()` will not list them and `Document::canApplyOn()` will return false for them:
 - it may have an impact on transfer process,
 - cleaning process in `CommonDBTM::cleanRelationTable()` will not be able to clean `document_items` table properly.

If we declare them in `$CFG_GLPI['document_types']`, I do not see any other consequence than having ability to manually link an existing document to a followup or a solution.

trasher: OK, I've changed in commondbtm to have the same behavior on both places; and I've added possible missing types in CFG_GLPI

cedric-anne:
> Due to https://github.com/glpi-project/glpi/blob/9.5/bugfixes/inc/document_item.class.php#L131 , images pasted in Ticket description appears in the timeline.
> 
When replaying migration: 
image

trasher:
Display in timeline of embed images has been fixed. I do not understand the question about `users_id`.

I do not reproduce the update issue

cedric-anne:
> I do not understand the question about `users_id`.

`glpi_documents_items.users_id` is not defined in your migration. I think it should be defined using the value of `users_id` field of source element (they all have a `users_id` field or `users_id_recipient` for Change/Problem/Ticket)

install/update_94_95.php
```diff
@@ -1293,6 +1293,48 @@ function update94to95() {
...
+   ];
+   foreach (['Change', 'Problem', 'Ticket'] as $itiltype) {
+      $itemtypes[$itiltype] = 'content';
+      $itemtypes[$itiltype . 'Task'] = 'content';
+   }
+   $docs_input =[];
+   foreach ($itemtypes as $itemtype => $field) {
+      // Check ticket and child items (followups, tasks, solutions) contents
+      $regexPattern = 'document\\\.send\\\.php\\\?docid=[0-9]+';
+      $user_field = ($itemtype instanceof CommonITILObject) ? 'users_id_recipient' : 'users_id';
```
```suggestion
      $user_field = is_a($itemtype, CommonITILObject::class, true) ? 'users_id_recipient' : 'users_id';
```

install/update_94_95.php
```diff
@@ -1293,6 +1293,48 @@ function update94to95() {
...
+         ]
+      ]);
+      while ($data = $result->next()) {
+         preg_match('/document\\.send\\.php\\?docid=([0-9]+)/', $data[$field], $matches);
+         $docs_input[] = [
+            'documents_id'       => $matches[1],
+            'itemtype'           => $itemtype,
+            'items_id'           => $data['id'],
+            'timeline_position'  => CommonITILObject::NO_TIMELINE,
+            'users_id'           => $user_field
```
```suggestion
            'users_id'           => $data[$user_field],
```

PR #6600, opened 2019-11-15
send a fetch query instead xhr for unlocking object

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #5799

Fetch api seems to work on my side like explained in #5799, tested with chrome 78 and firefox 70
Added fallback for IE<=14 (Edge supports fetch api)
I need confirmation for this fallback (@tomolimo if you have internet explorer in production)

### DIFF
--- a/inc/objectlock.class.php
+++ b/inc/objectlock.class.php
@@ -310,15 +310,14 @@ class ObjectLock extends CommonDBTM {
                                           'users_id' => Session::getLoginUserID()])) {
          // add a script to unlock the Object
          echo Html::scriptBlock( "$(function() {
-                  $(window).on('beforeunload', function() {
-                     //debugger;
-                      $.ajax({
-                          url: '".$CFG_GLPI['root_doc']."/ajax/unlockobject.php',
-                          cache: false,
-                          data: 'unlock=1&id=$id'
-                          });
-                      });
-                  });" );
+            $(window).on('unload', function() {
+               fetch('".$CFG_GLPI['root_doc']."/ajax/unlockobject.php?unlock=1&id=$id', {
+                  method: 'POST',
+                  keepalive: true,
+                  cache: 'no-cache'
+               });
+            });
+         })" );
          $ret = true;
       } else { // can't add a lock as another one is already existing
          if (!$gotIt) {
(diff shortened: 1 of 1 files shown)


orthagh:
For information, i needed to use `unload` to have fetch api working on firefox, `beforeunload` event with this browser failed without detail, just a cryptic message (TypeError: NetworkError when attempting to fetch resource.)

tomolimo:
Hello @orthagh 
I made tests in our environment using this:
```
         $(function() {
               $(window).on('beforeunload', function() {
                  // will do an asynchroneous POST to the unlockobject script
                  var url = '".$CFG_GLPI['root_doc']."/ajax/unlockobject.php?unlock=1&id=$id';
                  if (navigator.sendBeacon) {
                     navigator.sendBeacon(url, '');
                  } else {
                     $.ajax({
                        method: 'POST',
                        url: url,
                        cache: false,
                        data: ''
                     });
                  }
               });
```
And I changed the reception for using $_REQUEST instead of $_GET

What do you think about this?

Thank you,
Regards,

Tomolimo

orthagh:
No. As we said before in the issue conversation, sendBeacon doesn't work well (fetch yes) as it can send the beacon after a page is fully refreshed (too late so).

For POST/GET/REQUEST, yes it's ok for me.

tomolimo:
Ok, then I'll do some tests with your proposal.
Just one thing: are you sure about the test: `window.fetch !== 'undefined'`? Shouldn't it be `typeof window.fetch !== 'undefined'`

orthagh:
good catch

tomolimo:
And then in ajax/unlockobject.php you should change $GET into $_REQUEST?
The fetch will do a POST, and the $.ajax a GET

orthagh:
in my current change, both parts send parameters in GET.
By the way, i still can clean the method: POST as it's not used anymore

tomolimo:
Hello,

I made a test with FF 70.0.1 and Chrome 78.0.3904.97, and in the network traces, one can see that the fetch is done after the get of the main ticket page:

image

and such the ticket is locked when doing a refresh of the page (F5)

....

the tests are ok with IE11 (fetch is not used in this case)

cedric-anne:
On Chromium Version 78.0.3904.97, when adding an item in timeline, ticket is marked as locked on page reload.

PR #6616, opened 2019-11-21
Add new criteria to rule ticket

stonebuzz:
Thisd PR allow to rule engine to add solution on ticket  according to solution template defined in criteria

image
 

| Q             | A
| ------------- | ---
| Bug fix?      | no| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #number

### DIFF
--- a/inc/ruleticket.class.php
+++ b/inc/ruleticket.class.php
@@ -215,6 +215,21 @@ class RuleTicket extends Rule {
                      $output['_'.$action->fields["field"]] = $action->fields["value"];
 
                   }
+
+                  // special case of itil solution template
+                  if ($action->fields["field"] == 'solution_template') {
+                     //load template
+                     $template = new SolutionTemplate();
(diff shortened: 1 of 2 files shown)


inc/ruleticket.class.php
```diff
@@ -215,6 +215,21 @@ function executeActions($output, $params, array $input = []) {
...
+
+                  // special case of itil solution template
+                  if ($action->fields["field"] == 'solution_template') {
+                     //load template
+                     $template = new SolutionTemplate();
+                     if ($template->getFromDB($action->fields["value"]) && $output['id'] != null) {
+                        $solution = new ITILSolution();
+                        $solution->add([
+                           "itemtype" => "Ticket",
+                           "content" => $template->getField('content'),
```
trasher: I guess you should escape content here; add a content with a quote in your tests ;)

stonebuzz: add code to escape content and TU to test with quote

PR #6647, opened 2019-11-27
Impact sidebar

AdrienClairembault:
Replace the toolbar by a new sidebar.
A lot of minors improvements, the main one being that we can now drag and drop assets directly into the graph.

image

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | pending

### DIFF
--- a/ajax/impact.php
+++ b/ajax/impact.php
@@ -47,41 +47,67 @@ Session::checkLoginUser();
 switch ($_SERVER['REQUEST_METHOD']) {
    // GET request: build the impact graph for a given asset
    case 'GET':
-      $itemtype =  $_GET["itemtype"]  ?? "";
-      $items_id =  $_GET["items_id"]  ?? "";
-      $view     =  $_GET["view"]      ?? "graph";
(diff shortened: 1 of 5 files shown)


ajax/impact.php
```diff
@@ -47,41 +47,67 @@
...
+      switch ($action) {
+         case "search":
+            $itemtype = $_GET["itemtype"] ?? "";
+            $used     = $_GET["used"]     ?? "[]";
+            $filter   = $_GET["filter"]   ?? "";
+            $page     = $_GET["page"]     ?? 0;
+
+            // Check required params
+            if (empty($itemtype)) {
+               Toolbox::throwError(400, "Missing itemtype");
```
trasher: Not related to current PR, but I think the name of the method is not correct, since it only handle JSON related (also this method miss phpdoc)

AdrienClairembault: I've added phpdoc and an additional param to specify the format (default is json).

PR #6664, opened 2019-12-03
Impact context

AdrienClairembault:
### Main change
Add "context" to graph.
Before this change, node positions were shared between each graph -> modifying any graph would impact every graph using one of it's nodes.

Now, context is delegated to a "context" object which contains node positions, camera position, zoom level, visibility settings (impact, depends and depth) and edge color settings.

Any graph have one of the following:
- No context
- His own context (master)
- The context of another item (slave)

Rules for updating or creating context objets:
- If the item has a context (master), update it with the new data.
- If the item has no context or is a slave, store the data in a new context (master). Any linked node that also doesn't have a context or is a slave take this context as their own (slaves).

Rules for setting node positions:
- If the item has no context, generate graph position using the directed acyclic graph algorithm
- If the item has a context, use the saved position. If positions are missing for some nodes, they will be generated using a simple algorithm (try to put the node at a set distance and angle, change angle and increase distance until success).

### Minor changes
Make selected toolbar items more distinct:
Beforeimage   After      image

Delete/update impact objects when an asset is purged from GLPI.

Take entities restrictions into account when selecting assets to add in the graph.

Highlight starting node of the graph:
image

A few bugfixes.

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

### DIFF
--- a/inc/impact.class.php
+++ b/inc/impact.class.php
@@ -150,7 +150,7 @@ class Impact extends CommonGLPI {
 
       // Displays views
       self::displayGraphView($item);
-      self::displayListView($item, $graph);
+      self::displayListView($item, $graph, true);
 
       // Select view
       echo Html::scriptBlock("
@@ -190,15 +190,18 @@ class Impact extends CommonGLPI {
     */
    public static function displayListView(
       CommonDBTM $item,
-      array $graph
+      array $graph,
+      bool $scripts = false
    ) {
       global $CFG_GLPI;
 
       $impact_item = ImpactItem::findForItem($item);
+      $impact_context = ImpactContext::findForImpactItem($impact_item);
 
-      // Should not happen, $impact_item is created before
-      if (!$impact_item) {
-         throw new \InvalidArgumentException("No ImpactItem found");
+      if (!$impact_context) {
+         $max_depth = self::DEFAULT_DEPTH;
+      } else {
+         $max_depth = $impact_context->fields['max_depth'];
       }
 
       echo '<div id="impact_list_view">';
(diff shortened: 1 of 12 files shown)


AdrienClairembault:
Test should be rerun as their failure is not caused by theses changes:
```In file /home/circleci/project/tests/functionnal/TicketRecurrent.php on line 454, string() failed for data set [16] of data provider computeNextCreationDateProvider: string(19) '2020-01-09 12:00:00' is not identical to string(19) '2020-01-06 12:00:00'```

File: inc/impact.class.php [hunk 1]
```diff
@@ -1449,4 +1472,77 @@ public static function printImpactForm() {
       echo "<br><br>";
       Html::closeForm();
    }
+
+   /**
+    * Clean impact records for a given item that has been purged form the db
+    *
+    * @param CommonDBTM $item The item being purged
+    */
+   public static function clean(\CommonDBTM $item) {
```
trasher: You should add test on this

AdrienClairembault: Functional test added in e88544d.

tests/functionnal/Impact.class.php
```diff
@@ -280,4 +280,151 @@ public function testBuildGraph_complex() {
...
+
+      $computer1 = getItemByTypeName('Computer', '_test_pc01');
+      $computer2 = getItemByTypeName('Computer', '_test_pc02');
+      $computer3 = getItemByTypeName('Computer', '_test_pc03');
+      $computer4 = getItemByTypeName('Computer', '_test_pc11');
+      $computer5 = getItemByTypeName('Computer', '_test_pc12');
+      $computer6 = getItemByTypeName('Computer', '_test_pc13');
+
+      // Set compounds
+      $compound01Id = $impactCompoundManager->add([
```
trasher: You should check if add are successfulll

orthagh:
As seen together, please use a different way to indicate the main item without involving color (set label as bold, underline, etc)

AdrienClairembault:
> As seen together, please use a different way to indicate the main item without involving color (set label as bold, underline, etc)

image

AdrienClairembault:
Tests failure doens't seem to be linked to the last changes (TicketRecurrent.php).

PR #6674, opened 2019-12-04
Fixed : object types cannot contain numbers

flonou:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | none

I suggest this modification because some plugins add asset types and this line prevented them from working when the type contained a number

### DIFF
--- a/inc/search.class.php
+++ b/inc/search.class.php
@@ -1365,7 +1365,7 @@ class Search {
 
             // Parse datas
             foreach ($newrow['raw'] as $key => $val) {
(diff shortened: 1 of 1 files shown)


trasher:
Hi, this has already been discussed, see https://github.com/glpi-project/glpi/pull/5495#issuecomment-469220090

Thank you :)

PR #6821, opened 2020-01-20
Fix search to restore a feature removed permit to have only '^' and s…

ddurieux:
…o return only filled field (so exclude empty and null fields)

This restore the feature search with only _^_ and return all rows have the fields filled (so not empty / null).

Old code for reference is here: https://github.com/glpi-project/glpi/blob/9.2/bugfixes/inc/search.class.php#L6353-L6371

### DIFF
--- a/inc/search.class.php
+++ b/inc/search.class.php
@@ -7589,6 +7589,11 @@ JAVASCRIPT;
             ($matches[1] != '^' ? '%' : '') .
             trim($matches[2]) .
             ($matches[3] != '$' ? '%' : '');
+      } else if (isset($matches[1])
+            && strlen(trim($matches[1])) == 1
+            && (!isset($matches[3]) || empty($matches[3]))) {
+         // this case is for search with only ^, so mean the field is not empty / not null
+         $search = '%';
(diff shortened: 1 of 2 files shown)


inc/search.class.php
```diff
@@ -7589,6 +7589,10 @@ static function makeTextSearchValue($val) {
             ($matches[1] != '^' ? '%' : '') .
             trim($matches[2]) .
             ($matches[3] != '$' ? '%' : '');
+      } else if (isset($matches[1])
```
trasher: Maybe add a comment on the whole code to explain different cases

ddurieux: Ok, will do it

trasher:
Talking about that with other team members; it appears that this feature is not documented (https://glpi-project.org/DOC/FR/glpi/navigate_search.html).

Also, it seems that something like `^%$` should currently do the same (undocumented aswell... could be added).

inc/search.class.php
ddurieux: I added a comment line 7595

PR #6833, opened 2020-01-22
Fix/planning step 3

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

**Note: this current pr need to be rebased/merged after #6827.**

After customer review of planning last changes, here is a batch of fixes/enhancements:

- Manage series / instances actions for recurrent events : 
  - deleting a recurerrent event can be done with two buttons (one for the serie, one for the instance)
    image
  - dragging or rezising this kind of event will ask to choose also
   image
- when a user planning is added and a group of user plannings is also added, it the user is present in the group, previously, it was not displayed in the view, now, events key are retrieved for the both plannings (in different) colors
image
- when creating an external event, we have the now the choice of user (previously `input[type=hidden]`) and we can CC the new event to another users
image


inc/planningevent.class.php
```diff
@@ -236,6 +247,66 @@ function pre_updateInDB() {
...
+   /**
+    * Clone recurrent event into a non recurrent event
+    * (and add an exception to the orginal one)
+    *
+    * @param int $id of the serie
+    * @param string $start the new start for the event (in case of dragging)
+    *
+    * @return object the new object
+    */
+   public function createInstanceClone(int $id = 0, string $start = "") {
```
trasher: can you add a test on cloning?

PR #6933, opened 2020-02-12
Feature/marketplace

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes


remicollet:
Being able to execute code from a "web server" writable directory is a terrible idea, which is often being the cause of tons of security issue.

At least, please provide a global define such as "DISABLE_MARKET", so this feature can be totally removed from downstream distribution.

inc/marketplace/api/plugins.class.php
```diff
@@ -0,0 +1,259 @@
...
+      $GLPI_CACHE->set('marketplace_valid_plugins', $valid_plugins, HOUR_TIMESTAMP);
+      return $valid_plugins;
+   }
+
+
+   function matchGlpiVersion(string $constraint = ""):bool {
+      $glpi_version  = defined('GLPI_PREVER') ? GLPI_PREVER : GLPI_VERSION;
+      $glpi_version = "9.4";
+
+      return Semver::satisfies($glpi_version, $constraint);
```
remicollet: No.
GLPI doesn't follow SemVer (e.g. method removed in 9.4 so not compatible with 9.3)

remicollet:
> please provide a global define such as "DISABLE_MARKET", 

or allow `GLPI_MARKETPLACE_DIR `defined as `false`.

inc/autoload.function.php
```diff
@@ -287,8 +287,17 @@ function glpi_autoload($classname) {
...
-      $dir      = GLPI_ROOT . "/plugins/$plugname/inc/";
       $item     = str_replace('\\', '/', strtolower($plug['class']));
+
+      //try to load from marketplace folder
+      $dir = GLPI_MARKETPLACE_DIR . "/$plugname/inc/";
+
+      //then load from former plugins folder
+      if (!is_dir($dir)) {
+         $dir = GLPI_ROOT . "/plugins/$plugname/inc/";
+      }
```
remicollet: IMHO, plugins directory (proper installtion) should have higher priority than maketplace one.

inc/marketplace/controller.class.php
```diff
@@ -0,0 +1,175 @@
...
+   static function getAPI(): PluginsApi {
+      return self::$api ?? (self::$api = new PluginsApi());
+   }
+
+
+   function downloadPlugin():int {
+      $api      = self::getAPI();
+      $plugin   = $api->getPlugin($this->plugin_key, true);
+      $url      = $plugin['installation_url'] ?? "";
+      $file     = file_get_contents($url);
```
cedric-anne: This may use a huge amount of memory depending on archive size.
Using cURL to save it directly inside a file would consume less memory (code snippets here: https://stackoverflow.com/questions/1006604/saving-file-using-curl-and-php).

inc/marketplace/api/plugins.class.php
cedric-anne: Indeed.

The idea here is not to tells that GLPI repect Semver, but to use an existing tool to check compatibility declared in plugins API against GLPI version.
So if a plugins tells it is compatible with `~9.3.2 | ~9.4.0`, we will be able to validate this constraint even if GLPI does not follow Semver.

inc/autoload.function.php
cedric-anne: Can be done by redefining `PLUGINS_DIRECTORIES`.

inc/marketplace/api/plugins.class.php
remicollet: but `^9.4` won't work

cedric-anne: Indeed, but we expect plugin developpers to know that GLPI does not follow Semver and use strong constraints to handle with that.
Other projects does not follow Semver but can still be managed using its notation. For example, Laravel prior to version 6 was not following semver, but it was still possible to fetch its components using a composer.json configuration file.

inc/autoload.function.php
remicollet: No, default values have to be sane !
So order have to be change

remicollet:
BTW, end of the discussion for me on this one
I hate this feature

So will be removed from RPM packages, properly is possible, using local patch of needed.

cedric-anne:
> BTW, end of the discussion for me on this one
> I hate this feature
> 
> So will be removed from RPM packages, properly is possible, using local patch of needed.

I added a `GLPI_MARKETPLACE_ENABLE_DIRECT_INSTALL` constant. If defined to `false`, then installation from UI will be deactivated, but a download link will be proposed to the user. With this link, user will be able to download the plugin archive to its computer and manually install the plugin (building its own RPM for instance).

If you do not agree with this, please propose an alternative solution that can be, if possible, compatible with the ability to display the list of registered/popular/whatever plugins inside GLPI.

trasher:
Disabling the installation feature can be as simple as put the directory in read only mode. When writting is not possible, we can just propose the user to download the plugin and install it just like all regular plugins. Adding aslo a constant would duplicates this with no real benefit.

cedric-anne:
> Disabling the installation feature can be as simple as put the directory in read only mode. When writting is not possible, we can just propose the user to download the plugin and install it just like all regular plugins. Adding aslo a constant would duplicates this with no real benefit.

Indeed. I replace addition of `GLPI_MARKETPLACE_ENABLE_DIRECT_INSTALL` constant by a simple check on writable state of `GLPI_MARKETPLACE_DIR`. If it is not writable, installation/update button is replaced by a link to download the archive.

inc/autoload.function.php
cedric-anne: Order has been fixed.

ajax/marketplace.php
```diff
@@ -0,0 +1,109 @@
...
+
+   echo $_SESSION['marketplace_dl_progress'][$_REQUEST['key']] ?? 0;
+   exit;
+}
+
+
+// get common marketplace action, load GLPI framework
+include ("../inc/includes.php");
+
+Session::checkRight("config", READ);
```
trasher, line 52:
```suggestion
Session::checkRight("config", UPDATE);
```

inc/commonglpi.class.php
```diff
@@ -93,6 +93,37 @@ static function getType() {
...
+    * Same signature as CommonDBTM::can but in case of this class, we don't check instance rights
+    * so, id and input parameters are unused.
+    *
+    * @param integer $ID    ID of the item (-1 if new item)
+    * @param mixed   $right Right to check : r / w / recursive / READ / UPDATE / DELETE
+    * @param array   $input array of input data (used for adding item) (default NULL)
+    *
+    * @return boolean
+   **/
+   function can($ID, $right, array &$input = null) {
```
trasher, line 107: `$input` is never used

inc/glpinetwork.class.php
```diff
@@ -36,6 +36,184 @@
...
+
+      echo "<tr class='tab_bg_2'>";
+      echo "<td><label for='glpinetwork_registration_key'>" . __('Registration key') . "</label></td>";
+      echo "<td>" . Html::textarea(['name' => 'glpinetwork_registration_key', 'value' => $registration_key, 'display' => false]) . "</td>";
+      echo "</tr>";
+
+      if ($registration_key !== "") {
+         if (!empty($informations['validation_message'])) {
+            echo "<tr class='tab_bg_2'>";
+            echo "<td></td>";
```
trasher, line 75: to drop, or drop colspan

inc/commonglpi.class.php
orthagh: yes, due to inheritance

ajax/marketplace.php
```diff
@@ -0,0 +1,110 @@
...
+
+if (isset($_REQUEST['key'])) {
+   $marketplace_ctrl = new MarketplaceController($_REQUEST['key']);
+   if ($_REQUEST["action"] == "download_plugin"
+      || $_REQUEST["action"] == "update_plugin") {
+      $marketplace_ctrl->downloadPlugin();
+   }
+   if ($_REQUEST["action"] == "clean_plugin") {
+      if ($marketplace_ctrl->cleanPlugin()) {
+         echo "cleaned";
```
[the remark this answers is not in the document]
orthagh: used in marketplace.js to know if we can drop li (or not)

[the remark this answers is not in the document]
orthagh: not displayed so

inc/marketplace/controller.class.php
```diff
@@ -0,0 +1,347 @@
...
+            break;
+         }
+      }
+
+      return self::hasWriteAccess();
+   }
+
+
+   /**
+    * Check if a ginve plugin has on update online
```
trasher, line 182:
```suggestion
    * Check if a given plugin has on update online
```

inc/based_config.php
```diff
@@ -83,6 +83,9 @@
       'GLPI_NETWORK_SERVICES'             => 'https://services.glpi-network.com', // GLPI Network services project URL
       'GLPI_NETWORK_REGISTRATION_API_URL' => '{GLPI_NETWORK_SERVICES}/api/registration/',
       'GLPI_MARKETPLACE_PLUGINS_API_URI'  => '{GLPI_NETWORK_SERVICES}/api/glpi-plugins/',
+      'GLPI_MARKETPLACE_PLUGINS_API_URI'  => '{GLPI_NETWORK_SERVICES}/api/glpi-plugins/',
```
cedric-anne, line 86:
```suggestion
```

PR #6936, opened 2020-02-12
[9.5] distinguish input names for file uploads 

btry:
I need for Formcreator to handle image uploads in several textarea fields. To solve this I need to update how are handled uplaods for ticket itemtype

This PR is a proposal to distinguish input names when there areseveral textarea (with tinyMCE) in a single form.

Please note that this PR is also the base for an other which shall fixs the loss of uploads when a form is rejected then displayed again. This occurs when a user creates a ticket with an uploaded document or screenshot, and the form does not meets requirements (like a mandatory field).

- upload in ticket
  - helpdesk
    - [x] upload of a single screenshot
    - [x] upload of a single file
    - [x] upload of both
    - [x] respect o mandatory document (upload of either a file or a screenshot)
    - [x] uploads of a screenshot when creating / updating a task
    - [x] uploads of a screenshot in a solution
    - [x] reminder and KB item are not modified and still working on creation and update
  - simplified interface
    - [x] upload of a single screenshot
    - [x] upload of a single file
    - [X] upload of both
    - [X] respect of mandatory document (upload of either a file or a screenshot)
- [X] import of inline images in emails (mailcollector)
- [X] import of attachments in emails (mailcollector)

EDIT: Added support of form reload without loosing the uploaded screenshots. The uploads are displayed out of the file input. This allows to cut the functional relation between the file input and the textarea,

image

EDIT: 

Removed the file upload area which only supports screenshots. Other file types are ignored.

image

EDIT :

Found possible unwanted behaviors:
- saving a solution templates does not generates a document. (no addFiles() call). A screenshot is saved as a base64 encoded inline image, and when used to fill a solution, the screenshot remains as is. No document generated in a solution created from a solutin template
- Images in the body of an email are saved as attachment. A ticket generated from an email does not contains the image in the content, but only as an attachment.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes (allows several textareas with uploads in  the same form)
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | ???
| Fixed tickets | #number


inc/ticket.class.php
```diff
@@ -4082,6 +4082,8 @@ static function getDefaultValues($entity = 0) {
                'type'                      => $type,
                '_documents_id'             => [],
                '_tasktemplates_id'         => [],
+               '_content'                  => [],
+               'tag_content'               => [],
```
trasher: missing leading `_|`?

btry: yes, missing

btry:
Yes, adding a pic in a new task is broken. Confirmed

inc/commonitilobject.class.php
```diff
@@ -1745,6 +1760,7 @@ function post_addItem() {
 
       // Add document if needed, without notification
       $this->input = $this->addFiles($this->input, ['force_update' => true]);
+      $this->input = $this->addFiles($this->input, ['name' => 'content', 'force_update' => true]);
```
cedric-anne: Can you explain why there is 2 calls to addFiles here (in a code comment) ?

btry: There are 2 calls because the file area in a ticket handles the uploads dropped in its own area, but also screenshots dropped in the textarea. The textarea is named "content", and the file area is named "filename". 

The first call handles the uploads found in $this->input['_filename'] 

The second call handles files uploads (aka screenshots) found in this->['_content']

trasher: Names seems confusing to me

btry: I agree with you, that's whiy I'm cautious with this PR.

However, I believe I can continue the PR to completely cut the functional link between the file input and the textarea. This would slightly change the UI, and solve this mess.

inc/html.class.php
```diff
@@ -5731,13 +5731,15 @@ static function textarea($options = []) {
       $p['cols']              = 100;
       $p['rows']              = 15;
       $p['multiple']          = true;
+      $p['required']          = false;
```
btry: useful to remove the raw HTML in Ticket::showForm()

inc/ticket.class.php
```diff
@@ -4865,18 +4867,60 @@ function showForm($ID, $options = []) {
          $content = Html::cleanPostForTextArea($content);
       }
 
-      $content = Html::setRichTextContent(
-         $content_id,
-         $content,
-         $rand,
-         !$canupdate
-      );
```
btry: Cannot call this method because the inner call to Html::clean() breaks reload of an invalid form submission

inc/ticket.class.php
```diff
@@ -4865,18 +4867,60 @@ function showForm($ID, $options = []) {
...
-               $content."</textarea></div>";
+         Html::textarea([
+            'name'            => 'content',
+            'editor_id'       => $content_id,
+            'required'        => $tt->isMandatoryField('content'),
+            'rows'            => $rows,
+            'enable_richtext' => true,
+            'value'           => $content,
+         ]);
+         if (isset($this->input['_content'])) {
```
btry: in this if() is displayed the uploaded file names and a button to remove them (see fileupload.js, function displayUploadedFile()

inc/ticket.class.php
```diff
@@ -4865,18 +4867,60 @@ function showForm($ID, $options = []) {
...
+               echo "<img src='$extensionIcon' title='$extension'>&nbsp;";
+               echo "<b>" . $upload['display'] . "</b>&nbsp;(" . Toolbox::getSize($upload['size']) . ")";
+               echo Html::hidden('_content[' . $uploadId . ']', ['value' => $upload['name']]);
+               echo Html::hidden('_prefix_content[' . $uploadId . ']', ['value' => $upload['prefix']]);
+               echo Html::hidden('_tag_content[' . $uploadId . ']', ['value' => $tag['name']]);
+               $deleteUpload = "deleteImagePasted({0:'" . $upload['id'] . "', 1:'" . $upload['id'] . "'+'2'}, '" . $tag['tag'] . "', tinymce.get('$content_id'))";
+               echo '<span class="fa fa-times-circle pointer" onclick="' . $deleteUpload . '"></span>';
+               echo "</p>";
+            }
+         }
```
btry: the above should move to Html::textarea() to properly handle uploads everywhere in GLPI (and plugins). This would require a nex key in the options array with a sub array of 3 keys :  _field, _tag_field and  _content_field

inc/ticket.class.php
trasher: All changes on the way rich text is handled always have been consequences....
I'm still against any changes on that.

btry: call to setRichTextContent is replaced by its content, with one difference : the call to Html::clean() below is done only if the ticket is displayed from an instace saved in DB. In other cases it is displayed from a user input (after a refresh of the form or after the form is displayed again after a mandatory field is not properly filled (i.e : category).

inc/commonitiltask.class.php
```diff
@@ -1463,7 +1468,7 @@ function showForm($ID, $options = []) {
                       'value'             => $this->fields["content"],
                       'rand'              => $rand_text,
                       'editor_id'         => $content_id,
-                      'enable_fileupload' => true,
+                      'enable_fileupload' => false,
```
cedric-anne: Uploading files (not necessarly inlined images) during task may be a used feature, and seems to work correclty. You should not remove it.
```suggestion
```

inc/html.class.php
```diff
@@ -5770,6 +5789,96 @@ static function textarea($options = []) {
...
+
+      //merge default options with options parameter
+      $p = array_merge($p, $options);
+
+      // div who will receive and display file list
+      $display = "<div id='".$p['filecontainer']."' class='fileupload_info'>";
+      if (isset($p['uploads']['_' . $p['name']])) {
+         foreach ($p['uploads']['_' . $p['name']] as $uploadId => $upload) {
+            $prefix  = substr($upload, 0, 23);
+            $displayName = substr($upload, 23);
```
cedric-anne: An in-code comment would be appreciate for this.

inc/html.class.php
```diff
@@ -5770,6 +5789,96 @@ static function textarea($options = []) {
...
+            $display .= Html::hidden($name, ['value' => $tag['name']]);
+
+            // show button to delete the upload
+            $getEditor = 'null';
+            if ($p['editor_id'] != '') {
+               $getEditor = "tinymce.get('" . $p['editor_id'] . "')";
+            }
+            $textTag = $tag['tag'];
+            $domItems = "{0:'" . $upload['id'] . "', 1:'" . $upload['id'] . "'+'2'}";
+            $deleteUpload = "deleteImagePasted($domItems, '$textTag', $getEditor)";
```
cedric-anne: It works by just passing only one element. What is this `'" . $upload['id'] . "'+'2'` ?
```suggestion
            $deleteUpload = "deleteImagePasted(['{$upload['id']}'], '$textTag', $getEditor)";
```

inc/ticket.class.php
cedric-anne: Whatever changes made here, it should be reported to Problem and Change forms.

inc/ticket.class.php
```diff
@@ -4865,18 +4898,33 @@ function showForm($ID, $options = []) {
...
-               ($tt->isMandatoryField('content') ? " required='required'" : '') . ">" .
-               $content."</textarea></div>";
+         $uploads = [];
+         if (isset($this->input['_content'])) {
+            $uploads['_content'] = $this->input['_content'];
+            $uploads['_tag_content'] = $this->input['_tag_content'];
+         }
+         Html::textarea([
+            'name'            => 'content',
+            'filecontainer'   => 'cotnent_info',
```
```suggestion
            'filecontainer'   => 'content_info',
```

inc/ticket.class.php
```diff
@@ -3916,17 +3918,46 @@ function showFormHelpdesk($ID, $ticket_template = false) {
...
-               " . ($tt->isMandatoryField('content') ? " required='required'" : '') .">".
-                $content."</textarea></div>";
+         $uploads = [];
+         if (isset($options['_content'])) {
+            $uploads['_content'] = $options['_content'];
+            $uploads['_tag_content'] = $options['_tag_content'];
+         }
+         Html::textarea([
+            'name'            => 'content',
+            'filecontainer'   => 'cotnent_info',
```
```suggestion
            'filecontainer'   => 'content_info',
```

inc/ticket.class.php
cedric-anne: We could apply following patch:
```diff
diff --git a/inc/html.class.php b/inc/html.class.php
index 180ffba60b..ae6be92125 100644
--- a/inc/html.class.php
+++ b/inc/html.class.php
@@ -95,6 +95,13 @@ class Html {
          ]
       );
 
+      // Special case : remove the 'denied:' for base64 img in case the base64 have characters
+      // combinaison introduce false positive
+      foreach (['png', 'gif', 'jpg', 'jpeg'] as $imgtype) {
+         $value = str_replace('src="denied:data:image/'.$imgtype.';base64,',
+               'src="data:image/'.$imgtype.';base64,', $value);
+      }
+
       $value = str_replace(["\r\n", "\r"], "\n", $value);
       $value = preg_replace("/(\n[ ]*){2,}/", "\n\n", $value, -1);
 
```

This is already done on `Toolbox::unclean_html_cross_side_scripting_deep()`. Should be mutualized.

inc/html.class.php
btry: This method mimics the HTML snippet produced by displayUploadedFile() in fileupload.js . I did not found the reason of this concatenation in the original JS code.

inc/html.class.php
```diff
@@ -5753,6 +5770,15 @@ static function textarea($options = []) {
                         });
                      ");
       }
+      if (!$p['enable_fileupload'] && $p['enable_richtext']) {
```
trasher: What are those new options? I'm not sur I understand the condition here.

btry: The code in the if() statement generates an area where is displayed the file uploads. Example when a user crates a ticket

image

Instead of displaying the uploads of the textarea (aka screenshots)  in the file input, those uplaods are displayed just under the textarea. 

For backward compatibility, just in case, I avoid to generate the file uploads area if a textarea is explicitly linked to a file input via the options. This makes sense : when enable_fileuploads is false, only screenshots are allowed. If true a file input will be generated by the call to textarea()

btry: additional note : the latest case exists when creating a ticket  task

tests/functionnal/Ticket.php
```diff
@@ -2912,4 +2916,79 @@ public function testMerge() {
...
+         'value' => $content
+      ]);
+      $_SESSION['glpiactiveprofile']['tickettemplates_id'] = $ticketTemplate->getID();
+
+      $this->output(
+         function () {
+            $instance = new \Ticket();
+            $instance->showForm('0');
+         }
+      )->contains('src="data:image/png;base64,' . $this->inlineImage . '"');
```
cedric-anne: It is preferable to restore previous state of session before doing a assertions.
```suggestion
      $session_tpl_id_back = $_SESSION['glpiactiveprofile']['tickettemplates_id'];
      $_SESSION['glpiactiveprofile']['tickettemplates_id'] = $ticketTemplate->getID();

      $this->output(
         function () {
            $instance = new \Ticket();
            $instance->showForm('0');
            $_SESSION['glpiactiveprofile']['tickettemplates_id'] = $session_tpl_id_back;
         }
      )->contains('src="data:image/png;base64,' . $this->inlineImage . '"');
```

cedric-anne:
I found some bugs during tests:
 - pasted images in a followup remains on base64,
 - uploaded file in a followup does not appears in timeline,
 - uploaded file in a task does not appears in timeline.

I did not test every textarea, should be done.

btry:
OK, thanks : i'll add tests for them and fix it

inc/mailcollector.class.php
```diff
@@ -929,7 +929,7 @@ function buildTicket($uid, \Laminas\Mail\Storage\Message $message, $options = []
       // max size = 0 : no import attachments
       if ($this->fields['filesize_max'] > 0) {
          if (is_writable(GLPI_TMP_DIR)) {
-            $tkt['_filename'] = $this->getAttached($message, GLPI_TMP_DIR."/", $this->fields['filesize_max']);
+            $tkt['_content'] = $this->getAttached($message, GLPI_TMP_DIR."/", $this->fields['filesize_max']);
```
cedric-anne: I may be wrong, but inline attachement (images in a rich text) only should be pushed in `_content`. For attached documents (a PDF file for instance), should it be pushed in `_filename` ?

btry:
@cedric-anne 

-    pasted images in a followup remains on base64, => fixed
-    uploaded file in a followup does not appears in timeline, => fixed
-    uploaded file in a task does not appears in timeline. => is already true in 9.5/bugfixes. Also, I'm not sure that a file added in a task should go in the timeline.

inc/mailcollector.class.php
btry: I notice that the base branch has a bug here : inline images generate an incorrect img tag with "denied:" in the src attribute. I cancel this change which has no effect. The bug should be fixed with an other PR.

tests/functionnal/ITILFollowup.php
```diff
@@ -265,4 +264,122 @@ public function testDoNotComputeTakeintoaccount() {
       $_SESSION['glpiset_default_tech']      = $oldConf['glpiset_default_tech'];
       $_SESSION['glpiset_default_requester'] = $oldConf['glpiset_default_requester'];
    }
+
+   public function testScreenshotConvertedIntoDocument() {
+      // Test uploads for item creation
+      $ticket = new \Ticket();
+      $ticket->add([
```
trasher: You must test add/update/etc result

tests/functionnal/SolutionTemplate.php
```diff
@@ -0,0 +1,65 @@
...
+ */
+
+namespace tests\units;
+
+use \DbTestCase;
+
+/* Test for inc/itilsolution.class.php */
+
+class SolutionTemplate extends DbTestCase {
+   public function testScreenshotConvertedIntoRawBase64() {
```
cedric-anne: This test does not make sense for me. Pasting images in SolutionTemplate should be blocked. See #7058 .

btry: test removed

btry:
@cedric-anne : 
1 - the patch shall fix it. The latest rebases probably broke it, then I need to check again the whole use cases described in the PR.

2 - I did tests by comparing 9.4, 9.5 and 9.5 with this pr. This was weeks ago now, and I remember I noticed an issue. This should be this one. I propose to re-check if this occurs in GLPI 9.4 and handle this if needed in an other PR, after this one is merged (because I really need it, and its fix may be not impacted by this PR).

btry:
@cedric-anne : I tried to reproduce the first point you mentioinnen, and it worked well for me. An image in the ticket's content is properly reloaded when the mandatory category is not filled.

image

cedric-anne:
> @cedric-anne : I tried to reproduce the first point you mentioinned, and it worked well for me. An image in the ticket's content is properly reloaded when the mandatory category is not filled.
> 
> image

This is due to another bug, not related to this PR, I will create an issue for this one.

inc/html.class.php
trasher: Still not sure I understand, but `enable_richtext` should probably not be used.

btry: Look the occurences of calls to Html::uploadedFiles()

There is an occurence in Html::file() itself called by Html::textare() when enable_fileuploads is true.

When the Html::textarea is called with enable_fileuploads == false then we may still need to support uploads of screenshots embedded in the text. 

This is needed in Formcreator to allow rich text in a texstarea without making visible a file upload area (drag  /drop area + browse button).

Moreover it may be an interesting target to remove all raw textarea in GLPI code (see ```git grep "<textarea"```) . I did the pr with this target in mind. At least it will help me to reach this goal in plugins.

trasher:
As said on our internal chat; I've tested and do not find issues; but I'm absolutely not sure I've covered all cases. Ok for me so far.

cedric-anne:
@btry 
We are waiting for your GO.
When you will consider this PR ready to be merged, please rebase on 9.5/bugfixes and squash commit.

btry:
@cedric-anne : rebased (1 conflict was solved in ticket.class.php, line 5007)

PR #6946, opened 2020-02-17
Fix web tests

AdrienClairembault:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

Fix missing ips range in test database (use '127.0.0.1' instead of numeric value as it will be handled by ip2long).

### DIFF
--- a/tests/APIBaseClass.php
+++ b/tests/APIBaseClass.php
@@ -88,8 +88,8 @@ abstract class APIBaseClass extends \atoum {
          (int)$apiclient->add([
             'name'             => 'test app token',
             'is_active'        => 1,
-            'ipv4_range_start' => 2130706433,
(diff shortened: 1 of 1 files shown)


AdrienClairembault:
This does not prevent the tests from working, however the ip values are currently ignored and set to null in the database because they are parsed by the ip2long function :

```sh
php > echo ip2long("127.0.0.1");
2130706433
php > echo ip2long(2130706433);
false
```

It may be possible that this was accidently fixed by @cedric-anne since he changed it that we that we do not pass the value to ip2long if it's already an integer (not sure if this was merged or not).

If this is the case then both ways work, however `127.0.0.1` is way more "human readable" than `2130706433` so it may be best to use it.

PR #6947, opened 2020-02-17
Add GetUser(s)Picture(s)Url endpoint to REST API

AdrienClairembault:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

Add two new endpoints:
* getUserPictureUrl
* getUsersPicturesUrl

These endpoints are needed to allow self service users to load the author's picture of a given followup.

Examples:

#### getUserPictureUrl
```http
GET glpi/apirest.php/getUsersPicturesUrl/2  

"http://glpi/front/document.send.php?file=_pictures/fc/2_5e4a6ff3282fc.png"
```
#### getUsersPicturesUrl
```http
GET glpi/apirest.php/getUsersPicturesUrl?ids[0]=2

{
    "2": "http://glpi/front/document.send.php?file=_pictures/fc/2_5e4a6ff3282fc.png"
}
```


remicollet:
Retrieved info is not a URL but a local path... which is not accessible via any URI
Doesn't look like something useful for a REST client...

remicollet:
Again, no value.

If the client use the API, it is connected for the API (using the token), so is NOT authenticated in GLPI, so this URL won't be allowed.

If you need the picture, return the picture.

AdrienClairembault:
@orthagh @canne 
I checked the possibility to access document.send.php with the API token as we discussed last week.
While it has the benefit of also allowing access to inline images, it felt hacky to reuse the api session logic into another different component.

I've tried the alternate solution (new endpoint) in #6963 with a workaround for inline images.
This seem better for me (let the API handle all the API stuff).

Let me know which solution I should continue with.

PR #6991, opened 2020-02-28
API support for deprecated itemtypes

AdrienClairembault:
Keep support of deprecated itemtypes in the API.

Each API calls that may concern a deprecated item (getItem, getItems, createItems, updateItems, deleteItems, listSearchOptions and search) will check the itemtype at the beginning of their execution.

If the itemtype is deprecated, it will be corrected (e.g. TicketFollowup -> ITILFollowup) and the API will remember that there is an ongoing depreciation.

Later during the call, some data will retrofitted before output to the deprecated specs and / or upgraded to the current spec before being used.

API stuff was moved to an API folder and a `Glpi\Api` namespace.

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

### DIFF
--- a/inc/api/api.class.php
+++ b/inc/api/api.class.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
  * ---------------------------------------------------------------------
  * GLPI - Gestionnaire Libre de Parc Informatique
@@ -34,8 +35,36 @@
  * @since 9.1
  */
 
+namespace Glpi\Api;
+
+use APIClient;
+use Auth;
+use Change;
+use CommonDevice;
+use CommonGLPI;
+use Config;
+use Contract;
+use Document;
+use Dropdown;
 use Glpi\Exception\ForgetPasswordException;
 use Glpi\Exception\PasswordTooWeakException;
+use Html;
+use Infocom;
+use Item_Devices;
+use Log;
+use Michelf\MarkdownExtra;
+use NetworkEquipment;
+use NetworkPort;
+use Notepad;
+use Problem;
+use QueryExpression;
+use SavedSearch;
+use Search;
+use Session;
+use Software;
+use Ticket;
+use Toolbox;
+use User;
 
 abstract class API extends CommonGLPI {
 
@@ -49,6 +78,7 @@ abstract class API extends CommonGLPI {
    protected $ipnum         = "";
    protected $app_tokens    = [];
    protected $apiclients_id = 0;
+   protected $deprecated_item = null;
 
    /**
     * First function used on api call
@@ -490,7 +520,6 @@ abstract class API extends CommonGLPI {
    }
 
 
-
    /**
     * Return the instance fields of itemtype identified by id
     *
@@ -521,6 +550,7 @@ abstract class API extends CommonGLPI {
       global $CFG_GLPI, $DB;
 
       $this->initEndpoint();
+      $itemtype = $this->handleDepreciation($itemtype);
 
       // default params
       $default = ['expand_dropdowns'  => false,
@@ -1088,6 +1118,14 @@ abstract class API extends CommonGLPI {
          );
       }
 
+      // Convert fields to the format expected by the deprecated type
+      if ($this->isDeprecated()) {
+         $fields = $this->deprecated_item->mapCurrentToDeprecatedFields($fields);
+         $fields["links"] = $this->deprecated_item->mapCurrentToDeprecatedHateoas(
+            $fields["links"] ?? []
+         );
+      }
+
       return $fields;
    }
 
@@ -1132,6 +1170,7 @@ abstract class API extends CommonGLPI {
       global $DB;
 
       $this->initEndpoint();
+      $itemtype = $this->handleDepreciation($itemtype);
 
       // default params
       $default = ['expand_dropdowns' => false,
@@ -1194,8 +1233,9 @@ abstract class API extends CommonGLPI {
           && isset($this->parameters['parent_id'])) {
 
          // check parent itemtype
-         if (!class_exists($this->parameters['parent_itemtype'])
-             || !is_subclass_of($this->parameters['parent_itemtype'], 'CommonDBTM')) {
+         if (!Toolbox::isCommonDBTM($this->parameters['parent_itemtype'])
+            && !Toolbox::isAPIDeprecated($this->parameters['parent_itemtype'])
+         ) {
             $this->returnError(__("parent itemtype not found or not an instance of CommonDBTM"),
                                400,
                                "ERROR_ITEMTYPE_NOT_FOUND_NOR_COMMONDBTM");
@@ -1331,6 +1371,15 @@ abstract class API extends CommonGLPI {
             }
          }
       }
+      // Break reference
+      unset($fields);
+
+      // Map values for deprecated itemtypes
+      if ($this->isDeprecated()) {
+         $found = array_map(function($fields) {
+            return $this->deprecated_item->mapCurrentToDeprecatedFields($fields);
+         }, $found);
+      }
 
       return array_values($found);
    }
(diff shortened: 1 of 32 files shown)


inc/api/api.class.php
```diff
@@ -1537,14 +1577,23 @@ protected function searchItems($itemtype, $params = []) {
...
+         }
+      }
 
       // Check the criterias are valid
       if (isset($params['criteria']) && is_array($params['criteria'])) {
 
          // use a recursive closure to check each nested criteria
-         $check_message = "";
-         $check_criteria = function($criteria) use (&$check_criteria, $soptions, $check_message) {
+         $check_criteria = function($criteria) use (&$check_criteria, $soptions) {
```
AdrienClairembault: Bugfix here, the error message was never displayed.

inc/api/apirest.class.php
```diff
@@ -338,9 +342,19 @@ private function getItemtype($index = 0, $recursive = true, $all_assets = false)
                $itemtype                            = $additional_itemtype;
             }
```
AdrienClairembault: Fix a bug introduced by #6850, the special case "AllAssets" is not a real class and can't be instantiated by reflection.

inc/api/apirest.class.php
```diff
@@ -338,9 +342,19 @@ private function getItemtype($index = 0, $recursive = true, $all_assets = false)
...
+
+            // Load namespace for deprecated
+            if (\Toolbox::isAPIDeprecated($itemtype)) {
+               $itemtype = "Glpi\Api\Deprecated\\$itemtype";
+            }
+
             // Get case sensitive itemtype name
             $rc = new \ReflectionClass($itemtype);
-            $itemtype = $rc->getName();
+            $itemtype = $rc->getShortName();
```
AdrienClairembault: Use getShortName to get the class name without namespace.

inc/itilfollowup.class.php
```diff
@@ -535,6 +535,14 @@ function rawSearchOptions() {
...
       ];
 
+      $tab[] = [
+         'id'                 => '6',
+         'table'              => $this->getTable(),
+         'field'              => 'itemtype',
+         'name'               => __('Request source'),
+         'datatype'           => 'dropdown'
+      ];
+
```
AdrienClairembault: New searchOption to search ITILFollowups by type (needed for /search/TicketFollowup).

inc/search.class.php
```diff
@@ -3338,7 +3339,7 @@ static function addDefaultToView($itemtype, $params) {
 
       // Add entity view :
       if (Session::isMultiEntitiesMode()
-          && $item->isEntityAssign()
+          && $entity_check
```
AdrienClairembault: Fix bug introduced by #6966, the special case "AllAssets" is not a real object and can't be instantiated.

tests/router.php
```diff
@@ -33,4 +33,11 @@
 define('GLPI_CONFIG_DIR', __DIR__);
 define('GLPI_PICTURE_DIR', __DIR__ . '/files/_pictures');
 
+// Avoid warnings because of missing globals
+$DEBUG_SQL = [
+    'queries' => [],
+    'errors'  => [],
+    'times'   => [],
+ ];
+
```
AdrienClairembault: When running webtests, this was generating a lot of warnings in the "php -S" server used to run the web tests.

@cedric-anne 
Maybe any warning or error emitted by this web server should mark the tests as failed (they are currently ignored) ?
No idea if this is possible.

tests/web/APIRest.php
```diff
@@ -76,8 +77,12 @@ protected function query(
       $params = [],
       $expected_code = 200,
       $expected_symbol = '',
-      bool $file = false
+      bool $no_decode = false
    ) {
+      if (!is_array($expected_code)) {
+         $expected_code = [$expected_code];
+      }
+
```
AdrienClairembault: Changed expected_code to an array as some call may return different valid codes like the search:
- 200
- 206 for partial content

Since the items created during the web tests are not removed from the database after their execution, we may not know if have enough items to get a 206.

AdrienClairembault:
@orthagh Here is the working example we discussed for TicketFollowup, waiting for your GO to add the others deprecated types and the missing doc blocs.

tests/web/APIRest.php
```diff
@@ -76,8 +77,12 @@ protected function query(
       $params = [],
       $expected_code = 200,
       $expected_symbol = '',
-      bool $file = false
+      bool $no_decode = false
    ) {
+      if (!is_array($expected_code)) {
+         $expected_code = [$expected_code];
```
orthagh: maybe, you should rename the parameter to $expected_code**s** (with plural)

orthagh:
Great start.

here is a list a next candidates for depreciation handling: 

- 9.5 :
   *  `Computer_SoftwareLicense` replaced by `Item_SoftwareLicense`
   * `Computer_SoftwareVersion` replaced by `Item_SoftwareVersion`

i will complete this list next (if i find any examples)

AdrienClairembault:
Now supports:
* Computer_SoftwareLicense
* Computer_SoftwareVersion
* ITILFollowup

I haven't found any other itemtypes that were replaced in the changelog, I think it's only these 3.

inc/api/apirest.class.php
trasher: Please open another PR with fixes, this one must be applied on 9.4/bf

trasher:
You must add license header in every new file

inc/api/deprecated/computer_softwarelicense.class.php
```diff
@@ -0,0 +1,73 @@
+<?php
+
+namespace Glpi\Api\Deprecated;
+
+if (!defined('GLPI_ROOT')) {
+   die("Sorry. You can't access this file directly");
+}
+
+/**
+ * @since 9.4.0
```
trasher: Wrong. Change has been done in 9.4, but the @since tag is used to tell when method has been added (9.5 here)

inc/api/apirest.class.php
AdrienClairembault: Done in #7080

tests/web/APIRest.php
```diff
@@ -74,10 +90,14 @@ protected function doHttpRequest($verb = "get", $relative_uri = "", $params = []
    protected function query(
       $resource = "",
       $params = [],
-      $expected_code = 200,
+      $expected_codes = 200,
```
trasher, line 93:
```suggestion
      $expected_codes = [200],
```

tests/web/deprecated/Computer_SoftwareLicense.json
```diff
@@ -0,0 +1,50 @@
+{
```
trasher, line 1: Data files for tests are usually in their own directory (emails-tests, fixtures, ...)

trasher:
`tests/web/error.log` should not have been commited

AdrienClairembault:
Requested changes are in 75fe68e and 22f75f6 (rebase was for conflicts).

inc/toolbox.class.php
```diff
@@ -3234,4 +3234,31 @@ public static function getHeader(string $name) {
...
       return $_SERVER[$name] ?? null;
    }
+
+   /**
+    * Check if the given class exist and extends CommonDBTM
+    *
+    * @param string $class
+    * @return bool
+    */
+   public static function isCommonDBTM(string $class): bool {
```
trasher, line 3244: This one and the following should be added to toolbox unit tests

tests/router.php
```diff
@@ -33,4 +33,13 @@
...
 define('GLPI_PICTURE_DIR', __DIR__ . '/files/_pictures');
 
+// Avoid warnings because of missing globals
+$DEBUG_SQL = [
+    'queries' => [],
+    'errors'  => [],
+    'times'   => [],
+];
+
+ini_set("error_log", "tests/web/error.log");
```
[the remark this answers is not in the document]
AdrienClairembault: The point of this is to make the tests fail if there is a warning or an error on the server.

The php built-in server (php -S) used for these web tests output errors in stdout/stderr as you said so the user is free to display it if he wants to but there is no repercussion in the actual result of the test case.

In many case, the user running the tests will not check the output, a few examples: 
- The CI env redirect both descriptors to /dev/null
- On my end, I use a zsh function for starting tests that will automatically start the web server process if the test target web/* (and kill it after). Both descriptors are also redirected to /dev/null in this case.

I think any error in the web server should trigger a test failure, it should not be up to the user to rely on external data (web server output) to determine the test success (especially for the CI env were the output will never be available and thus any warning or error will always be ignored).

AdrienClairembault:
Force pushed for conflict resolution.

PR #7048, opened 2020-03-16
dashboard return cache

orthagh:
missing return.

Not tested (yet), my 9.5 env si currently broken


orthagh:
finally, i was able to set up correctly my 9.5 env . and so the pr seems ok

PR #7051, opened 2020-03-18
To force focus on the button

tomolimo:
To prevent any event (especially the space or enter keys) on the background

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |

### DIFF
--- a/inc/html.class.php
+++ b/inc/html.class.php
@@ -5560,6 +5560,7 @@ class Html {
             dialogClass: 'glpi_modal',
             open: function(event, ui) {
                $(this).parent().prev('.ui-widget-overlay').addClass('glpi_modal');
+               $(this).next('div').find('button').focus();
             },
             close: function(){
                $(this).remove();


orthagh:
Olivier, could you describe with a sort the text the original you're trying to solve.
I think i guess regarding the code change, but to ease other searching for a similar issue, it be good to describe it a little.

Best Regards

tomolimo:
Yes, sorry, it's a bit too short.

Steps to reproduce:
1. Start a ticket creation
1. Do not input any description
1. Click on the 'Add' button: a dialog should show that "description field is mandatory"
1. Press on the space key (or enter key)
1. Then you'll see that overlay is darken and darken (due to multiple dialogs appearing).

When setting the focus on the button of the dialog it prevents such behaviours.

Thank you
Regards,
Tomolimo

PR #7118, opened 2020-04-01
Datacenter passive equipments

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

Add passive equipement "matching" Other equipements from former rack plugin.

I tested manual adding / updating, not the migration script atm. Will do with customer data.

Address these suggestions:

 - https://glpi.userecho.com/communities/1/topics/541-add-patch-panel-in-network-asses?comment_form_activate=true
- https://glpi.userecho.com/communities/1/topics/940-passive-painel-to-rack

### DIFF
--- a/front/passivedcequipment.form.php
+++ b/front/passivedcequipment.form.php
@@ -1,0 +1,113 @@
+<?php
+/**
+ * ---------------------------------------------------------------------
+ * GLPI - Gestionnaire Libre de Parc Informatique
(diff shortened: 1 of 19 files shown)


orthagh:
Thanks for the addition. i guess we can merge now.

PR #7159, opened 2020-04-22
Fix CTRL+enter for textareas

AdrienClairembault:
The "submit" input was replaced by a button in 9.5 so we need to update this selector.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

### DIFF
--- a/js/common.js
+++ b/js/common.js
@@ -665,7 +665,7 @@ var submitparentForm = function(input) {
    var form = $(input).closest('form');
 
    // find submit button(s)
-   var submit = form.find('button[type=submit]').filter('[name=add], [name=update]');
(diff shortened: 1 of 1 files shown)


js/common.js
```diff
@@ -665,7 +665,7 @@ var submitparentForm = function(input) {
    var form = $(input).closest('form');
 
    // find submit button(s)
-   var submit = form.find('input[type=submit]').filter('[name=add], [name=update]');
+   var submit = form.find('button[type=submit]').filter('[name=add], [name=update]');
```
trasher, line 668: why not just:
```suggestion
   var submit = form.find('[type=submit]').filter('[name=add], [name=update]');
```

PR #7162, opened 2020-04-23
Fix markup diplay when term match with html entity

stonebuzz:
With the Select2 search function, when we search term "gt" (to find Wellington for example) and if GLPI is configured to display the completename

Select2 escapes ">" and "<" when occurences is find and display ```&gt;``` on result set because templateResult underline matches on text

image

This PR fix this 

image

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | internal 19509

### DIFF
--- a/js/common.js
+++ b/js/common.js
@@ -936,11 +936,6 @@ var templateResult = function(result) {
       }
 
       var text = result.text;
-      if (text.indexOf('>') !== -1 || text.indexOf('<') !== -1) {
-         // escape text, if it contains chevrons (can already be escaped prior to this point :/)
-         text = jQuery.fn.select2.defaults.defaults.escapeMarkup(text);
-      };
-
       if (!result.id) {
          // If result has no id, then it is used as an optgroup and is not used for matches
          _elt.html(text);
--- a/js/common.js
+++ b/js/common.js
@@ -906,17 +906,17 @@ function markMatch (text, term) {
    }
 
    // Put in whatever text is before the match
-   _result.html(text.substring(0, match));
+   _result.html(escapeMarkupText(text.substring(0, match)));
 
    // Mark the match
    var _match = $('<span class=\'select2-rendered__match\'></span>');
-   _match.html(text.substring(match, match + term.length));
+   _match.html(escapeMarkupText(text.substring(match, match + term.length)));
 
    // Append the matching text
    _result.append(_match);
 
    // Put in whatever is after the match
-   _result.append(text.substring(match + term.length));
+   _result.append(escapeMarkupText(text.substring(match + term.length)));
 
    return _result.html();
 }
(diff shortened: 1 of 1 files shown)


trasher:
Does not sounds good, see #5646 at least

stonebuzz:
> Does not sounds good, see #5646 at least

seems good for me

image

cedric-anne:
This escaping was made to prevent XSS.

I may be wrong, but maybe escaping (`jQuery.fn.select2.defaults.defaults.escapeMarkup()` used on text and term should be moved:
 1. inside `if (!result.id) {` just before outputing the resut;
 2. inside `markMatch` function, after match has been found.

See #5646, #5468, #6094 and #6109 to be sure to not reintroduce bugs.

stonebuzz:
refactor code to escape markup when is needed

check on location
image

check on itil categories
image

and groups for display 
image

and search

image

cedric-anne:
> and search
> 
> image

You fixed display, but I think `gt` should not match. Escaping of searched term should be done after match has been found instead of being done before calling `markMatch` .

stonebuzz:
> > and search
> > image
> 
> You fixed display, but I think `gt` should not match. Escaping of searched term should be done after match has been found instead of being done before calling `markMatch` .

gt should match ie: Welligton and select2 match correctly

image

cedric-anne:
> > > and search
> > > image
> > 
> > 
> > You fixed display, but I think `gt` should not match. Escaping of searched term should be done after match has been found instead of being done before calling `markMatch` .
> 
> gt should match ie: Welligton and select2 match correctly
> 
> image

My bad. I will make some tests to ensure all previously fixed cases works correctly.

cedric-anne:
There is a security issue with your changes (this issue was fixed by #5468).

To reproduce:
 - create a location with a name like `My location <script>alert('PWNED');</script>`,
 - open a location for and display parent location dropdown values.

js/common.js
```diff
@@ -1007,3 +1002,7 @@ var getTextWithoutDiacriticalMarks = function (text) {
    // They are removed to keep only chars without their diacritical mark.
    return text.replace(/[\u0300-\u036f]/g, '');
 }
+
+var escapeMarkupText = function (text) {
+   return jQuery.fn.select2.defaults.defaults.escapeMarkup(text);
```
cedric-anne, line 1007: Data created in old GLPI versions may be "twice" escaped in DB, so you should keep this check to prevent reescaping an already escaped string.
```suggestion
   if (text.indexOf('>') !== -1 || text.indexOf('<') !== -1) {	
      // escape text, if it contains chevrons (can already be escaped prior to this point :/)
      text = jQuery.fn.select2.defaults.defaults.escapeMarkup(text);
   }
   return text;
```

PR #7192, opened 2020-04-29
Feature/new planning timeline view

gguib:
Hello,

Here a new feature that implements a new planning view by day. Less information cases 

| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes (directly in the view, for every type of tasks)


gguib:
Bad glpi version

Issue #1159, opened 2016-10-13
Ticket Timeline Enhancement : Display total of ticket tasks actiontime


tsmr:
![image](https://cloud.githubusercontent.com/assets/13178158/19343870/533e6ca0-9138-11e6-9f4a-c9a56dca17a1.png)

Like this.

I send a PR ?

tomolimo:
yes, good idea

orthagh:
i agree also.

Some suggestions: 
- Align right (you have some padding)
- maybe fade the color, currently it attract eye (same color and size as titles) for a secondary information

flegastelois:
Hi Xavier, good idea yes! 
What do you think if also add other quick statistics like "% Done tasks" / "% ToDo tasks", maybe on tooltip ? I don't know the best way for displaying.
Cheers,

tsmr:
Like this ?

![image](https://user-images.githubusercontent.com/13178158/80604867-93269a80-8a32-11ea-8ac0-7858bca6d92c.png)

PR #7199, opened 2020-04-29
Ticket Tasks Enhancement

tsmr:
An old ticket :P https://github.com/glpi-project/glpi/issues/1159

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1159

image

& hover

total-hover

### DIFF
--- a/inc/commonitilobject.class.php
+++ b/inc/commonitilobject.class.php
@@ -6660,66 +6660,54 @@ abstract class CommonITILObject extends CommonDBTM {
       //total_actiontime stat
       if (Session::getCurrentInterface() != 'helpdesk') {
          echo "<div class='timeline_stats'>";
-         echo "<h3>";
+
+         $taskClass  = $objType . "Task";
+         $task_table = getTableForItemType($taskClass);
+         $foreignKey = static::getForeignKeyField();
+
          $total_actiontime = 0;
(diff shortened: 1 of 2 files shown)


inc/commonitilobject.class.php
```diff
@@ -6657,7 +6657,71 @@ function viewEditSubitem" . $this->fields['id'] . "$rand(e, itemtype, items_id,
 
       echo "</ul>"; // timeline_choices
       echo "<div class='clear'>&nbsp;</div>";
-
+      //total_actiontime stat
+      if (Session::getCurrentInterface() != 'helpdesk') {
+         echo "<div class='timeline_stats'>";
+         echo "<h3>";
+         $total_actiontime = 0;
+         foreach ($DB->request("glpi_tickettasks",
```
trasher, line 6665: Do not call `request` inside of a loop; store the restul in a variable before and use that variable in the loop (for all loops).

trasher: Also, do **not** use raw SQL.

tsmr: Better ?

trasher: :+1:

PR #7239, opened 2020-05-05
factorize showFormHeader for itil classes

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

Internal ref 19703

This prevent change of entity when editing changes/problems.

`$options` passed to `showFormHeader` contains a `entities_id` (set by `getDefaultValues`) key who force the change of entities of change/problem

Ticket doesn't have the issue because it had a custom header and doesn't rely on standard `showFormHeader` (it it was the case, Ticket would have the same issue as the `getDefaultValues` is also present.

I moved (and adapt a little)  the custom header of Ticket into a CommonItilObject::showFormHeader to share the same base code between the 3 itemtypes (and fix the issues)

The else part on the end prevent display hazard in split view (same reason as Ticket don't use `showFormButtons` but less impactant currently).

### DIFF
--- a/inc/commonitilobject.class.php
+++ b/inc/commonitilobject.class.php
@@ -7419,6 +7419,65 @@ abstract class CommonITILObject extends CommonDBTM {
       }
    }
 
+
+   function showFormHeader($options = []) {
+      $ID   = $this->fields['id'];
+      $rand = mt_rand();
+
+      if (!$options['template_preview']) {
+         echo "<form method='post' name='form_ticket' enctype='multipart/form-data' action='".
+                Ticket::getFormURL()."'>";
+         if (isset($options['_projecttasks_id'])) {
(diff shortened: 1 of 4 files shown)


inc/commonitilobject.class.php
```diff
@@ -7419,6 +7419,65 @@ static function showSubForm(CommonDBTM $item, $id, $params) {
...
    }
 
+
+   function showFormHeader($options = []) {
+      $ID   = $this->fields['id'];
+      $rand = mt_rand();
+
+      if (!$options['template_preview']) {
+         echo "<form method='post' name='form_ticket' enctype='multipart/form-data' action='".
+                Ticket::getFormURL()."'>";
```
trasher, line 7429:
```suggestion
                static::getFormURL()."'>";
```

PR #7255, opened 2020-05-11
Fix Kanban counters and add form

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #number

The source of a card drag was the Kanban column body instead of the column itself, while the target was the column. This caused the new column counter to update correctly, while the original column would not update.

### DIFF
--- a/js/kanban.js
+++ b/js/kanban.js
@@ -627,7 +627,7 @@
                } else {
                   list_item += "<input type='checkbox'/>";
                }
-               list_item += "<span class='kanban-color-preview' style='background-color: "+column['header_color']+"'/>";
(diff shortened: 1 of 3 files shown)


cconard96:
Kanban still isn't good. The new item form is broken now too for some reason.

js/kanban.js
```diff
@@ -1297,7 +1298,7 @@
                if (value !== undefined) {
                   add_form += " value='" + value + "'";
                }
-               add_form += "/>";
+               add_form += "></input>";
```
trasher, line 1301: `<input />` was correct

PR #7256, opened 2020-05-12
Enhance Status Endpoint

### LINKED ISSUE #5922: status.php check IMAP authentication
status.php display Problem checking Mail authentication  (IMAP in my case).
The problem is due not setting $host address from $param table;

I solve adding this line 
$host = $param['address'];
after line 129 .

------------------------------
         $param = Toolbox::parseMailServerConnectString($method['connect_string'], true);
         $host = $param['address'];     ### Added line
         if ($param['ssl'] === true) {
------------------------------

Bye

**Describe the bug**

**Page(s) URL**

If applicable, page(s) URL where the bug happens.

**To reproduce**

**Expected behavior**

**Logs**

```
paste logs here
Find them in *-error.log files under glpi/files/_log/
```

**Screenshots**

```
paste here
```

**Additional context**

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | maybe
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #5922

Added ability to get JSON format for status endpoint in addition to the "legacy" plain-text format. Allows for providing more information to monitoring systems/scripts (I added the DB replication delay as an example).
To provide BC, the plain-text format is the default and is output in the same way it was before. Extra information like the replication delay is not shown.

Fixed #5922 while I was at it (I don't have an IMAP server anymore for dev. Someone will have to test this to be sure. Wasn't sure if it mattered if I used the host from the DB or connection string.).

I think eventually this endpoint should be added to the API instead of a public file in the root.

Need to think about how plugins report their status. Currently it looks like they just change the flag for GLPI_OK/GLPI_PROBLEM. With a the new format, they could report their own status and the status of their dependencies.

### DIFF
--- a/inc/console/system/checkstatuscommand.class.php
+++ b/inc/console/system/checkstatuscommand.class.php
@@ -1,0 +1,72 @@
+<?php
+/**
+ * ---------------------------------------------------------------------
+ * GLPI - Gestionnaire Libre de Parc Informatique
+ * Copyright (C) 2015-2018 Teclib' and contributors.
+ *
+ * http://glpi-project.org
+ *
+ * based on GLPI - Gestionnaire Libre de Parc Informatique
+ * Copyright (C) 2003-2014 by the INDEPNET Development Team.
+ *
+ * ---------------------------------------------------------------------
+ *
+ * LICENSE
+ *
+ * This file is part of GLPI.
+ *
+ * GLPI is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * GLPI is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with GLPI. If not, see <http://www.gnu.org/licenses/>.
+ * ---------------------------------------------------------------------
+ */
+
+namespace Glpi\Console\System;
+
+if (!defined('GLPI_ROOT')) {
+   die("Sorry. You can't access this file directly");
+}
+
+use Glpi\Console\AbstractCommand;
+use Glpi\System\Status\StatusChecker;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+
+class CheckStatusCommand extends AbstractCommand {
+
+   protected function configure() {
+      parent::configure();
+
+      $this->setName('glpi:system:status');
+      $this->setAliases(['system:status']);
+      $this->setDescription(__('Check system status'));
+      $this->addOption('format', 'f', InputOption::VALUE_OPTIONAL,
+         'Output format [plain or json]', 'plain');
+      $this->addOption('private', 'p', InputOption::VALUE_NONE,
+         'Status information publicity. Private status information may contain potentially sensitive information such as version information.');
+   }
+
+   protected function execute(InputInterface $input, OutputInterface $output) {
+
+      $format = strtolower($input->getOption('format'));
+      $status = StatusChecker::getFullStatus(!$input->getOption('private'), $format === 'json');
+
+      if ($format === 'json') {
+         $output->writeln(json_encode($status, JSON_PRETTY_PRINT));
+      } else {
+         $output->writeln($status);
+      }
+
+      return 0; // Success
+   }
+}
--- a/inc/system/status/statuschecker.class.php
+++ b/inc/system/status/statuschecker.class.php
@@ -1,0 +1,516 @@
+<?php
+/**
+ * ---------------------------------------------------------------------
+ * GLPI - Gestionnaire Libre de Parc Informatique
+ * Copyright (C) 2015-2019 Teclib' and contributors.
+ *
+ * http://glpi-project.org
+ *
+ * based on GLPI - Gestionnaire Libre de Parc Informatique
+ * Copyright (C) 2003-2014 by the INDEPNET Development Team.
+ *
+ * ---------------------------------------------------------------------
+ *
+ * LICENSE
+ *
+ * This file is part of GLPI.
+ *
+ * GLPI is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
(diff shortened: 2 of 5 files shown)


trasher:
> Fixed #5922 while I was at it (I don't have an IMAP server anymore for dev. Someone will have to test this to be sure. Wasn't sure if it mattered if I used the host from the DB or connection string.).

There are imap tests on the CI, it is possible to get that tested.

> I think eventually this endpoint should be added to the API instead of a public file in the root.

In addition, yes, instead, no. This would require the API to be up and running, status checks must be available in all cases. Code can be refactored to be used in both places.
It also may be added as a cli command, we already have one for dependencies/modules/etc checks.

cedric-anne:
> > I think eventually this endpoint should be added to the API instead of a public file in the root.
> 
> In addition, yes, instead, no. This would require the API to be up and running, status checks must be available in all cases. Code can be refactored to be used in both places.
> It also may be added as a cli command, we already have one for dependencies/modules/etc checks.

I agree with the fact that a status endpoint should be available even if API is disabled.

- Having a public endpoint returning plain text make it easier to use with just a simple "wget + grep" script.

- Having a private endpoint returning JSON can be usefull if you want to use advanced tools that will likely prefer the JSON format. Even if it sound weird to me to put this kind of endpoint into the REST or XMLRPC API, I do not have a better solution to propose with current state of GLPI. (My thought is that the right place for it would be in an API dedicated to backend services on which authentication would be made using some kind of "services accounts").

- Having a CLI endoint could be usefull (i.e. `php bin/console glpi:system:status`), but it should propose, at least, a human readable format for manual checks and a parsable format to be able to use it in an automated process.

cconard96:
You are right, there needs to be a public endpoint for at least basic status testing since the API is optional. Also, me calling the plan-text output "legacy" is not correct. I didn't realize that there were some tools that handled it better than JSON or that humans would be reading it directly. I'm not sure I really want to touch the API right now. I don't like how it is set up currently and I am holding out hope that the changes in version 10 makes it a bit better or easier to reimplement.

I can look at getting a command set up.

Also, I noticed the status endpoint is not tested at all. Atoum doesn't like me setting up a test class that doesn't correlate to a PHP class. Does anyone have a solution for that or should the logic just get moved into a class? I guess that would make it easier to have a public and private endpoint in the future.

trasher:
No problem if you cannot add the api endpoint, any API client can call the status.php with proper headers. Same for command line, it is an option, you do not have to worry with that ;)

As of atoum... Yes, you need a real class. I did not find the way to bypass that, but real cases are quite rare so I did not spend much time on this.
I generally create an empty class like https://github.com/glpi-project/glpi/blob/9.4/bugfixes/tests/Autoload.php when it is not possible to have a real class, and refactor code when it is possible.

inc/console/system/checkstatuscommand.class.php
```diff
@@ -0,0 +1,72 @@
...
+
+      if ($format === 'json') {
+         $output->writeln(json_encode($status));
+      } else {
+         $output->writeln($status);
+      }
+
+      return 0; // Success
+   }
+}
```
orthagh, line 72:
```suggestion
}

```

inc/system/status/statuschecker.class.php
```diff
@@ -0,0 +1,516 @@
...
+         }
+      } else {
+         $output .= "Crontasks_OK\n";
+      }
+
+      // Overall Status
+      $output .= "\nGLPI_{$status['glpi']['status']}\n";
+      return $output;
+   }
+}
```
orthagh, line 516:
```suggestion
}

```

status.php
```diff
@@ -39,197 +41,16 @@
...
-
-if ($ok) {
-   echo "GLPI_OK\n";
+if ($_SERVER['HTTP_ACCEPT'] === 'application/json') {
+   echo json_encode(StatusChecker::getFullStatus(true, true));
 } else {
-   echo "GLPI_PROBLEM\n";
-}
+   echo StatusChecker::getFullStatus(true, false);
+}
```
orthagh, line 56:
```suggestion
}

```

tests/functionnal/Glpi/System/Status/StatusChecker.php
```diff
@@ -0,0 +1,168 @@
...
+      // We have a non-existent IMAP server. Verify the status is PROBLEM.
+      $this->string($status['imap']['status'])->isEqualTo(GlpiStatusChecker::STATUS_PROBLEM);
+
+      // Make sure there are two stuck tasks
+      $this->string($status['crontasks']['status'])->isEqualTo(GlpiStatusChecker::STATUS_PROBLEM);
+      $this->array($status['crontasks']['stuck'])->size->isEqualTo(2);
+
+      // afterTestMethod will rollback the DB changes for us
+   }
+}
```
orthagh, line 168:
```suggestion
}

```

inc/console/system/checkstatuscommand.class.php
```diff
@@ -0,0 +1,72 @@
...
+
+class CheckStatusCommand extends AbstractCommand {
+
+   protected function configure() {
+      parent::configure();
+
+      $this->setName('glpi:system:status');
+      $this->setAliases(['system:status']);
+      $this->setDescription(__('Check system status'));
+      $this->addOption('format', 'f', InputArgument::OPTIONAL,
```
orthagh, line 53: It could be better here to have possible options. During my tests, i didn't what to set to have json format.

inc/console/system/checkstatuscommand.class.php
```diff
@@ -0,0 +1,72 @@
...
+         'Output format', 'plain');
+      $this->addOption('private', 'p', InputArgument::OPTIONAL,
+         'Status information publicity. Private status information may contain potentially sensitive information such as version information.', false);
```
orthagh, line 56: it seems the `private` option requires a value, is is necessary ?

image

inc/console/system/checkstatuscommand.class.php
```diff
@@ -0,0 +1,72 @@
...
+   }
+
+   protected function execute(InputInterface $input, OutputInterface $output) {
+
+      $format = strtolower($input->getOption('format'));
+      $status = StatusChecker::getFullStatus(!$input->getOption('private'), $format === 'json');
+
+      if ($format === 'json') {
+         $output->writeln(json_encode($status));
```
orthagh, line 65: i think we can add an additional option `pretty` to set `JSON_PRETTY_PRINT` in this function

cedric-anne: Not sure that a new option is mandatory, maybe json outpout can always be done with `JSON_PRETTY_PRINT`.

cedric-anne:
I rebased to update test suite.

PR #7257, opened 2020-05-12
Fix Server Room events

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

Event type should be 'dcrooms' not 'serverroms'. Causes missing Source in log list and broken ID column link.

### DIFF
--- a/inc/update.class.php
+++ b/inc/update.class.php
@@ -462,7 +462,6 @@ class Update extends CommonGLPI {
          case "9.4.5":
             include_once "{$updir}update_945_946.php";
             update945to946();
-            break;
 
          case "9.4.6":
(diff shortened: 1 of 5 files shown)


trasher:
See CI results, the migration is not run at all; there are others changes to be done; check previous add of a new update file.

inc/update.class.php
```diff
@@ -464,6 +464,11 @@ public function doUpdates($current_version = null) {
             update945to946();
             break;
```
trasher, line 465:
```suggestion
```

PR #7260, opened 2020-05-12
User synchro: remove double escaping for manager field

AdrienClairembault:
Internal ref: 19550.

When synchronizing an user, the value used to search for his manager was escaped twice.
If the manager DN contained any escaped chars like `\` it would then never be found.

First escaping in `User::getFromLDAP()`  
```php
$val = Toolbox::addslashes_deep($val);
```
Second escaping in `User::getIdByField()`
```php
$iterator = $DB->request([
         'SELECT' => 'id',
         'FROM'   => self::getTable(),
         'WHERE'  => [$field => addslashes($value)]
]);
```

I've removed the second escaping for this specific case and kept it as a default parameter for the function since it is used in others places that may need the escaping.

----

Concrete example:

Manager DN to search for: `cn=Clairembault\2C Adrien,cn=abcd,dc=nodomain`
-> `\2C` is an encoded coma here, the displayed CN in the LDAP is `Clairembault, Adrien`

With double escaping, we end up doing this request which will never find anything:
```sql
SELECT `id`
FROM `glpi_users`
WHERE `user_dn` = 'cn=Clairembault\\\\2C Adrien,cn=abcd,dc=nodomain'
```

---

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

### DIFF
--- a/inc/user.class.php
+++ b/inc/user.class.php
@@ -1593,7 +1593,7 @@ class User extends CommonDBTM {
                      break;
 
                   case 'users_id_supervisor':
-                     $this->fields[$k] = self::getIdByField('user_dn', $val);
+                     $this->fields[$k] = self::getIdByField('user_dn', $val, false);
                      break;
 
                   default :
@@ -4403,13 +4403,17 @@ class User extends CommonDBTM {
     *
     * @return integer
     */
-   static function getIdByField($field, $value) {
+   static function getIdByField($field, $value, $escape = true) {
       global $DB;
+
+      if ($escape) {
+         $value = addslashes($value);
+      }
 
       $iterator = $DB->request([
          'SELECT' => 'id',
          'FROM'   => self::getTable(),
-         'WHERE'  => [$field => addslashes($value)]
+         'WHERE'  => [$field => $value]
       ]);
 
       if (count($iterator) == 1) {
--- a/tests/LDAP/AuthLdap.php
+++ b/tests/LDAP/AuthLdap.php
@@ -46,9 +46,10 @@ class AuthLDAP extends DbTestCase {
       //make sure bootstrapped ldap is active and is default
       $this->boolean(
          $this->ldap->update([
-            'id'           => $this->ldap->getID(),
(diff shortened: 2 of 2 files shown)


trasher:
I guess a better solution would be to fix encoding coming from ldap, what is that kind of encoding?

By the way, it would be usefull to add tests cases with you case, `'`, etc.

AdrienClairembault:
I've updated the description to be more precise, there should be no SQL injections since values are still escaped (but only once).

trasher:
> > Wait... No. If there are some values that should be escaped in the name, this will open a SQL injection possibility (and query will fail anyways)
> 
> I've updated the description to be more precise, there should be no SQL injections since values are still escaped (but only once).

Oh, OK, yes, it's exploicitely called on the value passed, did not see this. Just look if a test case with your issue can be added then and it's ok for me ;)

AdrienClairembault:
> I guess a better solution would be to fix encoding coming from ldap, what is that kind of encoding?

For the encoding, it seems consistent with ldap_escape so it should be standard: 
```sh
php > echo ldap_escape(',');
\2c
```

AdrienClairembault:
> > > Wait... No. If there are some values that should be escaped in the name, this will open a SQL injection possibility (and query will fail anyways)
> > 
> > 
> > I've updated the description to be more precise, there should be no SQL injections since values are still escaped (but only once).
> 
> Oh, OK, yes, it's exploicitely called on the value passed, did not see this. Just look if a test case with your issue can be added then and it's ok for me ;)

Ok, working on the test then.

trasher:
> > I guess a better solution would be to fix encoding coming from ldap, what is that kind of encoding?
> 
> For the encoding, it seems consistent with ldap_escape so it must be standard:
> 
> ```shell
> php > echo ldap_escape(',');
> \2c
> ```

Good to know, thank you :)

PR #7269, opened 2020-05-14
Master

Vinyxp1:
Good Morning,
It would be nice to have a data dictionary available directly in the application.
It would also be interesting to have a simpler way to connect the glpi to the powerbi or similar to make comparisons between calls and working time to buy if slas are being long.


trasher:
Please use [forums](http://forum.glpi-project.org) to get support or [suggestion website](http://suggest.glpi-project.org) for RFE.

PR #7270, opened 2020-05-14
No reply configuration for email notifications

AdrienClairembault:
Internal ref: 17298.

* Add a new configuration option for notification to indicate whether or not they should allow a response ('allow_response' param).
* If a notification doesn't expect a response, the `"To answer by email, write above this line"` headers and footers won't be printed in the email
* Add a new general configuration for notifications (no reply address)
* If a notification doesn't expect a response AND a no reply address is configured, use it as the sender address for the notification.

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

### DIFF
--- a/install/update_94_95.php
+++ b/install/update_94_95.php
@@ -1826,12 +1826,10 @@ HTML
       ]);
    }
 
-   if (!isset(Config::getConfigurationValues('core')['admin_email_noreply'])) {
-      $migration->addConfig([
-         'admin_email_noreply'      => "",
-         'admin_email_noreply_name' => "",
-      ]);
-   }
+   $migration->addConfig([
+      'admin_email_noreply'      => "",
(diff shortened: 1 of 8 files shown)


install/update_94_95.php
```diff
@@ -1819,6 +1819,21 @@ function update94to95() {
...
    }
 
+   // No-reply notifications
+   if (!$DB->fieldExists(Notification::getTable(), 'allow_response')) {
+      $migration->addField(Notification::getTable(), 'allow_response', 'bool', [
+         'value' => 1
+      ]);
+   }
+
+   if (!isset(Config::getConfigurationValues('core')['admin_email_noreply'])) {
```
trasher, line 1829: I think you do not have to check if values exist on your side

PR #7280, opened 2020-05-18
Fix display form for multiple classes

tsmr:
domain
group
user
operating system

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | none

### DIFF
--- a/inc/group.class.php
+++ b/inc/group.class.php
@@ -275,6 +275,7 @@ class Group extends CommonTreeDropdown {
 
       echo "<tr class='tab_bg_1'>";
       echo "<td class='subheader' colspan='2'>".__('Can contain');
+      echo "</td><td colspan='2' class='center'>";
       echo "</td></tr>";
 
       echo "<tr class='tab_bg_1'>";
--- a/inc/group.class.php
+++ b/inc/group.class.php
@@ -218,8 +218,8 @@ class Group extends CommonTreeDropdown {
       echo "<td>";
       Html::autocompletionTextField($this, "name");
       echo "</td>";
-      echo "<td rowspan='10' class='middle'>".__('Comments')."</td>";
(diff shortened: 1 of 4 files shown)


inc/group.class.php
```diff
@@ -275,6 +275,7 @@ function showForm($ID, $options = []) {
 
       echo "<tr class='tab_bg_1'>";
       echo "<td class='subheader' colspan='2'>".__('Can contain');
+      echo "</td><td colspan='2' class='center'>";
```
cedric-anne, line 278: Previous row seems to contains only 2 `td`, so existing one using `colspan='2'` was already using the 2 columns.

inc/user.class.php
```diff
@@ -2160,22 +2160,23 @@ function () {
          echo "<td><label for='password'>" . __('Password')."</label></td>";
          echo "<td><input id='password' type='password' name='password' value='' size='20'
                     autocomplete='new-password' onkeyup=\"return passwordCheck();\"></td>";
-         echo "<td rowspan='2'>";
          if ($CFG_GLPI["use_password_security"]) {
```
cedric-anne, line 2163: This is still not correct on my instance:
image

inc/group.class.php
tsmr: Previous rows are linked to rowspan='10' of comment field so we can update to rowspan='13' ?

trasher: I guess all empty columns should be dropped, and rowspan adapted accordingly. By the way, it seems there are 12 rows total, not 13; and this one is the 10th; so @cedric-anne was right

tsmr: I do it

Issue #7093, opened 2020-03-27
[GLPI 9.5] Impact Analysis - Add link to KnowbaseItem


tsmr:
Hello
We want to add link to Knobase Item into Impact Graph

Goal :
![image](https://user-images.githubusercontent.com/13178158/77754109-f5067580-702a-11ea-9a7d-fb41f145fcae.png)

Can you modify impact.class.php like this ? for allow sql resquest to Knowbase tables ?

```
diff --git a/inc/impact.class.php b/inc/impact.class.php
index 1ea2fcbffb..4587d13fc8 100644
--- a/inc/impact.class.php
+++ b/inc/impact.class.php
@@ -856,8 +856,95 @@ class Impact extends CommonGLPI {
 
       // Search for items
       $filter = strtolower($filter);
+      $join = [];
+      if ($itemtype == 'KnowbaseItem') {
+         $is_public_faq_context = !Session::getLoginUserID() && $CFG_GLPI["use_public_faq"];
+         $has_session_groups = isset($_SESSION["glpigroups"]) && count($_SESSION["glpigroups"]);
+         $has_active_profile = isset($_SESSION["glpiactiveprofile"])
+                               && isset($_SESSION["glpiactiveprofile"]['id']);
+         $has_active_entity = isset($_SESSION["glpiactiveentities"])
+                              && count($_SESSION["glpiactiveentities"]);
+         $forceall = true;
+         $join = [
+            'glpi_knowbaseitems_users' => [
+               'ON' => [
+                  'glpi_knowbaseitems_users' => 'knowbaseitems_id',
+                  'glpi_knowbaseitems'       => 'id'
+               ]
+            ]
+         ];
+         if ($forceall || $has_session_groups) {
+            $join['glpi_groups_knowbaseitems'] = [
+               'ON' => [
+                  'glpi_groups_knowbaseitems' => 'knowbaseitems_id',
+                  'glpi_knowbaseitems'       => 'id'
+               ]
+            ];
+         }
+         if ($forceall || $has_active_profile) {
+            $join['glpi_knowbaseitems_profiles'] = [
+               'ON' => [
+                  'glpi_knowbaseitems_profiles' => 'knowbaseitems_id',
+                  'glpi_knowbaseitems'       => 'id'
+               ]
+            ];
+         }
+         if ($forceall || $has_active_entity || $is_public_faq_context) {
+            $join['glpi_entities_knowbaseitems'] = [
+               'ON' => [
[...]

PR #7295, opened 2020-05-20
Fix KnowBase In ImpactAnalysis

tsmr:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #7093

### DIFF
--- a/inc/impact.class.php
+++ b/inc/impact.class.php
@@ -885,7 +885,10 @@ class Impact extends CommonGLPI {
             $base_request,
             $itemtype::getVisibilityCriteria()
          );
-      } else {
+      }
+
+      $item = new $itemtype();
+      if ($item->isEntityAssign()) {
          $base_request = array_merge_recursive(
             $base_request,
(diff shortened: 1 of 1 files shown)


inc/impact.class.php
```diff
@@ -877,14 +877,19 @@ public static function searchAsset(
...
-         ] + getEntitiesRestrictCriteria($itemtype::getTable()),
+         ],
       ];
 
       if (is_subclass_of($itemtype, "ExtraVisibilityCriteria", true)) {
          $base_request = array_merge_recursive(
             $base_request,
             $itemtype::getVisibilityCriteria()
          );
+      } else {
```
AdrienClairembault, line 888: We should indeed check if the itemtype has an entities_id field before adding the result of `getEntitiesRestrictCriteria`.

```suggestion
      }

      $item = new $itemtype();
      if ($item->isEntityAssign()) {
```

PR #7312, opened 2020-05-26
Bulk add items from Kanban

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

Also in commit:
Fix minor height different with columns without a header color (No Status column)

Peek 2020-05-25 20-10

### DIFF
--- a/ajax/kanban.php
+++ b/ajax/kanban.php
@@ -52,7 +52,7 @@ if (!isset($_REQUEST['action'])) {
 }
 $action = $_REQUEST['action'];
 
-$nonkanban_actions = ['update', 'add_item', 'move_item'];
+$nonkanban_actions = ['update', 'bulk_add_item', 'add_item', 'move_item'];
 if (isset($_REQUEST['itemtype'])) {
    $traits = class_uses($_REQUEST['itemtype'], true);
    if (!in_array($_REQUEST['action'], $nonkanban_actions) && (!$traits || !in_array('Kanban', $traits))) {
@@ -84,7 +84,7 @@ if (isset($itemtype)) {
          return;
       }
    }
-   if (in_array($action, ['add_item'])) {
+   if (in_array($action, ['bulk_add_item', 'add_item'])) {
       if (!$item->canCreate()) {
          // Missing rights
          http_response_code(403);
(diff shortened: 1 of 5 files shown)


orthagh:
some UX concerns:

- I missed, at first viewing of your gif, the placeholder attribute showing the "One item per line", i think it should be above/under the textarea, maybe in italic or/and with with a small font size. 
- Number of icons in header start to be high, i think it's the time to implement a "more" menu. Ex in github kanban:
image
  Delete and bulk actions could be moved to this burger context menu.
  More, this would allow to add full text labels beside icons and have a better understanding at first glance.
  And, it will ease addition of future actions.

Regards.

orthagh:
Thanks for the follow.

Imo, we can keep the icons left of the full text (sorry it was not clear in my previous message and the github screenshot didn't help).

Another thing i found while testing is related to "view all" right for projects? If it's unchecked, bulk (or not) addition of project(s) effectilvely create project(s) but don't display them anywhere, not in the kanban, nor the global search list. We have no error as it successfuly create the sub items.
I think to address this case, we could auto-fill supervisor with the current user id from session, no ?

ajax/kanban.php
```diff
@@ -118,6 +118,19 @@
...
+   $checkParams(['inputs']);
+   $item = new $itemtype();
+   $inputs = [];
+   parse_str($_REQUEST['inputs'], $inputs);
+
+   $bulk_item_list = preg_split('/\r\n|[\r\n]/', $inputs['bulk_item_list']);
+   if (!empty($bulk_item_list)) {
+      unset($inputs['bulk_item_list']);
+      foreach ($bulk_item_list as $item_entry) {
+         $item->add($inputs + ['name' => $item_entry]);
```
cedric-anne, line 131: You should add a test to prtevent created items from empty lines
```suggestion
         $item_entry = trim($item_entry);
         if (!empty($item_entry)) {
            $item->add($inputs + ['name' => $item_entry]);
         }
```

PR #7360, opened 2020-06-03
Fix 'notification_to_myself' setting for mailgate

AdrienClairembault:
Internal ref: 19947.

The setting 'notification_to_myself' allow a user to enable/disable notifications triggered by his own events.

However, this setting is ignored when actions are handled by the mailgate since the notification system need a connected user to check this setting.

This was fixed by allowing to specify an additional "emitter" parameter to the notification so it can use it instead of the current connected user (= use the sender of the email for mailgate).

This fix should be good enough for now, but if there is a notification remodel in the future it should maybe use a concrete notion of emitter/source and never rely on the current user.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

### DIFF
--- a/inc/notificationevent.class.php
+++ b/inc/notificationevent.class.php
@@ -137,9 +137,23 @@ class NotificationEvent extends CommonDBTM {
             $template->resetComputedTemplates();
 
             $notify_me = false;
+            $emitter = null;
+
             if (Session::isCron()) {
                // Cron notify me
                $notify_me = true;
+
+               // If mailcollector_user is set, use the given user preferences
+               if (isset($_SESSION['mailcollector_user'])) {
+                  $user = new User();
+                  $res = $user->getFromDB($_SESSION['mailcollector_user']);
+
+                  if ($res) {
+                     $user->computePreferences();
+                     $notify_me = $user->fields['notification_to_myself'];
+                     $emitter = $_SESSION['mailcollector_user'];
+                  }
+               }
             } else {
                // Not cron see my pref
                $notify_me = $_SESSION['glpinotification_to_myself'];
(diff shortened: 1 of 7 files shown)


inc/notificationtarget.class.php
```diff
@@ -167,15 +167,24 @@ function getFromDBForTarget($notifications_id, $type, $ID) {
...
     * @return boolean
    **/
-   function validateSendTo($event, array $infos, $notify_me = false) {
+   function validateSendTo($event, array $infos, $notify_me = false, $emitter = null) {
+      $users_id = Session::getLoginUserID(false);
+
+      // Override session ID with emitter ID if supplied
+      if (!is_null($emitter)) {
+         $users_id = $emitter;
+      }
```
trasher, line 182:
```suggestion
      // Override session ID with emitter ID if supplied
      $users_id = $emitter ?? Session::getLoginUserID(false);
```

inc/notificationevent.class.php
```diff
@@ -160,7 +174,8 @@ static function raiseEvent($event, $item, $options = [], $label = '') {
                   $data,
                   $notificationtarget->setEvent($eventclass),
                   $template,
-                  $notify_me
+                  $notify_me,
+                  $emitter
```
trasher, line 178: As far as I understand, `$emitter` will always be null here, so it is never used

inc/notificationtarget.class.php
AdrienClairembault: Still can't do that on 9.4 PR, need PHP 7 :(

trasher: Oh, you're right, sorry

inc/notificationevent.class.php
AdrienClairembault: I see what you mean, it's the github diff which stop at a weird place and make it looks like `$emitter` is in the `else` block but that's not the case:
image

trasher: Oh, OK, indeed, you're right; forgive me :)

PR #7370, opened 2020-06-04
Fix 'notification_to_myself' setting for mailgate 2

AdrienClairembault:
Following #7360.

One case was not covered by the previous changes: anonymous users.
Since they have no real account we must use:
- the global configuration settings for $notify_me
- theirs emais instead of theirs IDs when comparing with a potential email target

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

### DIFF
--- a/inc/notificationtarget.class.php
+++ b/inc/notificationtarget.class.php
@@ -167,9 +167,11 @@ class NotificationTarget extends CommonDBChild {
     * @param boolean $notify_me notify me on my action ?
     *                           ($infos contains users_id to check if the target is me)
     *                           (false by default)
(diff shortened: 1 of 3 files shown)


inc/notificationtarget.class.php
```diff
@@ -178,7 +180,19 @@ function validateSendTo($event, array $infos, $notify_me = false, $emitter = nul
 
       // Override session ID with emitter ID if supplied
       if (!is_null($emitter)) {
```
trasher, line 182: Maybe just `if (is_int) {} else if (!is_null)`?

PR #7374, opened 2020-06-05
Anonymous helpdesk: improve entity check

AdrienClairembault:
Following #7015.
Internal ref: 20024.

A little improvement of the "Anonymous helpdesk" feature.
Instead of checking the current entity of the user, we now check the entity of the ticket.

Example: 
We have two entities A and B (B being a sub-entity of A).
A -> Anonymous helpdesk **disabled**
B -> Anonymous helpdesk **enabled**

Before the change, a user on the A entity will be able to see the names of all the tech on any tickets from A or B.

After the change, this same user will still see the technicians' names for tickets on A but not for tickets on B.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

### DIFF
--- a/inc/commonitilobject.class.php
+++ b/inc/commonitilobject.class.php
@@ -2881,7 +2881,8 @@ abstract class CommonITILObject extends CommonDBTM {
             $k = $d['groups_id'];
             echo "$mandatory$groupicon&nbsp;";
             if ($group->getFromDB($k)) {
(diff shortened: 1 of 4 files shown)


trasher:
On 9.5 only?

AdrienClairembault:
> On 9.5 only?

This feature was added in 9.5 (original PR #7015).

PR #7383, opened 2020-06-09
Fix Kanban compat with Feature/traits #7296

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

At least the AJAX code file needs tests made.

### DIFF
--- a/ajax/kanban.php
+++ b/ajax/kanban.php
@@ -30,6 +30,8 @@
  * ---------------------------------------------------------------------
  */
 
+use Glpi\Features\Kanban;
+
 $AJAX_INCLUDE = 1;
 
 include ('../inc/includes.php');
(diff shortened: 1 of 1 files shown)


ajax/kanban.php
```diff
@@ -30,6 +30,8 @@
  * ---------------------------------------------------------------------
  */
 
+use Glpi\Features\Kanban;
```
trasher, line 33:
```suggestion
```
This is not required, you must work with class name and not trait one anyways.

trasher: OK, nevermind, I've just understood the fix

ajax/kanban.php
```diff
@@ -55,7 +57,7 @@
 $nonkanban_actions = ['update', 'bulk_add_item', 'add_item', 'move_item'];
 if (isset($_REQUEST['itemtype'])) {
    $traits = class_uses($_REQUEST['itemtype'], true);
-   if (!in_array($_REQUEST['action'], $nonkanban_actions) && (!$traits || !in_array('Kanban', $traits))) {
+   if (!in_array($_REQUEST['action'], $nonkanban_actions) && (!$traits || !in_array(Kanban::class, $traits, true))) {
```
trasher, line 60: I'm not sure the 3rd parameter is needed

trasher:
Hi @cconard96 , I do not know why this was still a draft, but since this really fix the issue, I merge so it can be included in the rc to come.
If there were other things to fix or enhance, it can be done on another pr ;)

Issue #7395, opened 2020-06-10
[9.5] Error if linked object to ticket is a plugin (& plugin is not present)


tsmr:
Update 9.5 from 9.4.5
No plugins activated

If i go in a existent ticket linked with a plugin object & because plugins is not activated i have : 

```
[2020-06-10 16:26:35] glpiphplog.CRITICAL:   *** Uncaught Exception Error: Class 'PluginResourcesResource' not found in /var/www/html/glpi/inc/ticket.class.php at line 7350
  Backtrace :
  inc/ticket.class.php:727                           Ticket->getLinkedItems()
  inc/commonitilobject.class.php:7719                Ticket->getTabNameForItem()
  inc/ticket.class.php:845                           CommonITILObject->defineDefaultObjectTabs()
  inc/commonglpi.class.php:294                       Ticket->defineTabs()
  inc/commonglpi.class.php:859                       CommonGLPI->defineAllTabs()
  inc/commonglpi.class.php:1230                      CommonGLPI->showTabsContent()
  front/ticket.form.php:221                          CommonGLPI->display()
```

Resolution

```
diff --git a/inc/ticket.class.php b/inc/ticket.class.php
index 75ff96a91c..cca2305f68 100644
--- a/inc/ticket.class.php
+++ b/inc/ticket.class.php
@@ -7347,6 +7347,9 @@ class Ticket extends CommonITILObject {
       if ($addNames) {
          foreach ($assets as $key => $asset) {
             /** @var CommonDBTM $item */
+            if (!(getItemForItemtype($asset['itemtype']))) {
+               continue;
+            }
             $item = new $asset['itemtype'];
             $item->getFromDB($asset['items_id']);
```

trasher:
The fix is correct, but similar problem may happen ni changes and problems (and code should be factorized probably)

PR #7403, opened 2020-06-11
Do not try to create missing objects

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #7395

### DIFF
--- a/inc/commonitilobject.class.php
+++ b/inc/commonitilobject.class.php
@@ -8139,4 +8139,63 @@ abstract class CommonITILObject extends CommonDBTM {
 
       return $status == CommonITILObject::INCOMING;
    }
+
+   /**
+    * Retrieve linked items table name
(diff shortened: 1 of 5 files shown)


inc/commonitilobject.class.php
```diff
@@ -8139,4 +8139,63 @@ protected function isNew() {
...
+
+   /**
+    * Retrieve linked items table name
+    *
+    * @since 9.5.0
+    *
+    * @return string
+    */
+   public static function getItemsTable() {
+      switch (static::getType()) {
```
AdrienClairembault, line 8151: Maybe you could use polymorphism here instead of a switch on itemtype ?

Issue #7405, opened 2020-06-11
[9.5] Error with RuleImportComputer->findWithGlobalCriteria


tsmr:
On OCSNG plugin we use hook _getSqlRestriction (for see computers to link / import)

```
[2020-06-11 11:20:01] glpiphplog.ERROR:   *** PHP User Error (256): Missing table name in C:\wamp64\www\glpi\inc\dbmysqliterator.class.php at line 277
  Backtrace :
  inc\dbmysqliterator.class.php:277                  trigger_error()
  inc\dbmysqliterator.class.php:94                   DBmysqlIterator->buildQuery()
  inc\dbmysql.class.php:849                          DBmysqlIterator->execute()
  inc\ruleimportcomputer.class.php:373               DBmysql->request()
  inc\rule.class.php:1577                            RuleImportComputer->findWithGlobalCriteria()
  inc\rule.class.php:1444                            Rule->checkCriterias()
  inc\rulecollection.class.php:1500                  Rule->process()
  ...ins\ocsinventoryng\inc\ocsserver.class.php:2692 RuleCollection->processAllRules()
  plugins\ocsinventoryng\front\ocsng.link.php:124    PluginOcsinventoryngOcsServer::showComputersToAdd()
```

PR #7406, opened 2020-06-11
Rule import computer fix

tsmr:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #7405

### DIFF
--- a/inc/ruleimportcomputer.class.php
+++ b/inc/ruleimportcomputer.class.php
@@ -365,8 +365,8 @@ class RuleImportComputer extends Rule {
                $sql_results = Plugin::doOneHook($plugin, "ruleImportComputer_getSqlRestriction",
                                                 $params);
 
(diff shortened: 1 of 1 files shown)


inc/ruleimportcomputer.class.php
```diff
@@ -304,9 +304,7 @@ function findWithGlobalCriteria($input) {
 
       $it_criteria = [
          'SELECT' => 'glpi_computers.id',
-         'FROM'   => '', //to fill
          'WHERE'  => [], //to fill
-         'ORDER'  => 'glpi_computers.is_deleted ASC'
```
trasher, line 309: Is that expected? As far as I know, OCS is the only one to use this piece of code currently.

tsmr: I was just wondering what this ORDER was for

PR #7408, opened 2020-06-11
Add ImpactAnalysis Tab For Change & Problem

tsmr:
Add Impact Analysis Tab For Change & Problem

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

PR #7409, opened 2020-06-12
Factorize checks and add impact tab on Change and Problem

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

Quite same as #7408, with factorization.

### DIFF
--- a/inc/commonitilobject.class.php
+++ b/inc/commonitilobject.class.php
@@ -8203,4 +8203,19 @@ abstract class CommonITILObject extends CommonDBTM {
 
       return $assets;
    }
+
+   /**
+    * Should impact tab be displayed? Check if there is a valid linked item
+    *
+    * @return boolean
+    */
+   protected function showImpactTab() {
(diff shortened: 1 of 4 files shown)


inc/commonitilobject.class.php
```diff
@@ -8203,4 +8203,19 @@ public function getLinkedItems(bool $addNames = true) :array {
 
       return $assets;
    }
+
+   /**
+    * Should impact tab be displayed? Check if there is a valid linked item
+    *
+    * @return boolean
+    */
+   protected function showImpactTab() {
```
orthagh, line 8212: hasImpactTab ?

orthagh: or maybe directly addImpactTab which do the checks and call directly addStandardTab

trasher: Since this is really specific for ITIL, I guess commonitil is not wrong

Issue #7196, opened 2020-04-29
[9.5] Setup menu always displayed even have no rights


tsmr:
1. Create a profile with only rights into tickets (assistance)

2. Log on GLPI with this profile 

3. Result: the Setup menu is displayed without any submenu

![image](https://user-images.githubusercontent.com/13178158/80581052-de2db700-8a0c-11ea-944d-8c8821f8eff9.png)

tsmr:
It cames from plugin.class.php

```
static function getAdditionalMenuOptions() {
      return [
         'marketplace' => [
            'icon'  => MarketplaceView::geticon(),
            'title' => MarketplaceView::getTypeName(),
            'page'  => MarketplaceView::getSearchURL(false),
         ]
      ];
   }
```

radiaips:
Bonjour, 
J'ai le meme problème avec le menu Management!
   Auriez vous une idée de la class concernée  sur la version GLPI 9.3.3  Svp?

trasher:
@radiaips use english only, non english comments will be removed.

radiaips:
Je vous remercie, mais j'ai besoin de savoir la classe concernée si c'est possible !

stdglpi:
@tsmr I confirm, I have same problem on the GLPI 9.4.6 version.

trasher:
Well, this code has been added in 9.5; that is not possible you have the same problem on 9.4.6. If you found a similar issue, please open another bug report with all requested details.

stdglpi:
@trasher Thank you for the job !

tsmr:
Php-error :

[2020-06-23 15:41:07] glpiphplog.WARNING:   *** PHP Warning (2): count(): Parameter must be an array or an object that implements Countable in C:\wamp64\www\glpi\inc\plugin.class.php at line 101
  Backtrace :
  inc\html.class.php:1577                            Plugin::getMenuContent()
  inc\html.class.php:6871                            Html::generateMenuSession()
  inc\html.class.php:1696                            Html::displayMainMenu()
  front\central.php:81                               Html::header()

![image](https://user-images.githubusercontent.com/13178158/85425006-ef142800-b578-11ea-8a35-aa963597bd70.png)

To resolve it : 

```
diff --git a/inc/plugin.class.php b/inc/plugin.class.php
index 48b3e061f8..13b1de0e1e 100644
--- a/inc/plugin.class.php
+++ b/inc/plugin.class.php
[...]

PR #7410, opened 2020-06-12
Fix 7196

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #7196

Need logout/login after patch has been applied.

### DIFF
--- a/inc/plugin.class.php
+++ b/inc/plugin.class.php
@@ -123,13 +123,15 @@ class Plugin extends CommonDBTM {
 
 
    static function getAdditionalMenuOptions() {
-      return [
-         'marketplace' => [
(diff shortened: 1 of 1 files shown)


inc/plugin.class.php
```diff
@@ -106,6 +106,9 @@ static function getMenuContent() {
 
 
    static function getAdditionalMenuLinks() {
+      if (!static::canView()) {
```
trasher, line 109: Not sure this one may cause a problem; sounds logic to me having this check there

PR #7611, opened 2020-07-07
Add missing 'is_recursive' search option to assets

AdrienClairembault:
Internal ref: 20271.

Add missing 'is_recursive' search option to most assets (it was only set in printer, software and network equipment).
This search options is very useful for massive actions.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes


trasher:
Since this is provided in CommonDBTM::rawSearchOptions, does not some of those changes just miss a call to parent method?

Issue #7612, opened 2020-07-07
Map view get "Undefined variable" error


tizianosartori:
When click on Map view of object, GLPI get this error on all assets pages:

`PHP Notice (8): Undefined variable: data in /var/www/glpi/inc/search.class.php at line 134`

`GLPI 9.5.0 ( => /var/www/glpi) Installation mode: TARBALL`

PR #7616, opened 2020-07-07
Fix map search view

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #7612 

Verified no regression on previous Global search issue.

### DIFF
--- a/inc/search.class.php
+++ b/inc/search.class.php
@@ -129,7 +129,9 @@ class Search {
          'searchtype'   => 'contains',
          'value'        => 'NULL'
       ];
-      self::displayData(self::getDatas($itemtype, $params));
+
+      $data = self::getDatas($itemtype, $params);
+      self::displayData($data);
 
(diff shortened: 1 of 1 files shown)


inc/search.class.php
```diff
@@ -129,7 +129,9 @@ static function showMap($itemtype, $params) {
          'searchtype'   => 'contains',
          'value'        => 'NULL'
       ];
-      self::displayData(self::getDatas($itemtype, $params));
+
```
trasher, line 132:
```suggestion
    
```

inc/search.class.php
```diff
@@ -129,7 +129,9 @@ static function showMap($itemtype, $params) {
...
```
trasher, line 132:
```suggestion

```

PR #7658, opened 2020-07-11
Search historical data

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

Useful when you know an old value for an item such as its name, but don't know its new name.

https://glpi.userecho.com/communities/1/topics/1062-search-into-historical-of-all-computer

Screenshot from 2020-07-11 16-31-53

### DIFF
--- a/front/log.php
+++ b/front/log.php
@@ -1,0 +1,41 @@
+<?php
+/**
+ * ---------------------------------------------------------------------
+ * GLPI - Gestionnaire Libre de Parc Informatique
+ * Copyright (C) 2015-2020 Teclib' and contributors.
+ *
+ * http://glpi-project.org
+ *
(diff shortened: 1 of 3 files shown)


trasher:
Working with logs table sounds a very bad idea That was used in the past in tickets, and it always lend to performances issues.

cconard96:
Right now performance is about the same as the Computer search for me and uses less memory. Even if searching historical data from a central place is a bit slower, is it not worth it if users have the need to search it?

Also, is it seems weird we can only audit a small fraction of the actions taken in GLPI (Events).

PR #7701, opened 2020-07-16
RuleTicket: add tasks from template

AdrienClairembault:
Internal ref: 19836.

Add an new action for ticket rules that create a task from a given template:
image

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

### DIFF
--- a/inc/ruleticket.class.php
+++ b/inc/ruleticket.class.php
@@ -778,10 +778,10 @@ class RuleTicket extends Rule {
       $actions['solution_template']['table']                 = 'glpi_solutiontemplates';
       $actions['solution_template']['force_actions']         = ['assign'];
 
-      $actions['task_template']['name']          = _n('Task template', 'Task templates', 1);
-      $actions['task_template']['type']          = 'dropdown';
-      $actions['task_template']['table']         = TaskTemplate::getTable();
-      $actions['task_template']['force_actions'] = ['assign'];
+      $actions['task_template']['name']                      = _n('Task template', 'Task templates', 1);
+      $actions['task_template']['type']                      = 'dropdown';
+      $actions['task_template']['table']                     = TaskTemplate::getTable();
+      $actions['task_template']['force_actions']             = ['assign'];
 
       return $actions;
    }
--- a/inc/ruleticket.class.php
+++ b/inc/ruleticket.class.php
@@ -140,7 +140,6 @@ class RuleTicket extends Rule {
 
 
    function executeActions($output, $params, array $input = []) {
-
       if (count($this->actions)) {
          foreach ($this->actions as $action) {
             switch ($action->fields["action_type"]) {
@@ -235,6 +234,40 @@ class RuleTicket extends Rule {
                            "content" => Toolbox::addslashes_deep($template->getField('content')),
                            "status" => CommonITILValidation::WAITING,
                            "items_id" => $output['id']]);
+                     }
+                  }
+
+                  if ($action->fields["field"] == 'task_template') {
+                     // Init special input '_tickettasks' if empty
+                     if (!isset($output["_tickettasks"])) {
+                        $output["_tickettasks"] = [];
+                     }
+
+                     $template = new TaskTemplate();
+                     if ($template->getFromDB($action->fields["value"])) {
+                        $values = [
+                           "tasktemplates_id"  => $template->getId(),
+                           "content"           => Toolbox::addslashes_deep($template->getField('content')),
+                           "taskcategories_id" => $template->getField('taskcategories_id'),
+                           "users_id_tech"     => $template->getField('users_id_tech'),
+                           "groups_id_tech"    => $template->getField('groups_id_tech'),
+                           "is_private"        => $template->getField('is_private'),
+                           "actiontime"        => $template->getField('actiontime'),
+                           "name"              => $template->getField('name'),
+                        ];
+
+                        if (!$output['id']) {
+                           // Target ticket is not yet created, data need to be
+                           // moved to special input '_tickettasks' so it can
+                           // be handled after the ticket creation
+                           $output["_tickettasks"][] = $values;
+                        } else {
+                           // Target ticket already exist, add the task
+                           $values['tickets_id'] = $output['id'];
+                           $task = new TicketTask();
+                           $task->add($values);
+                        }
+
                      }
                   }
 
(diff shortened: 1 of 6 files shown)


AdrienClairembault:
> I may be wrong, but there is already a mechanism in `CommonItilObject::post_addItem` that checks for `_tasktemplates_id` in input to create corresponding task. Maybe this piece of code should be reused.

I was not aware of that, I've switched to this param and everything seems to be ok.

inc/ruleticket.class.php
```diff
@@ -238,6 +238,34 @@ function executeActions($output, $params, array $input = []) {
...
+                  if ($action->fields["field"] == 'task_template') {
+                     $template = new TaskTemplate();
+                     if ($template->getFromDB($action->fields["value"])) {
+                        $ticket_id = $output['id'] ?? 0;
+                        if (!$ticket_id) {
+                           // Target ticket is not yet created, data need to be
+                           // moved to special input '_tasktemplates_id' so it
+                           // can be handled after the ticket creation
+                           $output["_tasktemplates_id"][] = $template->getId();
+                        } else {
```
orthagh, line 250: In the same spirit as @cedric-anne remark, we should have only `$output["_tasktemplates_id"][] = $template->getId();` also for update.

The code here is the same as `CommonItilObject::post_addItem`.

So, i think the solution is to add a new method in `CommonItilObject` to manage creation of task from `_tasktemplates_id` input and call it in both functions `post_xxxItem`

You will be also able to do real **unit** test (currently you do a functional one based on rules)

AdrienClairembault: > In the same spirit as @cedric-anne remark, we should have only `$output["_tasktemplates_id"][] = $template->getId();` also for update.
> 
> The code here is the same as `CommonItilObject::post_addItem`.
> 
> So, i think the solution is to add a new method in `CommonItilObject` to manage creation of task from `_tasktemplates_id` input and call it in both functions `post_xxxItem`
> 
> You will be also able to do real **unit** test (currently you do a functional one based on rules)

Done for the code change, I agree.

For the unit test, at this moment it does not seem possible for me to have a correct one as this method use multiple "repository" objects and we lack dependency injection to mock them properly.

inc/ruleticket.class.php
```diff
@@ -768,6 +778,11 @@ function getActions() {
       $actions['solution_template']['table']                 = 'glpi_solutiontemplates';
       $actions['solution_template']['force_actions']         = ['assign'];
 
+      $actions['task_template']['name'] = _n('Task template', 'Task templates', 1);
```
orthagh, line 781: could you align `=` here ?

tests/functionnal/RuleTicket.php
```diff
@@ -561,4 +565,95 @@ public function testGRoupRequesterAssignFromDefaultUser() {
...
+      $ticket_tasks = $ticket_task_em->find([
+         'tickets_id' => $ticket_id
+      ]);
+
+      $this->array($ticket_tasks)->hasSize(1);
+      $task_data = array_pop($ticket_tasks);
+      $this->array($task_data)->hasKey('content');
+      $this->string($task_data['content'])->isEqualTo('test content');
+   }
 }
```
orthagh, line 659:
```suggestion
}

```

AdrienClairembault: Done

tests/functionnal/RuleTicket.php
```diff
@@ -561,4 +565,96 @@ public function testGRoupRequesterAssignFromDefaultUser() {
...
+      $rule_ticket_em = new \RuleTicket();
+      $rule_ticket_id = $rule_ticket_em->add($ruletinput = [
+         'name'         => "test to assign ITILSolution",
+         'match'        => 'OR',
+         'is_active'    => 1,
+         'sub_type'     => 'RuleTicket',
+         'condition'    => \RuleTicket::ONADD + \RuleTicket::ONUPDATE,
+         'is_recursive' => 1,
+      ]);
+      $this->integer($rule_ticket_id);
```
trasher, line 589:
```suggestion
      $this->integer($rule_ticket_id)->isGreaterThan(0);
```

Same for all add checks

tests/functionnal/RuleTicket.php
```diff
@@ -561,4 +565,96 @@ public function testGRoupRequesterAssignFromDefaultUser() {
...
+
+   public function testTaskTemplateAssignFromRule() {
+      $this->login();
+
+      // Create solution template
+      $task_template = new TaskTemplate();
+      $task_template_id = $task_template->add([
+         'content' => "test content"
+      ]);
+      $this->integer($task_template_id);
```
trasher, line 577:
```suggestion
      $this->integer($task_template_id)->isGreaterThan(0);
```

tests/functionnal/RuleTicket.php
```diff
@@ -561,4 +565,96 @@ public function testGRoupRequesterAssignFromDefaultUser() {
...
+      $ticket_tasks = $ticket_task_em->find([
+         'tickets_id' => $ticket_id
+      ]);
+
+      $this->array($ticket_tasks)->hasSize(1);
+      $task_data = array_pop($ticket_tasks);
+      $this->array($task_data)->hasKey('content');
+      $this->string($task_data['content'])->isEqualTo('test content');
+   }
+
```
trasher, line 659:
```suggestion
```

inc/ruleticket.class.php
```diff
@@ -768,6 +778,11 @@ function getActions() {
       $actions['solution_template']['table']                 = 'glpi_solutiontemplates';
       $actions['solution_template']['force_actions']         = ['assign'];
 
+      $actions['task_template']['name']          = _n('Task template', 'Task templates', 1);
+      $actions['task_template']['type']          = 'dropdown';
+      $actions['task_template']['table']         = TaskTemplate::getTable();
+      $actions['task_template']['force_actions'] = ['assign'];
```
orthagh, line 784: 1. Align `=`with above actions.

2. i think, it could be better to set `append` action instead `assign` (for understanding  purpose) you need to check if the process is really different. 

3. Could it be possible to assign multiple time a task template to the same rule ? To have the feature complete, we should do it if it's possible.

AdrienClairembault: 1. Done

2. Dev is already in production, if we change it it will break when the client update.
This dev was inspired by the `solution_template` rule action that also use `assign`.
IMO it would be better to have a separate "code improvement" PR that move then both to `append` and migrate the data if needed.

3. Does not seem possible with current rule system (you can only add each actions type once).
IMO need to be another PR if we want to allow multiple usage of rules.

orthagh (approved):

PR #7724, opened 2020-07-17
Some performance fixes

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

Change strtotime("now") to time().
Move some function calls out of for loop condition.
Merge a str_replace call when it doesn't affect readability of code.
Replace array_push with $array[] when pushing a single item.

### DIFF
--- a/inc/computer.class.php
+++ b/inc/computer.class.php
@@ -138,7 +138,8 @@ class Computer extends CommonDBTM {
       global $DB, $CFG_GLPI;
 
       $changes = [];
-      for ($i=0; $i<count($this->updates); $i++) {
+      $update_count = count($this->updates ?? []);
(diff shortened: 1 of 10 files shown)


cconard96:
It isn't something I was focusing on. I ran inspections on the inc folder in PHPStorm to try and get a list of PHPDoc issues and bugs between the IDE itself and one of the plugins I use for extended checks. As I was going through, I saw these recommended changes and felt they were relatively low-risk to change and pretty quick to implement. I'm more focused on bug fixing than these small performance improvements.

Issue #7730, opened 2020-07-18
[GLPI 9.5.1] html tags will show up in sub-projects preview on hover


joomlafun:
//PHP 7.3.16 ; Centos 8.1 stable
//Chrome & Firefox latest versions

Hi,

![image](https://user-images.githubusercontent.com/63743077/87855691-28db1300-c91a-11ea-8364-7f701fa08a68.png)

PR #7734, opened 2020-07-20
decode Html before display it in project task tootlip

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #7730


orthagh:
For informatio, i search a little in GLPI sources to try to find an existing example of HTML from database displayed in a such tooltip without success..

If you have a better way, let me know

PR #7760, opened 2020-07-23
Use friendly names in impact analysis; fix #7729

### LINKED ISSUE #7729: [9.5.1] Impact analysis - Add & Search users
When you want to add a user into the Analysis Impact graph, glpi shows login of users & not their fistname & Lastname

image

AdrienClairembault:
Use friendly names in graph view, list view and asset search.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  |no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #7729

### DIFF
--- a/inc/user.class.php
+++ b/inc/user.class.php
@@ -5558,6 +5558,47 @@ JAVASCRIPT;
       return $expiration_time < time();
    }
 
+   public static function getFriendlyNameSearchCriteria(string $filter): array {
+      $table     = self::getTable();
+      $login     = DBmysql::quoteName("$table.name");
+      $firstname = DBmysql::quoteName("$table.firstname");
+      $lastname  = DBmysql::quoteName("$table.realname");
+
+      $filter = strtolower($filter);
+      $filter_no_spaces = str_replace(" ", "", $filter);
+
+      return [
+         'OR' => [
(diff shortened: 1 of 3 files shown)


inc/user.class.php
```diff
@@ -5558,6 +5558,47 @@ public function hasPasswordExpired() {
...
+      $filter = strtolower($filter);
+      $filter_no_spaces = str_replace(" ", "", $filter);
+
+      return [
+         'OR' => [
+            ['RAW' => ["LOWER($login)" => ['LIKE', "%$filter%"]]],
+            ['RAW' => ["LOWER(CONCAT($firstname, $lastname))" => ['LIKE', "%$filter_no_spaces%"]]],
+            ['RAW' => ["LOWER(CONCAT($lastname, $firstname))" => ['LIKE', "%$filter_no_spaces%"]]],
+         ]
+      ];
```
cedric-anne, line 5576: It is common to have spaces in firstnames or lastnames, so you should not remove spaces here.

AdrienClairembault: I've removed spaces from the concat result instead.
The goal is to have a more forgiving search result so you can still match  "Adrien Clairembault" with "adrienclairembault".

After removing spaces from concat, it works fine for names with spaces:
image

PR #7815, opened 2020-08-01
Fix clone massive action targeting

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

### DIFF
--- a/inc/commondbrelation.class.php
+++ b/inc/commondbrelation.class.php
@@ -1852,4 +1852,10 @@ abstract class CommonDBRelation extends CommonDBConnexity {
 
       throw new \RuntimeException('Cannot guess ');
    }
+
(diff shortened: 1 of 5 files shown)


cconard96:
I found this has a BC break. Tickets, Changes, Problems, and possibly more are meant to be clonable but don't use the Clonable trait. I made this PR a draft for now.

cconard96:
Cloning is too fragmented between the Clonable trait and CommonDBTM for a clean, global solution. For now, I just forbid the clone action on the specific itemtype I noticed an issue with.
A good solution going forward would be moving all clone functions into the Clonable trait.

PR #7874, opened 2020-08-21
check if called url is the one setup for api

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | internal ref 20579


orthagh:
seems not sufficient for reporter, wait before merge

PR #7897, opened 2020-08-30
ITIL Object Kanban Views

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

Expand supported Kanban types to include ITIL Objects to offer a new way to view service requests, changes, and problems.

Done:
- [x] Basic global Kanban support including:
  - Drag and drop between columns to change status
  - Built-in state saving for card ordering
  - Show/hide columns based on the ITIL Object's statuses
- [x] Column header colors

To do:
- [x] Fix ability to add items from the Kanban.
- [x] Teams (badges shown on the cards). Use Assigned users, groups and suppliers.
- [x] Handle child tickets
- [ ] Handle planned tasks?
- [x] Fix card actions

~A quick warning if testing: There is currently no limit to the number of cards shown in the closed column so having a large number of closed items will cause performance issues.~
~Also, since this was initially done before the Modern UI work, it relies on jQuery dialogs as does the existing Kanban code to view the items in a Kanban. You won be able to click the title of the cards to open the items.~

Screenshot from 2020-08-30 18-49-25

### DIFF
--- a/ajax/kanban.php
+++ b/ajax/kanban.php
@@ -51,8 +51,7 @@ $action = $_REQUEST['action'];
 
 $nonkanban_actions = ['update', 'bulk_add_item', 'add_item', 'move_item', 'show_card_edit_form', 'delete_item'];
 if (isset($_REQUEST['itemtype'])) {
-   $traits = class_uses($_REQUEST['itemtype'], true);
-   if (!in_array($_REQUEST['action'], $nonkanban_actions) && (!$traits || !in_array(Kanban::class, $traits, true))) {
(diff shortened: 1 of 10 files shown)


orthagh:
@cconard96 Could you rework this pr after the recent merge.
I'll do a full review just after as you mention in the description the work is done.

orthagh:
Hello @cconard96 

I fixed some UI issues similar to ones in #9320.
I found some others i didn't succeed to fix:

- massive addition of ticket doesn't seem to be functional (in fact not projects/tasks)
- after remove of the "close status" column, re-adding it triggers a dbmysqliterator error.

Apart these, we're ready to merge i think

cconard96:
Both issues should now be fixed

PR #7992, opened 2020-09-23
Fix #7980 Bug : Notification template block post_addItem of ticket

### LINKED ISSUE #7980: [9.4 / 9.5] Important Bug : Notification template block post_addItem of ticket
I noticed this bug in a plugin because the generated ticket id was not returned & it's really annoying

https://github.com/glpi-project/glpi/blob/9.5/bugfixes/inc/notificationtargetcommonitilobject.class.php#L1283

into notificationtargetcommonitilobject.class.php there is a count :

$data["##$objettype.numberofunresolved##"]
               = countElementsInTableForEntity($item->getTable(), $this->getEntity(), $restrict);

This count use this fonction
https://github.com/glpi-project/glpi/blob/a50ae4bb29b4aa20ea37893bb59b06671a64618f/inc/dbutils.class.php#L402

But itil objects are not recursive : so the query hangs the rest of the ticket process

The solution is to add this parameter

```
$data["##$objettype.numberofunresolved##"]
               = countElementsInTableForEntity($item->getTable(), $this->getEntity(), $restrict, **false**);
```

tsmr:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #7980

### DIFF
--- a/inc/notificationtargetcommonitilobject.class.php
+++ b/inc/notificationtargetcommonitilobject.class.php
@@ -1281,7 +1281,7 @@ abstract class NotificationTargetCommonITILObject extends NotificationTarget {
          }
 
          $data["##$objettype.numberofunresolved##"]
-               = countElementsInTableForEntity($item->getTable(), $this->getEntity(), $restrict);
+               = countElementsInTableForEntity($item->getTable(), $this->getEntity(), $restrict, false);
 
          // Document
          $iterator = $DB->request([


trasher:
Some tests are already present for `NotificationTargetTicket`; I wonder if that case should not be added

trasher:
I took a look, but I've not been able to reproduce using unit tests, results are always the same with or without fix

trasher (approved):

cedric-anne (approved):

PR #8020, opened 2020-09-30
fix(plugin): sanitize after plugin override

stonebuzz:
since a while, GLPI sanitize $_GET from ```include.class.php```

But this change takes place before the plugins are loaded.

if plugin want to override some data into $_GET, it is not taken into account because plugin are called after.

This Pr prevent this

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #number

### DIFF
--- a/inc/includes.php
+++ b/inc/includes.php
@@ -72,10 +72,7 @@ if (isset($_POST)) {
    }
    $_POST = Toolbox::sanitize($_POST);
 }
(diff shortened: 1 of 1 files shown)


orthagh:
An easier solution found : see mattermost

Issue #8056, opened 2020-10-05
[GLPI 9.5.X] No Empty Choice for User Dropdown


tsmr:
In 9.5 : 
![image](https://user-images.githubusercontent.com/13178158/95052799-d676ea80-06ef-11eb-9af8-7887b9021c1a.png)

In 9.4 :
![image](https://user-images.githubusercontent.com/13178158/95052911-fd352100-06ef-11eb-954e-80f545feca87.png)

So there is no default value ?

PR #8057, opened 2020-10-05
Add Default Value For User Dropdown

tsmr:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #8056

### DIFF
--- a/inc/user.class.php
+++ b/inc/user.class.php
@@ -3926,25 +3926,26 @@ JAVASCRIPT;
 
       // Default values
       $p = [
-         'name'             => 'users_id',
-         'value'            => '',
-         'values'           => [],
-         'right'            => 'id',
-         'all'              => 0,
-         'on_change'        => '',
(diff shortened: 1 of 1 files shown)


cedric-anne (approved):

trasher:
I'm not sure changing default is a good solution, problem seems to be located in only one form; but this change will affect all dropdowns that may want a different default (and also from plugins).

tsmr:
> I'm not sure changing default is a good solution, problem seems to be located in only one form; but this change will affect all dropdowns that may want a different default (and also from plugins).

Precisely, I have a problem with a plugin using the User::Dropdown function, because it does not provide a default value.

This behavior changed in 9.5

The problem exist on actors too

9.5
image

9.4
image

orthagh (approved):

PR #8061, opened 2020-10-05
Use League/Csv

AdrienClairembault:
Issue ref : #7988.

Goals of these changes :
- Use `League/Csv` instead of `fputcsv` or other ways of manually building CSV data.
- One liner to output CSV in front files (`CsvResponse`) to avoid having the same boilerplate code in each front file.
- Simplify adding new CSV exports (`ExportToCsvInterface`).
- Simplify adding new CSV unit tests (`CsvTestCase`).

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #7988


trasher (commented):
I was wondering if `*Csv` classes should move to `Glpi\Csv` namespace (would require to handle the migration of existing `PlanningCsv`).

Another point; there is still a specifi CSV export in the `Stat` class; and in the `Search` class that should proably use the new lib as well.

trasher:
> Another point; there is still a specific CSV export in the `Stat` class; and in the `Search` class that should probably use the new lib as well.

Since Search may be a huge job; I guess it would be ok to do it in another PR. No idea for stats, this may require a bit of work.

AdrienClairembault:
> I was wondering if `*Csv` classes should move to `Glpi\Csv` namespace (would require to handle the migration of existing `PlanningCsv`).

I've though about it but I wasn't really sure.
Assuming everything get namespaces in the future, we could also have for example a `Glpi\Planning` namespace containing all planning related stuff, including the csv export.

What's the best way to do this ? Put all Csv export in the Csv name space or put them in their "associated feature" namespace ?

> Another point; there is still a specific CSV export in the `Stat` class; and in the `Search` class that should probably use the new lib as well.

I'll look for `Stat`, guess I missed it.
I've seen the Csv in search but it seemed hard to convert so I left it out for now... maye it's better to do it in another PR like you said.

trasher:
> > I was wondering if `*Csv` classes should move to `Glpi\Csv` namespace (would require to handle the migration of existing `PlanningCsv`).
> 
> I've though about it but I wasn't really sure.
> Assuming everything get namespaces in the future, we could also have for example a `Glpi\Planning` namespace containing all planning related stuff, including the csv export.
> 
> What's the best way to do this ? Put all Csv export in the Csv namespace or put them in their "associated feature" namespace ?

I do not think there is a "best" way, on another project, I tend to group all my *pdf on the same namespace, and all my *events in another; but it should indeed be done differently.
But for now, either we place them in the Csv namespace, either we keep default one for now; it's worth to create several empty new namespaces for just add one class into it.

> 
> > Another point; there is still a specific CSV export in the `Stat` class; and in the `Search` class that should probably use the new lib as well.
> 
> I'll look for `Stat`, guess I missed it.
> I've seen the Csv in search but it seemed hard to convert so I left it out for now... maye it's better to do it in another PR like you said.

OK for Search, let's see that later ;)

trasher:
By the way, I've tested CSV exports of impacts, planning and logs; I did not see any issues.

inc/csv/logcsvexport.class.php
```diff
@@ -0,0 +1,91 @@
...
+   }
+
+   public function getFileName(): string {
+      $name = $this->item->getFriendlyName();
+      $date = date('Y_m_d', time());
+
+      // Avoid whitespace in name
+      $name = str_replace(' ', '_', $name);
+
+      return "{$name}_$date.csv";
```
cedric-anne, line 64: Spaces in filenames is not really more a problem, but if you want to "clean" filenames, you should use `Toolbox::slugify()`.

Also `getFriendlyName()` can return an empty string (name may not be set), it may be better to keep type and id: 

```suggestion
      return Toolbox::slugify("{$this->item->getTypeName(1)}_{$this->item->getId()}_{$name}_$date.csv";
```
Do not forget to add the `use \Toolbox;` directive.

tests/functionnal/CsvResponse.php
```diff
@@ -0,0 +1,74 @@
...
+         'name'        => 'testExportToCsv 1',
+         'entities_id' => getItemByTypeName('Entity', '_test_root_entity', true),
+      ]);
+
+      $this->integer($id)->isGreaterThan(0);
+      $this->boolean($computer->getFromDB($id))->isTrue();
+
+      // Output CSV
+      Core_CsvResponse::output(new CsvLogCsvExport($computer, []));
+
```
cedric-anne, line 58: I would prefer that ob* methods encapsulate the minimum PHP code to be sue that any failure on test have a minimal impact. Indeed, I am not sure that atoum correctly close an output buffer that would be opened in a test and not closed.
```suggestion
      // Output CSV
      ob_start();
      Core_CsvResponse::output(new CsvLogCsvExport($computer, []));
      $csv_response = ob_get_clean();
```

tests/functionnal/PlanningCsv.php
```diff
@@ -30,34 +30,15 @@
  * ---------------------------------------------------------------------
  */
 
-namespace tests\units;
+namespace tests\units\Glpi\Csv;
```
cedric-anne, line 33: You should move this file into a `Csv` subdirectory.

tests/units/LogCsvExport.php
```diff
@@ -0,0 +1,133 @@
...
+ * along with GLPI. If not, see <http://www.gnu.org/licenses/>.
+ * ---------------------------------------------------------------------
+*/
+
+namespace tests\units\Glpi\Csv;
+
+use Computer;
+use Glpi\Csv\LogCsvExport as Core_LogCsvExport;
+
+class LogCsvExport extends \CsvTestCase {
```
cedric-anne, line 38: You should move this file into a `Csv` subdirectory.

inc/csv/logcsvexport.class.php
trasher: hum... `use Toolbox;` no? (no starting backslash)

AdrienClairembault:
All points should be OK with the two latest commits.

cedric-anne (commented):
Posting the whole series data to be able to generate a CSV for stats is a bit weird, and may be long on slow connections (if series data are important).
Refactoring this whole part could be a real pain, but maybe we should keep the original way to do, i.e.:
 - storing serialized series into a temporaty file (and name it using a checksum of data),
 - serve file (unserialize and transform it into CSV) when user clicks on a `/front/graph.send.php?checksum=$checksum`.

What do you think of it ?

AdrienClairembault:
> 
> 
> Posting the whole series data to be able to generate a CSV for stats is a bit weird, and may be long on slow connections (if series data are important).
> Refactoring this whole part could be a real pain, but maybe we should keep the original way to do, i.e.:
> 
>     * storing serialized series into a temporaty file (and name it using a checksum of data),
> 
>     * serve file (unserialize and transform it into CSV) when user clicks on a `/front/graph.send.php?checksum=$checksum`.
> 
> 
> What do you think of it ?

I think I would rather clean the front file and generate the data "properly" in the model.
I need to look at the file again but I don't think it would be too long do to.

inc/stat/data/graph/statdatanumber.class.php
```diff
@@ -0,0 +1,93 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with GLPI. If not, see <http://www.gnu.org/licenses/>.
+ * ---------------------------------------------------------------------
+ */
+
+// Using sglobal instead of global as it is a PHP keyword.
+// This is fixed in php 8 so to be changed back when we no longer support php 7.
```
cedric-anne, line 34: This comment do not apply on this file.

inc/stat/data/graph/statdatanumber.class.php
```diff
@@ -0,0 +1,93 @@
...
+use Session;
+
+class StatDataNumber extends StatDataGraph
+{
+   public function __construct(array $params) {
+      parent::__construct($params);
+
+      $avgsolved     = $this->getDataByType($params, "inter_avgsolvedtime");
+      $avgclosed     = $this->getDataByType($params, "inter_avgclosedtime");
+      $avgactiontime = $this->getDataByType($params, "inter_avgactiontime");
```
cedric-anne, line 47: Unless I am wrong, average data should be computed by `StatDataAverageTime` and totals by `StatDataNumber`.

inc/stat/data/sglobal/statdatanumber.class.php
```diff
@@ -0,0 +1,72 @@
...
+ */
+
+// Using sglobal instead of global as it is a PHP keyword.
+// This is fixed in php 8 so to be changed back when we no longer support php 7.
+
+namespace Glpi\Stat\Data\Sglobal;
+
+use Session;
+
+class StatDataNumber extends StatDataGlobal
```
cedric-anne, line 40: This class handles only counters of tickets by type, its name should not be so generic.

inc/stat/data/sglobal/statdataaveragetime.class.php
```diff
@@ -0,0 +1,81 @@
...
+ */
+
+// Using sglobal instead of global as it is a PHP keyword.
+// This is fixed in php 8 so to be changed back when we no longer support php 7.
+
+namespace Glpi\Stat\Data\Sglobal;
+
+use Session;
+
+class StatDataAverageTime extends StatDataGlobal
```
cedric-anne, line 40: This class handles only average times spent on tickets, its name should not be so generic.

inc/stat/data/location/statdataclosed.class.php
```diff
@@ -0,0 +1,51 @@
...
+ * along with GLPI. If not, see <http://www.gnu.org/licenses/>.
+ * ---------------------------------------------------------------------
+ */
+
+namespace Glpi\Stat\Data\Location;
+
+use Session;
+use Ticket;
+
+class StatDataClosed extends StatDataLocation
```
cedric-anne, line 38: Even if there is a technical reason (I did not checked), extending something related to `Location` for a counter of closed tickets seems strange.

cedric-anne (commented):
I did not reviewed precisely this commit, but here is my global remarks.

It seems that the only difference between `Sglobal` and `Graph` classes is 
that how the `Stat::constructEntryValues()` is called. On `Sglobal`, last 3 params (`$param`, `$value` and `$value2`) are ommited, and on `Graph`, they are passed. I guess it could be possible to have only one class that handle both `Sglobal` and `Graph` case.

inc/stat/data/graph/statdatanumber.class.php
AdrienClairembault: I don't think we can do that, the current stat system doesn't seem to work this work (every stats is independent and calculated from `Stat::getDataByType()` with a different key.
Definitely a possible improvement for another PR tho :)

inc/stat/data/location/statdataclosed.class.php
AdrienClairembault: I used the names of the front file : `StatDataLocation` contains the stats previously found in `front/stats.location.php` which have a "Closed" graph.

AdrienClairembault:
All your point should be fixed .

cedric-anne (commented):
Code seems OK, although naming inside inc/stats are not always perfects.

I did not tested the stat part.

cedric-anne (approved):

orthagh:
Functional review:

- not new, but exporting stat from icon side of graphics don't use a correct content-type and libreoffice is not suggested on my computer
image

- not new also (is present in 9.5) stat.item doesn't seems to be correct :

image

- no new again, why stat.location is named so ? it doesn't involve at all location stuff
   - stat.item seems to be a global report about number of ticket by imte
   - stat.location, same, but detailed as opened/closed etc

front/graph.send.php
```diff
@@ -30,25 +30,25 @@
...
-if (isset($_GET["switchto"])) {
-   $_SESSION['glpigraphtype'] = $_GET["switchto"];
-   Html::back();
-}
+include ('../inc/includes.php');
 
-if (($uid = Session::getLoginUserID(false))
-    && isset($_GET["file"])) {
+// Check rights
+Session::checkRight("statistic", READ);
```
orthagh, line 40: for this change of right check (before it seems we just checked session), i approve but i'm not very sure to forget a place in glpi self-service where this file could be called

AdrienClairembault: All modified stats were only accessible through front files like `stat.location.php` which use the same `Session::checkRight("statistic", READ);` right check.

AdrienClairembault:
> not new, but exporting stat from icon side of graphics don't use a correct content-type and libreoffice is not suggested on my computer

From the lib sources :
```php
header('Content-Type: text/csv');
```

I think this is the correct header ?

AdrienClairembault:
> not new also (is present in 9.5) stat.item doesn't seems to be correct :
Seems to be a issue with missing linebreaks (everything is stuck on the second line).

I think this export is handled by `Search` (which we agreed in the previous comments to not modify for this).

AdrienClairembault:
> no new again, why stat.location is named so ? it doesn't involve at all location stuff
> * stat.item seems to be a global report about number of ticket by imte
> * stat.location, same, but detailed as opened/closed etc

The label for this stat is "By hardware characteristics": 
image

Maybe we can rename it `stat.hardware.characteristics.php` so it match the label ?

trasher (approved):
Looks ok to me as is. As far as I see, all remaining "issues" were already present; fixing them should not be a blocker for this one.

@AdrienClairembault could you please rebase and squash so tests can be run with our latest changes?

cedric-anne (approved):

PR #8078, opened 2020-10-07
Fix associated items in change and problems templates

AdrienClairembault:
Following #8072, I wanted to check if the issue was also affecting changes and problems.

It turns out that the dropdown wasn't even rendering and was replaced by a simple text field:
image

The UI was only working for tickets because of three functions found in `Item_Ticket`: 
- `dropdownAllDevices`
- `getSpecificValueToDisplay`
- `getSpecificValueToSelect`

I've moved these functions to a new `CommonItilObject_Item` class and made it the parent of `Item_Ticket`, `Item_Problem` and `Change_Item`.

After that UI was OK but we still had the same issue as in #8072 for changes and problems so I moved the fix from `front/tickettemplatepredefinedfield.form.php ` to `front/itiltemplatefield.form.php`.

Template page was fine after this but when creating a new change or problem the associated items were ignored.

This was because changes and problems didn't have an `items_id` input for this like tickets does so I moved the code from `Ticket` that create and handle this input into two news methods in `CommonItilObject` in order to use them in `Change` and `Problem`.

I've also had to add  `items_id` to `getDefaultValues` for changes and problems (just like it's done in tickets).

I ended up losing a bit of time because I didn't notice that the default template for change and problem was not enabled by default in the root entity config.

I've enabled them by default to have the same behavior that for tickets (and because you expect a template named "default" to be enabled by default :D).

I've tested associated elements on tickets, changes and problems and they are all working now.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes


front/itiltemplatefield.form.php
```diff
@@ -46,6 +46,11 @@
 $item_class = $itiltype . 'Template' . $fieldtype . 'Field';
 $item = new $item_class;
 
+$item_field = $item_class::$items_field;
```
cedric-anne, line 49: I would have put this `$items_field` in `CommonITILObject` and its subclasses and name it `$itemlinkclass` (as we already have `$userlinkclass`, `$grouplinkclass` and `$supplierlinkclass`).

So you could asse it using :
```suggestion
$itil_type = $item_class::$itiltype;
$item_field = $itil_type::$itemlinkclass;
```

By the way, your code will fail if `$fieldtype` is different from `Predefined`.


EDIT: You could also use `getForeignKeyFieldForItemType($itil_type::getItemClass())`

inc/commonitilobject.class.php
```diff
@@ -8239,4 +8239,31 @@ public static function getOpenCriteria(): array {
          ]
       ];
    }
+
+   public function displayHiddenItemsIdInput(array $options): void {
+      foreach ($options['items_id'] ?? [] as $itemtype => $items) {
```
cedric-anne, line 8244: If `$options['items_id']` is equal to `'0'`, it will try to do a foreach on a string.
I guess you should keep old fashionned code that check for emptyness `if (!empty($options['items_id'])) { } `.

inc/commonitilobject.class.php
```diff
@@ -8239,4 +8239,31 @@ public static function getOpenCriteria(): array {
...
+         foreach ($items as $items_id) {
+            echo "<input type='hidden' name='items_id[$itemtype][$items_id]' value='$items_id'>";
+         }
+      }
+   }
+
+   public function handleItemsIdInput(): void {
+      if (!empty($this->input['items_id'])) {
+         $item_class = static::getItemClass();
+         $item_ticket = new $item_class();
```
cedric-anne, line 8254: Maybe `$item_ticket` should be renamed to ... (no idea, but is not only applies to tickets).

inc/commonitilobject.class.php
```diff
@@ -8239,4 +8239,31 @@ public static function getOpenCriteria(): array {
...
+                  'itemtype'                    => $itemtype,
+                  static::getForeignKeyField()  => $this->fields['id'],
+                  '_disablenotif'               => true
+               ]);
+            }
+         }
+      }
+   }
+
+   abstract public static function getItemClass(): string;
```
cedric-anne, line 8268: `getItemClass` seems too generic for me.
```suggestion
   abstract public static function getItemLinkClass(): string;
```

inc/commonitilobject.class.php
AdrienClairembault: Done in c962154.

front/itiltemplatefield.form.php
AdrienClairembault: Done in c962154.

inc/commonitilobject.class.php
AdrienClairembault: Done in c962154.

inc/commonitilobject.class.php
AdrienClairembault: Done in c962154.

cedric-anne (commented):
Not tested yet, but code seems OK.

trasher (commented):
Looks good to me aswell. Will try tomorrow

inc/commonitilobject.class.php
trasher, line 8244: On a new problem form from a fresh install, I get:
```
[16-Oct-2020 06:15:23 UTC] PHP Warning:  Invalid argument supplied for foreach() in /var/www/webapps/glpi/inc/commonitilobject.class.php on line 8244
[16-Oct-2020 06:15:23 UTC] PHP Stack trace:
[16-Oct-2020 06:15:23 UTC] PHP   1. {main}() /var/www/webapps/glpi/ajax/common.tabs.php:0
[16-Oct-2020 06:15:23 UTC] PHP   2. CommonGLPI::displayStandardTab() /var/www/webapps/glpi/ajax/common.tabs.php:92
[16-Oct-2020 06:15:23 UTC] PHP   3. Problem->showForm() /var/www/webapps/glpi/inc/commonglpi.class.php:627
[16-Oct-2020 06:15:23 UTC] PHP   4. Problem->displayHiddenItemsIdInput() /var/www/webapps/glpi/inc/problem.class.php:1208
```

trasher:
Trying to add a predefined field on a new problem template:
```
[16-Oct-2020 06:19:34 UTC] PHP Notice:  Undefined variable: itil_type in /var/www/webapps/glpi/front/itiltemplatefield.form.php on line 50
[16-Oct-2020 06:19:34 UTC] PHP Stack trace:
[16-Oct-2020 06:19:34 UTC] PHP   1. {main}() /var/www/webapps/glpi/front/problemtemplatepredefinedfield.form.php:0
[16-Oct-2020 06:19:34 UTC] PHP   2. include() /var/www/webapps/glpi/front/problemtemplatepredefinedfield.form.php:35
```

install/empty_data.php
```diff
@@ -2102,6 +2102,8 @@
       'autofill_delivery_date'               => 0,
       'autofill_order_date'                  => 0,
       'tickettemplates_id'                   => 1,
+      'changetemplates_id'                   => 1,
```
trasher, line 2105: Should that require a database update?

AdrienClairembault: This will make the "default" templates enabled by default on fresh installs.
It doesn't impact anything as the default template is empty.

I don't think we want to impact existing GLPI instances with a database update.
They may have modified or deleted this template (or they may be using a different template).

trasher: If the value is not strictly needed for the templates to work, ok for me

AdrienClairembault: The value is for this setting (Root entity):
image

inc/commonitilobject.class.php
AdrienClairembault: @cedric-anne reported it but I missed it.
Should be good now, fixed in a0fabc3.

AdrienClairembault:
> 
> 
> Trying to add a predefined field on a new problem template:
> 
> ```
> [16-Oct-2020 06:19:34 UTC] PHP Notice:  Undefined variable: itil_type in /var/www/webapps/glpi/front/itiltemplatefield.form.php on line 50
> [16-Oct-2020 06:19:34 UTC] PHP Stack trace:
> [16-Oct-2020 06:19:34 UTC] PHP   1. {main}() /var/www/webapps/glpi/front/problemtemplatepredefinedfield.form.php:0
> [16-Oct-2020 06:19:34 UTC] PHP   2. include() /var/www/webapps/glpi/front/problemtemplatepredefinedfield.form.php:35
> ```

Fixed in a0fabc3, my bad forgot a var when implementing @cedric-anne suggested changes.

trasher (approved):

PR #8093, opened 2020-10-09
Add unique for reservation items + drop is_deleted

AdrienClairembault:
Internal ref: 19770.
Followup of #7318.

Items have 3 states regarding reservation:
*  reservable (= there is ONE entry in `glpi_reservations_items ` with is_active = 1)
*  reservable but not active (= there is ONE entry in `glpi_reservations_items` with is_active = 0)
*  not reservable (= no entry in glpi_reservations_items).

If for some reason there is multiple entry for one item in `glpi_reservations_items ` then reservations are broken for this item.
-> Adding a unique key to prevent this.

I've also dropped the is_deleted field as it never used (there is no trashbin for this itemtype).

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

### DIFF
--- a/install/update_95_xx.php
+++ b/install/update_95_xx.php
@@ -46,8 +46,9 @@ function update95toXX() {
    $migration->setVersion('x.x.x');
 
    require __DIR__ . '/update_95_xx/softwares.php';
-   include __DIR__ . '/update_95_xx/domains.php';
+   require __DIR__ . '/update_95_xx/domains.php';
    require __DIR__ . '/update_95_xx/devicebattery.php';
+   require __DIR__ . '/update_95_xx/reservationitem.php';
 
    //add is_fqdn on some domain records types
    $fields = [
--- a/install/update_95_xx/reservationitem.php
+++ b/install/update_95_xx/reservationitem.php
@@ -1,0 +1,83 @@
+<?php
+/**
+ * ---------------------------------------------------------------------
+ * GLPI - Gestionnaire Libre de Parc Informatique
+ * Copyright (C) 2015-2020 Teclib' and contributors.
+ *
+ * http://glpi-project.org
+ *
+ * based on GLPI - Gestionnaire Libre de Parc Informatique
+ * Copyright (C) 2003-2014 by the INDEPNET Development Team.
+ *
+ * ---------------------------------------------------------------------
+ *
+ * LICENSE
+ *
+ * This file is part of GLPI.
+ *
+ * GLPI is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * GLPI is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with GLPI. If not, see <http://www.gnu.org/licenses/>.
+ * ---------------------------------------------------------------------
+ */
+
+/**
+ * @var DB $DB
+ * @var Migration $migration
+ */
+
+$migration->displayMessage("Adding unicity key to reservationitem");
+$table = ReservationItem::getTable();
+
+// Drop is_deleted
+$migration->dropKey($table, 'is_deleted');
+$migration->dropField($table, "is_deleted");
+
+// Find duplicates by itemtype/items_id couple
+$duplicates_by_itemtype = $DB->request([
+   'SELECT'  => ['itemtype', 'items_id'],
+   'COUNT'   => 'cpt',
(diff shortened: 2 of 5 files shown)


install/update_95_xx.php
```diff
@@ -48,6 +48,10 @@ function update95toXX() {
    require __DIR__ . '/update_95_xx/softwares.php';
    include __DIR__ . '/update_95_xx/domains.php';
 
+   // Add unicity key to reservationitem
+   require_once __DIR__ . '/update_95_xx/reservationitem.php';
+   reservationItemAddUnicity($migration);
```
cedric-anne, line 53: I would move this call directly in `/update_95_xx/reservationitem.php`.

install/update_95_xx/reservationitem.php
```diff
@@ -0,0 +1,83 @@
...
+      // LIMIT 5000 is there because we need a LIMIT clause to use OFFSET
+      // We dont expect more than a couple duplicates so 5000 is fine
+      $to_delete = $DB->request([
+         'SELECT' =>  'id',
+         'FROM' => $table,
+         'WHERE' => [
+            'itemtype' => $duplicates['itemtype'],
+            'items_id' => $duplicates['items_id'],
+         ],
+         'LIMIT' => 5000,
```
cedric-anne, line 63: I guess you could use this instead of a limit pulled out of a hat:
```suggestion
         'LIMIT' => $duplicates['cpt'],
```

install/update_95_xx.php
AdrienClairembault: I've done it this way because I need to pass `$migration` to the function.

cedric-anne: $migration should be accessible in included file.

By the way, we have to choose which function we use for inclusion :
image

AdrienClairembault: I've switched them to `require` since we don't expect the file to be missing.

cedric-anne (approved):

AdrienClairembault:
Rebased for conflicts.

trasher (approved):

orthagh (changes requested):
There exists a very convenient way to add an unique index AND drop duplicates in the same query:

```SQL
ALTER IGNORE TABLE tablename
ADD UNIQUE INDEX idx_name (xxx, yyy);
```

Note the `IGNORE` keyword.

I didn't check if it's possible with migration class, but i'm sure the update file your suggest is way more complex than adding an option to `Migration::addKey` to ignore warnings.
More the `IGNORE` method is very fast (i'm not sure about your proposal) :)

AdrienClairembault:
> 
> 
> There exists a very convenient way to add an unique index AND drop duplicates in the same query:
> 
> ```sql
> ALTER IGNORE TABLE tablename
> ADD UNIQUE INDEX idx_name (xxx, yyy);
> ```
> 
> Note the `IGNORE` keyword.
> 
> I didn't check if it's possible with migration class, but i'm sure the update file your suggest is way more complex than adding an option to `Migration::addKey` to ignore warnings.
> More the `IGNORE` method is very fast (i'm not sure about your proposal) :)

`ALTER IGNORE TABLE` was only supported by MySQL and was removed 5 years ago : https://dev.mysql.com/worklog/task/?id=7395.

orthagh:
Hum, your're right.

Another suggestion so :
- rename the old table
- create it again with good keys
- do an `INSERT IGNORE`

Not sure about performances, but it should be better then select, delete

install/update_95_xx/reservationitem.php
```diff
@@ -38,46 +38,29 @@
...
+// Insert without duplicates
+$quote_tmp_table = $DB->quoteName($tmp_table);
+$select = $DB->request([
+   'FROM' => $table
+])->getSql();
+$DB->query("INSERT IGNORE INTO $quote_tmp_table $select");
+
+// Replace table with the new version
+$migration->dropTable($table);
+$migration->renameTable($tmp_table, $table);
```
orthagh, line 66:
```suggestion
$migration->renameTable($tmp_table, $table);

```

orthagh (approved):

install/update_95_xx/reservationitem.php
```diff
@@ -38,46 +38,29 @@
...
    'unicity',
    'UNIQUE'
 );
+
+// Insert without duplicates
+$quote_tmp_table = $DB->quoteName($tmp_table);
+$select = $DB->request([
+   'FROM' => $table
+])->getSql();
+$DB->query("INSERT IGNORE INTO $quote_tmp_table $select");
```
cedric-anne, line 62:
```suggestion
$DB->queryOrDie("INSERT IGNORE INTO $quote_tmp_table $select");
```

cedric-anne, line 62: Could you add a comment in code to explain why there is an usage of INSERT IGNORE ? It may help to understand this copy/drop/rename part.

cedric-anne (approved):

trasher (approved):

Issue #8114, opened 2020-10-15
9.5.2 - bad items count implementation


btry:
Here is an SQL error

```
[2020-10-14 15:39:55] glpisqllog.ERROR: DBmysql::query() in /var/www/glpi/inc/dbmysql.class.php line 306
  * MySQL query error:
  SQL: SELECT COUNT(*) AS cpt FROM glpi_items_tickets INNER JOIN glpi_tickets ON (glpi_items_tickets.tickets_id = g
lpi_tickets.id) WHERE itemtype = '2008' AND items_id = '3 - J'ai une question' AND 
is_deleted = '0'
  Error: You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the righ
t syntax to use near 'ai une question' AND is_deleted = '0'' at line 1
  Backtrace :
  inc/dbmysqliterator.class.php:95
  inc/dbmysql.class.php:857                          DBmysqlIterator->execute()
  inc/dbutils.class.php:335                          DBmysql->request()
  inc/db.function.php:178                            DbUtils->countElementsInTable()
  inc/ticket.class.php:746                           countElementsInTable()
  inc/commonitilobject.class.php:7727                Ticket->getTabNameForItem()
  inc/ticket.class.php:843                           CommonITILObject->defineDefaultObjectTabs()
  inc/commonglpi.class.php:294                       Ticket->defineTabs()
  inc/commonglpi.class.php:786                       CommonGLPI->defineAllTabs()
  inc/commonglpi.class.php:1227                      CommonGLPI->showPrimaryForm()
  front/ticket.form.php:221                          CommonGLPI->display()
```

In GLPI 9.5.0 CommonITILObject has a new method getLinkedItems() which overrides an implementation in CommonDBTM

### CommonDBTM

- returns an empty array
- methods doesn't accepts any argument

![image](https://user-images.githubusercontent.com/14139801/96086573-d4b4e000-0ec2-11eb-83b2-e6dd63cfc9c9.png)

### Computer

- an implementation (in Computer, Printer for example) shows how is the returned array. First dimension uses itemtype as key, and the value is an array with an item ID as key and as value.

```php
   function getLinkedItems() {
      global $DB;

      $iterator = $DB->request([
         'SELECT' => ['itemtype', 'items_id'],
         'FROM'   => 'glpi_computers_items',
         'WHERE'  => ['computers_id' => $this->getID()]
      ]);

      $tab = [];
      while ($data = $iterator->next()) {
[...]

PR #8115, opened 2020-10-15
Fix itemcount in Ticket

btry:
I think I properly updated all uses of getLinkedItems to a CommonITILObject

| Q             | A
| ------------- | ---
| Bug fix?      | yes/
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #8114

### DIFF
--- a/inc/commonitilobject.class.php
+++ b/inc/commonitilobject.class.php
@@ -8206,8 +8206,8 @@ abstract class CommonITILObject extends CommonDBTM {
     * @return boolean
     */
    protected function hasImpactTab() {
-      foreach ($this->getLinkedItems() as $linkedItem) {
-         $class = $linkedItem['itemtype'];
+      foreach ($this->getLinkedItems() as $itemtype => $items) {
+         $class = $itemtype;
          if (Impact::isEnabled($class) && Session::getCurrentInterface() === "central") {
             return true;
          }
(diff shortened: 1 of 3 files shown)


inc/ticket.class.php
```diff
@@ -705,26 +705,24 @@ function getTabNameForItem(CommonGLPI $item, $withtemplate = 0) {
 
                   if (count($linkeditems)) {
                      foreach ($linkeditems as $type => $tab) {
-                        foreach ($tab as $ID) {
```
btry, line 708: diff is unclear : removed this foreach and de-indented 1 level its content

inc/commonitilobject.class.php
```diff
@@ -8177,33 +8177,27 @@ public static function getItemsTable() {
     *
     * @return array
     */
-   public function getLinkedItems(bool $addNames = true) :array {
```
trasher, line 8180: Should be removed from phpdoc

trasher (approved):

inc/commonitilobject.class.php
btry: CommonDBTM already has a docblock, then I removed it completely.

trasher (approved):

cedric-anne (approved):

PR #8136, opened 2020-10-19
Fix shared dashboards + redirect

AdrienClairembault:
1\)  Fix shared dashboards

If an user : 
- doesn't have the right to see all dashboards
- can see one or more shared dashboards
- can't see the default dashboard for the given page 

Then he get an empty page instead of being able to see the few shared dashboards he is able to see.
Fixed in e5e6a44.

Before:
image

After:
image

2\) Redirect if no dashboards

When clicking on the "assets" or "assistance" menu, if you can't see any dashboards for this page then you get an almost empty page.
I think it would be better to redirect the user to a useful page instead.

Fixed in 368c753 (url used to redirect were the one used in 9.4 before we added dashboards).

Before:
image

After:
image

3\) Default dashboard set by profiles

No change done on this, only a suggestion.
Shouldn't this setting be defined in each profiles instead of the general conf ?
Or even better, keep the global conf but allow profiles to overload it ?

image

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

### DIFF
--- a/front/dashboard_assets.php
+++ b/front/dashboard_assets.php
@@ -39,10 +39,15 @@ include ('../inc/includes.php');
 
 
 Session::checkCentralAccess();
+$default = Glpi\Dashboard\Grid::getDefaultDashboardForMenu('assets');
+
+// Redirect to "/front/computer.php" if no dashboard found
+if ($default == "") {
+   Html::redirect($CFG_GLPI["root_doc"] . "/front/computer.php");
+}
 
 Html::header(__('Assets Dashboard'), $_SERVER['PHP_SELF'], "assets", "dashboard");
 
-$default = Glpi\Dashboard\Grid::getDefaultDashboardForMenu('assets');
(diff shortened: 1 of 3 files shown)


orthagh (approved):

trasher (approved):

trasher:
> 3) Default dashboard set by profiles
> 
> No change done on this, only a suggestion.
> Shouldn't this setting be defined in each profiles instead of the general conf ?
> Or even better, keep the global conf but allow profiles to overload it ?

I'd say it could be set per profile in adition of the global configuration, yes; seems consistent with other parts working like this.

andrepetinga:
I think that the existing permission "All dashboards" should be improved.

image

In my opinion we should apply permission "Per Dashboard".

Because there are dashboards that doesn't make sense for all profiles. Or is this already possible and I'm missing something? :)

PR #8147, opened 2020-10-21
fix(ui): fix min dashboard css for lower resolution

stonebuzz:
For lower resolution (ie 1280 * 1024 or less), long label from mini dashboard (big number type) goes beyond the widget (see "tickets waiting your validation)

image

 I propose this : 

- put main-icon to top
- give 100% of width for label
- reduce number font size (3 to 2.5)

And the result

image

This affect only min dashboard for screen with width lower of 1280 

Best regards

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | internal 20871

Issue #8110, opened 2020-10-14
[GLPI 9.5.2]Dashboard: faulty data display & missing translations


CupidSG:
Hello,

**Bug**

- In a widget, data are missing when displayed with " Computer by manufacturer " options and "horizontal bars" or "bars" display options. You can see some data when resizing the widget (cf screenshot) but only during the actions.

- A lot of words and strings on dashboard and widgets are displayed in french even if language is different.

**Page(s) URL**

/front/central.php

**To reproduce**

1. Create a personal dashboard
2. Create a widget
3. in this widget, choose to display "Computer by manufacturer"
4. Choose horizontal bars or bars
5. Save
6. See errors (data are missing or poorly displayed)

Steps to try to fix:

1. Resize widget (with mouse) (cf screenshot)
2. Language modification fix the problem

**Expected behavior**

I except to see all datas with the right display and the right language.

**Logs**

I find nothing in logs about this problem.

**Screenshots**

Data displayed directly after the first setup of the widget:

![image](https://user-images.githubusercontent.com/58808962/96012340-4e59b900-0e44-11eb-8da0-9aaaf64445d4.png)

i'm trying to resize the widget :

![image](https://user-images.githubusercontent.com/58808962/96012370-54e83080-0e44-11eb-9438-8a986120c7d6.png)

after language change:

![image](https://user-images.githubusercontent.com/58808962/96013779-ed32e500-0e45-11eb-8e33-b60a917f86db.png)

Extra problem, Translations:

![image](https://user-images.githubusercontent.com/58808962/96012415-5e719880-0e44-11eb-997b-3bf5d53a92a3.png)

**Your GLPI setup (you can find it in Setup > General menu, System tab)**

```
{
    "glpi": {
        "uuid": "hw1KIi7JduiXdwh7dCVWBSMdbQIqQhHvg5Xq2Pv6",
        "version": "9.5.2",
        "plugins": [
            {
                "key": "news",
                "version": "1.7.0"
            },
            {
                "key": "anonymize",
                "version": "2.1.5"
            },
            {
                "key": "approvalbymail",
                "version": "2.0.3"
            },
            {
                "key": "branding",
                "version": "1.1.0"
            },
            {
                "key": "collaborativetools",
                "version": "1.1.0"
            },
            {
                "key": "credit",
                "version": "1.8.0"
            },
            {
                "key": "datainjection",
                "version": "2.8.0"
            },
            {
                "key": "escalade",
                "version": "2.6.0"
[...]

PR #8151, opened 2020-10-21
fix dashboards (overlapping with search bar, cache issue on language, bars graph hidden on chrome)

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | replace #8147, internal !20871, #8110

### DIFF
--- a/css/dashboard.scss
+++ b/css/dashboard.scss
@@ -66,7 +66,7 @@ $break_tablet: 1400px;
       padding-top: 0;
       width: calc(80% + 28px);
       margin: 0 auto;
-      margin-bottom: -35px;
+      margin-bottom: -20px;
       z-index: 1;
 
       & + .search_page {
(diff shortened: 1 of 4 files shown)


stonebuzz:
tested and approved

orthagh:
For commit 1cf1cc7, 2 things done:
- i removed cache on cards collection, it didn't improve response time in fact, and on development processes, it's annoying
- i added language on cache key footprint
- i added card cache key on xhr request (and add the language specifity also), because the above point was nullified by browser cache.

orthagh:
For bb5ca7d and the related #8110 issue:

The bug triggers only on chromium based browser (not in firefox, my main browser...) and it seems on bars graph, the height attribute computing of the parent svg is affected by animation (i don't know precisely why), and i decided to fix by forcing a min height on it (and hide the issue in fact).

I think about a regression/bug in the browser rather than in the lib

trasher (approved):

cedric-anne (approved):

Issue #8132, opened 2020-10-17
9.5.2 - When add consumables it does not add the Financial and administrative information


enlanebulosa:
**Describe the bug**

When add consumables it does not add the Financial and administrative information that is found in the Consumable Management tab

**To reproduce**

1. Go to 'Assets --> Consumables'
2. Click on 'Item'
3. Go to 'Consumables' tab
4. Click on 'Add consumable'
5. Go to icon show Financial and administrative information

**Expected behavior**

Add Financial and administrative information what was added to the item

**Screenshots**

![Screenshot (2)](https://user-images.githubusercontent.com/22381246/96351281-b3095380-1090-11eb-9f3e-4e24386fef44.png)
![Screenshot (4)](https://user-images.githubusercontent.com/22381246/96351288-bb618e80-1090-11eb-92d4-4ce3cfb0b1c8.png)
![Screenshot (5)](https://user-images.githubusercontent.com/22381246/96351290-bf8dac00-1090-11eb-8eae-a32bbd499272.png)

**Additional context**

I replied this error on GLPI Network Cloud (https://hec2.with.glpi-network.cloud)

PR #8152, opened 2020-10-21
Fix inheritance of consumable financial information

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #8132

### DIFF
--- a/inc/commondbchild.class.php
+++ b/inc/commondbchild.class.php
@@ -452,6 +452,7 @@ abstract class CommonDBChild extends CommonDBConnexity {
     * @return void
    **/
    function post_addItem() {
+      global $CFG_GLPI;
 
       if ((isset($this->input['_no_history']) && $this->input['_no_history'])
           || !static::$logs_for_parent) {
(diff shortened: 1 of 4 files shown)


cedric-anne (commented):
Unless I am wrong, for now, this inheritance is done only on `SoftwareLicence` (inheritance from `Software`). If we add it for `Consumable`, we should also add it for `Cartridge`.

Do you think it is possible to do a global logic like this ?
```
if ($this instanceof CommonDBChild && in_array(static::class, $CFG_GLPI["infocom_types"]) && in_array(static::$itemtype, $CFG_GLPI["infocom_types"])) {
   // inherit infocom
}
```

It would not handle the `SoftwareLicence` case as this class is not a `CommonDBChild`, but it is another problem.

cconard96:
> Unless I am wrong, for now, this inheritance is done only on `SoftwareLicence` (inheritance from `Software`).

I wasn't aware of that functionality with software licenses. I just knew my workflow for adding a bunch of new consumables had been to set the financial info on the ConsumableItem, add the consumables, then disable the financial info on the ConsumableItem to prevent accidentally adding it to consumables from a different order.
This functionality was removed when cloning was added (1de713e6777edc0c57139246f373d7abda1e5b00). The `post_addItem` in consumable was removed. I re-added it and changed it to use the new clone function.

Looking more at the commit, there are more instances of deleted `post_addItem` that will need tested and re-implemented.

cedric-anne (approved):

trasher (approved):

PR #8174, opened 2020-10-27
Fix ticket dashboards titles

AdrienClairembault:
Introduced by https://github.com/glpi-project/glpi/pull/8151.

Tickets dashboard titles are broken on latest 9.5/bugfixes:
image

This is caused by a font size increase (11px -> 2em in #8151).
To fix this I've allowed the line to use 100% of the available width and added a text-ellipsis in case the text may still be too long in some other languages (in this case the user can hover on the text to see the full title thanks to the new title attribute).

After (with example of a long title for the last one) : 
image

FYI the labels were much smaller in 9.5.1 (font size is em based, 

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

### DIFF
--- a/inc/dashboard/widget.class.php
+++ b/inc/dashboard/widget.class.php
@@ -273,6 +273,7 @@ class Widget extends CommonGLPI {
          ? "href='{$p['url']}'"
          : "";
 
+      $label = $p['label'];
       $html = <<<HTML
       <style>
          #{$p['id']} {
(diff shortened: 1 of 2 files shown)


inc/dashboard/widget.class.php
```diff
@@ -290,7 +291,7 @@ public static function bigNumber(array $params = []): string {
          class="card big-number $class"
          title="{$p['alt']}">
          <span class="content">$formatted_number</span>
-         <div class="label">{$p['label']}</div>
```
orthagh, line 293: this change is probably not needed, no ?

AdrienClairembault: This is for when to text is shortened by the ellipsis: 
image

orthagh (approved):

inc/dashboard/widget.class.php
orthagh: sorry, i didn't see the title part

cedric-anne (approved):

trasher (approved):

PR #8258, opened 2020-11-12
Increase max items per contract

AdrienClairembault:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !21058


trasher (approved):

orthagh (approved):
I agree with this change.

To note, i'm not satisfied at all with the current number input.
The max value is mandatory, a dropdown is not very convenient.

I'll suggest we work, onto `master`branch, on a overhaul of these dropdowns and move to a `input[type=number]`, removing max value when it's not needed (the majority i think)

cedric-anne (approved):
Same remark

PR #8274, opened 2020-11-14
Try using current location as map default

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

Try to make it easier to assign geolocation data by defaulting the map view to the current location (Or as close as we can approximate) if the user allows us to use their location. If not, fall back on the previous defaults.
If we can get geolocation data, the map zoom level is adjusted based on the reported accuracy.
This only works over HTTPS or localhost. If those requirements don't apply to the user, they will see the original functionality.

Tested on a desktop (very-low accuracy) and a cell phone (high accuracy) and the zoom levels seemed appropriate.

### DIFF
--- a/inc/mapgeolocation.class.php
+++ b/inc/mapgeolocation.class.php
@@ -83,58 +83,84 @@ trait MapGeolocation {
       }
 
       $(function(){
-         map_elt = initMap($('#setlocation_container'), 'setlocation', '200px');
-
-         var osmGeocoder = new L.Control.OSMGeocoder({
-            collapsed: false,
-            placeholder: '".__s('Search')."',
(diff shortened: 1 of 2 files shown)


trasher:
I've not tested yet, code sounds correct. On which browsers did you test?

orthagh (approved):

cconard96:
Tested on:
- Edge (Chromium version on Windows)
- Firefox (Windows, Linux)
- Chrome (Windows, Linux, Android)

trasher (approved):

PR #8291, opened 2020-11-18
Fix Object/Int comparison in templates

AdrienClairembault:
Fix this warning: 

image

| Q             | A
| ------------- | ---
| Bug fix?      | yes/
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !21085


orthagh (approved):

cedric-anne (changes requested):
Problem is not the comparison, but the fact that the `$options[$key]` should contains the previous template id, not the new one.

See #8292 .

PR #8322, opened 2020-11-24
Few minor changes/improvements

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

### DIFF
--- a/inc/dropdown.class.php
+++ b/inc/dropdown.class.php
@@ -822,274 +822,164 @@ class Dropdown {
       if (is_null($optgroup)) {
          $optgroup = [
              __('Common') => [
-                 'Location'               => Location::getTypeName(Session::getPluralNumber()),
-                 'State'                  => _n('Status of items',
-                                                'Statuses of items',
-                                                Session::getPluralNumber()),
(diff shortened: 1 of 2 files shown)


trasher:
To be reviwed per commit.

* 1st one is a new possibility not yet used, but I'll need it on another branch
* 2nd one is a small improvement
* 3rd one use cases changed for 2nd

inc/dropdown.class.php
```diff
@@ -822,274 +822,164 @@ static function getStandardDropdownItemTypes() {
       if (is_null($optgroup)) {
          $optgroup = [
              __('Common') => [
-                 'Location'               => Location::getTypeName(Session::getPluralNumber()),
```
trasher, line 825: All entries were in facts using the getTypeName, either directly or copy/pasted. I've kept an array with null value to be compliant with existing (for plugins); doing differently would certainly require an extra foreach on all entries.

orthagh (approved):

PR #8367, opened 2020-12-02
Prevent object to be sanitized; closes #8345

### LINKED ISSUE #8345: [GLPI 9.5] $tkt['_message'] object is corrupted Toolbox::addslashes_deep()
Hello,

**Describe the bug**

When getting an email, the buildTicket function returns an array which contains the key '_message' in which is a Laminas\Mail\Storage\Message object, but unfortunately this object is corrupted by the Toolbox::addslashes_deep() which is called just before return.

**To reproduce**

1. Send an email to GLPI
1. Set a breakpoint on the line 1218 in the mailcollector.class.php file
1. Click on the get emails from the receiver page
1. See the value of the $tkt['_message']: it's an object
1. execute the line
1. See that the value of $tkt['_message'] is no longer an object: it's now a string

**Expected behavior**

To keep the object value in the '_message' key of the $tkt variable, so that it will be available later on in the hooks like _pre_item_add_

**Logs**

Nothing is present in the logs

**Screenshots**
$tkt['_message'] is an object
image

The new value of the $tkt['_message']: it's now a string
image

**Suggestion**

Could it be possible to:
1. Either to apply the addslashes_deep only to the $tkt['content'] which really needs to be treated,
or
2. Either add a condition in the addslashes_deep to keep _as is_ the objects like  it is done for the ressources.

Thank you,
Regards,

Tomolimo

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #8345

### DIFF
--- a/inc/mailcollector.class.php
+++ b/inc/mailcollector.class.php
@@ -1059,7 +1059,6 @@ class MailCollector  extends CommonDBTM {
       } catch (Laminas\Mail\Storage\Exception\InvalidArgumentException $e) {
          $subject = null;
       }
-      $tkt['_message']  = $message;
 
       if (!Toolbox::seems_utf8($body)) {
          $tkt['content'] = Toolbox::encodeInUtf8($body);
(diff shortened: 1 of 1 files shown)


cedric-anne (approved):

tomolimo (approved):
Hello
I made a test using this fix, and it's OK.
You may merge this.
Thank you,
Regards,
Tomolimo

Issue #8303, opened 2020-11-19
Public saved search are saved as private (9.5.2)


fralla2:
**Describe the bug**

When creating a saved search from tickets or assets page (possibly others too), if I set it to public on creation, it is saved as private search.

If I go to the front/savedsearch.php page (Tools -> Saved search), edit the search and change it to public from there, it saves as public correctly

**Page(s) URL**

front/ticket.php
front/computer.php
front/savedsearch.php

**To reproduce**

Go to ticket.php page
Create a search filter
Click on the little star to created a saved search
Change the visibility to "Public"
Save the filter
Go to the filter list (top corner)
The saved search is listed as private

**Expected behavior**

The saved search should be saved accordingly to the visibility dropdown

**Logs**

```
No Logs
```

**Screenshots**

![image](https://user-images.githubusercontent.com/38400135/99685554-b48fb800-2a50-11eb-917a-8fb9d7f7d823.png)
![image](https://user-images.githubusercontent.com/38400135/99685673-d6893a80-2a50-11eb-939b-cc230697946a.png)
![image](https://user-images.githubusercontent.com/38400135/99685838-01738e80-2a51-11eb-830b-37edbe2805ae.png)
![image](https://user-images.githubusercontent.com/38400135/99686010-28ca5b80-2a51-11eb-9ee4-653905068319.png)

**Your GLPI setup (you can find it in Setup > General menu, System tab)**

```
[code]   GLPI 9.5.2 (/glpi => /var/www/html/glpi-9.5.2) Installation mode: TARBALL
[...]

PR #8435, opened 2020-12-21
Fix creation of public saved search

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #8303

There was a duplicate `is_private` field shown for technicians who could create private searches and it was overriding the user's selection and setting it back to the default which is private.
This hidden input now only shows when the user cannot create a public search and therefore only has one `is_private` field in either case now.

### DIFF
--- a/inc/savedsearch.class.php
+++ b/inc/savedsearch.class.php
@@ -426,7 +426,9 @@ class SavedSearch extends CommonDBTM implements ExtraVisibilityCriteria {
       }
       if ($ID <= 0) { // add
          echo Html::hidden('users_id', ['value' => $this->fields['users_id']]);
-         echo Html::hidden('is_private', ['value' => $this->fields['is_private']]);
+         if (!self::canCreate()) {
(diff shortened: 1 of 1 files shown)


inc/savedsearch.class.php
```diff
@@ -426,7 +426,9 @@ function showForm($ID, $options = []) {
       }
       if ($ID <= 0) { // add
          echo Html::hidden('users_id', ['value' => $this->fields['users_id']]);
-         echo Html::hidden('is_private', ['value' => $this->fields['is_private']]);
+         if (!self::canCreate()) {
+            echo Html::hidden('is_private', ['value' => $this->fields['is_private']]);
```
trasher, line 430: Maybe:
```suggestion
            echo Html::hidden('is_private', ['value' => 1]);
```

orthagh (approved):

PR #8504, opened 2021-01-07
Fix associated elements in ticket list

xacobofg:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #4107

I don't know exactly why it fails depending on the database you have.
In my case it was with 10.3.27-MariaDB-1:10.3.27+maria~stretch-log.
When I executed the SQL with that database the result was:

ITEM_Ticket_13 | ITEM_Ticket_13_itemtype
1507$#$4631$$##$$1891$#$4632 | Computer$#$4632$$##$$Monitor$#$4631

When it should be:

ITEM_Ticket_13 | ITEM_Ticket_13_itemtype
1507$#$4631$$##$$1891$#$4632 | Computer$#$4631$$##$$Monitor$#$4632

To avoid this I forced to also order by the id the field "ITEM_Ticket_13_itemtype"

### DIFF
--- a/inc/search.class.php
+++ b/inc/search.class.php
@@ -3494,7 +3494,7 @@ JAVASCRIPT;
                 || (isset($searchopt[$ID]["forcegroupby"]) && $searchopt[$ID]["forcegroupby"])) {
                $ADDITONALFIELDS .= " IFNULL(GROUP_CONCAT(DISTINCT CONCAT(IFNULL(`$table$addtable`.`$key`,
                                                                          '".self::NULLVALUE."'),
-                                                   '".self::SHORTSEP."', $tocomputeid) SEPARATOR '".self::LONGSEP."'), '".self::NULLVALUE.self::SHORTSEP."')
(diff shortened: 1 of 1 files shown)


cedric-anne (approved):
Now I wonder why there are some GROUP_CONCAT that are not ordered. Same kind of issue may exist for them.

trasher (approved):

OscarBeiro:
> Now I wonder why there are some GROUP_CONCAT that are not ordered. Same kind of issue may exist for them.

We've been almost a couple of days looking for the origin, and we couldn't find it. I'm a hundred per cent sure this issue affects lots of installations, but it will only pop up when more than 2 assets are linked to a ticket. 

We have a particular server with all the tickets with 2 assets linked, which became a real problem. 

Anyway, forcing group by in every search query it looks, at least, consistent.

PR #8506, opened 2021-01-08
Changed attached filename retrieve order, first get filename from content-type and then from content-disposition

tomolimo:
If header is present will get it from content-disposition

Added a test to be sure that the file name will be accepted by the underlying OS, and if not, then will generate a new file name (keeping the file extension).

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #8495

### DIFF
--- a/inc/mailcollector.class.php
+++ b/inc/mailcollector.class.php
@@ -1535,18 +1535,14 @@ class MailCollector  extends CommonDBTM {
             $subpart = 1;
          }
 
-         $filename = '';
+         // Try to get filename from Content-Type header
+         $filename = $content_type_header->getParameter('name') ?? '';
 
          // Try to get filename from Content-Disposition header
          if (empty($filename)
              && $part->getHeaders()->has('content-disposition')
              && ($content_disp_header = $part->getHeader('content-disposition')) instanceof ContentDisposition) {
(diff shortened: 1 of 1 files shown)


trasher:
Existing tests are failing; and a new one must be added to check that specific case

tomolimo:
@trasher, ok I'll check that

tomolimo:
Hello @trasher 
I'm blocked with the `Laminas\Mail\Protocol\Exception\RuntimeException: cannot connect to host; error = stream_socket_client(): unable to connect to tcp://127.0.0.1:143 (No connection could be made because the target machine actively refused it.` error message.
Is the imap server included in the tests folder, or do use something else? And which one? 
Thank you,
Regards,
Tomolimo

trasher:
I guess any one can be used; CI relies on a dovecot in which we inject eml files (see CI configuration for the exact command).
You can avoid installing a local mail server if you rely on github actions from another working branch to write your tests; you can disable other tests and check temporary in the CI conf for faster results (because mail tests are not run first).

trasher:
Replaced by #8531

PR #8531, opened 2021-01-14
Fixed mailcollector with multibyte filename (#8495)

edgardmessias:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #8495

https://github.com/laminas/laminas-mail/pull/128


trasher:
Sounds like the fix is correct, thank you :+1: 
Anyways, we need to add the case in GLPI tests, by adding a .eml file that permit to reproduce the initial issue.

tomolimo:
Hello @edgardmessias 

Here is the eml file,
[22-multibyte_attachment_name.zip](https://github.com/glpi-project/glpi/files/5819336/22-multibyte_attachment_name.zip)

Thank you
Regards,
Tomolimo

tomolimo:
Hello @edgardmessias 

I made a test with your solution: it's OK, and I think it's better than mine :smiley: 

However, could you add a test to be sure that `file_put_contents($path.$filename, $contents)` will be accepted by OS?
Who knows if the attachment name has invalid chars for the OS hosting GLPI...

```
         // try to save file to disk, to be sure file name is valid (= accepted by OS)
         $ret = @file_put_contents($path.$filename, $contents);
         if ($ret === false) {
            // if not, then create a valid file name
            $ext = pathinfo($filename, PATHINFO_EXTENSION);
            $strip = ["*", "\\", "|", ":", "\"", "<", ">", "/", "?"];
            $filename = "doc_{$subpart}_(renamed)_"
               . trim(str_replace($strip, "", pathinfo($filename, PATHINFO_FILENAME)))
               . ".$ext";
            // retry to save the file
            $ret = file_put_contents($path.$filename, $contents);
         }
         if ($ret) {
```

Thank you
Regards,

Tomolimo

tomolimo:
Hello @edgardmessias 
Here is the eml for the bad chars for OS:
[23-bad_attachment_name_for_OS.zip](https://github.com/glpi-project/glpi/files/5820727/23-bad_attachment_name_for_OS.zip)
Thank you,
Regards,
Tomolimo

edgardmessias:
Hello @tomolimo .
Big thanks.

As possible, I will write the unit tests

trasher:
Thank you very much @edgardmessias! Please feel free to ping us if you have issues with tests.

glensc:
btw, you may want to use this package for filename normalizing:
- https://packagist.org/packages/glen/filename-normalizer

edgardmessias:
@trasher @tomolimo , Could you help me to fix the test? In my local environment is working fine.

trasher:
Hi @edgardmessias I'll try to take a look today to understand what's going wrong

inc/mailcollector.class.php
```diff
@@ -1507,14 +1508,18 @@ function getTotalMails() {
    private function getRecursiveAttached(\Laminas\Mail\Storage\Part $part, $path, $maxsize, $subject, $subpart = "") {
       if ($part->isMultipart()) {
          $index = 0;
-         foreach (new RecursiveIteratorIterator($part) as $mypart) {
+         // while instead of foreach to supress header error in laminas-mail
+         $iterator = new RecursiveIteratorIterator($part);
+         while (@$iterator->valid()) {
```
trasher, line 1513: Is that strictly required? I do not like `@` usage

cedric-anne: If there is an error triggered by laminas-mail on some emails, we should, at least, open an issue on their side.

I do not like suppressing errors, especially when the called method can have many unpredictible reasons to trigger an error. Indeed, you may want to suppress a specific error that may not affect us, but you will also mute all other errors related to any email parsing.

trasher:
Error in tests is the following:
```
'2021-01-18 07:46:50 [@976e2e5af287]
Message is invalid: The input is not a valid email address. Use the basic format local-part@hostname<br/>2021-01-18 07:46:50 [@976e2e5af287]
Error during message parsing: Header with Name date or date not found<br/>2021-01-18 07:46:50 [@976e2e5af287]
Error during message parsing: str_split(): Argument #2 ($length) must be greater than 0<br/>'
```

Found it reading the `mailgate.log` file from tests

tomolimo:
Hello @edgardmessias @trasher 
On my environment I got the same kind or problem, the errors counter is not OK.
I believe, this is due to one of the emails that can or cannot (depending on the mail server) serves it.
I think (but not sure) that it's coming from the chat email (eml file name is: `05-empty-from.eml`)
Regards,
Thank you,
Tomolimo

inc/mailcollector.class.php
edgardmessias: The main problem in laminas-mails is `str_split(): Argument #2 ($length) must be greater than 0`.
I openned a PR to resolve it: https://github.com/laminas/laminas-mail/pull/128

trasher: OK, so we'll have to wait for new laminas-mail release - or at least for that PR to be approved

stonebuzz (changes requested):
this part
```php
         // Try to get filename from Content-Type header
         if (empty($filename)) {
            $filename = $content_type_header->getParameter('name') ?? '';
         }
```

need to be moved after the attempt to convert the filename with encoding value (espaclially with base64encoding) 
Otherwise the name of the file becomes the title of the email.

You can try with this file 

[DUC_SING WAH PAN (MA Chan Ho Ki).pdf](https://github.com/glpi-project/glpi/files/5848783/DUC_SING.WAH.PAN.MA.Chan.Ho.Ki.pdf)

Done by latest commit

stonebuzz:
Sorry @edgardmessias I have force-push to you PR :confused:

edgardmessias:
@stonebuzz , don't worry

.github/workflows/ci.yml
```diff
@@ -227,6 +227,13 @@ jobs:
           docker exec app bin/console glpi:database:configure --config-dir=./tests --no-interaction --reconfigure --db-name=glpi --db-host=db --db-user=root
           docker exec app vendor/bin/atoum -p 'php -d memory_limit=512M' --debug --force-terminal --use-dot-report --bootstrap-file tests/bootstrap.php --no-code-coverage --fail-if-skipped-methods --max-children-number 1 -d tests/database
           docker exec app bin/console glpi:database:configure --config-dir=./tests --no-interaction --reconfigure --db-name=glpi --db-host=db --db-user=root
+      - name: "IMAP tests"
```
cedric-anne, line 230: I put tests on the top, but this should be reverted before merge.

cedric-anne (commented):
@edgardmessias 

I applied your patch and adapted a bit tested emails, but I think there is still a bug in laminas component, in the method that returns the filename value. As I changed manually tested mails headers, Iit may be due to a bad manipulation, but filenames are correct when I open them with Thunderbird.

You can check test suite output.

.github/workflows/ci.yml
edgardmessias: Yes, I looking here what is the problem

edgardmessias:
@cedric-anne , I think it is a good idea to append this code for bad header:

```php
if (strncmp($filename, '=?', 2) === 0) {
   $filename = preg_replace('/\?\=\=\?[\w-]+\?\w\?/', '', $filename);
   var_dump($filename);
   $filename = \Laminas\Mail\Header\HeaderWrap::mimeDecodeValue($filename);
   var_dump($filename);
}
```

tests/imap/MailCollector.php
```diff
@@ -593,12 +596,16 @@ public function testCollect() {
       // Check creation of expected documents
       $expected_docs = [
          '00-logoteclib.png',
-         '01-Screenshot-2018-4-12 Observatoire - France très haut débit.png',
+         // Space is missing between "France" and "très" due to a bug in laminas-mail
+         '01-Screenshot-2018-4-12 Observatoire - Francetrès haut débit.png',
```
cedric-anne, line 600: @edgardmessias 

Did you saw that ?

I guess this is only a side effect of your fix. Indeed, now the space char is located at the beginning or end of a param line, and trimming done in Laminas removes it. Before your fix, this space was certainly located in the middle of second parameter line.

Param lines are extracted with these values:
```
array(2) {
  [0]=>
  string(45) "01-Screenshot-2018-4-12 Observatoire - France"
  [1]=>
  string(21) "très haut débit.png"
}
```

cedric-anne (approved):

edgardmessias:
@cedric-anne , I fixed the problem with malformed header: https://github.com/laminas/laminas-mail/pull/128/commits/2f81aedcb83de61e00fffd6aee4c8938272a1b24
This solve the problem with previous eml file

cedric-anne:
> @cedric-anne , I think it is a good idea to append this code for bad header:
> 
> ```
> if (strncmp($filename, '=?', 2) === 0) {
>    $filename = preg_replace('/\?\=\=\?[\w-]+\?\w\?/', '', $filename);
>    var_dump($filename);
>    $filename = \Laminas\Mail\Header\HeaderWrap::mimeDecodeValue($filename);
>    var_dump($filename);
> }
> ```

Did you already encounter this case ? Was it due to a malformed email, or to a laminas-mail bug ?

Trying to fix laminas-mail behaviour in our side should only be done if we are not able to propose them a fix. (This has been done to handle RFC5987 format for filenames only because I was not able to find how to fix that on their side).

Trying to fix malformed emails is not always a good idea, it may introduce side effects.

Unless you are able to quickly propose a fix and test(s) email(s) that validates the fix, you should open another issue/PR for this case, to not block merging of current one.

Regards

tomolimo:
Hello @edgardmessias 
image

I have some troubles with the `patch` command which is not available on Windows...
Could it be possible to use agnostic tools?
Thank you
Regards,
Tomolimo

cedric-anne:
> Hello @edgardmessias
> image
> 
> I have some troubles with the `patch` command which is not available on Windows...
> Could it be possible to use agnostic tools?
> Thank you
> Regards,
> Tomolimo

I would love to use agnostic tool for this. Do you know a tool that works both on Windows, Alpine linux, Debian and Fedora ?

tomolimo:
> I would love to use agnostic tool for this. Do you know a tool that works both on Windows, Alpine linux, Debian and Fedora ?

Currently I failed finding a patch tool for Windows...

edgardmessias:
We can use `git apply`

cedric-anne:
> We can use `git apply`

Not sure of this, if dependency was fetched using dist, then there is no git index and patch may not be applied.

tomolimo:
I found a way: I've used the WSL2 to run the `composer install` command

edgardmessias:
I do know if this work, but I will test it:
https://github.com/netresearch/composer-patches-plugin
https://github.com/vaimo/composer-patches

trasher (approved):

PR #8594, opened 2021-01-28
Fix kanban

stonebuzz:
First problem

The bulk and individual add form uses the same css class ```kanban-add-form```
and the javascript code that add only one task is triggered during bulk add

Second problem
add task or sub project project from kanban view, use current active entity, rather than the problem entity

This PR fix this

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | internal 21417

### DIFF
--- a/inc/project.class.php
+++ b/inc/project.class.php
@@ -2385,7 +2385,7 @@ class Project extends CommonDBTM implements ExtraVisibilityCriteria {
                ],
                'entities_id' => [
                   'type'   => 'hidden',
-                  'value'  => $_SESSION['glpiactive_entity']
+                  'value'  => $project->getField("entities_id"),
                ],
                'is_recursive' => [
                   'type'   => 'hidden',
@@ -2420,7 +2420,7 @@ class Project extends CommonDBTM implements ExtraVisibilityCriteria {
                ],
                'entities_id' => [
                   'type'   => 'hidden',
-                  'value'  => $_SESSION['glpiactive_entity']
+                  'value'  => $project->getField("entities_id"),
                ],
                'is_recursive' => [
                   'type'   => 'hidden',
--- a/inc/project.class.php
+++ b/inc/project.class.php
@@ -2385,7 +2385,7 @@ class Project extends CommonDBTM implements ExtraVisibilityCriteria {
                ],
                'entities_id' => [
                   'type'   => 'hidden',
-                  'value'  => isset($project->fields["entities_id"]) ? $project->fields["entities_id"] : 0, //undefined in global kanban but manage by 'supported_itemtypes'
+                  'value'  => $ID > 0 ? $project->fields["entities_id"] : $_SESSION['glpiactive_entity'],
                ],
                'is_recursive' => [
                   'type'   => 'hidden',
@@ -2420,7 +2420,7 @@ class Project extends CommonDBTM implements ExtraVisibilityCriteria {
                ],
                'entities_id' => [
                   'type'   => 'hidden',
-                  'value'  => isset($project->fields["entities_id"]) ? $project->fields["entities_id"] : 0, //undefined in global kanban but manage by 'supported_itemtypes'
+                  'value'  => $ID > 0 ? $project->fields["entities_id"] : $_SESSION['glpiactive_entity'],
                ],
                'is_recursive' => [
                   'type'   => 'hidden',
@@ -2448,7 +2448,7 @@ class Project extends CommonDBTM implements ExtraVisibilityCriteria {
 
       echo "<div id='kanban' class='kanban'></div>";
       $darkmode = ($_SESSION['glpipalette'] === 'darker') ? 'true' : 'false';
-      $canadd_item = json_encode($project->canEdit($ID) && $project->can($ID, UPDATE));
+      $canadd_item = json_encode($ID > 0 ? $project->canEdit($ID) && $project->can($ID, UPDATE) : self::canCreate() || ProjectTask::canCreate());
       $canmodify_view = json_encode(($ID == 0 || $project->canModifyGlobalState()));
       $cancreate_column = json_encode((bool)ProjectState::canCreate());
       $limit_addcard_columns = $canmodify_view !== 'false' ? '[]' : json_encode([0]);
--- a/inc/project.class.php
+++ b/inc/project.class.php
@@ -2448,7 +2448,7 @@ class Project extends CommonDBTM implements ExtraVisibilityCriteria {
 
       echo "<div id='kanban' class='kanban'></div>";
       $darkmode = ($_SESSION['glpipalette'] === 'darker') ? 'true' : 'false';
-      $canadd_item = json_encode(self::canCreate() || ProjectTask::canCreate());
+      $canadd_item = json_encode($project->canEdit($ID) && $project->can($ID, UPDATE));
       $canmodify_view = json_encode(($ID == 0 || $project->canModifyGlobalState()));
       $cancreate_column = json_encode((bool)ProjectState::canCreate());
       $limit_addcard_columns = $canmodify_view !== 'false' ? '[]' : json_encode([0]);
--- a/inc/project.class.php
+++ b/inc/project.class.php
@@ -2385,7 +2385,7 @@ class Project extends CommonDBTM implements ExtraVisibilityCriteria {
                ],
                'entities_id' => [
                   'type'   => 'hidden',
-                  'value'  => $project->getField("entities_id"),
+                  'value'  => isset($project->fields["entities_id"]) ? $project->fields["entities_id"] : 0, //undefined in global kanban but manage by 'supported_itemtypes'
                ],
                'is_recursive' => [
                   'type'   => 'hidden',
@@ -2420,7 +2420,7 @@ class Project extends CommonDBTM implements ExtraVisibilityCriteria {
                ],
                'entities_id' => [
                   'type'   => 'hidden',
-                  'value'  => $project->getField("entities_id"),
+                  'value'  => isset($project->fields["entities_id"]) ? $project->fields["entities_id"] : 0, //undefined in global kanban but manage by 'supported_itemtypes'
                ],
                'is_recursive' => [
                   'type'   => 'hidden',
--- a/js/kanban.js
+++ b/js/kanban.js
@@ -1469,7 +1469,6 @@
          add_form += form_header;
 
          add_form += "<div class='kanban-item-content'>";
-         add_form += "<textarea name='bulk_item_list'></textarea>";
          $.each(self.supported_itemtypes[itemtype]['fields'], function(name, options) {
             var input_type = options['type'] !== undefined ? options['type'] : 'text';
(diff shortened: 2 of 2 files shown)


js/kanban.js
```diff
@@ -1461,7 +1461,7 @@
 
          var uniqueID = Math.floor(Math.random() * 999999);
          var formID = "form_add_" + itemtype + "_" + uniqueID;
-         var add_form = "<form id='" + formID + "' class='kanban-add-form kanban-form no-track'>";
+         var add_form = "<form id='" + formID + "' class='kanban-bulk-add-form kanban-form no-track'>";
```
cedric-anne, line 1464: I see no occurence of `kanban-bulk-add-form` in current code. Are you sure this does not break some JS binding ?

cconard96:
Adding a Project in the Global Kanban now places it in the root entity instead of the current.
It seems the global Kanban ignores entity restrictions and would show that new Project, but you get an access denied error when trying to open it.

js/kanban.js
cconard96: kanban-add-form is used as a selector to preserve new item forms during a column refresh. Without it, bulk item forms are removed during the background refresh.
Perhaps the selector for the event for adding single items needs to be changed instead of this.

stonebuzz:
> Adding a Project in the Global Kanban now places it in the root entity instead of the current.
> It seems the global Kanban ignores entity restrictions and would show that new Project, but you get an access denied error when trying to open it.

maybe we should have the same behaviour as the 'project task' tab, i.e. hide the 'add' button if the user is in a sub-entity.

image

js/kanban.js
stonebuzz: ```preserveNewItemForms``` updated to match with new class ```.kanban-bulk-add-form
```

File: js/kanban.js [hunk 2]
```diff
@@ -371,7 +371,7 @@
          self.temp_forms = {};
          var columns = $(self.element + " .kanban-column");
          $.each(columns, function(i, column) {
-            var forms = $(column).find('.kanban-add-form');
+            var forms = $(column).find('.kanban-add-form').find('.kanban-bulk-add-form');
```
cedric-anne, line 374: Should it be this ?
```suggestion
            var forms = $(column).find('.kanban-add-form, .kanban-bulk-add-form');
```

stonebuzz: I have doubt about the comma -> https://stackoverflow.com/a/39565294

stonebuzz:
> > Adding a Project in the Global Kanban now places it in the root entity instead of the current.
> > It seems the global Kanban ignores entity restrictions and would show that new Project, but you get an access denied error when trying to open it.
> 
> maybe we should have the same behaviour as the 'project task' tab, i.e. hide the 'add task' button if the user is in a sub-entity.
> 
> image
> 
> Idem for add sub project from kanban view, with 'project' tab if user in a sub-entity it's impossible to add sub project

Shouldn't we remove the possibility to add a task or sub-project from the kanban view when the project is viewed from a sub-entity?

As the current behaviour of the "Project tasks" and "Sub-project" tabs

stonebuzz:
I update ```canadd_item``` check 

I have replaced the old checks by those used to add a sub-project or a task in order to have the same behaviour as the tabs.

js/kanban.js
cedric-anne: `$(column).find('.kanban-add-form').find('.kanban-bulk-add-form');` will return all `'.kanban-bulk-add-form'` that are childs of `'.kanban-add-form'`, which correspond to the CSS selector `.kanban-add-form .kanban-bulk-add-form`

`$(column).find('.kanban-add-form, .kanban-bulk-add-form');` will return all `'.kanban-bulk-add-form'` and `'.kanban-add-form'`, which correspond to the CSS selector `.kanban-add-form, .kanban-bulk-add-form`

cconard96: The case with the comma works as intended for me. Chaining finds does not because the `kanban-add-form` class isn't on the `.kanban-bulk-add-form` elements or as a parent.

cconard96 (changes requested):
There are still issues with the bulk form disappearing after a background refresh and the styles on the bulk form aren't working.
I propose the following changes to add the `kanban-bulk-add-form` class to the bulk form in addition to the `kanban-add-form` class to maintain compatibility with the styles, background refresh preservation, and maybe more. We should just exclude the bulk form from the selector on the single item submit handler. This could be cleaned up during the refactor in the next version with less risk of breakage.

```diff
diff --git a/js/kanban.js b/js/kanban.js
index fc859a41e7..f0f649db1b 100644
--- a/js/kanban.js
+++ b/js/kanban.js
@@ -371,7 +371,7 @@
          self.temp_forms = {};
          var columns = $(self.element + " .kanban-column");
          $.each(columns, function(i, column) {
-            var forms = $(column).find('.kanban-add-form').find('.kanban-bulk-add-form');
+            var forms = $(column).find('.kanban-add-form');
             if (forms.length > 0) {
                self.temp_forms[column.id] = [];
                $.each(forms, function(i2, form) {
@@ -704,7 +704,7 @@
             );
          }
 
-         $(self.element + ' .kanban-container').on('submit', '.kanban-add-form', function(e) {
+         $(self.element + ' .kanban-container').on('submit', '.kanban-add-form:not(.kanban-bulk-add-form)', function(e) {
             e.preventDefault();
             var form = $(e.target);
             var data = {};
@@ -1461,7 +1461,7 @@
 
          var uniqueID = Math.floor(Math.random() * 999999);
          var formID = "form_add_" + itemtype + "_" + uniqueID;
-         var add_form = "<form id='" + formID + "' class='kanban-bulk-add-form kanban-form no-track'>";
+         var add_form = "<form id='" + formID + "' class='kanban-add-form kanban-bulk-add-form kanban-form no-track'>";
          var form_header = "<div class='kanban-item-header'>";
          form_header += "<span class='kanban-item-title'>"+self.supported_itemtypes[itemtype]['name']+"</span>";
          form_header += "<i class='fas fa-times' title='Close' onclick='$(this).parent().parent().remove()'></i>";
```

stonebuzz:
> There are still issues with the bulk form disappearing after a background refresh and the styles on the bulk form aren't working.
> I propose the following changes to add the `kanban-bulk-add-form` class to the bulk form in addition to the `kanban-add-form` class to maintain compatibility with the styles, background refresh preservation, and maybe more. We should just exclude the bulk form from the selector on the single item submit handler. This could be cleaned up during the refactor in the next version with less risk of breakage.
> 
> ```diff
> diff --git a/js/kanban.js b/js/kanban.js
> index fc859a41e7..f0f649db1b 100644
> --- a/js/kanban.js
> +++ b/js/kanban.js
> @@ -371,7 +371,7 @@
>           self.temp_forms = {};
>           var columns = $(self.element + " .kanban-column");
>           $.each(columns, function(i, column) {
> -            var forms = $(column).find('.kanban-add-form').find('.kanban-bulk-add-form');
> +            var forms = $(column).find('.kanban-add-form');
>              if (forms.length > 0) {
>                 self.temp_forms[column.id] = [];
>                 $.each(forms, function(i2, form) {
> @@ -704,7 +704,7 @@
>              );
>           }
>  
> -         $(self.element + ' .kanban-container').on('submit', '.kanban-add-form', function(e) {
> +         $(self.element + ' .kanban-container').on('submit', '.kanban-add-form:not(.kanban-bulk-add-form)', function(e) {
>              e.preventDefault();
>              var form = $(e.target);
>              var data = {};
> @@ -1461,7 +1461,7 @@
>  
>           var uniqueID = Math.floor(Math.random() * 999999);
>           var formID = "form_add_" + itemtype + "_" + uniqueID;
> -         var add_form = "<form id='" + formID + "' class='kanban-bulk-add-form kanban-form no-track'>";
> +         var add_form = "<form id='" + formID + "' class='kanban-add-form kanban-bulk-add-form kanban-form no-track'>";
>           var form_header = "<div class='kanban-item-header'>";
>           form_header += "<span class='kanban-item-title'>"+self.supported_itemtypes[itemtype]['name']+"</span>";
>           form_header += "<i class='fas fa-times' title='Close' onclick='$(this).parent().parent().remove()'></i>";
> ```

Fixed

cconard96 (approved):

trasher (approved):

orthagh (approved):

inc/project.class.php
```diff
@@ -2420,7 +2420,7 @@ static function showKanban($ID) {
                ],
                'entities_id' => [
                   'type'   => 'hidden',
-                  'value'  => $_SESSION['glpiactive_entity']
+                  'value'  => $project->getField("entities_id"),
```
cedric-anne: In the global Kanban, value is not correct:
image

cedric-anne (changes requested):

inc/project.class.php
stonebuzz: fixed

inc/project.class.php
```diff
@@ -2385,7 +2385,7 @@ static function showKanban($ID) {
                ],
                'entities_id' => [
                   'type'   => 'hidden',
-                  'value'  => $_SESSION['glpiactive_entity']
+                  'value'  => isset($project->fields["entities_id"]) ? $project->fields["entities_id"] : 0, //undefined in global kanban but manage by 'supported_itemtypes'
```
cedric-anne, line 2388:
```suggestion
                  'value'  => $ID > 0 ? $project->fields["entities_id"] : $_SESSION['glpiactive_entity'],
```

inc/project.class.php
```diff
@@ -2420,7 +2420,7 @@ static function showKanban($ID) {
...
```
cedric-anne, line 2423:
```suggestion
                  'value'  => $ID > 0 ? $project->fields["entities_id"] : $_SESSION['glpiactive_entity'],
```

inc/project.class.php
```diff
@@ -2448,7 +2448,7 @@ static function showKanban($ID) {
 
       echo "<div id='kanban' class='kanban'></div>";
       $darkmode = ($_SESSION['glpipalette'] === 'darker') ? 'true' : 'false';
-      $canadd_item = json_encode(self::canCreate() || ProjectTask::canCreate());
+      $canadd_item = json_encode($project->canEdit($ID) && $project->can($ID, UPDATE));
```
cedric-anne, line 2451:
```suggestion
      $canadd_item = json_encode($ID > 0 ? $project->canEdit($ID) && $project->can($ID, UPDATE) : self::canCreate() || ProjectTask::canCreate());
```

inc/project.class.php
stonebuzz: done

inc/project.class.php
stonebuzz: done

inc/project.class.php
stonebuzz: done

cedric-anne (approved):

PR #8666, opened 2021-02-15
Do no delete default profiles if no rules were applied

AdrienClairembault:
When users have only one dynamic profiles that match the default profile of GLPI, this profile should not be deleted by the right rules engine if no rules match this user.

Concrete example (!21527) :  
- An user is created through the API
- The default profile is auto affected as dyamic
- There is no right rules targeting this user
-> When this user log in, he should keep the current profile (currently the profile is removed and he can't login).

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

### DIFF
--- a/inc/profile_user.class.php
+++ b/inc/profile_user.class.php
@@ -100,12 +100,6 @@ class Profile_User extends CommonDBRelation {
          return false;
       }
 
-      if (isset($input['profiles_id']) && !empty($input['profiles_id'])) {
-         if ($input['profiles_id'] == Profile::getDefault()) {
-            $input['is_default_profile'] = true;
-         }
-      }
-
       return parent::prepareInputForAdd($input);
    }
 
--- a/inc/profile_user.class.php
+++ b/inc/profile_user.class.php
@@ -99,6 +99,13 @@ class Profile_User extends CommonDBRelation {
                                           false, ERROR);
          return false;
       }
+
+      if (isset($input['profiles_id']) && !empty($input['profiles_id'])) {
+         if ($input['profiles_id'] == Profile::getDefault()) {
+            $input['is_default_profile'] = true;
+         }
+      }
+
       return parent::prepareInputForAdd($input);
    }
 
--- a/inc/user.class.php
+++ b/inc/user.class.php
@@ -679,6 +679,7 @@ class User extends CommonDBTM {
             $profile                   = Profile::getDefault();
             // Default right as dynamic. If dynamic rights are set it will disappear.
             $affectation['is_dynamic'] = 1;
+            $affectation['is_default_profile'] = 1;
          }
 
          if ($profile) {
@@ -1052,10 +1053,12 @@ class User extends CommonDBTM {
             $return = true;
          } else if (count($dynamic_profiles) == 1) {
             $dynamic_profile = reset($dynamic_profiles);
-            $default_profile = Profile::getDefault();
+
             // If no rule applied and only one dynamic profile found, check if
             // it is the default profile
             if ($dynamic_profile['is_default_profile'] == true) {
+               $default_profile = Profile::getDefault();
+
                // Remove from to be deleted list
                $dynamic_profiles = [];
 
--- a/inc/user.class.php
+++ b/inc/user.class.php
@@ -1050,6 +1050,25 @@ class User extends CommonDBTM {
             unset($this->input["_ldap_rules"]);
 
             $return = true;
+         } else if (count($dynamic_profiles) == 1) {
+            $dynamic_profile = reset($dynamic_profiles);
+            $default_profile = Profile::getDefault();
+            // If no rule applied and only one dynamic profile found, check if
+            // it is the default profile
+            if ($dynamic_profile['is_default_profile'] == true) {
+               // Remove from to be deleted list
+               $dynamic_profiles = [];
+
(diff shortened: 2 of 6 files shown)


trasher (commented):
Sounds good, tests should be added

tests/units/RuleRight.php
```diff
@@ -152,4 +152,35 @@ public function testLocalAccount() {
...
+
+   public function testLocalAccountNoRules() {
+      $testuser = [
+         'name'      => 'testuser',
+         'password'  => 'test',
+         'password2' => 'test',
+      ];
+
+      $user = new \User();
+      $users_id = $user->add($testuser);
```
trasher, line 164:
```suggestion
      $users_id = $user->add($testuser);
      $this->integer($users_id)->isGreaterThan(0);
```

trasher (approved):

AdrienClairembault:
Rebased + forcepushed since my 9.5 branches was a bit too old.

As we discussed this morning, added a new "is_default_profile" field to handle case where the default profile was modified.

inc/profile_user.class.php
```diff
@@ -99,6 +99,13 @@ function prepareInputForAdd($input) {
                                           false, ERROR);
          return false;
       }
+
+      if (isset($input['profiles_id']) && !empty($input['profiles_id'])) {
+         if ($input['profiles_id'] == Profile::getDefault()) {
+            $input['is_default_profile'] = true;
+         }
+      }
```
cedric-anne, line 107: IMHO, this `is_default_profile` flag should only be added from the current `User::post_addItem()`. Indeed if this profile is added manually or via right rules, then this profile would have been explicitely added, and then should not be considered as a default value.

inc/user.class.php
```diff
@@ -1050,6 +1050,25 @@ function applyRightRules() {
             unset($this->input["_ldap_rules"]);
 
             $return = true;
+         } else if (count($dynamic_profiles) == 1) {
+            $dynamic_profile = reset($dynamic_profiles);
+            $default_profile = Profile::getDefault();
```
cedric-anne, line 1055: This should be moved inside the `if ($dynamic_profile['is_default_profile'] == true) {` block, to prevent an useless query if this condition is not reached.

tests/units/RuleRight.php
```diff
@@ -152,4 +152,63 @@ public function testLocalAccount() {
...
+
+      // Get rights again, should not have changed
+      $pu = \Profile_User::getForUser($users_id, true);
+      $this->array($pu)->hasSize(1);
+      $right2 = array_pop($pu);
+      $this->array($right2)->isEqualTo($right);
+
+      $_SESSION['glpiactiveprofile']['user'] = 7327;
+      $_SESSION['glpiactiveprofile']['profile'] = 23;
+
```
cedric-anne, line 187:
```suggestion
```

inc/user.class.php
```diff
@@ -1050,6 +1050,25 @@ function applyRightRules() {
...
+               $dynamic_profiles = [];
+
+               // Update profile if need to match the current default profile
+               if ($dynamic_profile['profiles_id'] !== $default_profile) {
+                  $pu = new Profile_User();
+                  $dynamic_profile['profiles_id'] = $default_profile;
+                  $pu->add($dynamic_profile);
+                  $pu->delete([
+                     'id' => $dynamic_profile['id']
+                  ]);
```
cedric-anne, line 1069: Why did you not use an update instead ?

tests/units/RuleRight.php
AdrienClairembault: Does not work without, seems like there is some rights checks inside the DB engine for this case :/

inc/user.class.php
AdrienClairembault: Update doesn't work due to some hidden right checks (same is done inside `applyRightRules`,  it doesn't update anything, only remove and add).

tests/units/RuleRight.php
cedric-anne: Where does thes values comes ?

AdrienClairembault: empty_data.php, profiles rights for a super admin:
image

cedric-anne: Use `$this->login();` to login with TU_USER which has these rights, instead of forcing value.

AdrienClairembault: I've tried it before and it didn't work but it seemed weird since as you said TU_USER have the two rights that I was adding.
I just found out that you also need to be the one that created the user to be allowed to edit his profiles, so adding a second `$this->login();` before the creating the user at line 164 make it works.

Is it normal to have these rights issues ?
Shouldn't rights be managed in an upper layer (front files, controllers, ...) and not intervene in this kind of situations ?

cedric-anne: > I just found out that you also need to be the one that created the user to be allowed to edit his profiles

Are you sure of that ?

I think it was failing before because, and it is normal, a user cannot add itself a profile that gives him more rights. As in your test, you were connected as the tested user `$this->login($testuser['name'], $testuser['password'], false);`, it is normal that you were not able to add a profile for this user.

cedric-anne (approved):

orthagh (approved):

PR #8688, opened 2021-02-23
API: Add 'with_networkports' option to getItems endpoint

AdrienClairembault:
Option `with_networkports` was only available on the `getItem` endpoint, I've moved the code to a function and called it from `getItem` AND `getItems`.

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !21585


inc/api/api.class.php
```diff
@@ -2798,6 +2644,187 @@ protected function getFriendlyNames(
       return $_names;
    }
 
+   /**
+    * Get network ports
+    *
+    * @since 9.5
```
cedric-anne, line 2650:
```suggestion
    * @since 9.5.4
```

cedric-anne (commented):
`inc/api/api.class.php.orig` should be removed.

inc/api/api.class.php
trasher: x.x.x if we target master

apirest.md
```diff
@@ -605,6 +605,7 @@ Note: To download a document see [Download a document file](#download-a-document
   * *searchText* (default NULL): array of filters to pass on the query (with key = field and value the text to search)
   * *is_deleted* (default: false): Return deleted element. Optional.
   * *add_keys_names*: Retrieve friendly names. Array containing fkey(s) and/or "id". Optional.
+  * *with_networkports*: Retrieve all network's connections and advanced network's informations. Optional.
```
trasher, line 608:
```suggestion
  * *with_networkports*: Retrieve all network connections and advanced information. Optional.
```

inc/api/api.class.php
```diff
@@ -1158,6 +999,7 @@ protected function arrayRightError() {
     * - 'searchText'       (default: NULL): array of filters to pass on the query (with key = field and value the search)
     * - 'is_deleted'       (default: false): show trashbin. Optionnal
     * - 'add_keys_names'   (default: []): insert raw name(s) for given itemtype(s) and fkey(s)
+    * - 'with_networkports':Retrieve all network connections and advanced network informations. Optionnal.
```
trasher, line 1002:
```suggestion
    * - 'with_networkports':Retrieve all network connections and advanced information. Optionnal.
```

inc/api/api.class.php
```diff
@@ -2798,6 +2644,187 @@ protected function getFriendlyNames(
...
+    * @param array  $data
+    * @param array  $params         API parameters
+    * @param string $self_itemtype  Itemtype the API was called on
+    *
+    * @return array
+    */
+   protected function getNetworkPorts(
+      int $id,
+      string $itemtype
+   ) {
```
trasher, line 2661:
```suggestion
   ): array {
```

inc/api/api.class.php
```diff
@@ -2798,6 +2644,187 @@ protected function getFriendlyNames(
...
+      global $DB;
+
+      $_networkports = [];
+
+      if (!NetworkEquipment::canView()) {
+         $_networkports = $this->arrayRightError();
+      } else {
+         foreach (NetworkPort::getNetworkPortInstantiations() as $networkport_type) {
```
trasher, line 2669: Probably better to store `getNetworkPortInstantiations` result in a variable used for iteration

AdrienClairembault:
> 
> 
> `inc/api/api.class.php.orig` should be removed.

Went a bit foo fast when committing ;)
Should be all good now!

apirest.md
AdrienClairembault: Ok, I've updated others places where this args was documented so we have the same everywhere.

inc/api/api.class.php
AdrienClairembault: Is it a bad practice to use a function result as a loop ? 
It feels more readable that way.

(I didn't write this, just moved the code).

trasher: Doing this can lower performances; depending on what the called method is really doing (I'm not sure why, but this will kind of call the method at each iteration).

cedric-anne (approved):

inc/api/api.class.php
AdrienClairembault: Maybe you are thinking of this ? 

image

I've changed it anyway.

cedric-anne (approved):

trasher (approved):

Issue #8595, opened 2021-01-28
9.5.3 - Invisible icons on main tab of user - darker mode 


btry:
Hi 

The icons exist, but their foreground color is probably the same as the background color. Trying to hover them reveals their existence.

![image](https://user-images.githubusercontent.com/14139801/106147653-6990d000-6178-11eb-9712-e67c74797f7b.png)

![image](https://user-images.githubusercontent.com/14139801/106147708-7e6d6380-6178-11eb-8f71-0a69650463db.png)

PR #8815, opened 2021-03-11
Fix some darker theme and general style issues

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #8595

Fix issue where the box around the notes field could overflow the form area. Removing the static width fixes the issue.
Fixed button/icon colors in the form header like the user vCard and impersonate buttons.
Fixed issue where the timeline filter was nearly invisible unless you hovered over it.
Fixed the color of the box around the note field to better match the darker theme and not a light theme.

### DIFF
--- a/css/palettes/_darker.scss
+++ b/css/palettes/_darker.scss
@@ -436,6 +436,10 @@ body #page .mce-btn-group:not(:first-child) {
    color: #262626;
 }
 
+.timeline_history .filter_timeline {
+   opacity: 1;
+}
+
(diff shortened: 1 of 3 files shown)


cedric-anne (approved):
I added a commit to fix note box overflowing. Your solution was not preventing the overflow, at least, on Chrome.

trasher (approved):

PR #8826, opened 2021-03-12
Newlines are removed in preformatted text

falco2001:
Email signatures are in text format (15 perfect lines). They appear on the same line in the follow-ups making it impossible to read the follow-up!
Since the transition to rich text, it is no longer possible to modify the content and description of the notificationtargetcommonitilobject.class file because the effects on the operation of GLPI are significant.

I did the following tests with a signature of 15 lines! :

- ticket creation by mail tool -> the signature is displayed as a block, it's ok all the same
- follow-up by email -> ok now signature ok follow-up with the follow-up before that is displayed

### DIFF
--- a/inc/mailcollector.class.php
+++ b/inc/mailcollector.class.php
@@ -1232,7 +1232,7 @@ class MailCollector  extends CommonDBTM {
       // Replace plain text line breaks to marker if content is not html
       // and rich text mode is enabled (otherwise remove them)
       $string = str_replace(
-         ["\r\n", "\n", "\r"],
+         ["\r\n", "\n\n", "\r"],
          $this->body_is_html ? ' ' : $br_marker,
          $string
       );


trasher:
Hi, thank you for your contribution.

If your PR is related to an existing ticket, please refer to it in your commit message; it'll be closed automatically if the PR is merged.

Also, tests are currently broken, they must be fixed; and we may need an new test case for that change.

cedric-anne (changes requested):
This fix does not seems correct to me. With the current state of code, all newlines should be handled the same way.
If there is a rendering issue on notifications, the then fix has to be applied on the rendering part.
If the rendering part cannot be patched, the the fix has to be more global, in order to fix all places where this is rendered.

trasher:
Ne feedback from a while; I close

falco2001:
Hello,
with that part of the code. This has been working for 1 month without too much problem. Of course, the old tickets have imperfections, but the new ones with 10 lines signature everything is ok whether it is in email, in webmail, in creation of closed ticker and in response

```

Ligne 1243 de  mailcollector.class.php 
 // Replace plain text line breaks to marker if content is not html
 // and rich text mode is enabled (otherwise remove them)
      
      $string = str_replace(
        // Avant  ["\r\n", "\n", "\r"],
         ["\r\n", "\n\n", "\r"], //YD
         $this->body_is_html ? ' ' : $br_marker,
         $string
      );
      $string = str_replace(["\n"],$br_marker,$string); //YD


```

PR #8850, opened 2021-03-18
UI/timeline2

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

### DIFF
--- a/ajax/actors.php
+++ b/ajax/actors.php
@@ -39,6 +39,7 @@ switch ($_REQUEST['action']) {
    case "getActors":
       header("Content-Type: application/json; charset=UTF-8");
       Html::header_nocache();
+      session_write_close(); // don'l lock session to permits parallel calls
       echo Dropdown::getDropdownActors($_POST);
       break;
 }
--- a/ajax/timeline.php
+++ b/ajax/timeline.php
@@ -30,6 +30,8 @@
  * ---------------------------------------------------------------------
  */
 
+use Glpi\Application\View\TemplateRenderer;
+
 include ('../inc/includes.php');
 
 Session::checkLoginUser();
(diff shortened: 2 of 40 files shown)


orthagh:
Some issues i found while testing : 

 - Some fields in task doesn't seems to have `'icon_label': true,` option like `state`, `duration`  and it broke alignment of the form
image

- A class is missing when adding a new task: the yellow background is missing
image

- task.duration icon is not the good one, it should be `fa-stopwatch`
image

- for user mention feature, we miss the css. see : https://github.com/glpi-project/glpi/pull/8751/files#diff-76628e774773cc79d896ef83f3617a00239f1a6a477a0b537f6612de5dc7e4a6R350-R361
  The css rule should be moved into a separate file included in `css/style.scss`

- the can add document should be removed completely i think, we though the "add document" action is useless. (Display/"edit of existing" actions should be kept)
  
- there is on the `feat/modern-ui` branch deprecated calls of `Html::setRichTextContent` polluting the display of timeline. I asked @cedric-anne to fix this.

-  Maybe due to the last point, i couldn't check display of documents in timeline. But a rapid read shows you still have raw code 
 (`Sample Image.png`, `Screenshot1.png`)

cconard96:
> * Some fields in task doesn't seems to have `'icon_label': true,` option like `state`, `duration`  and it broke alignment of the form
Fixed

> * A class is missing when adding a new task: the yellow background is missing
Fixed. The CSS uses "ITILTask" so when I added the Timeline feature and switched to using the actual class "CommonITILTask", it broke.

> * for user mention feature, we miss the css. see : https://github.com/glpi-project/glpi/pull/8751/files#diff-76628e774773cc79d896ef83f3617a00239f1a6a477a0b537f6612de5dc7e4a6R350-R361
I just started adding this and then realized it is already included in `css/includes/components/richtext.scss`

> * the can add document should be removed completely i think, we though the "add document" action is useless. (Display/"edit of existing" actions should be kept)
I'm not sure what you mean about this.

> * i couldn't check display of documents in timeline. But a rapid read shows you still have raw code
That template wasn't meant to be included. I was prototyping UX for adding new documents to the timeline.

PR #8852, opened 2021-03-18
Fix URL validation

AdrienClairembault:
New regex introduced in #7733 is a bit too strict: URL containing "?" are considered to be not valid (so any URL with a query string would be marked as invalid).
 
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

### DIFF
--- a/inc/toolbox.class.php
+++ b/inc/toolbox.class.php
@@ -3488,8 +3488,9 @@ HTML;
       // Verify absence of known disallowed characters.
       // It is still possible to have false positives, but a fireproof check would be too complex
       // (or would require usage of a dedicated lib).
+      // return filter_var($url, FILTER_VALIDATE_URL) !== false;
       return (preg_match(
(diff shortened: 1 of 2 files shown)


trasher (approved):

inc/toolbox.class.php
```diff
@@ -3488,8 +3488,9 @@ public static function isValidWebUrl($url): bool {
       // Verify absence of known disallowed characters.
       // It is still possible to have false positives, but a fireproof check would be too complex
       // (or would require usage of a dedicated lib).
+      // return filter_var($url, FILTER_VALIDATE_URL) !== false;
```
orthagh, line 3491: comment to clean ?

cedric-anne (approved):

inc/toolbox.class.php
trasher, line 3491:
```suggestion
```

PR #8947, opened 2021-04-09
Add link between Contrat and Tickets

AdrienClairembault:
Added a new "Contracts" tab on tickets: 
image

Added a new "Tickets" tab on contracts:
image

Added a new "Contract" field on the tickets creation form:
image

This field support hidden/predefined/mandatory values in templates: 
image

Added two new search options for tickets:
image

Added a new ticket rule criteria (contract type):
image

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !21598

### DIFF
--- a/front/ticketcontract.form.php
+++ b/front/ticketcontract.form.php
@@ -1,0 +1,63 @@
+<?php
+/**
+ * ---------------------------------------------------------------------
+ * GLPI - Gestionnaire Libre de Parc Informatique
+ * Copyright (C) 2015-2021 Teclib' and contributors.
+ *
+ * http://glpi-project.org
+ *
+ * based on GLPI - Gestionnaire Libre de Parc Informatique
+ * Copyright (C) 2003-2014 by the INDEPNET Development Team.
+ *
+ * ---------------------------------------------------------------------
+ *
+ * LICENSE
+ *
+ * This file is part of GLPI.
+ *
+ * GLPI is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * GLPI is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with GLPI. If not, see <http://www.gnu.org/licenses/>.
+ * ---------------------------------------------------------------------
+ */
+
+include ('../inc/includes.php');
+
+Session ::checkLoginUser();
+
+$item = new TicketContract();
+
+if (isset($_POST["add"])) {
+   if (!empty($_POST['tickets_id']) && empty($_POST['contracts_id'])) {
+      $message = sprintf(
+         __('Mandatory fields are not filled. Please correct: %s'),
+         Contract::getTypeName(1)
+      );
+      Session::addMessageAfterRedirect($message, false, ERROR);
+      Html::back();
+   }
+   if (empty($_POST['tickets_id']) && !empty($_POST['contracts_id'])) {
+      $message = sprintf(
+         __('Mandatory fields are not filled. Please correct: %s'),
+         Ticket::getTypeName(1)
+      );
+      Session::addMessageAfterRedirect($message, false, ERROR);
+      Html::back();
+   }
+   $item->check(-1, CREATE, $_POST);
+   $item->add($_POST);
+
+   Html::back();
+
+}
+
+Html::displayErrorAndDie("lost");
--- a/inc/commonitilobject.class.php
+++ b/inc/commonitilobject.class.php
@@ -6323,7 +6323,12 @@ abstract class CommonITILObject extends CommonDBTM {
                                  $align_desc." width='150'");
          } else {
             echo Search::showItem($p['output_type'], $name_column, $item_num, $p['row_num'], $align_desc." width='200'");
-            $takeintoaccountdelay_column = Html::timestampToString($item->fields['takeintoaccount_delay_stat']);
+
+            $takeintoaccountdelay_column = "";
+            // Show only for tickets taken into account
+            if ($item->fields['takeintoaccount_delay_stat'] > 0) {
+               $takeintoaccountdelay_column = Html::timestampToString($item->fields['takeintoaccount_delay_stat']);
+            }
             echo Search::showItem($p['output_type'], $takeintoaccountdelay_column, $item_num, $p['row_num'], $align_desc." width='150'");
 
             $solvedelay_column = "";
--- a/inc/commonitilobject.class.php
+++ b/inc/commonitilobject.class.php
@@ -6044,6 +6044,8 @@ abstract class CommonITILObject extends CommonDBTM {
          'row_num'                => 0,
          'type_for_massiveaction' => 0,
          'id_for_massiveaction'   => 0,
+         'followups'              => false,
+         'ticket_stats'           => false,
       ];
 
       if (count($options)) {
@@ -6197,56 +6199,14 @@ abstract class CommonITILObject extends CommonDBTM {
 
          echo Search::showItem($p['output_type'], $fifth_col, $item_num, $p['row_num'], $align);
 
-         // Sixth Colum
-         // Ticket : simple link to item
-         $sixth_col  = "";
-         $is_deleted = false;
-         $item_ticket = new Item_Ticket();
-         $data = $item_ticket->find(['tickets_id' => $item->fields['id']]);
-
-         if ($item->getType() == 'Ticket') {
-            if (!empty($data)) {
-               foreach ($data as $val) {
-                  if (!empty($val["itemtype"]) && ($val["items_id"] > 0)) {
-                     if ($object = getItemForItemtype($val["itemtype"])) {
-                        if ($object->getFromDB($val["items_id"])) {
-                           $is_deleted = $object->isDeleted();
-
-                           $sixth_col .= $object->getTypeName();
-                           $sixth_col .= " - <span class='b'>";
-                           if ($item->canView()) {
-                              $sixth_col .= $object->getLink();
-                           } else {
-                              $sixth_col .= $object->getNameID();
-                           }
-                           $sixth_col .= "</span><br>";
-                        }
-                     }
-                  }
-               }
-            } else {
-               $sixth_col = __('General');
-            }
-
-            echo Search::showItem($p['output_type'], $sixth_col, $item_num, $p['row_num'], ($is_deleted ? " class='center deleted' " : $align));
-         }
-
-         // Seventh column
-         echo Search::showItem($p['output_type'],
-                               "<span class='b'>".
-                                 Dropdown::getDropdownName('glpi_itilcategories',
-                                                           $item->fields["itilcategories_id"]).
-                               "</span>",
-                               $item_num, $p['row_num'], $align);
-
          // Eigth column
-         $eigth_column = "<span class='b'>".$item->getName()."</span>&nbsp;";
+         $name_column = "<span class='b'>".$item->getName()."</span>&nbsp;";
 
          // Add link
          if ($item->canViewItem()) {
-            $eigth_column  = sprintf(
+            $name_column  = sprintf(
                __('%1$s (%2$s)'),
-               "<a id='".$item->getType().$item->fields["id"]."$rand' href=\"".$item->getLinkURL()."\">$eigth_column</a>",
+               "<a id='".$item->getType().$item->fields["id"]."$rand' href=\"".$item->getLinkURL()."\">$name_column</a>",
                sprintf(
                   __('%1$s - %2$s'),
                   $item->numberOfFollowups($showprivate),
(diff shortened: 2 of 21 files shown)


inc/commonitilobject.class.php
```diff
@@ -6263,66 +6222,126 @@ static function showShort($id, $options = []) {
...
-                                    Html::showToolTip($planned_infos,
-                                                      ['display' => false,
-                                                            'applyto' => $item->getType().
-                                                                           $item->fields["id"].
-                                                                           "planning".$rand]));
+            echo Search::showItem($p['output_type'], $tenth_column, $item_num, $p['row_num'],
+                                 $align_desc." width='150'");
+         } else {
+            echo Search::showItem($p['output_type'], $name_column, $item_num, $p['row_num'], $align_desc." width='200'");
+            $takeintoaccountdelay_column = Html::timestampToString($item->fields['takeintoaccount_delay_stat']);
```
cedric-anne, line 6332: It may happen that this value is not yet set, it should be handled.

inc/commonitilobject.class.php
```diff
@@ -6263,66 +6222,126 @@ static function showShort($id, $options = []) {
...
+            echo Search::showItem($p['output_type'], $takeintoaccountdelay_column, $item_num, $p['row_num'], $align_desc." width='150'");
+
+            $solvedelay_column = "";
+            // Show only for solved tickets
+            if ($item->fields['solve_delay_stat'] > 0) {
+               $solvedelay_column = Html::timestampToString($item->fields['solve_delay_stat']);
+            }
+            echo Search::showItem($p['output_type'], $solvedelay_column, $item_num, $p['row_num'], $align_desc." width='150'");
+
+            $waiting_duration_column = Html::timestampToString($item->fields['waiting_duration']);
```
cedric-anne, line 6342: It may happen that this value is not yet set, it should be handled.

inc/commonitilobject.class.php
```diff
@@ -6356,12 +6380,20 @@ static function commonListHeader($output_type = Search::HTML_OUTPUT, $mass_id =
...
+         }
+         $items[__('Category')]           = "glpi_itilcategories.completename";
+         $items[__('Title')]              = "name";
+         $items[__('Planification')]      = "glpi_tickettasks.begin";
+      } else {
+         $items[__('Title')]             = "name";
+         $items[__('Take into account')] = "takeintoaccount_delay_stat";
+         $items[__('Resolution')]        = "solve_delay_stat";
+         $items[__('Pending')]           = "waiting_duration";
       }
```
cedric-anne, line 6396: It is really hard to read/understand the code that produce the list and its headers, and I guess it will be hard to maintain it.
Maybe `$ticket_stats` value should only act on corresponding columns display.

inc/dropdown.class.php
```diff
@@ -2258,7 +2258,6 @@ public static function getDropdownValue($post, $json = true) {
       if (!isset($post['emptylabel']) || ($post['emptylabel'] == '')) {
          $post['emptylabel'] = Dropdown::EMPTY_VALUE;
       }
-
```
cedric-anne, line 2261: Should be reverted.

inc/ticketcontract.class.php
```diff
@@ -0,0 +1,192 @@
...
+
+   public static $itemtype_1 = 'Ticket';
+   public static $items_id_1 = 'tickets_id';
+
+   public static $itemtype_2 = 'Contract';
+   public static $items_id_2 = 'contracts_id';
+   public static $checkItem_2_Rights = self::HAVE_VIEW_RIGHT_ON_ITEM;
+
+   public static function getTypeName($nb = 0) {
+      return _n('fsefesfsfe', 'fsefsefsef', $nb);
```
cedric-anne, line 47: `fsefesfsfe` ? :)

inc/ticketcontract.class.php
```diff
@@ -0,0 +1,192 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with GLPI. If not, see <http://www.gnu.org/licenses/>.
+ * ---------------------------------------------------------------------
+ */
+
+if (!defined('GLPI_ROOT')) {
+   die("Sorry. You can't access this file directly");
+}
+
+class TicketContract extends CommonItilObject_Item {
```
cedric-anne, line 37: Is it necessary to extends `CommonItilObject_Item` instead of `CommonDBRelation``?

cedric-anne, line 37: Class should be named `Ticket_Contract` to follow GLPI conventions.

inc/ticketcontract.class.php
```diff
@@ -0,0 +1,192 @@
...
+         if ($_SESSION['glpishow_count_on_tabs']) {
+            $nb = count(self::getItemsForItem($item));
+         }
+         return self::createTabEntry(Ticket::getTypeName(Session::getPluralNumber()), $nb);
+      } else {
+         return '';
+      }
+   }
+
+   public static function getItemsForItem(CommonDBTM $item) {
```
cedric-anne, line 68: I guess existing `getListForItem` could be used.

install/migrations/update_9.5.x_to_10.0.0/ticketcontract.php
```diff
@@ -0,0 +1,56 @@
...
+$default_collation = DBConnection::getDefaultCollation();
+
+if (!$DB->tableExists('glpi_ticketcontracts')) {
+   $query = "CREATE TABLE `glpi_ticketcontracts` (
+      `id` int NOT NULL AUTO_INCREMENT,
+      `tickets_id` int NOT NULL DEFAULT '0',
+      `contracts_id` int NOT NULL DEFAULT '0',
+      PRIMARY KEY (`id`),
+      UNIQUE KEY `unicity` (`tickets_id`,`contracts_id`),
+      -- KEY `tickets_id` (`tickets_id`),
```
cedric-anne, line 52:
```suggestion
```

install/mysql/glpi-empty.sql
```diff
@@ -8701,4 +8701,15 @@ CREATE TABLE `glpi_manuallinks` (
...
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci ROW_FORMAT=DYNAMIC;
 
+DROP TABLE IF EXISTS `glpi_ticketcontracts`;
+CREATE TABLE `glpi_ticketcontracts` (
+  `id` int NOT NULL AUTO_INCREMENT,
+  `tickets_id` int NOT NULL DEFAULT '0',
+  `contracts_id` int NOT NULL DEFAULT '0',
+  PRIMARY KEY (`id`),
+  UNIQUE KEY `unicity` (`tickets_id`,`contracts_id`),
+  -- KEY `tickets_id` (`tickets_id`),
```
cedric-anne, line 8711:
```suggestion
```

tests/functionnal/MassiveAction.php
```diff
@@ -490,4 +491,100 @@ public function testProcessMassiveActionsForOneItemtype_resolveTickets(
...
+         'name'    => 'test',
+         'content' => 'test',
+      ]);
+      $ticket->getFromDB($id);
+
+      $contract = new Contract();
+      $contract_id = $contract->add([
+         'name'        => 'test',
+         'entities_id' => getItemByTypeName('Entity', '_test_root_entity', true),
+      ]);
```
cedric-anne, line 507: You should check that add operations worked as expected.

tests/functionnal/MassiveAction.php
```diff
@@ -490,4 +491,100 @@ public function testProcessMassiveActionsForOneItemtype_resolveTickets(
...
+   ) {
+      $this->login();
+
+      // Set up session rights
+      $old_session = $_SESSION['glpiactiveprofile'][Ticket::$rightname] ?? 0;
+      if ($has_right) {
+         $_SESSION['glpiactiveprofile'][Ticket::$rightname] = UPDATE;
+      } else {
+         $_SESSION['glpiactiveprofile'][Ticket::$rightname] = 0;
+      }
```
cedric-anne, line 560: Maybe you could rely on existing users (tech, post-only, ...).

front/ticketcontract.form.php
```diff
@@ -0,0 +1,63 @@
...
+      Html::back();
+   }
+   if (empty($_POST['tickets_id']) && !empty($_POST['contracts_id'])) {
+      $message = sprintf(
+         __('Mandatory fields are not filled. Please correct: %s'),
+         Ticket::getTypeName(1)
+      );
+      Session::addMessageAfterRedirect($message, false, ERROR);
+      Html::back();
+   }
```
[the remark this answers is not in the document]
AdrienClairembault: This is pasted code from similar links, you can find it in:
* `change_ticket.form.php`
* `contract_item.form.php`
* `itm_softwarelicense.form.php`
* `item_ticket.form.php`
* `problem_ticket.form.php`
* `ticketcontract.form.php`

If we want to do things differently it should be updated everywhere in another PR (also merging all the code in one place to avoid pasting the same thing over and over would be nice...)

inc/commonitilobject.class.php
AdrienClairembault: Isn't it always set here ?
image

inc/commonitilobject.class.php
AdrienClairembault: I think this one is different that the two above since it is not a distinct "milestone" in the ticket lifecycle but a representation of the ticket spent in one specific state.

It would be bad to display the `solve_delay_stat` value before the ticket was solved because it would read as '0 seconds happened before the ticket was solved' which is not true.

However for `waiting_duration`  it read as 'This ticket has been pending for 0 seconds' which is true.

inc/commonitilobject.class.php
AdrienClairembault: I think I understand what you mean, this issue is that we need to remove some fields to make space (category, planification) but keep the field in the middle of them (title).
I actually think checking column by column would end up more messy.

IMO the real issue is that this method is very hard to override in it's current state and force us to use the $params array to deal with custom code.
Maybe the method should be split in two: one part should be in commondbtm and handle the "framework" (calls to `showNewLine` / `showHeaderItem`) with a new `fields` parameter that would contain the labels to display.
The default fields could be placed in another method.
Do you see what I mean ?

inc/ticketcontract.class.php
AdrienClairembault: Yeah maybe we should not keep that :D 
To my defense I don't think it is actually displayed anywhere so it didn't catch it when testing the feature.

inc/ticketcontract.class.php
AdrienClairembault: I didn't realize `CommonItilObject_Item` was only for items that are `CommonItilObject` (and not items of `CommonItilObject`).
Fixed.

install/migrations/update_9.5.x_to_10.0.0/ticketcontract.php
AdrienClairembault: I kept it because it wanted to discuss it with you.

Unit tests ask me to remove it because they field already exist in `unicity`.
However `contracts_id` is also in `unicity` and this time the unit test ask me to keep my separate key.

Why is it working this way ? IMO we either need both keys or none ?

AdrienClairembault: Nvm, forgot composite index work for the first key.

install/mysql/glpi-empty.sql
```diff
@@ -2711,6 +2711,7 @@ CREATE TABLE `glpi_entities` (
   `autofill_decommission_date` varchar(255) NOT NULL DEFAULT '-2',
   `suppliers_as_private` int NOT NULL DEFAULT '-2',
   `anonymize_support_agents` int NOT NULL DEFAULT '-2',
+  `default_contracts_id` int NOT NULL DEFAULT '-2',
```
cedric-anne, line 2714: To be compatible with foreign keys manipulation methods, it should be named `contracts_id_default`.

orthagh (approved):
Ok with the pr but some comments (as discussed together):

- There is some warning after update in entity tab (maybe a bad init of data? )
- regarding searchoptions, merging them with asset's one could be a good idea with Contract::rawSearchOptionsToAdd method (should be adapted to ensure relation is done correctly for tickets)
- maybe also, replace SO by meta : this could enable all contract searchoptions directly **but** remove the possibility to set them as columns in displaypreferences (we tried recently to permits, but we failed)

AdrienClairembault:
>     * There is some warning after update in entity tab (maybe a bad init of data? )

Fixed in 2b0bfee.

>     * regarding searchoptions, merging them with asset's one could be a good idea with Contract::rawSearchOptionsToAdd method (should be adapted to ensure relation is done correctly for tickets)

I like the idea but there will be some ids conflict as `Contract::rawSearchOptionsToAdd` use multiple searchoptions ids that are already used by `Ticket`.
Not sure what we can do here, modifying the ids in the `Contract::rawSearchOptionsToAdd` method would kinda break existing bookmarks/savedsearch/API search request.

inc/ticket.class.php
```diff
@@ -4356,6 +4515,7 @@ function showForm($ID, $options = []) {
             // If entity is not in the list of user's entities,
             // then use as default value the first value of the user's entites list
             $this->fields["entities_id"] = $this->userentities[0];
+            $this->input["entities_id"] = $this->userentities[0];
```
cedric-anne, line 4518: IMHO, forcing input is not a good idea.
Unless I am wrong, passing `$options['entities_id']` to `$this->getDefaultContract()` should work.

AdrienClairembault: I thought too but it seems like it doesn't change anything in the case.
`$this->input["entities_id"]` will be set directly in an HTML input from the value in `$this->fields["entities_id"]` by this code in `CommonITILObject::showActorsPartForm()`:
```php
if ($this->userentity_oncreate
     && isset($this->countentitiesforuser)
     && ($this->countentitiesforuser > 1)) {
    echo "<br>";
    $rand = Entity::dropdown(['value'     => $this->fields["entities_id"],
                                   'entity'    => $this->userentities,
                                   'on_change' => 'this.form.submit()']);
 } else {
    echo "<input type='hidden' name='entities_id' value='".$this->fields["entities_id"]."'>";
 }
```

Not sure what to think of this, can change if you insist but why shouldn't we set the input ?

cedric-anne: As a developper, I expect `$this->input` to be set by user input, and eventually by business rules. If some values are forced by a really specific logic, most of developpers will not be aware of this (it may make debugging harder, code may be unvoluntary altered/removed, ...).
As `$options['entities_id']` is already set and is used to adapt proposed values on some fields (proposed actors, default ticket type, ...), it would be logical to use it to adapt default contract type.
Also, `$options['entities_id']` is set to original item entity when a ticket is created from an item.

AdrienClairembault:
Rebased for conflicts

inc/ticket.class.php
```diff
@@ -4892,6 +5051,26 @@ function showForm($ID, $options = []) {
...
 
+      if ($this->isNewItem() && !$tt->isHiddenField('_contracts_id')) {
+         echo "<tr class='tab_bg_1'>";
+         echo "<th>".$tt->getBeginHiddenFieldText('_contracts_id');
+         printf(__('%1$s%2$s'), Contract::getTypeName(1), $tt->getMandatoryMark('_contracts_id'));
+         echo $tt->getEndHiddenFieldText('_contracts_id')."</th>";
+         echo "<td>";
+         echo $tt->getBeginHiddenFieldValue('_contracts_id');
+
+         $contract_value = $this->getDefaultContract($options['entities_id'] ?? null) ?: $tt->predefined['_contracts_id'] ?? 0;
```
cedric-anne, line 5062:
```suggestion
         $contract_value = $this->getDefaultContract($options['entities_id'] ?? $this->fields['entities_id']) ?: $tt->predefined['_contracts_id'] ?? 0;
```

AdrienClairembault:
> With these changes, `getDefaultContract()` method could be moved in in `Entity` class. It would be more logical as test is done on `tests/functionnal/Entity.php` file.

Done in 8d394954e9.

AdrienClairembault:
Rebased + forced for conflicts

inc/ticket.class.php
```diff
@@ -1630,6 +1652,21 @@ function prepareInputForAdd($input) {
          $input['itilcategories_id_code'] = ITILCategory::getById($cat_id)->fields['code'];
       }
 
+      // Set default contract if not specified
+      if (!isset($input['_contracts_id'])) {
+         $input['_contracts_id'] = Entity::getDefaultContract($this->input['entities_id'] ?? $this->fields['entities_id'] ?? false);
```
cedric-anne, line 1657: 1. `$this->fields` will never be set in `prepareInputForAdd`.
2. If `$input['entities_id']` is not defined, ticket will be created in root entity, so you could fallback to `0`.
```suggestion
         $input['_contracts_id'] = Entity::getDefaultContract($input['entities_id'] ?? 0);
```

inc/ticket.class.php
```diff
@@ -4891,6 +5050,26 @@ function showForm($ID, $options = []) {
...
 
+      if ($this->isNewItem() && !$tt->isHiddenField('_contracts_id')) {
+         echo "<tr class='tab_bg_1'>";
+         echo "<th>".$tt->getBeginHiddenFieldText('_contracts_id');
+         printf(__('%1$s%2$s'), Contract::getTypeName(1), $tt->getMandatoryMark('_contracts_id'));
+         echo $tt->getEndHiddenFieldText('_contracts_id')."</th>";
+         echo "<td>";
+         echo $tt->getBeginHiddenFieldValue('_contracts_id');
+
+         $contract_value = Entity::getDefaultContract($options['entities_id'] ?? $this->fields['entities_id']) ?: $tt->predefined['_contracts_id'] ?? 0;
```
cedric-anne, line 5061: IMHO, if the ticket template provides a predefined value for `_contracts_id`, it should be used prior to the default value defined in entity.

inc/entity.class.php
```diff
@@ -3415,7 +3457,65 @@ static function getIcon() {
...
+
    public static function getAnonymizeConfig(?int $entities_id = null) {
       return Entity::getUsedConfig('anonymize_support_agents', $entities_id);
    }
+
+   public static function getDefaultContract(int $entities_id): int {
+      // Entity isn't set, can't load config
+      if ($entities_id === false) {
+         return 0;
+      }
```
cedric-anne, line 3493: Due to `$entities_id` param type hinting, it is not possible to have a `false` value here.
```suggestion
```

inc/entity.class.php
```diff
@@ -3415,7 +3457,65 @@ static function getIcon() {
...
+    * Get values for contracts_id_default field
+    *
+    * @since 9.5
+    *
+    * @return array
+    */
+   public static function getDefaultContractValues($entities_id): array {
+      $values = [
+         self::CONFIG_PARENT => __('Inheritance of the parent entity'),
+         -1 => __('Contract in ticket entity'),
```
cedric-anne, line 3470: Can you confirm that this will result in usage of first found not expired contract from the entity of the newly created ticket ?

inc/entity.class.php
```diff
@@ -3415,7 +3457,65 @@ static function getIcon() {
       return "fas fa-layer-group";
    }
 
+   /**
+    * Get values for contracts_id_default field
+    *
+    * @since 9.5
```
cedric-anne, line 3463:
```suggestion
    * @since 10.0.0
```

inc/entity.class.php
AdrienClairembault: Yes.

inc/entity.class.php
AdrienClairembault: Done in a5da574 .

inc/ticket.class.php
AdrienClairembault: Done in a5da574 .

inc/ticket.class.php
AdrienClairembault: Done in a5da574 .

AdrienClairembault:
Waiting for https://github.com/glpi-project/glpi/pull/9090 to rebase and integrate it in the new Contract::getExpiredCriteria()

inc/entity.class.php
cedric-anne: I usually do not like this kind of logic, as it may produce unpredictable result in many situations.

If we keep this possibility, label should indicate this unpredictable state:
```suggestion
         -1 => __('First found valid contract in ticket entity'),
```

@orthagh @trasher  ?

cedric-anne (approved):

trasher (approved):

trasher:
Could you please rebase to fix conflicts?

AdrienClairembault:
> 
> 
> Could you please rebase to fix conflicts?

I need this one to be merged first as it will impact some of the code : https://github.com/glpi-project/glpi/pull/9090

orthagh (approved):

trasher:
> I need this one to be merged first as it will impact some of the code : #9090

It's just been done ;)

PR #8957, opened 2021-04-13
Don't allow modifying racks without update right

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #8951

### DIFF
--- a/ajax/rack.php
+++ b/ajax/rack.php
@@ -37,6 +37,10 @@ Html::header_nocache();
 
 Session::checkLoginUser();
 
+if (!Session::haveRight('datacenter', UPDATE)) {
+   http_response_code(403);
+   die;
+}
 if (isset($_REQUEST['action'])) {
    $answer = [];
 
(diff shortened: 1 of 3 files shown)


trasher (approved):

orthagh (approved):

orthagh (commented):
i remove my previous review, as i have a question.
issue #8957 explains the grid can be moved without the adequate right.

With the current PR, the move is still permitted bu will reverted after the ajax query.
I think it could be better, if possible, to prevent movement before (with setting `data-gs-no-move="true"` to all items in grid)

cconard96:
> I think it could be better, if possible, to prevent movement before (with setting `data-gs-no-move="true"` to all items in grid)

I didn't see that option when looking for a way to make the gridstack readonly. When I set that though, it still has the grab cursor if you try moving an item but it doesn't move. Instead trying to drag an item is the same as clicking it so it opens the asset form.

Actually it seems unrelated to the addition of the attribute. The mouse-up event anywhere in the rack seems to trigger opening an asset form...

cconard96:
Should be working properly now

orthagh (approved):

trasher (approved):

PR #8958, opened 2021-04-13
initial commit

SpectrumZX81:


orthagh (changes requested):
Hello @SpectrumZX81 

Here is a more detailed review.

PHP stuff:

- All classes in the `inc/gantt/` folder should have a `namespace Glpi\Gantt;`
- this way, all `require|require_once` become superflu
- all the calls to these classes could be done with a `use Glpi\Gantt;` on top ot the file.
- none of `include ('../inc/includes.php');` in top of classes are needed
- there a reference to a class named `ProjectTaskLink`, the class file doesn't exists in the current PR

JS: 

- your paths in the bundle are broken, you don't need `node_modules/` prefix
- ajax url path in the js/gantt-helper is absolute and is broken on my instance, you should use maybe `var url = '../ajax/gantt.php';` or use CFG_GLPI.root_doc var
- You have TODO/disabled comments present in your code

CS: there still a lot of issue here

- there's trailing spaces everywhere
- you should have an empty eof line in each file
- GLPI currently use 3 spaces indentation (this is subject to change in a near future, but we must have a good indentation before merge of this pr)

You may find the documentation about our current standards:
https://glpi-developer-documentation.readthedocs.io/en/master/codingstandards.html

CSS:

- use scss features : nested classes

More globally, i had an hard time to get the feature working on my side to at least have the view displaying.

orthagh:
closed in favor of #5212

Issue #655, opened 2016-05-17
[90.3] Doesn't delete the null inscription in ticket


AircraftRu:
Hello. Yesterday I found an error when deleting a documemt. The document was removed but the inscription has remained and when I click on the Delete permanently button doesn't delete the inscription. When I click on the Edit opens a document with null id.

![glpi - tickets - google chrome 2016-05-16 16 18 31](https://cloud.githubusercontent.com/assets/17082164/15315401/1623aef8-1c20-11e6-9ddb-651196231d60.png)
![glpi - tickets - google chrome 2016-05-16 16 19 16](https://cloud.githubusercontent.com/assets/17082164/15315407/196652f0-1c20-11e6-94a9-fddb01ed8500.png)
![glpi - tickets - google chrome 2016-05-16 17 03 02](https://cloud.githubusercontent.com/assets/17082164/15315410/1c18c6d6-1c20-11e6-8c0d-8f8bff3bd58d.png)

orthagh:
Hello @AircraftRu 

I can't replicate your issue at all.
Could you precise the glpi version  and if you can, a way to always trigger theses errors ?

I think it can a cache issue.

AircraftRu:
In MySQL database not deleted link with a non-existent document. I corrected it. 
May was local problem to php. Maybe it's a random bug.

orthagh:
Ok So i close this issue now.
Feels free to re-open and comments if you trigger again and had replication steps .

Regards

PR #8978, opened 2021-04-20
9.5/gantt integration

SpectrumZX81:


inc/gantt/taskdao.class.php
```diff
@@ -0,0 +1,78 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with GLPI. If not, see <http://www.gnu.org/licenses/>.
+ * ---------------------------------------------------------------------
+ */
+
+namespace Glpi\Gantt;
+
```
trasher, line 35:
```suggestion

use \Exception;

```

ajax/gantt.php
```diff
@@ -0,0 +1,199 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with GLPI. If not, see <http://www.gnu.org/licenses/>.
+ * ---------------------------------------------------------------------
+ */
+namespace Glpi\Gantt;
+use \ProjectTaskLink;
+use \Exception;
+
```
trasher, line 36: Only classes should be namespaced; not front files.
```suggestion
```

inc/gantt/datafactory.class.php
```diff
@@ -0,0 +1,164 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with GLPI. If not, see <http://www.gnu.org/licenses/>.
+ * ---------------------------------------------------------------------
+ */
+
+namespace Glpi\Gantt;
+
+use Glpi\Gantt\Item;
+use Glpi\Gantt\LinkDAO;
+
```
trasher, line 37: Same namespace, this is not needed
```suggestion
```

inc/gantt/datafactory.class.php
```diff
@@ -0,0 +1,164 @@
...
+   /**
+    * Recursive function used to get all subprojects and tasks of a project
+    *
+    * @param array $itemArray Array holding the items
+    * @param integer $projectId ID of the parent project
+    *
+    */
+   function getSubprojects(&$itemArray, $projectId) {
+      global $DB;
+      foreach ($DB->request('glpi_projects', ['projects_id' => $projectId, 'is_deleted' => 0]) as $record) {
```
trasher, line 112: For performance reasons, avoid calling "complex" methods in any loop
```suggestion
      $iterator = $DB->request('glpi_projects', ['projects_id' => $projectId, 'is_deleted' => 0]);
      while ($record = $iterator->next()) {
```

inc/project.class.php
```diff
@@ -1827,135 +1827,74 @@ static function getDataToDisplayOnGantt($ID, $showall = true) {
...
+     * Show GANTT diagram for a project
+     * @param $ID ID of the project
    */
    static function showGantt($ID) {
-      global $DB;
+      echo "<div id=\"gantt-container\" style=\"width:100%; height:63vh;\"></div>";
+      echo "<div id=\"gantt-features\">
+               <ul class=\"gantt-controls\">
+                  <li class=\"gantt-menu-item gantt-menu-item-right\">
+                  <a href=\"#\" class=\"fullscreen\" onclick=\"gantt.ext.fullscreen.toggle();\"><img src=\"../pics/icones/ic_fullscreen_24.png\">Fullscreen</a>
```
trasher, line 1839: Should certainly be replaced with a fontawesome icon

trasher:
Remaining CS issues can be fixed with phpcbf tool (see https://github.com/squizlabs/PHP_CodeSniffer); with the a command like: 
```
vendor/bin/phpcbf --standard=vendor/glpi-project/coding-standard/GlpiStandard/ inc/gantt ajax/gantt.php
```

Many IDE provides codesniffer integration also.

trasher:
Table creation is missing in the installation SQL file (`install/mysql/glpi-empty.sql`)

js/gantt-api.js
```diff
@@ -0,0 +1,1057 @@
+(function(){
+
+var apiUrl = "https://export.dhtmlx.com/gantt";
```
trasher, line 3: That does not sounds correct to call an external API; there certainly are limitations.

SpectrumZX81: Thanks for your feedback, I agree. But standalone modules require payed license, so I decided to use the online export service which is claimed to be free of charge on the vendor's website.
See below
https://dhtmlx.com/docs/products/dhtmlxGantt/export.shtml

trasher: Well, some may care about privacy issues, there may be limit on the number of calls that can be done (there are many GLPI instances); ...
Also, what if the free service disapears?

Exporting should certainly be done on GLPI side. 
Adding this is not a blocker for the PR, I'm not sure current gantt implementation permit such exports (and you should keep as is on your side if you want).
But I guess removing the external call will be a blocker :)

cconard96: I don't like having data sent to an external service just for an export. At the very least, I think there needs to be a disclaimer and make it so a GLPI admin has to acknowledge that their project data will be sent to this service to be able to enable exports. Otherwise, it should be disabled.
Doing exports natively in GLPI may be possible later too.

SpectrumZX81: Would it be appropriate to have a warning message like this, until GLPI PNG/PDF exports are implemented ?
image

inc/project.class.php
SpectrumZX81: it's done

js/gantt-api.js
SpectrumZX81: Hello @trasher, @cconard96,
I have removed the export function from this PR, could you please check if there is anything else to address ?
Many thanks.

trasher: Hi, sure, I'll take another look at the code today or tomorrow.
There are still some CS issues on JS (you've fixed all PHP ones already ;))

I'm not confident with gantt, someone else will do a functionnal review.

js/gantt-helper.js
```diff
@@ -0,0 +1,610 @@
...
+                  progress
+               }
+            },
+            success: function(resp) {
+               var json = JSON.parse(resp);
+               if (json.ok) {
+                  task.progress = progress;
+                  gantt.updateTask(task.id);
+               } else {
+                  console.log('Could not update Task[' + id + ']: ' + json.error);
```
trasher, line 201: Error should be show to the user, not in the console, I guess with:
```suggestion
                  gantt.alert('Could not update Task[' + id + ']: ' + json.error);
```

Same remark for all other `console` usages. This will also fix some of the remaining lint issue.

trasher (changes requested):

orthagh:
During a meeting with Nicolas Coudière this morning, we found some issues:

- Project(s) line(s) doesn't care at all about their planned start and end dates in the gantt view. It may be normal but it disturbing at usage. It can be explained by the way so if we can't indicate to dhtmlx to display real date on project line, it will be ok for me.

- Other thing, i think we need to disable completely double-click on projects line, as the fields are not editable by the way. Only delete button works and i think we can pass on having this button here (as it exists on main form).

- If a project's task is created (like via kanban) without planned dates (start or end), the gantt view struggle to permits edition of it. You can't move it, edit duration

Regards

SpectrumZX81:
Hello,
    
    The thing is, these issues are in fact default behavior of this
    plugin. Gantt displays projects on the timeline based on its child
    items start/end dates.
    
    https://docs.dhtmlx.com/gantt/desktop__task_types.html#projecttasks
    
    And this may be the reason why projects can't be dragged, this
    functionality being disabled by default according to the
    documentation and it can be enabled in the Pro version.
    https://docs.dhtmlx.com/gantt/api__gantt_drag_project_config.html
    
    If we double click a project, Gantt does not allow editing the
    fields - also a default behavior. But we can of course disable
    opening the dialog when double clicking them, this is not a problem.
    Regarding tasks, if they don't have start or end dates Gantt will
    have issues handling them because of missing parameters. This can be
    addressed by adding default start/end dates either during creation
    from kanban or before sending them to Gantt to display on the
    timeline.
    
    Regards
    
    On 05/05/2021 11:47 AM, Alexandre
      Delaunay wrote:
    
    
      
      During a meeting with Nicolas Coudière this morning, we found
        some issues:
      
        
          Project(s) line(s) doesn't care at all about their planned
            start and end dates in the gantt view. It may be normal but
            it disturbing at usage. It can be explained by the way so if
            we can't indicate to dhtmlx to display real date on project
            line, it will be ok for me.
        
        
          Other thing, i think we need to disable completely
            double-click on projects line, as the fields are not
            editable by the way. Only delete button works and i think we
            can pass on having this button here (as it exists on main
            form).
        
        
          If a project's task is created (like via kanban) without
            planned dates (start or end), the gantt view struggle to
            permits edition of it. You can't move it, edit duration
        
      
      Regards
      —
        You are receiving this because you authored the thread.
        Reply to this email directly, view it on GitHub, or unsubscribe.
      [

{

***@***.***": "http://schema.org",

***@***.***": "EmailMessage",

"potentialAction": {

***@***.***": "ViewAction",

"target": "https://github.com/glpi-project/glpi/pull/8978#issuecomment-832522534",

"url": "https://github.com/glpi-project/glpi/pull/8978#issuecomment-832522534",

"name": "View Pull Request"

},

"description": "View this Pull Request on GitHub",

"publisher": {

***@***.***": "Organization",

"name": "GitHub",

"url": "https://github.com"

}

}

]

orthagh:
For project line, can we simply remove it from the view ? 

For missing dates, i would prefer the case taken into account by our code before sending them to the gantt view (in ajax query? ), this could be better to handle all possible cases we could miss.

inc/gantt/datafactory.class.php
```diff
@@ -153,10 +173,11 @@ function populateGanttItem($record, $type) {
       $item->parent = ($type == "root-project") ? 0 : (($type == "project") ? $record['projects_id'] : ($record["projecttasks_id"] > 0 ? $parentTaskUid : $record['projects_id']));
       $item->linksource_id = ($item->type != "project") ? $record["projecttasks_id"] : 0;
       $item->linktask_id = ($item->type != "project") ? $record["id"] : 0; // parent task id to search for by child->linksource_id
-      $item->start_date = $record['plan_start_date'];
-      $item->end_date = $record['plan_end_date'];
+      $item->start_date = isset($record['plan_start_date']) ? $record['plan_start_date'] : $_SESSION['glpi_currenttime'];
+      $item->end_date = isset($record['plan_end_date']) ? $record['plan_end_date'] : date('Y-m-d H:i:s', strtotime($item->start_date . ' + 1 day'));
```
trasher, line 177:
```suggestion
      $item->start_date = $record['plan_start_date'] ?? $_SESSION['glpi_currenttime'];
      $item->end_date = $record['plan_end_date'] ?? date('Y-m-d H:i:s', strtotime($item->start_date . ' + 1 day'));
```

inc/gantt/datafactory.class.php
```diff
@@ -153,10 +173,11 @@ function populateGanttItem($record, $type) {
...
       $item->text = $record['name'];
-      $item->note = isset($record['code']) ? $record['code'] : "";
+      $item->content = isset($record['content']) ? html_entity_decode($record['content']) : "";
+      $item->comment = isset($record['comment']) ? $record['comment'] : "";
```
trasher, line 180:
```suggestion
      $item->comment = $record['comment'] ?? "";
```

inc/gantt/taskdao.class.php
```diff
@@ -44,21 +45,118 @@
...
       $t->getFromDB($task->id);
 
       $t->update([
          'id' => $task->id,
          'plan_start_date' => $task->start_date,
+         'real_start_date' => ($task->type == "milestone" && $t->fields["real_start_date"] == null) ? $task->start_date : $t->fields["real_start_date"],
          'plan_end_date' => $task->end_date,
          'percent_done' => ($task->progress * 100),
-         'name' => (isset($task->text) ? $task->text : $t->fields['name'])
+         'name' => (isset($task->text) ? $task->text : $t->fields['name']),
```
trasher, line 87:
```suggestion
         'name' => $task->text ?? $t->fields['name']),
```

SpectrumZX81:
Known issue with filtering: if the parent task is collapsed its child items are not displayed when we filter them, only visible items can be filtered.
As a possible fix, dhtmlx suggests expanding all nodes before calling the filter function.
Please let me know your opinion about it, other suggestions are welcomed too.

trasher:
> As a possible fix, dhtmlx suggests expanding all nodes before calling the filter function.

I may be wrong, but loading all nodes may be huge on large databases I guess :/

SpectrumZX81:
Gantt data is loaded only once, during page load. Expanding nodes after doesn't cause additional traffic, only the view is changed.

orthagh:
Some remarks after testing the last commits:

- global gantt view (available in breadcrumb) seems to be broken
image

I have no idea if its difficult to fix or not.

- alerts from glpi are not called by ajax queries callback. You can see the issue by doing some stuff in gantt view (resize, move, create a task) ans then go to a task form (by double clicking). All pending alerts will show in bottom right of glpi.
Ex:
image

You can fix this by calling `displayAjaxMessageAfterRedirect()` js fonction after just before finish processing a gantt event.
Should be easy to fix

- radio buttons in bottom-right have a strange ui. here is some fast suggestions to improve them:
   - add a bottom border on :checked state like `border-bottom: 2px solid #655;`
   - hide radio input and keep only label on fieldset radios

image

Should be easy to fix

- is parent column usefull ? Imo, it can be removed

image

SpectrumZX81:
> Some remarks after testing the last commits:
> 
>     * global gantt view (available in breadcrumb) seems to be broken
>       image
> 
> 
> I have no idea if its difficult to fix or not.
> 

If the problem persist after clearing the browser cache (Ctrl+F5) and there are no plugin errors in the browser console, there might be inconsistency issues with the data being sent to Gantt.
If it's possible, please provide your ajax payload, by email or pm, to let me analyze the structure of your data. (Network tab -> gantt.php -> getData() function response).

>     * alerts from glpi are not called by ajax queries callback. You can see the issue by doing some stuff in gantt view (resize, move, create a task) ans then go to a task form (by double clicking). All pending alerts will show in bottom right of glpi.
>       Ex:
>       image
> 
> 
> You can fix this by calling `displayAjaxMessageAfterRedirect()` js function after just before finish processing a gantt event.
> Should be easy to fix

Ok, I will place this function call in $(document).ajaxStop( ) event handler and see how it works.
Thanks.

>     * radio buttons in bottom-right have a strange ui. here is some fast suggestions to improve them:
>       
>       * add a bottom border on :checked state like `border-bottom: 2px solid #655;`
>       * hide input[type=radio] and keep only labels in fieldsets
> 
> 
> image
> 
> Should be easy to fix

Ok 

>     * is parent column usefull ? Imo, it can be removed
> 
> 
> image

Yes, there is a customer request for it to be there.

orthagh:
For global gantt, my bad, i forgot about the "Show on global GANTT" field.
It's working well in fact.

for others correction -> great, it's now very good.

For the column "parent", i think it's not very usefull but i can live with and it's a minor point.
So let's advance on this pr.

css/glpi-gantt.scss
```diff
@@ -0,0 +1,225 @@
...
+   margin: -3px 3px 0 7px;
+}
+
+fieldset {
+   color: #655;
+   border: solid 1px #655;
+   border-radius: 5px;
+}
+
+.gantt-container {
```
cedric-anne, line 48: This class is not used so I guess this style is useless.

css/glpi-gantt.scss
```diff
@@ -0,0 +1,225 @@
+/**
```
cedric-anne, line 1: It would be preferable to encapsulate all these styles into a `.gantt-wrapper` class, or something like that, to ensure styles will be only applied to the gantt view.

cedric-anne (commented):
Rights checks have to be added on read/write operations, on PHP side, and maybe some operations have to be blocked on JS side when user has only project/projecttask read rights.

js/gantt-helper.js
```diff
@@ -0,0 +1,893 @@
...
+         gantt.templates.tooltip_date_format(start) +
+         "<br/><b>End date:</b> " + gantt.templates.tooltip_date_format(end) +
+         "<br/><b>Progress:</b> " + parseInt(task.progress * 100) + "%";
+      if (task.content && task.content.length > 0) {
+         text += "<br/><b>Description:</b><div style=\"padding-left:25px\">" + task.content + "</div>";
+      }
+      if (task.comment && task.comment.length > 0) {
+         text += "<br/><b>Comment:</b><div style=\"padding-left:25px\"" + task.comment + "</div>";
+      }
+      return text;
```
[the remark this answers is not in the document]
SpectrumZX81: Ok, for ajax messages if I put something like
`gantt.alert(__('Could not update Task[%s]: ', id) + json.error);`
how can I update the language files (mo, po) to have this message working ?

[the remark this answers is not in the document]
cedric-anne: In current PR, you do not have to update po/mo files. It will be done separately.

For your GLPI instance, you can manually create a file using POEdit, and store corresponding mo file in `files/core/xx_YY.mo`. You may have to clear GLPI cache and restart your PHP server.

inc/gantt/datafactory.class.php
```diff
@@ -0,0 +1,198 @@
...
+         $parentTask->getFromDB($record["projecttasks_id"]);
+         $parentTaskUid = $parentTask->fields["uuid"];
+      }
+
+      $item = new Item();
+      $item->id = ($type == "project" || $type == "root-project") ? $record['id'] : $record['uuid'];
+      $item->type = ($type == "root-project") ? "project" : $type;
+      $item->parent = ($type == "root-project") ? 0 : (($type == "project") ? $record['projects_id'] : ($record["projecttasks_id"] > 0 ? $parentTaskUid : $record['projects_id']));
+      $item->linksource_id = ($item->type != "project") ? $record["projecttasks_id"] : 0;
+      $item->linktask_id = ($item->type != "project") ? $record["id"] : 0; // parent task id to search for by child->linksource_id
```
cedric-anne, line 188: These properties does not exist in `Glpi\Gantt\Item` and are not used anywhere.
```suggestion
```

inc/gantt/datafactory.class.php
```diff
@@ -0,0 +1,198 @@
...
+      $item->start_date = $record['plan_start_date'] ?? $_SESSION['glpi_currenttime'];
+      $item->end_date = $record['plan_end_date'] ?? date('Y-m-d H:i:s', strtotime($item->start_date . ' + 1 day'));
+      $item->text = $record['name'];
+      $item->content = isset($record['content']) ? html_entity_decode($record['content']) : "";
```
cedric-anne, line 192: To get safe HTML, you have to use.
```suggestion
      $item->content = isset($record['content']) ? \Html::clean($record['content'], false, 1) : "";
```

inc/gantt/datafactory.class.php
```diff
@@ -0,0 +1,198 @@
...
+      $item->comment = $record['comment'] ?? "";
```
cedric-anne, line 193: This property does not exist, did you mean `$item->note` ?

js/gantt-helper.js
```diff
@@ -0,0 +1,872 @@
...
+               var task = gantt.getTask(id);
+               task.text = item.text;
+               task.type = item.type;
+               task.start_date = item.start_date;
+               task.end_date = item.end_date;
+               gantt.updateTask(id);
+               gantt.hideLightbox();
+               displayAjaxMessageAfterRedirect();
+            } else
+               gantt.alert(__('Could not update Task[%s]: ', item.text) + json.error);
```
cedric-anne, line 598:
```suggestion
               gantt.alert(__('Could not update Task[%s]: ').replace('%s', item.text) + json.error);
```

js/gantt-helper.js
```diff
@@ -0,0 +1,872 @@
...
+               progress
+            }
+         },
+         success: function(json) {
+            if (json.ok) {
+               task.progress = progress;
+               gantt.updateTask(task.id);
+               displayAjaxMessageAfterRedirect();
+            } else {
+               gantt.alert(__('Could not update Task[%s]: ', id) + json.error);
```
cedric-anne, line 625:
```suggestion
               gantt.alert(__('Could not update Task[%s]: ').replace('%s', id) + json.error);
```

js/gantt-helper.js
```diff
@@ -0,0 +1,872 @@
...
+         },
+         success: function(json) {
+            if (json.ok) {
+               var project = gantt.getTask(id);
+               project.text = item.text;
+               gantt.updateTask(id);
+               gantt.hideLightbox();
+               displayAjaxMessageAfterRedirect();
+            } else {
+               gantt.alert(__('Could not update Project[%s]: ', item.text) + json.error);
```
cedric-anne, line 727:
```suggestion
               gantt.alert(__('Could not update Project[%s]: ').replace('%s', item.text) + json.error);
```

js/gantt-helper.js
```diff
@@ -0,0 +1,872 @@
...
+
+         $("input[name='branch_state']").on('click', function() {
+            expandCollapse($("input[name='branch_level']:checked").val());
+         });
+
+         gantt.attachEvent("onBeforeRowDragMove", function(id) {
+            $('#hf_gantt_item_state').val(id + "|" + gantt.getTask(id).parent);
+         });
+
+         gantt.attachEvent("onBeforeRowDragEnd", function(id, target) {
```
cedric-anne, line 232: 1. An error appears when trying to move a project to the root level. Doing this should be possible.

2. An error appears when trying to move a task to the root level. This action should not be possible, so it should just be ignored and no error should be shown.

SpectrumZX81:
Hello,
Thanks for pointing out the bug at the root level (!) - it's been fixed.
Regarding deleteTask(), putInTrashbin() functions: although the Delete button is no longer available, I considered leaving these functions in place inside gantt-helper.js, gantt.php and DAO classes, in case there is need to adapt them for mass-delete functionality.
I removed them now from this PR, keeping a backup on my dev site. 
Please let me know if there is something else to address.

inc/gantt/projectdao.class.php
```diff
@@ -0,0 +1,99 @@
...
+
+   function addProject($project) {
+
+      if (!\Project::canCreate()) {
+         throw new \Exception(__('Not enough rights'));
+      }
+
+      $input = [
+         'name' => $project->text,
+         'comment' => $project->note,
```
cedric-anne, line 53:
```suggestion
         'comment' => $project->comment,
```

cedric-anne:
It has been merged in master in commit d45f662f5ac776ae8420927331059242b204a2a4.

Thanks for the contribution.

SpectrumZX81:
Sounds great, thanks for reviewing.

PR #9022, opened 2021-04-30
Copy names to clipboard

AdrienClairembault:
Internal ref: 21809.

image

Useful to export search results (only the name column) and paste them somewhere else.

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |

### DIFF
--- a/js/common.js
+++ b/js/common.js
@@ -1243,7 +1243,7 @@ function _showMessage(message, css_class, title) {
 }
 
 function showInfoMessage(message) {
-   _showMessage(message, 'info_msg', __('Information'));
+   _showMessage(message, 'info_msg', _n('Information', 'Information', 1));
 }
 
 function showWarningMessage(message) {
--- a/js/common.js
+++ b/js/common.js
@@ -1212,3 +1212,44 @@ function getFlatPickerLocale(language, region) {
       return language;
    }
 }
+
+function _showMessage(message, css_class, title) {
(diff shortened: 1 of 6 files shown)


js/common.js
```diff
@@ -1212,3 +1212,44 @@ function getFlatPickerLocale(language, region) {
...
+      show: {
+         effect: 'slide',
+         direction: 'down',
+         'duration': 800
+      }
+   });
+}
+
+function showInfoMessage(message) {
+   _showMessage(message, 'info_msg', __('Information'));
```
AdrienClairembault, line 1246: Lint is failing because I am not using plural form to get the 'Information' locale.
However it looks like we don't support plural right now in JS files as [ __n](https://www.npmjs.com/package/i18n#i18n__n) is undefined ?

AdrienClairembault: Found it, we use _n instead.

AdrienClairembault:
Up, would need at least one review to move forward with the internal ref.

trasher (commented):
Sounds good to me globally. The only question would be on the `_showMessage`, I had in mind there were already something to do that (but I may be wrong).

cedric-anne (approved):

trasher (approved):

PR #9058, opened 2021-05-07
Fluid Search Experience

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

Create a more fluid searching experience.

Major changes:
- The column headers are now always visible when scrolling (Sticky headers)
- You can sort by more than one field at a time (Multi-column sorting)
- Search results update through AJAX requests instead of a full page reload

This should improve perceived performance when searching. It also fixes the issue where the automatic ticket list refresh could reload the full page while you are trying to change your search or perform massive actions.

### DIFF
--- a/ajax/search.php
+++ b/ajax/search.php
@@ -31,6 +31,7 @@
  */
 
 // Direct access to file
+
 include ('../inc/includes.php');
 header("Content-Type: text/html; charset=UTF-8");
 Html::header_nocache();
@@ -51,6 +52,26 @@ switch ($_REQUEST['action']) {
       ]);
 
       echo json_encode(['success' => $success]);
+      break;
+
+   case 'display_results':
+      if (!isset($_REQUEST['itemtype'])) {
+         http_response_code(400);
+         die;
+      }
+
+      /** @var CommonDBTM $itemtype */
+      $itemtype = $_REQUEST['itemtype'];
+      if (!$itemtype::canView()) {
+         http_response_code(403);
+         die;
+      }
+
+      $search_params = Search::manageParams($itemtype, $_REQUEST);
+
+      $results = Search::getDatas($itemtype, $search_params);
+      $results['searchform_id'] = $_REQUEST['searchform_id'] ?? null;
+      Search::displayData($results);
       break;
 }
 
--- a/inc/search.class.php
+++ b/inc/search.class.php
@@ -373,8 +373,8 @@ class Search {
       // Default values of parameters
       $p['criteria']            = [];
       $p['metacriteria']        = [];
-      $p['sort']                = '1'; //
-      $p['order']               = 'ASC';//
+      $p['sort']                = ['1'];
+      $p['order']               = ['ASC'];
       $p['start']               = 0;//
       $p['is_deleted']          = 0;
       $p['export_all']          = 0;
@@ -395,15 +395,25 @@ class Search {
       foreach ($params as $key => $val) {
          switch ($key) {
             case 'order':
-               if (in_array($val, ['ASC', 'DESC'])) {
-                  $p[$key] = $val;
+               if (!is_array($val)) {
+                  // Backward compatibility with GLPI < 10.0 links
+                  if (in_array($val, ['ASC', 'DESC'])) {
+                     $p[$key] = [$val];
+                  }
+                  break;
                }
+               $p[$key] = $val;
                break;
             case 'sort':
-               $p[$key] = intval($val);
-               if ($p[$key] < 0) {
-                  $p[$key] = 1;
+               if (!is_array($val)) {
+                  // Backward compatibility with GLPI < 10.0 links
+                  $val = (int) $val;
+                  if ($val >= 0) {
+                     $p[$key] = [$val];
+                  }
+                  break;
                }
+               $p[$key] = $val;
                break;
             case 'is_deleted':
                if ($val == 1) {
@@ -522,9 +532,8 @@ class Search {
       }
 
       // Add order item
-      if (!in_array($p['sort'], $data['toview'])) {
-         array_push($data['toview'], $p['sort']);
-      }
+      $to_add_view = array_diff($p['sort'], $data['toview']);
+      array_push($data['toview'], ...$to_add_view);
 
       // Special case for Ticket : put ID in front
       if ($itemtype == 'Ticket') {
@@ -714,14 +723,20 @@ class Search {
 
       //// 4 - ORDER
       $ORDER = " ORDER BY `id` ";
-      foreach ($data['tocompute'] as $val) {
-         if ($data['search']['sort'] == $val) {
-            $ORDER = self::addOrderBy(
-               $data['itemtype'],
-               $data['search']['sort'],
-               $data['search']['order']
-            );
+      $sort_fields = [];
+      $sort_count = count($data['search']['sort']);
+      for ($i = 0; $i < $sort_count; $i++) {
+         foreach ($data['tocompute'] as $val) {
+            if ($data['search']['sort'][$i] == $val) {
+               $sort_fields[] = [
+                  'searchopt_id' => $data['search']['sort'][$i],
+                  'order'        => $data['search']['order'][$i] ?? null
+               ];
+            }
          }
+      }
+      if (count($sort_fields)) {
+         $ORDER = self::addOrderBy($data['itemtype'], $sort_fields);
       }
 
       $SELECT = rtrim(trim($SELECT), ',');
@@ -1521,7 +1536,13 @@ class Search {
             'criteria'     => Toolbox::stripslashes_deep($search['criteria']),
             'metacriteria' => Toolbox::stripslashes_deep($search['metacriteria'])
       ], '&');
-      $parameters = "sort={$search['sort']}&order={$search['order']}&{$globallinkto}";
+
+      $parameters = http_build_query([
+         'sort'   => $search['sort'],
+         'order'  => $search['order']
+      ]);
+
+      $parameters .= "&{$globallinkto}";
 
       if (isset($_GET['_in_modal'])) {
          $parameters .= "&_in_modal=1";
@@ -1547,8 +1568,8 @@ class Search {
          'union_search_type'   => $CFG_GLPI["union_search_type"],
          'rand'                => mt_rand(),
          'no_sort'             => $search['no_sort'] ?? false,
-         'order'               => $search['order'] ?? "ASC",
-         'sort'                => $search['sort'] ?? "",
+         'order'               => $search['order'] ?? [],
+         'sort'                => $search['sort'] ?? [],
          'start'               => $search['start'] ?? 0,
          'limit'               => $_SESSION['glpilist_limit'],
          'count'               => $data['data']['totalcount'] ?? 0,
@@ -2208,7 +2229,7 @@ class Search {
             }
 
             if ($p['showreset']) {
-               echo "<a class='btn btn-ghost-secondary btn-icon btn-sm me-1'
+               echo "<a class='btn btn-ghost-secondary btn-icon btn-sm me-1 search-reset'
                         data-bs-toggle='tooltip' data-bs-placement='bottom'
                         href='"
                   .$p['target']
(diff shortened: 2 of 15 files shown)


orthagh:
just tested the last commit:

- the overlay/spinner doesn't show anymore
- you have broken lint, in php but not mentionned by CI, i saw also js (indent = 4 space in Table.js for example)

cconard96:
>     * the overlay/spinner doesn't show anymore

I wrapped the jQuery load method in a window.setTimeout to artificially slow down the AJAX refresh and the spinner did show. I also added a new test case to ensure the initially created spinner as well as subsequent spinners are visible and it passed. Maybe it is refreshing too fast so it doesn't show?

>     * you have broken lint, in php but not mentionned by CI, i saw also js (indent = 4 space in Table.js for example)

I fixed the JS indentations, but don't see an issue in PHP even when running the `composer run-script lint` command locally.

.babelrc
```diff
@@ -0,0 +1,15 @@
...
+  "env": {
+    "test": {
+      "presets": ["@babel/env"],
+      "plugins": [
+        "@babel/plugin-proposal-class-properties",
+        "@babel/plugin-transform-modules-commonjs"
+      ]
+    }
+  }
+}
```
cedric-anne, line 15: I do not really know Jest. Can you explain why babel is required ?

jest.config.js
```diff
@@ -0,0 +1,9 @@
+module.exports = {
```
cedric-anne, line 1: This file shoule be placed in `tests/` directory (or `test/js/`). You should be able to do this by using the `--config` option in jest command call.

.babelrc
cconard96: Without Babel it doesn't understand modules and ES6 features, and will fail tests immediately at the import statement. There may be another way to get it to work though, but this is what I had working for Kanban at the time.

Jest encountered an unexpected token

    This usually means that you are trying to import a file which Jest cannot parse, e.g. it's not plain JavaScript.

    By default, if Jest sees a Babel config, it will use that to transform your files, ignoring "node_modules".

    Here's what you can do:
     • If you are trying to use ECMAScript Modules, see https://jestjs.io/docs/en/ecmascript-modules for how to enable it.
     • To have some of your "node_modules" files transformed, you can specify a custom "transformIgnorePatterns" in your config.
     • If you need a custom transformation specify a "transform" option in your config.
     • If you simply want to mock your non-JS modules (e.g. binary assets) you can stub them out with the "moduleNameMapper" config option.

    You'll find more details and examples of these config options in the docs:
    https://jestjs.io/docs/en/configuration.html

orthagh:
Hello @cconard96 

We tested another round with @trasher, we found some issues: 

- if a search return no record, the [Search] main button become inactive: no ajax request triggers even after changing some criteria
- on sorting event, the number indicator display a "random" number (11) before getting the correct (2). It's not blocking but disturbing.
- I'm still not sure about multi sorting as a default behavior, mainly for resetting sort (if i set 5 sorts, i need to click on reset on manually on each columns, both options are not easy), i still prefer to have main sorting by default and get the multisort by a key modifier (ctrl or shift)
- not your responsibility, but we need to check this pr against #8858 

Regards

orthagh:
i miss to mention, spinner overlay works now and it gives a great "impression"

cconard96:
All reported issues should be fixed now.
The new behavior for single/multi sort is that a click will allow changing the order of a column that is already sorted or the column will become the only sorted column. Control clicking a column will keep the other sorts and append it (Same as initial behavior).

ajax/search.php
```diff
@@ -52,6 +54,63 @@
...
+
+      $search_params = [
+         'sort'   => $sort
+      ];
+      unset($_REQUEST['sort'], $_REQUEST['order']);
+      $search_params = array_merge($search_params, $_REQUEST);
+
+      $results = Search::getDatas($itemtype, $search_params);
+
+      TemplateRenderer::getInstance()->display('components/search/display_data.html.twig', [
```
trasher, line 92: As far as I've seen, `Search::displayData()` should be used here

inc/search.class.php
```diff
@@ -3102,129 +3147,169 @@ static function addHaving($LINK, $NOT, $itemtype, $ID, $searchtype, $val) {
...
+   static function addOrderBy($sort_fields, $_id = null, $_order = 'ASC') {
       global $CFG_GLPI;
 
-      // Security test for order
-      if ($order != "ASC") {
-         $order = "DESC";
+      // BC parameter conversion
+      if (!is_array($sort_fields)) {
+         // < 10.0.0 parameters
+         //Toolbox::deprecated('The parameters for Search::addOrderBy have changed to allow sorting by multiple fields. Please update your calling code.');
```
trasher, line 3170:
```suggestion
         Toolbox::deprecated('The parameters for Search::addOrderBy have changed to allow sorting by multiple fields. Please update your calling code.');
```

templates/components/search/display_data.html.twig
```diff
@@ -23,19 +30,27 @@
...
-         {% if data['search']['as_map'] == 0 %}
-            {{ include('components/search/table.html.twig') }}
-         {% endif %}
          {% if data['display_type'] != constant('Search::GLOBAL_SEARCH') and data['search']['as_map'] == 0 %}
             <div class="card-footer search-footer">
                {{ include('components/search/pager.html.twig') }}
             </div>
          {% endif %}
-      {% else %}
+      {% elseif data['search']['as_map'] == 1 %}
```
trasher, line 42: I do not understand this change

cedric-anne (commented):
I have multiple remarks about the spinner.

1. The spinner `div` container is appended as a child of a `table` element. This is not a valid DOM structure and may lead to unexpected rendering depending on the browser rendering engine (I did not tested it, but I already had problems due to this kind of structure in the past).

2. The spinner position is not sticking to the displayed result table. If you have many displayed columns, you may not see the spinner. Thus, vertical centering seems to be relative to the page wrapper.
image

cedric-anne:
I also noticed that the stickyness of first column is not working properly
image

Maybe it should be done in another PR.

templates/components/search/display_data.html.twig
cconard96: If the search result view is a table, even if there are no results, the table template is rendered now (table header is hidden if there are no results) so that the JS code doesn't break. It needs the table element and columns intact. If the result view is a map, which is the only other option right now, and there are no results, it displays the "No items found" message. The display of the map when there are results isn't handled anywhere yet and there isn't a template for it.

cconard96:
I moved the loading overlay outside of the table element and tried centering it better. The best I got working was centering to the screen. I couldn't get anything working to center an element to the visible portion of a scrollable container.
The sticky first column issue is due to the table background color being transparent. I've changed it to be the same as the card background color (its container).

js/Search/Table.js
```diff
@@ -0,0 +1,237 @@
...
+         const sort_nums_obj = {};
+         sort_nums.forEach((v, k) => {
+            sort_nums_obj[v] = k;
+         });
+
+         // Clear sort-nums from all columns or change value
+         all_colums.each((i, c) => {
+            const col = $(c);
+            col.attr('data-sort-num', undefined);
+            if (sort_nums_obj.hasOwnProperty(col.attr('data-searchopt-id'))) {
```
cedric-anne, line 98: Not sure converting to an object was necessary here, I think it should be simplier to use something like
```js
const col_sort_num = sort_nums.indexOf(col.attr('data-searchopt-id'));
if (col_sort_num !== -1) {
   col.attr('data-sort-num', col_sort_num);
}
```

js/Search/Table.js
```diff
@@ -0,0 +1,237 @@
...
+               col.attr('data-sort-num', sort_nums_obj[col.attr('data-searchopt-id')]);
+            }
+         });
+      };
+
+      if (sort_num === undefined && new_order !== 'nosort') {
+         // Recalculate sort-num on other columns, then set new sort-num
+         recalulate_sort_nums();
```
cedric-anne, line 106: I do not understand why this one is necessary. Maybe it should be done on line 66.

cedric-anne:
Although adding test on javascript code is really something we have in a near future, I think adding them now is a bit early. Indeed, DOM structure may change a lot, ans as DOM is mocked inside your tests, we may accidentally arrive at a situation where the test is done on a DOM structure that differs from the real one.

inc/search.class.php
```diff
@@ -3102,129 +3147,169 @@ static function addHaving($LINK, $NOT, $itemtype, $ID, $searchtype, $val) {
...
+    *    - order: The sort direction (Default: ASC). Invalid sort directions will be replaced with the default option
+    * @param ?integer $_id    field to add (Deprecated)
+    * @param string  $_order  order define (Deprecated)
     *
-    * @return select string
+    * @return string ORDER BY query string
     *
    **/
-   static function addOrderBy($itemtype, $ID, $order) {
+   static function addOrderBy($sort_fields, $_id = null, $_order = 'ASC') {
```
cedric-anne, line 3164: You should add an entry in CHANGELOG for this. Some plugins are using this method (`fusioninventory` for instance).

orthagh:
Some note after functionnal test:
- fixed first columns are indeed better with a background. By the way, a z-index issue exists with fixed headers row. 
image
 
I'm inclined to leave aside this part if the work needed is too long (z-index issues can be a pain). I prefer you pass time fixing the issues mentioned by @cedric-anne first, and so we can merge, and later work on fixed columns. The current pr is very impacting and so we need to focus on the merge (Also the #8858  is impacted).

- i got warnings in php-errors log file
```
[2021-05-19 15:22:03] glpiphplog.WARNING:   *** PHP Warning (2): foreach() argument must be of type array|object, null given in /var/www/html/glpi/10.0.git/inc/search.class.php at line 749
```

- you have a php cs issue, remove the `use` in top of ajax/search file.

inc/search.class.php
```diff
@@ -3102,129 +3147,169 @@ static function addHaving($LINK, $NOT, $itemtype, $ID, $searchtype, $val) {
...
-
-         case "glpi_networkequipments.ip" :
-         case "glpi_ipaddresses.name" :
-            return " ORDER BY INET_ATON(`$table$addtable`.`$field`) $order ";
-      }
+         }
 
-      //// Default cases
+         if (isset($CFG_GLPI["union_search_type"][$itemtype])) {
+            $criterion = "`ITEM_{$itemtype}_{$ID}` $order";
```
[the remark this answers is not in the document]
cconard96: Each sort/order criterion is appended to an array on line 3309 `$orderby_criteria[] = $criterion ?? "`ITEM_{$itemtype}_{$ID}` $order";` and then imploded into a final ORDER BY statement on line 3319 `return ' ORDER BY '.implode(', ', $orderby_criteria).' ';`. There may have been an instance of the criterion being overwritten by a later case, so I've added a check to each test to only run if there is no criterion yet.

js/Search/Table.js
cconard96: Moving to within that if/then block breaks sort completely. Cannot add/set any sort.

js/Search/Table.js
cconard96: Done

inc/search.class.php
```diff
@@ -1521,7 +1554,19 @@ static function displayData(array $data) {
             'criteria'     => Toolbox::stripslashes_deep($search['criteria']),
             'metacriteria' => Toolbox::stripslashes_deep($search['metacriteria'])
       ], '&');
-      $parameters = "sort={$search['sort']}&order={$search['order']}&{$globallinkto}";
+
+      $sort_string = '';
+      foreach ($search['sort'] as $sort_field) {
+         $sort_string .= $sort_field['searchopt_id'].(isset($sort_field['order']) ? '_'.$sort_field['order'] : '');
```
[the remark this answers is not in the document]
cconard96: We would have to break out the search option and direction wouldn't we? Otherwise the order of the sorts could be wrong? It was already suggested to use `?sort[0]=1&sort[1]=5&order[0]=ASC&order[1]=DESC` instead where the value of each `sort` is a search option and the value of each `order` is the sort direction, which I am OK with.

This was my response to that suggestion:
Using `?sort[0]=1&sort[1]=5&order[0]=ASC&order[1]=DESC` is OK with me as long as there isn't a risk that the total URL length won't be an issue. I was just trying to consolidate the sort and order options while keeping order for BC. Breaking them out would be much easier to parse on the client and server side or at least not using a comma separated string like I was before.

We may not need to worry about URL length for regular searches anymore (The POC for multi-sort was done before the POC for AJAX refresh) but we do for saved searches. After some research, the only browser that the limit may be run into would be Internet Explorer which isn't supported anyways.
IE - 2048 characters
Firefox, Chrome - 64K characters max for display but actually supports over 100k characters

inc/search.class.php
```diff
@@ -522,9 +532,11 @@ static function prepareDatasForSearch($itemtype, array $params, array $forcedisp
       }
 
       // Add order item
-      if (!in_array($p['sort'], $data['toview'])) {
-         array_push($data['toview'], $p['sort']);
+      if (!is_array($p['sort'])) {
+         $p['sort'] = [$p['sort']];
       }
```
cedric-anne, line 537: This conversion is already done in `foreach ($params as $key => $val) {` block.
```suggestion
```

inc/search.class.php
```diff
@@ -373,7 +373,7 @@ static function prepareDatasForSearch($itemtype, array $params, array $forcedisp
       // Default values of parameters
       $p['criteria']            = [];
       $p['metacriteria']        = [];
-      $p['sort']                = '1'; //
+      $p['sort']                = []; //
       $p['order']               = 'ASC';//
```
cedric-anne, line 377:
```suggestion
      $p['sort']                = ['1'];
      $p['order']               = ['ASC'];
```

cedric-anne (approved):

trasher (approved):

orthagh (approved):

PR #9086, opened 2021-05-19
Document: handle duplicate links in associated items tab

AdrienClairembault:
The "Associated items" tab of documents may contain duplicate lines.
This is because some items like followups or tasks are shown as their parent in this tab (since we need a link and we don't have link to this kind of item yet).
This means that if you have a document linked both to a followup and to its parent ticket then the ticket will be shown twice in the list: 
image

My suggestion is to add a type hint for followup, solutions and task so the users have a bit more information and are aware it isn't a duplicated line but two different links:
image

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !22093

### DIFF
--- a/inc/document_item.class.php
+++ b/inc/document_item.class.php
@@ -426,13 +426,16 @@ class Document_Item extends CommonDBRelation{
             }
 
             while ($data = $iterator->next()) {
+               $linkname_extra = "";
                if ($item instanceof ITILFollowup || $item instanceof ITILSolution) {
+                  $linkname_extra = "(" . $item::getTypeName(1) . ")";
                   $itemtype = $data['itemtype'];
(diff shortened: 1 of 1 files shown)


trasher (approved):

OscarBeiro:
I don't know if it is a feasible scenario but apparently, with this PR, if you link a document to two different tasks, it won't work.

AdrienClairembault:
Hello Oscar,

What do you mean by that ?
Does this change cause unexpected issues on your side or do you just mean that both rows will be identical like "Ticket XXX (Task)"  ?

Best regards,
Adrien

PR #9094, opened 2021-05-21
Dashboard/TicketSummary: don't filter waiting_validation on current user

AdrienClairembault:
image

On the "Ticket summary" card that you can see above, the "To validate" column is filtered on the current logged in user.
This doesn't make sense as the other four columns only filter on status.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !22124

### DIFF
--- a/inc/dashboard/grid.class.php
+++ b/inc/dashboard/grid.class.php
@@ -1286,6 +1286,9 @@ HTML;
             'provider'   => "Glpi\\Dashboard\\Provider::nbTicketsGeneric",
             'args'       => [
                'case' => $case,
+               'params' => [
+                  'validation_check_user' => true,
+               ]
             ],
             'cache'      => false,
             'filters'    => [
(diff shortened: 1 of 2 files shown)


cedric-anne (commented):
If you remove this filter, you also have to change this label: `__("Tickets waiting your validation")`.

orthagh (changes requested):
I disagree with this change as it impact some cards and it shouldn't

Ex:
- "Number of Ticket waiting for your approval"
- "List of Ticket waiting for your approval"

these two (at least) are impacted by your change and i think they should be kept as current.

The cause is `Glpi\Dashboard\Provider::genericTicket` is genericaly used by many cards and one change can be destructive

AdrienClairembault:
After last commit:
image

AdrienClairembault:
There is an issue with the list display that seems to be caused by this changes, looking into it.

AdrienClairembault:
Rebased and fixed, should be all good now:
image

orthagh (approved):

cedric-anne (approved):

PR #9142, opened 2021-06-06
[Modern UI] Convert Some Asset Forms

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

Relies on #9109. Commits after ff617d1 are for this PR only.

Converted the following itemtype forms to use the generic asset form:
- Rack
- Enclosure
- PDU
- Passive DC Equipment
- Unmanaged
- Appliance
- Domain
- Cluster
- Data Center
- Certificate
- Line
- Contract


templates/asset_form.html.twig
```diff
@@ -367,6 +598,141 @@
...
+                  {{ fields.dropdownHoursField('week_begin_hour', item.fields['week_begin_hour'], __('Start')) }}
+                  {{ fields.dropdownHoursField('week_end_hour', item.fields['week_end_hour'], __('End')) }}
+
+                  {{ fields.smallTitle(__('On Saturday')) }}
+                  {{ fields.dropdownYesNo('use_saturday', item.fields['use_saturday'], __('Use Saturday')) }}
+                  {{ fields.dropdownHoursField('saturday_begin_hour', item.fields['saturday_begin_hour'], __('Start')) }}
+                  {{ fields.dropdownHoursField('saturday_end_hour', item.fields['saturday_end_hour'], __('End')) }}
+
+                  {{ fields.smallTitle(__('Sundays and holidays')) }}
+                  {{ fields.dropdownYesNo('use_monday', item.fields['use_monday'], __('Use Sunday')) }}
```
cconard96, line 731: I'm not sure about this label. The group label mentions Sunday and Holidays, but the old fields in the group refer to Monday. In the old form, the dropdown was unlabeled. Is it just supposed to reefer to the start of the week which would be Sunday or Monday?

templates/asset_form.html.twig
```diff
@@ -101,6 +138,14 @@
                   ) }}
                {% endif %}
 
+               {% if item.isField('date_expiration') %}
+                  {{ fields.datetimeField('date_expiration', item.fields['date_expiration'], __('Expiration date'), {
+                     'add_field_html': Html__showToolTip(__('Empty for infinite')|nl2br, {
+                        'display': false
+                     })
```
orthagh, line 145:
```suggestion
                     'helper': __('Empty for infinite'),
```

templates/asset_form.html.twig
```diff
@@ -367,6 +598,141 @@
...
+               {% if item.isField('alert') and Entity_getUsedConfig('use_contracts_alert') %}
+                  {% set alert_html %}
+                     {{ displayLastAlert('Contract', item.fields['id']) }}
+                  {% endset %}
+                  {{ fields.dropdownArrayField('alert', item.fields['alert'], item.getAlertName(), __('Email alarms'), {
+                     'add_field_html': "<span class='text-muted'>" ~ alert_html ~ "</span>"
+                  }) }}
+               {% endif %}
+
+               {% if item.getType == 'Contract' %}
```
orthagh, line 718: regarding the whole part in the if block.
I rather move it into in a separate template included at the bottom of the page (before `components/form/buttons.html.twig` include)

and also insert the html into a div with `mx-n2` class to stick border of the tab

css/includes/components/_asset-form.scss
```diff
@@ -61,4 +61,8 @@
          }
       }
    }
+
+   .form-field-ajax > .mb-3 {
```
orthagh, line 65: I'm not very fan of this override, could you pass a specific class into the html loaded by ajax, or better remove `mb-3` (maybe adding a flag) in case of ajax load

inc/projecttasktemplate.class.php
```diff
@@ -199,30 +199,35 @@ function rawSearchOptions() {
...
+   public function getSpecificTypeField(int $ID, array $field): string {
       switch ($field['type']) {
          case 'percent_done' :
-            Dropdown::showNumber("percent_done", ['value' => $this->fields['percent_done'],
-                                                  'min'   => 0,
-                                                  'max'   => 100,
-                                                  'step'  => 5,
-                                                  'unit'  => '%']);
-            break;
+            return Dropdown::showNumber("percent_done", [
```
orthagh, line 205: do this change from displaying html to return it will cause some break in other glpi displays ?

templates/asset_form.html.twig
```diff
@@ -42,6 +51,24 @@
...
+               {% if item.isField('num') %}
+                  {{ fields.autoNameField(
+                     "num",
+                     item,
+                     _x('phone', 'Number'),
+                     withtemplate
+                  ) }}
+               {% endif %}
+
+               {% if item.isField('power_supplies') %}
```
orthagh, line 63: if it's possible to change, a `field.numberField` has a better ux (it may require to add support of max/min/step keys)

templates/asset_form.html.twig
orthagh: ex : 
image

templates/asset_form.html.twig
```diff
@@ -189,6 +250,121 @@
...
+                     'id': '__VALUE__',
+                     'current': current_position,
+                     'rand': rand
+                  })|raw }}
+                  {{ updateItemOnSelectEvent('dropdown_dcrooms_id'~rand, 'dropdown_locations_id'~rand, config('root_doc')~"/ajax/dropdownLocation.php", {
+                     'items_id': '__VALUE__',
+                     'itemtype': 'DCRoom'
+                  })|raw }}
+
+                  {% set inital_field %}
```
orthagh, line 278: initial ?

templates/asset_form.html.twig
```diff
@@ -189,6 +250,121 @@
...
+                     (constant('Rack::ROOM_O_EAST')): __('East'),
+                     (constant('Rack::ROOM_O_SOUTH')): __('South'),
+                     (constant('Rack::ROOM_O_WEST')): __('West'),
+                  }, {
+                     'rand': rand
+                  }) }}
+               {% endif %}
+
+               {% if item.isField('number_units') %}
+                  {{ fields.dropdownNumberField('number_units', item.fields['number_units'], __('Number of units'), {
```
orthagh, line 313: same prefer `numberField`

templates/asset_form.html.twig
```diff
@@ -189,6 +250,121 @@
...
+               {% endif %}
+
+               {% if item.isField('mesured_power') %}
+                  {{ fields.textField('mesured_power', item.fields['mesured_power'], __('Measured power (in watts)')) }}
+               {% endif %}
+
+               {% if item.isField('max_weight') %}
+                  {{ fields.textField('max_weight', item.fields['max_weight'], __('Max. weight')) }}
+               {% endif %}
+
```
orthagh, line 345: all these fields could be with a `numberField` also

templates/asset_form.html.twig
```diff
@@ -367,6 +598,141 @@
...
+               {% if item.isField('accounting_number') %}
+                  {{ fields.autoNameField(
+                     "accounting_number",
+                     item,
+                     __('Account number'),
+                     withtemplate
+                  ) }}
+               {% endif %}
+
+               {% if item.isField('periodicity') %}
```
orthagh, line 658: to precise with the above comments with `numberFields`, this field is a perfect usage of `dropdownNumberField` as we add textual options

templates/asset_form.html.twig
orthagh: This is indeed strange, regarding the whole old table
we set opened hours for a support on 3 lines
- during week (not possible to disable)
- saturday, with a yes/no to disable 
- sunday and holidays, with also a yes/no

I have no idea why this is called use_monday
I'm note sure this is used for any hardcoded behavior, but rather just a field to save data.

I am for renaming the field to `use_sunday`

templates/asset_form.html.twig
orthagh: also the display is alterered compared to previous html.
I think we need to have one colum with the yes/no part and another with the hours fields

templates/components/form/fields_macros.html.twig
```diff
@@ -18,6 +18,22 @@
    </div>
 {% endmacro %}
 
+{% macro smallTitle(label, icon = '') %}
```
orthagh, line 21: good addition, but i think `mt-2` should be sufficient

templates/asset_form.html.twig
```diff
@@ -367,6 +598,141 @@
...
+                     {{ displayLastAlert('Contract', item.fields['id']) }}
+                  {% endset %}
+                  {{ fields.dropdownArrayField('alert', item.fields['alert'], item.getAlertName(), __('Email alarms'), {
+                     'add_field_html': "<span class='text-muted'>" ~ alert_html ~ "</span>"
+                  }) }}
+               {% endif %}
+
+               {% if item.getType == 'Contract' %}
+                  {# Support hours #}
+                  {{ fields.largeTitle(__('Support hours')) }}
```
orthagh, line 720: add `fas fa-business-time` icon

templates/components/form/fields_macros.html.twig
```diff
@@ -305,6 +400,100 @@
...
+         'rand': rand,
+         'width': '100%',
+         'class': 'form-select',
+         'value': value
+      }|merge(options)) }}
+   {% endset %}
+
+   {{ _self.field(name, field, label, options|merge({'id': 'dropdown_' ~ name ~ '_' ~ options.rand})) }}
+{% endmacro %}
+
```
orthagh, line 496: regarding the whole part of dropdown fields, could you move them into a separate macro template (included in the main one) ?

inc/datacenter.class.php
```diff
@@ -57,28 +59,11 @@ function defineTabs($options = []) {
...
-      Html::autocompletionTextField($this, "name", ['rand' => $rand]);
-      echo "</td>";
-
-      echo "<td><label for='dropdown_locations_id$rand'>".Location::getTypeName(1)."</label></td>";
-      echo "<td>";
-      Location::dropdown([
-         'value'  => $this->fields["locations_id"],
-         'entity' => $this->fields["entities_id"],
-         'rand'   => $rand
+      TemplateRenderer::getInstance()->display('asset_form.html.twig', [
```
orthagh, line 63: datacenter form layout is currently broken, dont' know why precisely

orthagh (changes requested):

orthagh:
To note, i merged #9109, because i needed dbextension on my side (and i wanted to avoid superflu rebases), comments are still valid for the current pr

templates/components/form/fields_macros.html.twig
cconard96: May be possible if fields were defined as blocks, but I don't know how to get macro inheritance to work properly. You have to import macros in the template that is going to use them.

inc/projecttasktemplate.class.php
cconard96: It is possible for some plugin to have issues with this if they are using it for their own forms, but I don't see any instances of other GLPI core code calling this method or in the plugins I have in my workspace.

templates/asset_form.html.twig
cconard96: Done. The field name `use_monday` will need updated in the datainjection plugin's hook file. I didn't see these fields used in other plugins.

inc/projecttasktemplate.class.php
orthagh: ok for me

orthagh (approved):

PR #9179, opened 2021-06-16
21861 dashboard filter in db and cards sla per tech group tech

btry:
- migrates dashboard filters to database
- add a private mode to dashboards. A private is invisible to other users except if it is explicitly shared.
- adds 2 dashboard cards to see 
  - count of tickets by SLA status and technician
  - count of tickets by SLA status and technician group

- adds 2 new graph types
  -  multiple horizontal bars
  - horizontal stacked bars

image

Private mode 
image


inc/dashboard/filter.class.php
```diff
@@ -164,7 +167,7 @@ static function user_tech(string $value = ""): string {
       ]);
    }
 
-   static function dropdown(
+   static function itemDropdown(
```
btry, line 170: Method renamed because of a conflict with the implementation of the ancestor. Before  this PR, the class does not have any ancestor.

I dislike this change, but it is necessary. Maybe I can improve it by finding a better name.

inc/dashboard/provider.class.php
```diff
@@ -413,6 +416,317 @@ static function nbTicketsGeneric(
...
+            "$userTable.id",
+         ],
+         'ORDER' => [
+            'sla_state DESC',
+            'cpt DESC',
+            "$userTable.id",
+         ]
+      ];
+
+      unset($params['apply_filters']['group_tech']);
```
btry, line 500: this card is not compatible with group_tech filter. To prevent false results, it is necessary to not apply the filter.

I did not find a proper way to do this. Comments welcome

inc/dashboard/provider.class.php
```diff
@@ -413,6 +416,317 @@ static function nbTicketsGeneric(
...
+            "$groupTable.id",
+         ],
+         'ORDER' => [
+            'sla_state DESC',
+            'cpt DESC',
+            "$groupTable.id",
+         ]
+      ];
+
+      unset($params['apply_filters']['user_tech']);
```
btry, line 650: this card is not compatible with user_tech filter. To prevent false results, it is necessary to not apply the filter.

I did not find a proper way to do this. Comments welcome

inc/dashboard/widget.class.php
```diff
@@ -878,20 +878,21 @@ private static function getBarsGraph(
...
+         'multiple'          => false,
+         'stacked'           => false,
+         'horizontal'        => false,
+         'distributed'       => false,
+         'use_gradient'      => false,
+         'point_labels'      => false,
+         'limit'             => 99999,
+         'filters'           => [],
+         'rand'              => mt_rand(),
+         'truncate_boundary' => 'beginning',
```
btry, line 895: new option : the 2 new dashboard cards rely on ordering to show the most important data. However, the code truncates the data on the wrong bound (because it was designed to work on data ordered by date / time).

orthagh:
## functionnal review

**Regarding the private dashboard feature:**

- private field should be a input[type=checkbox] available on dashboard creation:
image

- when the dashboard already exists, i think the control to toggle the privacy should be in the share modal:

image

when the "is private" is check, the share with user/group/entities should be disabled (grayed out).
I think adding a dedicated control in the dashboard toolbar is not necessary and pollute the UX

- In both place, the field should have an helper (icon displaying a tooltip) to explains inline what set a dashboard as private means.


**Regarding the filter in db:** nothing to say on functional side

**Regarding the new cards:**

I'll talk about the two card, as they display very similar data.
i think the current two widget options are not very convenient as the presented data are not time based.
I think, in this particular case, a vertical alternative for stackedBar and Multiplebar widget could be better

example with sla/status/group card (blue part is an obsfucation of data):
image

The addition of these two new widgets should be easy i think, it should consist to:

- add a new dedicated function:
```php
   public static function horizontalStackedBars(array $params = []): string {
      return self::StackedBars(array_merge($params, [
         'horizontal' => true,
      ]));
   }
```

- declare the functions in `Widget::getAllTypes`
- use it the card declaration (in `Grid::getAllDasboardCards`)
- add picture for new widget (maybe the hardest part 😁 )

inc/dashboard/dashboard.class.php
```diff
@@ -487,4 +519,22 @@ static function importFromJson($import = null) {
 
       return true;
    }
+
+   public function setPrivate($is_private) {
+      $this->load();
+
+      return $this->update([
+         'id'       => $this->fields['id'],
+         'key'     => $this->fields['key'],
```
orthagh, line 528:
```suggestion
         'key'      => $this->fields['key'],
```

inc/dashboard/filter.class.php
```diff
@@ -255,4 +258,47 @@ static function field(
 
       return $html.$js;
    }
+
+   /**
+    * Return rights for the provided dashboard
```
orthagh, line 263:
```suggestion
    * Return filters for the provided dashboard
```

inc/dashboard/provider.class.php
```diff
@@ -413,6 +414,301 @@ static function nbTicketsGeneric(
...
 
+   public static function nbTicketsByAgreementStatusAndTechnician(
+      string $case = "",
+      array $params = []
+   ):array {
+      $DBread = DBConnection::getReadConnection();
+
+      $default_params = [
+         'label'         => "",
+         'icon'          => 'far fa-clock',
```
orthagh, line 425:
```suggestion
         'icon'          => 'fas fa-stopwatch',
```

inc/dashboard/provider.class.php
```diff
@@ -413,6 +414,301 @@ static function nbTicketsGeneric(
...
+
+   public static function nbTicketsByAgreementStatusAndTechnicianGroup(
+      string $case = "",
+      array $params = []
+   ):array {
+      $DBread = DBConnection::getReadConnection();
+
+      $default_params = [
+         'label'         => "",
+         'icon'          => 'far fa-clock',
```
orthagh, line 577:
```suggestion
         'icon'          => 'fas fa-stopwatch',
```

orthagh (changes requested):

btry:
@orthagh : I changed the cards to use only horizontal bars (multiple or stacked). In this mode, boundary option I introduced truncates the data on the wrong boundary, then I reverted it.

Code cleanup, method naming are fixed. I'll work soon on the private feature to apply UI changes. If you have opportunity, please see if the 2 new horizontal display modes are working for you.

inc/dashboard/dashboard.class.php
```diff
@@ -426,11 +456,6 @@ static function convertRights(array $raw_rights = []): array {
     * @return bool
     */
    static function checkRights(array $rights = []): bool {
-      // check global (admin) right
-      if (self::canView()) {
-         return true;
-      }
-
```
btry, line 433: To be checked carefully : rights checks are spread on several places.

Here I removed the canView() call because CheckRights is called in CanViewCurrent(). This method already calls canView().

In getAll() I replaced checkRights() by a call to canViewCurrent() to properly check the private status. This changes makes taht the checks includes the view all dashboards right, which was not checked before. I expect validation from reviewers.

Finally, with all the right check changes, the method checkRights may be merged with canViewCurrent() or renamed into checkShareRights().

btry: ping @orthagh

orthagh: should be ok

orthagh (approved):
ok for me.

Need at least : 

- reword of private in "Personal"
- some tests on new providers

orthagh (approved):
ok for merge

tests/units/Glpi/Dashboard/Provider.php
```diff
@@ -0,0 +1,293 @@
...
+            'id' => $ticket->getID()
+        ]);
+        $ticket->getFromDB($ticket->getID());
+        $output = \Glpi\Dashboard\Provider::nbTicketsByAgreementStatusAndTechnicianGroup();
+        $this->integer($output['data']['series'][0]['data'][0])->isEqualTo(1);
+        $this->integer($output['data']['series'][1]['data'][0])->isEqualTo(0);
+        $this->integer($output['data']['series'][2]['data'][0])->isEqualTo(0);
+        $this->integer($output['data']['series'][3]['data'][0])->isEqualTo(0);
+    }
+}
```
orthagh, line 293:
```suggestion
}

```

inc/dashboard/filter.class.php
```diff
@@ -255,4 +258,47 @@ static function field(
...
+      global $DB;
+
+      $dr_iterator = $DB->request([
+         'FROM'  => self::getTable(),
+         'WHERE' => [
+            'dashboards_dashboards_id' => $dashboards_id,
+         ]
+      ]);
+
+      $settings = $dr_iterator->next()['filter'];
```
cedric-anne, line 279: Unless I am wrong, filter can be missing in DB. For instance, migration does not create a filter for every existing dashboard.
```suggestion
      $settings = $dr_iterator->count() === 1 ? $dr_iterator->next()['filter'] : null;
```

js/dashboard.js
```diff
@@ -396,9 +400,10 @@ var Dashboard = {
          filter.remove();
 
          // remove filter from storage and refresh cards
-         var filters = Dashboard.getFiltersFromStorage();
+         var filters = Dashboard.getFiltersFromDB();
          delete filters[filter_id];
-         Dashboard.setFiltersInStorage(filters);
+         Dashboard.setFiltersInDB(filters);
+         //Dashboard.setFiltersInStorage(filters);
```
cedric-anne, line 406:
```suggestion
```

js/dashboard.js
```diff
@@ -659,9 +664,10 @@ var Dashboard = {
 
    saveFilter: function(filter_id, value) {
       // store current filter in localStorage
-      var filters = Dashboard.getFiltersFromStorage();
+      var filters = Dashboard.getFiltersFromDB();
       filters[filter_id] = value;
-      Dashboard.setFiltersInStorage(filters);
+      Dashboard.setFiltersInDB(filters);
+      //Dashboard.setFiltersInStorage(filters);
```
cedric-anne, line 670:
```suggestion
```

js/dashboard.js
```diff
@@ -1043,64 +1049,63 @@ var Dashboard = {
...
+            var filters     = Dashboard.getFiltersFromDB();
             var new_filters = {};
             $.each(items_after, function() {
                var filter_id = $(this).data('filter-id');
                new_filters[filter_id] = filters[filter_id];
             });
 
-            Dashboard.setFiltersInStorage(new_filters);
+            Dashboard.setFiltersInDB(new_filters);
+            //Dashboard.setFiltersInStorage(new_filters);
```
cedric-anne, line 1060:
```suggestion
```

cedric-anne (approved):

trasher (approved):

trasher:
Could you please rebase to fix conflicts?

PR #9207, opened 2021-06-25
Dhtmlx Gantt integration

SpectrumZX81:
New feature for v9.5.5

### DIFF
--- a/ajax/gantt.php
+++ b/ajax/gantt.php
@@ -1,0 +1,294 @@
+<?php
+
+/**
+ * ---------------------------------------------------------------------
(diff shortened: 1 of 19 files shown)


trasher:
I do not understand the need for this one; it duplicates #8978.

SpectrumZX81:
Glpi v9.5.5 has some important fixes in its ticketing system compared to v9.5.4, therefore Smartwood is about to migrate to v9.5.5. 
They also need the last features added to Gantt, so they would like you to validate these changes before they switch version.

trasher:
As far as I've seen, this one does not differs at all from #8978; and all discussions are there, so I close this one.

PR #9210, opened 2021-06-25
Fix fields consistency on add/update

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

`CommonDBTM::$fields` property may be inconsistent. When adding or updating such an object; `fields` property is populated from `input` property, which is sanitized. When loading from DB, the `fields` property is loaded verbatim (ie. not sanitized).
This can cause difficult to handle cases, with required extra decoding.

The change may seem huge; but I do not think this would break. Indeed, this change would just make any extra `stripslashes` (and others) useless; calling such methods on clean data should not be a big issue; there were only a few problems in core right now.

### DIFF
--- a/inc/commondbtm.class.php
+++ b/inc/commondbtm.class.php
@@ -574,6 +574,7 @@ class CommonDBTM extends CommonGLPI {
 
       if (count($oldvalues)) {
          Log::constructHistory($this, $oldvalues, $this->fields);
+         $this->getFromDB($this->fields['id']);
       }
 
       return true;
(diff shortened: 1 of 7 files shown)


orthagh (approved):

btry (approved):

trasher:
I made extra tests on some cases on my side I did not see any issue (but maybe I've just been lucky :D)

cedric-anne (approved):

4lex11s:
Hello,

Thank you very much for your correction.
However if we use a particular attribute on user, the problem persists.
Could you also consider fixing inc /user.class.php or other please?

To reproduce the problem:
- LDAP connection
- take a user with the attribute department = "24-IT_department \ 01-IT_production"
- Assign to a group automatically

PR #9242, opened 2021-07-02
Allow validators to add followups

AdrienClairembault:
Validators aren't allowed to add followups to tickets.
It seems useful to allow them to add followups for the tickets they validate, in case they need to add some additional informations later.

Changes:
* Allow validator to add followups for tickets they validate (new function CommonITILObject::isValidator + tests)
* Improve readability of canAddFollowups conditions
* Add a new `createItems` function in tests to avoid writing the same boiler plate code every tests.

Side note as it isn't related to the PR changes but having to check "manually" if validation is enabled on an itemtype (`if (!$this instanceof Ticket && !$this instanceof Change)` and building "manually" the validation class (`$validation_class = static::class . "Validation";`) should IMO be done with more robust polymorphic functions like `$this->supportValidation()` and `$this->getValidationClass()`.
Didn't want to add them and replace the existing code here as it is a bugfix but maybe a possible improvement in master 😉 .

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !22325

### DIFF
--- a/inc/commonitilobject.class.php
+++ b/inc/commonitilobject.class.php
@@ -205,17 +205,56 @@ abstract class CommonITILObject extends CommonDBTM {
     * @return boolean
     */
    function canAddFollowups() {
-      return ((Session::haveRight("followup", ITILFollowup::ADDMYTICKET)
-               && ($this->isUser(CommonITILActor::REQUESTER, Session::getLoginUserID())
-                   || (isset($this->fields["users_id_recipient"])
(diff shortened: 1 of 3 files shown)


trasher (approved):

cedric-anne (approved):

orthagh (approved):

trasher:
Could you please rebase and fix conflict?

PR #9247, opened 2021-07-03
[Modern UI] High contrast

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

Re-implement high contrast option.

Current State (Needs some work for dark themes)
| | Auror | Dark Auror | Midnight |
| - | ------- | -------------- | ------------ |
| Computer Form (Disabled) | Screenshot from 2021-07-03 11-05-54 | Screenshot from 2021-07-03 11-05-17 | Screenshot from 2021-07-03 11-04-45 |
| Computer Form (Enabled) | Screenshot from 2021-07-06 21-54-17 | Screenshot from 2021-07-06 21-51-35 | Screenshot from 2021-07-06 21-58-35 |


orthagh:
on light theme, missing borders:
- top right controls
- buttons
- dropdown sub buttons

on dark themes, same but also inputs need more contrasted borders

cconard96:
Screenshots were updated after the last commit.

orthagh:
much better 👍🏻 

could you check also tinymce rendering, i found the border differs on light theme but relatively to this pr, we need to match also high contrast borders

orthagh:
after a short test of the current branch, here is a list of items which can be also bordered:

- pager links in list
- mini pager buttons on top of the page when viewing an asset for example
- actors custom input in itil object (i think it's generalized to multiple selects in fact - check "associable items" in asssistance tab of profiles
- accordion headers
- timeline items
- timeline mini buttons
- horizontal tabs

this may be from a legacy style to clean but in auror theme, preferences have some dark backgrounds in the content

cconard96:
> this may be from a legacy style to clean but in auror theme, preferences have some dark backgrounds in the content

It was the same in the old implementation. I just used the old styles and replaced the hard-coded colors with variables mostly. I think the dark background was meant only for making striped backgrounds for table rows and the user preferences form just happened to match the same selectors. The original styles had just a tab_bg_2 selector for this.
Since the new forms use flex for layout instead of tables, I think the best solution is continuing to convert forms from using tables for layout to flex.
For the original purpose, we no longer have alternating tab_bg_1 and tab_bg_2 classes on table rows it seems so we should just rely on odd or even selectors.
I think I've resolve this issue in my last commit.

orthagh:
Ok, we're almost done i think.

For table, stripped style is a little too much on light theme.
I suggest to keep as when the high contrast is not enabled but add border on tr (on search table)
Ex, just by adding table-striped:
image

blue highlight need to be addressed also.
For form-check, just add `border: 1px solid` without color to have something like that:
image

And last thing i saw, remove border radius here:
image

cconard96:
Did you want the search table to always have `table-striped` and then add a darker border for high contrast in addition?

orthagh:
> 
> 
> Did you want the search table to always have `table-striped` and then add a darker border for high contrast in addition?

Yes please

PR #9248, opened 2021-07-03
{Modern UI] Timeline images

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

Re-add thumbnails to timeline and re-add the ability to open images in a lightbox.

### DIFF
--- a/templates/components/itilobject/timeline/form_document_item.html.twig
+++ b/templates/components/itilobject/timeline/form_document_item.html.twig
@@ -13,11 +13,16 @@
             {% set is_image = isImage(entry_i['filepath']) %}
             <div class="col text-truncate">
                {% if is_image %}
+                  {% set img_size = getImageSize(entry_i['filepath']) %}
                   {% include 'components/photoswipe.html.twig' with {
                      'imgs': [{
                         'title': '',
                         'thumbnail_src': docpath ~ '&context=timeline',
+                        'thumbnail_w': 'auto',
+                        'thumbnail_h': 'auto',
                         'src': docpath,
+                        'w': img_size[0],
+                        'h': img_size[1],
                         'clearable': false,
                      }]
                   } %}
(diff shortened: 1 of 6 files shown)


templates/components/itilobject/timeline/form_document_item.html.twig
```diff
@@ -9,11 +9,30 @@
...
+                        'thumbnail_src': docpath ~ '&context=timeline',
+                        'thumbnail_w': 'auto',
+                        'thumbnail_h': 'auto',
+                        'src': docpath,
+                        'w': img_size[0],
+                        'h': img_size[1],
+                        'clearable': false,
+                     }]
+                  } %}
+                  <a href="{{ docpath }}" target='_blank'>{{ name }}</a>
```
orthagh, line 29: one thing here, in case of picture, i'm not sure having a plain text under is very usefull.
If needed, i think a title attribute should be sufficient (you can add tootltip toggle also if needed)

orthagh (commented):
I tested within dom inspector on my browser, it hink we can add to these specific image a `shadow` class to add a subtle separation from background.

On global design, i'm not convinced we should keep the list design used for classic documents.
I think i prefer something like that:
paint master :)

where:
- pictures are horizontally aligned
- controls are displayed vertically right of image (keep hover effect)
- in case of documents and images, we display first images then documents list

cconard96:
Images are now grouped together first in a single Photoswipe gallery, followed by a list of other documents.
Screenshot from 2021-07-09 19-08-50

First image is being hovered over. The images have a `shadow` class but this doesn't seem like a good thing for dark palettes. Currently, the shadows are always a partially transparent black which is not visible in the `midnight` palette and may be difficult to see with other dark palettes. I experimented with using a partially transparent version of the palette's text color as the shadow color but I don't think the glow looks good.

Screenshot from 2021-07-09 19-33-28

Should we make this a subtle border in dark palettes instead of a shadow?

orthagh:
no it's ok i think.

PR #9266, opened 2021-07-08
Enhance getById method

btry:
An improvement to set a return type on the method.

Add unit tests

It looks like that the method was added while implementing pending reasons. A lot of calls were found in  the code.

I examined all of them
- return value of some calls is is assumed an object, without any check. I assume is is explicitly expected the method suceeeds, thanks to the call context.
- return value is tested as a boolean in if () expressions. As null == false, there is no problem changing the failure return value into null.

### DIFF
--- a/inc/commondbtm.class.php
+++ b/inc/commondbtm.class.php
@@ -5661,17 +5661,17 @@ class CommonDBTM extends CommonGLPI {
     *
     * @param integer $ID ID of the item to get
     *
-    * @return static|boolean false on failure
+    * @return CommonDBTM|null false on failure
    */
-   public static function getById(?int $id) {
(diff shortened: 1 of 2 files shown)


AdrienClairembault (approved):

cedric-anne (approved):

orthagh (commented):
I'm not for this change, but if others agree, you could merge

btry:
Discussion about this PR shows that GLPI usually returns false on error, and making this method return null makes it unusual. It is already used then despite it will not break anything, extended unit tests are required.

Then it is better to wait that GLPI required PHP >= 8.0 to typehint with union types (bool|stdclass)

PR #9273, opened 2021-07-10
Kanban Filtering System

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

This PR is open for a review of the specification of this feature before the implementation is finalized.
Selection_003

The goal is to add a more in-depth filtering system to the Kanban than just a simple text search that only searches the card item's name and content. This is especially useful for "global" Kanbans where a lot of items are shown at once.

A filter system is implemented in the tasklists plugin on top of the copied Kanban code from GLPI 9.5.0. Adding filters natively should help reduce their plugin's code complexity and allow them to build off the feature without needing to maintain a copy of the Kanban codebase (bug and security fixes).

Filters are defined in each itemtype with the only default-defined filters being for `name` and `content`. All filterable data is passed in a `_metadata` property to the Kanban code. Filters are defined in groups (General is the default group for core filters) and there is a new plugin hook `kanban_filters` so plugins that add data to the Kanban like the Tag plugin can add their own filters (The tag filter is being developed in parallel to this PR).
Filters are displayed in the panel by group and each group can be collapsed or expanded.

The filter panel is empty by default and new filters can be added from a dropdown list. Filters that are added can be removed with a button. Filter criteria are displayed one per line in a way that is presented in  *mostly* natural way. In the example screenshot, the criteria are read naturally as "Name Contains Filter". All filters are currently evaluated together so that all criteria need to be met for a card to be visible.

The current "Search" textbox could stay as a "quick search" which acts as a `contains` or `equals` filter for any of the metadata properties. It could also act similarly to the search in Github for Issues and Pull Requests where filters are shown as a text equivalent alongside a text search such as "Kanban is:open is:pr label:enhancement". This way, the filter panel doesn't need to be opened to see the applied filters.

Supported Filter Types:
- [ ] Boolean
  - [ ] Equals
  - [ ] Does not equal
- [ ] Number
  - [ ] Equals
  - [ ] Not equal to
  - [ ] Less than
  - [ ] Greater than
  - [ ] Less than or equal to
  - [ ] Greater than or equal to
- [x] Text
  - [x] Contains
  - [x] Does not Contain
  - [x] Equals
  - [x] Does not equal
  - [x] Starts with
  - [x] Ends with
  - [x] Matches (Regex pattern matching)
- [ ] Dropdown
  - [ ] Equals
  - [ ] Does not equal
- [ ] Dropdown (Multiple)
  - [ ] Contains any
  - [ ] Contains all
  - [ ] Not contains any
  - [ ] Not contains all
- [ ] Datetime
  - [ ] Equals
  - [ ] Does not equal
  - [ ] On date (Ignore time part)
  - [ ] Before
  - [ ] After
- [ ] Duration
  - [ ] Same criteria types as numbers

Filterable Fields for Projects:
- [x] Name (text)
- [x] Content (text)
- [ ] Is Milestone (boolean)
- [ ] Planned Start/End Date (datetime)
- [ ] Real Start/End Date (datetime)
- [ ] Planned/Effective Duration (duration)
- [ ] Percent Done (number)

Some improvements that can be made in another PR:
- Build card content using information from the provided metadata (Use a twig template?) instead of passing static HTML content from a PHP function. Right now the name and content for items are both hard-coded into the card content and in the metadata.
- Separate filtering logic into its own JS class. The code for filtering can easily be applied to any array of properties so it is reusable by any front-end feature if needed (Impact graph, Gannt, etc). The Twig template code for the filters could be generalized too.


orthagh:
Hello curtis, i tested quickly this pr and though about the current ux/ui
I'm not sure it's very convenient atm and if we can continue in this direction.

After some discussions with the team on mattermost, maybe we can merge these filters in the current input[type=text] bar with some keywords/query language, autocompletion, and possibly tokenisation.

I tried this morning to find a js component fitting the need without much sucess. here is my findings:
- https://documentcloud.github.io/visualsearch/ -> very dated but ux is the one i seek
- https://yaireo.github.io/tagify/ -> recent, well maintened but only the part with tokenisation, i think, some example are interesting (mix mode, user list)

The last lib may me think we could start with some query language to match the filter and add tagify to render them
i'm not sure it's still feasible by the way. If you find any other component to do this, I'll take a look.

For example, discord search feature has the kind of behavior i seek

orthagh:
ex from discord: 
image

PR #9274, opened 2021-07-11
[Modern UI] Muliple FIxes

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

Multiple fixes for modern UI branch

Fixes:
- Plugin menu icon was not shown because the `fas` class was missing on the icon element
- The inclusion of some TInyMCE plugins was causing a JS console warning since they have been included in the TinyMCE core since 5.0 (3 years ago).
- Implementing ExtensionInterface on GLPI's Twig extensions is redundant because AbstractExtension already implements this interface.
- Use static closures when access to instance properties are not required to improve performance.
- Add/fix multiple parameter and return types
- Fixed ignored `entity_restrict` in `ItemtypeExtension::getMassiveActions`
- Replaced `width` HTML attribute which is deprecated with the equivalent style attribute value
- Fixed HTML attribute name `roel` to `role`

Best reviewed by commit.

### DIFF
--- a/inc/application/errorhandler.class.php
+++ b/inc/application/errorhandler.class.php
@@ -347,7 +347,7 @@ class ErrorHandler {
          $exit_code = $this->exit_code;
          register_shutdown_function(
             'register_shutdown_function',
-            function () use ($exit_code) { exit($exit_code); }
+            static function () use ($exit_code) { exit($exit_code); }
          );
       }
    }
--- a/inc/application/errorhandler.class.php
+++ b/inc/application/errorhandler.class.php
@@ -140,7 +140,7 @@ class ErrorHandler {
     *
     * @return void
     */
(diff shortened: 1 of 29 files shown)


orthagh:
i just recreated the branch from scratch, @cconard96 could you check i didn't missed any commits ?

orthagh:
Apart 085b900530da67a14c99096711e67f97c0a7c2f7 commit i don't understand (yet), the rest are ok and the PR ok for merge

cconard96:
> Apart 085b900530da67a14c99096711e67f97c0a7c2f7 commit i don't understand (yet), the rest are ok and the PR ok for merge

The entity_restrict input was being set but never used or returned.

All commits seem to be here.

PR #9283, opened 2021-07-13
Fix default picture on tickets

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes


cedric-anne (commented):
These are the only calls using this second parameter that was introduced by this PR. So if this change is functionnally correct, this second parameter should be removed from method calls and also from method signature itself.

inc/user.class.php
```diff
@@ -5212,7 +5211,7 @@ static function getThumbnailURLForPicture(string $picture = null, bool $send_pla
          }
       }
 
-      return $send_placeholder ? $CFG_GLPI["root_doc"]."/pics/picture_min.png" : "";
+      return $CFG_GLPI["root_doc"]."/pics/picture_min.png";
```
trasher, line 5214:
```suggestion
      return "";
```

orthagh (approved):

PR #9308, opened 2021-07-21
To prevent followup creation form after re-open of an item

tomolimo:
When an item is re-open (like a ticket) it proposes a followup to input reason, but after post of the followup, it proposes again a followup creation.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #


cedric-anne (commented):
I am OK with this change. Maybe it should also be applied to line 64.

orthagh (approved):

PR #9313, opened 2021-07-22
Fix duplicated saved searched

AdrienClairembault:
Before:
image

After:
image

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !22341


AdrienClairembault:
The fix caused the search engine to add an alias for this join, which made the code in the `addWhere` function fail as it can't use this alias.

Removing this function doesn't break the search options 11, it works just as good.

orthagh (approved):

PR #9320, opened 2021-07-23
migrate kanban to tabler style

orthagh:
Hello @cconard96 

I open this pr separately from modern-ui branch to let you validate the changes.
Works done:
- separate css file included in base style
- usage of bootstrap var in css and classes in generated html
- addition of a frontground color for column to have a good contrast on the badges

Ex with auror_dark theme:
image

I dedicated changes to only styles of kanban
but i found some issues:

- addition of tasks/projects triggers an error:

```
Uncaught TypeError: self.supported_itemtypes[itemtype] is undefined
    showAddItemForm kanban.js:1604
```

- click on a card title give an relative to jquery ui dialog (removed in modern-ui) a migration for a bootstrap dialog is needed here
```
Uncaught TypeError: $(...).dialog is not a function
```

I let you review this pr first but fixing the above issues is up to you i think (it will take a lot of time from me)


inc/project.class.php
```diff
@@ -2157,7 +2158,7 @@ static function getKanbanColumns($ID, $column_field = null, $column_ids = [], $g
          $card = [
             'id'              => "{$itemtype}-{$item['id']}",
             'title'           => '<span class="pointer">'.$item['name'].'</span>',
-            'title_tooltip'   => Html::resume_text(RichText::getTextFromHtml($item['content'], false, true, true), 100),
+            'title_tooltip'   => Html::resume_text(RichText::getTextFromHtml($item['content'] ?? "", false, true, true), 100),
```
orthagh, line 2161: i got errors here with `content` was equals to null

templates/components/table.html.twig
```diff
@@ -1,4 +1,4 @@
-<div class="table-responsive">
+<div class="table-responsive card-table">
```
orthagh, line 1: not related to kanban, if this bother you i can separate to another commit

cconard96:
The dialog issue was fixed in one of the other open PRs. I can add it here.

orthagh:
you can push to the current pr

cconard96:
I don't have permission to push so I made some changes in another branch.
https://github.com/cconard96/glpi/tree/feat/modern-ui%2Bkanban_UI

Changes:
- Fixed showing the new item form (Caused by a bad selector so clicking the text itself didn't work)
- Added a generic modal macro
- Moved display of Kanban to a template that includes a generic modal for the Kanban to use
- Fixed the progress bar colors

I approve all changes made in this PR.

cconard96 (approved):

orthagh:
pushed, one more thing i forgot in my first comment:
we have some references to old dark mode management we need to clean

cconard96:
Dark mode detection without looking for a specific palette:
https://github.com/cconard96/glpi/commit/f8cbdfb0b6e4c2c48ee79d2ef25fda1863f6bc7c

orthagh:
@cconard96 i removed your commit about progressbar to do a more bootstrap compliant one

PR #9325, opened 2021-07-23
Twig support in emplates: Fix massive actions

AdrienClairembault:
When adding #9291, we forgot a specific case: it is possible to solve a ticket using a template through the massive actions.

This is not supported by the initial PR because we render the template when it is loaded into the form but this is not possible for this case as we do not have a given item to use as parameter (instead we have multiples items and the template should be rendered differently for each of these items).

The proposed solution is, for this specific case, to not render the template in the form.
Instead, we do it in the `prepareInputForAdd` function.

Example:
1) Using massive actions:
image
2) Ticket 102:
image
3) Ticket 95: 
image

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |

### DIFF
--- a/ajax/itilfollowup.php
+++ b/ajax/itilfollowup.php
@@ -44,9 +44,15 @@ Html::header_nocache();
 Session::checkLoginUser();
 
 // Mandatory parameter: itilfollowuptemplates_id
-$itilfollowuptemplates_id = $_POST['itilfollowuptemplates_id'] ?? 0;
-if (!$itilfollowuptemplates_id) {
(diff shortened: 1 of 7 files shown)


trasher (approved):

orthagh (commented):
As discussed, we may need autocompletion for solution form in massive action

AdrienClairembault:
Done, also added another fix to handle better when the user select the "---" value of the template dropdown.

orthagh (approved):

PR #9329, opened 2021-07-26
Fix OS meta criteria

stonebuzz:
Meta criteria for OS no longer work

GLPI tries to make the relation (LEFT JOIN) from ```glpi_computers_items``` instead of ```glpi_items_operatingsystems```

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | Internal 22385

### DIFF
--- a/inc/search.class.php
+++ b/inc/search.class.php
@@ -5388,6 +5388,24 @@ JAVASCRIPT;
          return $JOIN;
       }
 
+      if ($from_referencetype === 'Computer' && $to_type === 'OperatingSystem') {
+         // From OperatingSystem to glpi_items_operatingsystems
+         $operatingsystemitems_alias = "glpi_items_operatingsystems{$alias_suffix}";
+         if (!in_array($operatingsystemitems_alias, $already_link_tables2)) {
+            array_push($already_link_tables2, $operatingsystemitems_alias);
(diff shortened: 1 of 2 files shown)


inc/search.class.php
```diff
@@ -5388,6 +5388,24 @@ static function addMetaLeftJoin($from_type, $to_type, array &$already_link_table
          return $JOIN;
       }
 
+      if ($from_referencetype === 'Computer' && $to_type === 'OperatingSystem') {
```
trasher, line 5391: Why only computer? Better to rely on presence in `$CFG_GLPI['operatingsystem_types']` I gues..

trasher (changes requested):

inc/search.class.php
stonebuzz: Good point, I didn't commit my final version.
Fixed

trasher (commented):
Sounds good to me. Do you think you can add a unit test on this case?

stonebuzz:
> Sounds good to me. Do you think you can add a unit test on this case?

no problem, I'm planning this =)

trasher (approved):

orthagh (approved):

stonebuzz:
see https://github.com/glpi-project/glpi/pull/9378

PR #9343, opened 2021-07-28
replicate somehow the entity filtering when selecting a new requester…

orthagh:
Some changes in the current fields panel of a new ticket form.
I try to replicate as far as possible the behavior of old actors component

On a new ticket:
- if a requester is selected, reload the form
- if requester(s) exists, try to find entities available for him(them)
- permits to change entity dropdown
- if entity dropdown changes, reload the form

During the fix dev, i found that a form reload didn't restore the previously selected actors.
So the first changes in `CommonITILObject::getActorsForType` are here to restore actors from post data.

If everyone can check i didn't miss anything in this pr, it's the last (complex) feature i think we missed from 9.x branches.

### DIFF
--- a/front/change.form.php
+++ b/front/change.form.php
@@ -40,8 +40,10 @@ if (empty($_GET["id"])) {
 
 Session::checkLoginUser();
 
+// as _actors virtual field stores json, bypass automatic escaping
 if (isset($_UPOST['_actors'])) {
    $_POST['_actors'] = json_decode($_UPOST['_actors'], true);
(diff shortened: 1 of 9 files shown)


cconard96:
Seems OK but triggers the unsaved changes alert. The form element needs "data-track-changes='false'" added or something else done for the inputs that trigger a reload.
This issue was already here for the Type dropdown but I didn't notice it until now.

cconard96 (approved):

trasher (approved):

cedric-anne (approved):

PR #9344, opened 2021-07-28
Twig content templates in place documentation

AdrienClairembault:
Related: #9291, #9325.

Added an "in place" documentation for the twig templates variables, similar to what we use for the API

The first section of the documentation contain the exposed "root" variables in the current form.
The following sections contains the dependencies of these root variables.
For example if the root variable have a `TicketParameter` then we have a `Ticket` dependencies with all the ticket fields displayed.
Each dependency may have its own dependencies (the entity of the ticket, the category, ...) that will also be displayed in their own sections.

image

image

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !22419

### DIFF
--- a/inc/contenttemplates/parameters/abstractparameters.class.php
+++ b/inc/contenttemplates/parameters/abstractparameters.class.php
@@ -53,7 +53,7 @@ abstract class AbstractParameters implements TemplatesParametersInterface
     *
     * @return \Glpi\ContentTemplates\Parameters\ParametersTypes\AbstractParameterType[]
     */
-   abstract protected function defineParameters(): array;
+   abstract public function defineParameters(): array;
 
    /**
     * To by defined in each subclasses, get the exposed values for a given item
@@ -73,6 +73,13 @@ abstract class AbstractParameters implements TemplatesParametersInterface
     * @return array
     */
    abstract protected function getTargetClasses(): array;
+
+   /**
+    * Get label for the target itemtype of this parameter
+    *
+    * @return string
+    */
+   abstract static function getObjectLabel(): string;
 
    public function getValues(CommonDBTM $item): array {
       $valid_class = false;
--- a/inc/contenttemplates/parameters/abstractparameters.class.php
+++ b/inc/contenttemplates/parameters/abstractparameters.class.php
@@ -53,7 +53,7 @@ abstract class AbstractParameters implements TemplatesParametersInterface
     *
     * @return \Glpi\ContentTemplates\Parameters\ParametersTypes\AbstractParameterType[]
     */
-   abstract public function defineParameters(): array;
+   abstract public function getAvailableParameters(): array;
 
    /**
     * To by defined in each subclasses, get the exposed values for a given item
@@ -89,15 +89,5 @@ abstract class AbstractParameters implements TemplatesParametersInterface
       }
 
       return $this->defineValues($item);
-   }
-
-   public function getAvailableParameters(): array {
-      $parameters = [];
-
-      foreach ($this->defineParameters() as $parameter) {
-         $parameters[] = $parameter->compute();
-      }
-
-      return $parameters;
    }
 }
--- a/inc/contenttemplates/parameters/abstractparameters.class.php
+++ b/inc/contenttemplates/parameters/abstractparameters.class.php
@@ -74,13 +74,6 @@ abstract class AbstractParameters implements TemplatesParametersInterface
     */
    abstract protected function getTargetClasses(): array;
 
-   /**
-    * Get label for the target itemtype of this parameter
-    *
-    * @return string
-    */
-   abstract static function getObjectLabel(): string;
-
    public function getValues(CommonDBTM $item): array {
       $valid_class = false;
       foreach ($this->getTargetClasses() as $class) {
--- a/inc/contenttemplates/parameters/abstractparameters.class.php
+++ b/inc/contenttemplates/parameters/abstractparameters.class.php
@@ -64,13 +64,6 @@ abstract class AbstractParameters implements TemplatesParametersInterface
     */
    abstract protected function getTargetClasses(): array;
 
-   /**
-    * Get the parameters values for a given item
-    *
-    * @param CommonDBTM $item
-    *
-    * @return array
-    */
    public function getValues(CommonDBTM $item): array {
       $valid_class = false;
       foreach ($this->getTargetClasses() as $class) {
--- a/inc/contenttemplates/parameters/abstractparameters.class.php
+++ b/inc/contenttemplates/parameters/abstractparameters.class.php
@@ -46,16 +46,6 @@ if (!defined('GLPI_ROOT')) {
 abstract class AbstractParameters implements TemplatesParametersInterface
 {
    /**
-    * To be defined in each subclasses, define all available parameters for one or more itemtypes.
-    * These parameters informations are meant to be used for autocompletion on the client side.
-    *
-    * Result will be returned by `self::getAvailableParameters()`.
-    *
-    * @return \Glpi\ContentTemplates\Parameters\ParametersTypes\AbstractParameterType[]
-    */
-   abstract public function getAvailableParameters(): array;
-
-   /**
     * To by defined in each subclasses, get the exposed values for a given item
     * These values will be used as parameters when rendering a twig template.
     *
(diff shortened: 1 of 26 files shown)


orthagh (commented):
Globally Ok with this.

Some remarks:
- the current icon for access documentation is not very visible, maybe we can use a small text under the textarea for that with a question icon. Something like "read documentation about usable variables"
- the documentation page may be introduced by a paragraph explaining the purpose of the feature
- `for` tags may have a matching closing tag : `{% for asset in ticket.assets %}...{% endfor %}
- maybe we can also set the `...` as clickable and targeting the corresponding section.

AdrienClairembault:
>  the current icon for access documentation is not very visible, maybe we can use a small text under the textarea for that with a question icon. Something like "read documentation about usable variables"

image

>  the documentation page may be introduced by a paragraph explaining the purpose of the feature

Its not really a documentation, it's a list of the variable available for a given page.
I've removed the "documentation" word and made clear the list is specific to a given context:

Context 1 :
image

Context 2:
image

>   `for` tags may have a matching closing tag : `{% for asset in ticket.assets %}...{% endfor %}

I though autocomplete would be enough for that and this way it does not "bloat" the documentation page with extra details.
Tell me if you want me to add it.

>   maybe we can also set the `...` as clickable and targeting the corresponding section.

".XXX" is inside \`\` code block, seems it can't be a link.

inc/contenttemplates/parameterspreset.class.php
```diff
@@ -0,0 +1,139 @@
...
+            $types = [
+               TicketTemplate::getTypeName(0),
+               TaskTemplate::getTypeName(0),
+               SolutionTemplate::getTypeName(0),
+            ];
+
+            return strtolower(implode("/", $types));
+
+         case self::TICKET_SOLUTION:
+            return __("ticket solutions");
```
cedric-anne, line 105:
```suggestion
            return ITILSolution::getTypeName(0);
```

inc/contenttemplates/parameterspreset.class.php
```diff
@@ -0,0 +1,139 @@
...
+    *
+    * @return string
+    */
+   public static function getContextByKey(string $key): string {
+      switch ($key) {
+         case self::ABSTRACT_TEMPLATE:
+            $types = [
+               TicketTemplate::getTypeName(0),
+               TaskTemplate::getTypeName(0),
+               SolutionTemplate::getTypeName(0),
```
cedric-anne, line 99: `0` may be a plural form in some languages.
```suggestion
               TicketTemplate::getTypeName(1),
               TaskTemplate::getTypeName(1),
               SolutionTemplate::getTypeName(1),
```

inc/contenttemplates/parameterspreset.class.php
```diff
@@ -0,0 +1,139 @@
...
+            ];
+
+            return strtolower(implode("/", $types));
```
cedric-anne, line 102: `strtolower` may produce invalid strings, depending on language. For instance, in German, all nouns are capitalized.

inc/contenttemplates/parameters/abstractparameters.class.php
```diff
@@ -53,7 +53,7 @@ abstract class AbstractParameters implements TemplatesParametersInterface
     *
     * @return \Glpi\ContentTemplates\Parameters\ParametersTypes\AbstractParameterType[]
     */
-   abstract protected function defineParameters(): array;
+   abstract public function defineParameters(): array;
```
cedric-anne, line 56: Changing it to public will result in exposing two methods that are almost doint the same thing, which may lead to bad usage of both of them.

I think that `ObjectParameter::compute()` should take care of parameters computation, with something like this:
```php
   public function compute(): array {
      return [
         'type'       => "ObjectParameter",
         'key'        => $this->key,
         'label'      => $this->template_parameters->getObjectLabel(),
         'properties' => TemplateManager::computeParameters($this->template_parameters->getAvailableParameters()),
      ];
   }
```

By doing this, we would be able to replace all `public function defineParameters(): array` to `public function getAvailableParameters(): array`, and use this method in `TemplateDocumentation::addSection()`.

inc/contenttemplates/parameters/abstractparameters.class.php
```diff
@@ -74,6 +74,13 @@ abstract protected function defineValues(CommonDBTM $item): array;
     */
    abstract protected function getTargetClasses(): array;
 
+   /**
+    * Get label for the target itemtype of this parameter
+    *
+    * @return string
+    */
+   abstract static function getObjectLabel(): string;
+
```
cedric-anne, line 83: This one is already declared in interface, so it should not be declared here.
```suggestion
```

inc/contenttemplates/parameters/parameterstypes/abstractparametertype.class.php
```diff
@@ -60,6 +62,40 @@ abstract class AbstractParameterType
...
+    * @return string
+    */
+   abstract function getDocumentationUsage(?string $parent = null): string;
+
+   /**
+    * Reference to others parameters for this parameter's documentation
+    *
+    * @return AbstractParameters|null
+    */
+   abstract function getDocumentationReferences(): ?AbstractParameters;
```
cedric-anne, line 98: 1) The `public` keyword is missing on all new methods.

2) This abstract class has no abstract logic, it is more an interface than an abstract class. Maybe method declarations should be moved into an interface.

3) I guess some abstract default logic can be produced, like 
```php
   public function __construct(string $key, string $label) {
      $this->key = $key;
      $this->label = $label;
   }

   public function getDocumentationField(): string {
      return $this->key;
   }

   public function getDocumentationLabel(): string {
      return $this->label;
   }
```

inc/contenttemplates/templatedocumentation.class.php
```diff
@@ -0,0 +1,169 @@
...
+         __("Usage"),
+         __("References"),
+      ]);
+
+      // Keep track of parameters needing aditionnal references
+      $references = [];
+
+      // Add a row for each parameters
+      foreach ($parameters as $parameter) {
+         /** @var AbstractParameterType $parameter */
```
cedric-anne, line 142:
```suggestion
         /** @var \Glpi\ContentTemplates\Parameters\ParametersTypes\AbstractParameterType $parameter */
```

inc/contenttemplates/templatemanager.class.php
```diff
@@ -176,4 +176,35 @@ public static function getSecurityPolicy(): SecurityPolicy {
...
+    * @return string
+    */
+   public static function generateMarkdownDocumentation(
+      int $preset_parameters_key
+   ): string {
+      $parameters = ParametersPreset::getByKey($preset_parameters_key);
+      $context = ParametersPreset::getContextByKey($preset_parameters_key);
+
+      $documentation = new TemplateDocumentation($context);
+      $documentation->addSection(__("Variables"), $parameters);
```
cedric-anne, line 194: All sections are containing variables, so using the `Root` keyword may help users.
```suggestion
      $documentation->addSection(__("Root variables"), $parameters);
```

inc/toolbox/markdownbuilder.class.php
```diff
@@ -0,0 +1,185 @@
...
+   /**
+    * Generated markdown content
+    *
+    * @var string
+    */
+   protected $content;
+
+   public function __construct() {
+      $this->content = "";
+   }
```
cedric-anne, line 49:
```suggestion
   protected $content = '';
```

inc/contenttemplates/parameters/abstractparameters.class.php
AdrienClairembault: Done c9a14bd914.

inc/contenttemplates/parameters/parameterstypes/abstractparametertype.class.php
AdrienClairembault: Done in 136e4f2fdb, not exactly as suggested since I realized that `$this->label` should not be part of the abstraction as it is not used for object parameters.

inc/contenttemplates/parameters/abstractparameters.class.php
```diff
@@ -53,7 +53,7 @@ abstract class AbstractParameters implements TemplatesParametersInterface
     *
     * @return \Glpi\ContentTemplates\Parameters\ParametersTypes\AbstractParameterType[]
     */
-   abstract protected function defineParameters(): array;
+   abstract public function getAvailableParameters(): array;
```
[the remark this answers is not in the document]
AdrienClairembault: Done 8353289.

cedric-anne:
@AdrienClairembault 

I added 2 commits.

The first one fixes documentation footer, which was displayed on the right side of the page.
The second one is to mutualize code.

If you are OK with both commits, we can merge this.

trasher (approved):

orthagh (approved):

PR #9370, opened 2021-08-01
Dashboard load time improvements

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

With debug mode on or off, loading the central dashboard (27 cards. Default + 1 plugin card) takes around 12 seconds. The GLPI and DB were both hosted on my laptop at the time so I expected it to be slow, but it seemed pretty bad in comparison to other parts of GLPI.
I profiled multiple portions of the code and found a few optimizations that could be made and got it down to about 3 seconds at most. I expect other users to see similar or better performance improvements.

On the client side, I saw that a synchronous request to get filter data was made for every card it tried to load that takes up to half a second each time. This request has nothing to do with the individual cards, so it only needs called once.per load/refresh. Also, it seems like this request blocks other JavaScript from running so you cannot open the menus on the modern UI branch at least while waiting for the request(s) to finish.

The next slow part I found was that each individual request for card HTML required rebuilding the catalog of dashboard cards in `\Glpi\Dashboard\Grid::getAllDashboardCards` and this is a relatively slow process (about 150-250ms). I combined all the card requests into a single request and cached the dashboard card catalog in a static variable so it only needs done once per refresh.

The final portion of the code I noticed could be optimized was widgets that use a gradient. The `\Glpi\Dashboard\Widget::getCssGradientPalette` function took around 50-70ms each time because it was recompiling SCSS each time. The compiled styles for each card don't change between refreshes of the dashboard so it can be cached.

I also noticed that a lot of time is spent checking that certain fields and tables exist when generating the dashboard card catalog. Since fields and tables existence are only "cached" per request and there are a lot of itemtypes it goes through to create cards, it can take a lot of time querying the DB to see if these fields/tables exist. I don't think this can be optimized without affecting other parts of the code and potentially breaking some expected behavior.

### DIFF
--- a/ajax/dashboard.php
+++ b/ajax/dashboard.php
@@ -43,7 +43,8 @@ if (!isset($_REQUEST['embed']) || !$_REQUEST['embed']) {
 
 } else if (!in_array($_REQUEST['action'], [
    'get_dashboard_items',
-   'get_card'
+   'get_card',
+   'get_cards'
 ])) {
    Html::displayRightError();
 }
@@ -106,6 +107,21 @@ switch ($_REQUEST['action']) {
    case 'get_card':
       session_write_close();
       echo $grid->getCardHtml($_REQUEST['card_id'], $_REQUEST);
+      break;
+
+   case 'get_cards':
+      session_write_close();
+      header("Content-Type: application/json; charset=UTF-8");
+      // Parse stringified JSON payload (Used to preserve integers)
+      $request_data = array_merge($_REQUEST, json_decode(html_entity_decode(stripslashes($_REQUEST['data'])), true));
+      unset($request_data['data']);
+      $cards = $request_data['cards'];
+      unset($request_data['cards']);
+      $result = [];
+      foreach ($cards as $card) {
+         $result[$card['card_id']] = $grid->getCardHtml($card['card_id'], array_merge($request_data, $card));
+      }
+      echo json_encode($result);
       break;
 
    case 'display_add_filter':
--- a/ajax/dashboard.php
+++ b/ajax/dashboard.php
@@ -113,7 +113,7 @@ switch ($_REQUEST['action']) {
       session_write_close();
       header("Content-Type: application/json; charset=UTF-8");
       // Parse stringified JSON payload (Used to preserve integers)
-      $request_data = array_merge($_REQUEST, json_decode(html_entity_decode(stripslashes($_REQUEST['data'])), true));
+      $request_data = array_merge($_REQUEST, json_decode($_UREQUEST['data'], true));
       unset($request_data['data']);
       $cards = $request_data['cards'];
       unset($request_data['cards']);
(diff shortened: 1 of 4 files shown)


js/dashboard.js
```diff
@@ -906,7 +906,10 @@ var Dashboard = {
    getCardsAjax: function(specific_one) {
       specific_one = specific_one || "";
 
-      var promises = [];
+      var filters = Dashboard.getFiltersFromDB();
```
btry, line 909: getFiltersFromDB only exists in master and replaces getFiltersFromStorage. I believe that you developed your pr on top of master branch. If I'm right, this PR will break dashboards; you need to call getFiltersFromStorage instead of getFiltersFromDB. See deleted line 924 below.

cconard96: It was. I added a commit to make it work with 9.5. Since the main issue is the synchronous AJAX request in getFiltersFromDB, this will need changed when applied to the master branch. The other difference is using `compile` vs `compileString` in the SCSS class. The first is deprecated in the version used on master but will still work. I switched it back to `compile` to work on both.

ajax/dashboard.php
```diff
@@ -108,6 +109,21 @@
       echo $grid->getCardHtml($_REQUEST['card_id'], $_REQUEST);
       break;
 
+   case 'get_cards':
+      session_write_close();
+      header("Content-Type: application/json; charset=UTF-8");
+      // Parse stringified JSON payload (Used to preserve integers)
+      $request_data = array_merge($_REQUEST, json_decode(html_entity_decode(stripslashes($_REQUEST['data'])), true));
```
cedric-anne, line 116: You can use the non sanitized request, I guess.
```suggestion
      $request_data = array_merge($_REQUEST, json_decode($_UREQUEST['data'], true));
```

trasher (approved):

js/dashboard.js
orthagh: question about this, as some enhancements have been done in master branch for the dashboard.
Is it mandatory for this pr to target 9.5/bugfixes ?

At least, can we retarget master and work on a backport if needed ?
This steps will ease the integration of the current and other PR. And it will also ease the work of cherry picking the pr in master for mainteners.

Another possibility could be to let this pr ending in 9.5 and you work on the adaptation of master branch in a separate pr.
Anyway, you see the issue i think, the current work targetting the stable branch overlap with existing evolutions.

cconard96:
This PR is the backport. The actual work was done against the modern UI
branch. The main work is done in a branch named the same as this one just
without the 9.5 suffix.

orthagh:
Oh, ok forget me so.

I'ill test quickly and give my approbation then

orthagh:
fyi @cconard96, since my return, i tested the feat/modern-ui branch and the dashboard page seems to load heavily.
It seems not the same issue, as i don't see the card's spinners.
I don't know where but the js code seems to have a very long time to init the grid.

If you have any idea, byt maybe it's a recent change of @cedric-anne and i'll with him in two weeks

orthagh (approved):
So ok for me

PR #9398, opened 2021-08-05
fix(inventory): fix remote inventory / status

stonebuzz:
Actually, GLPI try to get all addresses from computer to request a remote inventory / status

from a standard inventory GLPI return three addresses (or more depending of inventory) (```NetworkPortLocal``` are excluded)
```https://192.168.1.24:62354/status``` (by ipv4)
```https://2a01:e0a:87c:4ea0:f178:597:e709:8aea:62354/status``` (by ipv6)
```https://Desktop:62354/status``` (by name)

Only IPV4 work

This PR Restrict addresses to get only IPV4 compliant

Perhaps we could save the IP used by the agent during the ```PROLOG``` (into ```Agent``` object)

| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #number

### DIFF
--- a/inc/agent.class.php
+++ b/inc/agent.class.php
@@ -331,9 +331,16 @@ class Agent extends CommonDBTM {
          $input['tag'] = $metadata['tag'];
       }
 
+      if (!isset($metadata['httpd-port'])) {
+         $input['port'] = $metadata['httpd-port'];
+      }
+
       $input['ip_protocol'] = 'http';
-      if (!isset($metadata['httpd-plugins']['ssl'])) {
-         $input['ip_protocol'] = 'https';
+      if (isset($metadata['httpd-plugins']['ssl']['ports'])) {
+         $ssl_ports[] = $metadata['httpd-plugins']['ssl']['ports'];
+         if (in_array("0", $ssl_ports) || (isset($metadata['httpd-port']) && in_array($metadata['httpd-port'], $ssl_ports))) {
+            $input['ip_protocol'] = 'https';
+         }
       }
 
       if ($deviceid === 'foo') {
--- a/inc/agent.class.php
+++ b/inc/agent.class.php
@@ -62,8 +62,6 @@ class Agent extends CommonDBTM {
    static $rightname = 'computer';
    //static $rightname = 'inventory';
 
-   private static $found_adress = false;
-
    static function getTypeName($nb = 0) {
       return _n('Agent', 'Agents', $nb);
    }
@@ -139,10 +137,30 @@ class Agent extends CommonDBTM {
             'field'         => 'port',
             'name'          => _n('Port', 'Ports', 1),
             'datatype'      => 'integer',
+         ], [
+            'id'            => '15',
+            'table'         => $this->getTable(),
+            'field'         => 'ip_binary',
+            'name'          => __('Ip'),
+            'datatype'      => 'specific',
+            'nosearch'      => true,
+            'massiveaction' => false,
          ]
       ];
 
       return $tab;
+   }
+
+   static function getSpecificValueToDisplay($field, $values, array $options = []) {
+      if (!is_array($values)) {
+         $values = [$field => $values];
+      }
+      switch ($field) {
+         case 'ip_binary':
+         return inet_ntop($values[$field]);
+      }
+
+      return parent::getSpecificValueToDisplay($field, $values, $options);
    }
 
    /**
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by stonebuzz** [comment]
As discuss with @g-bougard 

When agent contact GLPI save into ```glpi_agents``` table three new informations

- IP ```$_SERVER['REMOTE_ADDR']``` use by agent to contact GLPI (store as binary with ```inet_pton```)
- IP version
- Protocol : ```http``` by default , ```https``` if plugin ```ssl``` is active agent side 

When user request ```/status```, ```/now``` (or other from plugins)
GLPI use this IP and make only one call.

File: inc/agent.class.php [hunk 1]
```diff
@@ -296,6 +314,15 @@ public function handleAgent($metadata) {
          'itemtype'     => $metadata['itemtype'] ?? 'Computer'
       ];
 
+      $input['ip_version'] = 4;
+      if (isset($_SERVER['REMOTE_ADDR']) && filter_var($_SERVER['REMOTE_ADDR'], FILTER_VALIDATE_IP, FILTER_FLAG_IPV6)) {
+         $input['ip_version'] = 6;
+      }
```
g-bougard, line 320: I don't see any usage of `ip_version` in your PR but there to just set it in database. I think you can just get rid of it.

inc/agent.class.php
```diff
@@ -304,6 +331,11 @@ public function handleAgent($metadata) {
          $input['tag'] = $metadata['tag'];
       }
 
+      $input['ip_protocol'] = 'http';
+      if (!isset($metadata['httpd-plugins']['ssl'])) {
+         $input['ip_protocol'] = 'https';
+      }
+
```
g-bougard, line 338: As discussed privately, you also will be able to set the port and you have to check set ports for ssl httpd-plugin before telling we should use `https`.
```suggestion
      if (!isset($metadata['httpd-port'])) {
         $input['port'] = $metadata['httpd-port'];
      }

      $input['ip_protocol'] = 'http';
      if (isset($metadata['httpd-plugins']['ssl']['ports'])) {
         $ssl_ports[] = $metadata['httpd-plugins']['ssl']['ports'];
         if (in_array("0", $ssl_ports) || (isset($metadata['httpd-port']) && in_array($metadata['httpd-port'], $ssl_ports))) {
            $input['ip_protocol'] = 'https';
         }
      }

```
The right logic is there but this is a suggestion and I didn't test it, so try it and fix it before merging.

This code is also following glpi agent commit: https://github.com/glpi-project/glpi-agent/commit/209f730be0f98797ce7bb5fa10ea63ee2da6339b
Here is an example of what you can obtain from [glpi-agent](https://github.com/glpi-project/glpi-agent) during CONTACT request:
```
{
   "action": "contact",
   "deviceid": "asus-2021-01-26-17-04-22",
   "enabled-tasks": [
      "inventory"
   ],
   "httpd-plugins": {
      "inventory": "disabled",
      "proxy": {
         "disabled": "no",
         "glpi_protocol": "yes",
         "local_store": "/srv/glpi-agent/proxy",
         "max_pass_through": 5,
         "max_proxy_threads": 10,
         "maxrate": 30,
         "maxrate_period": 3600,
         "no_category": "",
         "only_local_store": 0,
         "port": "8081",
         "prolog_freq": 24,
         "url_path": "/proxy"
      },
      "secondaryproxy": "disabled",
      "ssl": {
         "disabled": "no",
         "ports": [
            "0",
            "8081"
         ],
         "ssl_cert_file": "cert.pem",
         "ssl_key_file": "key.pem"
      },
      "test": "disabled",
      "toolbox": "disabled"
   },
   "httpd-port": "62354",
   "installed-tasks": [
      "collect",
      "deploy",
      "esx",
      "inventory",
      "netdiscovery",
      "netinventory",
      "wakeonlan"
   ],
   "name": "GLPI-Agent",
   "version": "1.0-dev"
}
```

inc/agent.class.php
```diff
@@ -139,12 +137,32 @@ function rawSearchOptions() {
...
       return $tab;
    }
 
+   static function getSpecificValueToDisplay($field, $values, array $options = []) {
+      if (!is_array($values)) {
+         $values = [$field => $values];
+      }
+      switch ($field) {
+         case 'ip_binary':
+         return inet_ntop($values[$field]);
```
g-bougard, line 160: You should probably handle the case `inet_ntop()` returns `false` in case of error.

inc/agent.class.php
```diff
@@ -296,6 +314,15 @@ public function handleAgent($metadata) {
          'itemtype'     => $metadata['itemtype'] ?? 'Computer'
       ];
 
+      $input['ip_version'] = 4;
+      if (isset($_SERVER['REMOTE_ADDR']) && filter_var($_SERVER['REMOTE_ADDR'], FILTER_VALIDATE_IP, FILTER_FLAG_IPV6)) {
+         $input['ip_version'] = 6;
+      }
+
+      if (isset($_SERVER['REMOTE_ADDR'])) {
+         $input['ip_binary'] = inet_pton($_SERVER['REMOTE_ADDR']);
```
g-bougard, line 323: You should probably handle the case `inet_pton()` returns `false` even if this should not happen.

g-bougard (changes requested):

inc/agent.class.php
```diff
@@ -489,41 +427,33 @@ public function getAgentURLs(): array {
     *
     * @return Response
     */
-   protected function requestAgent($endpoint): Response {
+   public function requestAgent($endpoint): Response {
```
g-bougard, line 430: This update comes first from #9395.

stonebuzz:
I have a question about local imports 
With this feature the IP registered will be the IP of the computer injecting the XML file and not the IP of the inventory

g-bougard:
In that case, this generally means we can't contact remote agent, so in that case, we should probably not set any remote status url.

Then we can also think about the proxy agent case. In that case, the targeted agent UUID should be set in GLPI-Agent-ID HTTP header so the proxy can do the job.

stonebuzz:
close this PR
see -> https://github.com/glpi-project/glpi/pull/10115

Issue #6902, opened 2020-02-04
Pending time not considered when SLA is changed


vaibhav12321:
**Describe the bug**

I put a ticket from Processing to Pending status and after some minutes I changed Urgency to Very High. This attach new SLA by Business rule and waiting time is not added to the Time to resolve after SLA changed.

Note: tested with GLPI 9.4.3 and GLPI 9.2.4

**To reproduce**

Please check screenshot with ticket history.

**Expected behavior**

As seen in ticket history it never came to Processing status. So it should not breach the SLA.

**Screenshots**

Ticket history:

![Screenshot from 2020-02-04 17-03-24](https://user-images.githubusercontent.com/13692091/73742802-3b4dff80-4773-11ea-887c-1b93c33fe2b1.png)

Statistics tab:

![Screenshot from 2020-02-04 17-02-32](https://user-images.githubusercontent.com/13692091/73742815-40ab4a00-4773-11ea-81cd-76ebb2189132.png)

Ticket page:

![Screenshot from 2020-02-04 17-01-28](https://user-images.githubusercontent.com/13692091/73742829-46a12b00-4773-11ea-904d-d00117b9e23d.png)

aagz:
Tested in 9.5.1. The same problem.

aagz:
Tested in 9.5.2 and 9.5.3. The bug still exists.

PR #9412, opened 2021-08-12
Prevent reset value of sla_waiting_duration field

vaibhav12321:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #6902


flegastelois:
The "sla_waiting_duration" field is linked to the applied SLA, so it must be reset to zero in the event of a change of SLA.

PR #9415, opened 2021-08-13
fix(database): missing raw search options

flegastelois:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a

### DIFF
--- a/inc/database.class.php
+++ b/inc/database.class.php
@@ -157,15 +157,6 @@ class Database extends CommonDBChild {
       ];
 
       $tab[] = [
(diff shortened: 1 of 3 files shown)


cedric-anne (commented):
You should also add default display preferences in default install/migration.

trasher:
> You should also add default display preferences in default install/migration.

Done

orthagh (approved):

PR #9422, opened 2021-08-18
retro compatibility

delcroip:
$itemtype is used only in case of stat, all the other cases don't need params, without default it create a compatibility issue with reports plugin (and I guess other piece of code)

| Q             | A
| ------------- | ---
| Bug fix?      | yes (but on master)
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | NA

### DIFF
--- a/inc/dropdown.class.php
+++ b/inc/dropdown.class.php
@@ -2126,7 +2126,7 @@ class Dropdown {
     *
     * @since 0.83
    **/
-   static function showOutputFormat($itemtype = NULL) {
+   static function showOutputFormat($itemtype = null) {
       global $CFG_GLPI;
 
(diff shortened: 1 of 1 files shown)


cconard96 (approved):

inc/dropdown.class.php
```diff
@@ -2126,7 +2126,7 @@ static function getGlobalSwitch($value = 0) {
     *
     * @since 0.83
    **/
-   static function showOutputFormat($itemtype) {
+   static function showOutputFormat($itemtype = NULL) {
```
cconard96, line 2129:
```suggestion
   static function showOutputFormat($itemtype = null) {
```

cconard96 (changes requested):

cedric-anne (approved):

trasher (approved):

PR #9454, opened 2021-08-30
Reduce search memory usage

AdrienClairembault:
Improve memory usage by a LOT.

The search module only used LIMIT for searches with no criteria (SELECT *).
These changes force every search request to respect the LIMIT criteria.

It does seem to lower a bit SQL performances (request seems to be between 10 or 20% longer) but completely negate the memory used by the SQL search results.

For example, on a test server with 250k tickets, the server memory limit of 256MB (!!) would be reached if the search returned more than 30k tickets.
Now it doesn't matter how many results are returned: the memory stays around 15MB as we only load the 200 items of the current page instead of up to 250k tickets.

| Q             | A
| ------------- | ---
| Bug fix?      | yes/
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !22511

### DIFF
--- a/inc/search.class.php
+++ b/inc/search.class.php
@@ -720,6 +720,11 @@ class Search {
       }
 
       $LIMIT = " LIMIT ".(int)$data['search']['start'].", ".(int)$data['search']['list_limit'];
+
+      // Can't use the one method "trick" to count + get limited results for this case, keep original code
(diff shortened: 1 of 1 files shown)


trasher:
Tests are broken

AdrienClairembault:
Not sure for the remaining tests that still fails, they run fine on my side.
I'll try to see tomorrow.

PR #9484, opened 2021-09-01
prevent unwanted single actions to be displayed on closed/solved ticket

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | internal ref !22529

### DIFF
--- a/inc/massiveaction.class.php
+++ b/inc/massiveaction.class.php
@@ -387,8 +387,6 @@ class MassiveAction {
             if (isset($POST['check_items_id'])) {
                if (!$this->check_item->getFromDB($POST['check_items_id'])) {
                   exit();
-               } else {
-                  $this->check_item->getEmpty();
                }
             }
          } else {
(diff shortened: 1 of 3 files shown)


inc/massiveaction.class.php
```diff
@@ -387,8 +387,6 @@ function getCheckItem($POST) {
             if (isset($POST['check_items_id'])) {
                if (!$this->check_item->getFromDB($POST['check_items_id'])) {
                   exit();
-               } else {
```
orthagh, line 390: This part was very strange.
We have at choice, an exit on a bad id passed, or a forced getempty. So the getFromDB doesn't have any usage

orthagh:
It should be done in `getForbiddenSingleMassiveActions` in fact, It will be done soon by another contributor

PR #9486, opened 2021-09-01
Fix empty value for user dropdown

AdrienClairembault:
I've noticed a few weird values in the users dropdown: 
image

It turns out to be a side effect of the new 'Myself' search criteria and PHP weird type casting for loose comparisons:
image

This issue is only present on PHP 7 as [saner string to number comparisons](https://www.php.net/releases/8.0/en.php#saner-string-to-number-comparisons) were introduced in PHP 8.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |

### DIFF
--- a/inc/user.class.php
+++ b/inc/user.class.php
@@ -4049,7 +4049,7 @@ JAVASCRIPT;
 
       $view_users = self::canView();
 
-      if (strval($p['value']) == 'myself') {
+      if ($p['value'] === 'myself') {
          $default = __("Myself");
       } else if (!empty($p['value']) && ($p['value'] > 0)) {
          $default = $user["name"];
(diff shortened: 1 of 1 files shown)


orthagh (approved):

trasher (approved):

inc/user.class.php
```diff
@@ -4049,7 +4049,7 @@ static function dropdown($options = []) {
 
       $view_users = self::canView();
 
-      if ($p['value'] == 'myself') {
+      if (strval($p['value']) == 'myself') {
```
trasher, line 4052: Should not this be solved like:
```suggestion
      if ($p['value'] === 'myself') {
```

trasher (approved):

PR #9495, opened 2021-09-02
Reduce search engine memory usage

AdrienClairembault:
* Add LIMIT to search queries

#### Current search engine

The search engine currently only use a `LIMIT` clause if there is no search criteria specified.
In this case, it will send a second `COUNT(*)` query after the main search to know the total number of results.

Any query with a defined criteria does not use the `LIMIT` clause.
The full results are send to PHP and we only iterate on the first X results (the number of results to display on one page).

This help execution time as  `COUNT` requests can be costly but is very bad for memory consumption on big databases.

 #### Changes to the search engine

Every search request is now accompanied by a second `COUNT` request.

The mandatory clauses to re-use from the main query are `FROM`, `GROUP BY` and `HAVING`.
The `SELECT` and `FROM` clauses must only contains mandatory fields and joins to get the fastest request as possible.

The simplest case is a search without a `HAVING` clause.
This will results in a very simple `COUNT` request:
```sql
SELECT COUNT(*) FROM {main_table} {mandatory joins for the conditions fields} WHERE {condition}
```

The more complicated case is when we have a `HAVING` clause as we it forces us to add fields in the `SELECT` clause (the `HAVING` clause filters on selected results and thus cannot be done without them in the query).
In this case, we end up with a request like this:
```sql
SELECT COUNT(*) FROM (
   SELECT {having fields} FROM {main_table} {mandatory joins for the conditions fields} WHERE {condition} HAVING ... GROUP BY ...
)
```

#### Benchmark 

Here is a quick benchmark on a ~140k computers database.
We are doing some searches on computers and checking the time used by the `Search::constructData()` function as well as the memory peak of the application.
I've put in bold some values that stand out as worse.

| Search criteria | 9.5 bugfixes | #9495|
| ------------- | ---|-----|
| * | 1.18s (7.64 Mio) | 1.16s(7.60 Mio) |
| name contains 'se'  | 0.35s (**10.36 Mio**) | **0.58s** (7.62 Mio) |
| alternate username contains 'a' and location is not "Caen" | 1.51s (**56.68 Mio**)  | 1.43s (7.63 Mio) |
| manufacturer is 'HP' and status is "Status A" | 0.83s (**36.38 Mio**) | 0.73s (7.64 Mio) |
| comments contains 'ab' or serial contains '44' or uuid contains '7' | 1.78s (**72.88 Mio**)| 1.57s (7.68 Mio) |

On the memory consumption side the results are excellent.
Before these changes the memory peak will scale depending on the numbers of returned results and the number of fields displayed (more fields = more data to send back to PHP).
After these change the memory peak will always stays around the same for a given page size.

On the execution time side the results are more difficult to analyze.
The gains / loses all depend on the `COUNT` query execution time.
If it is almost instantaneous then we gain a little time or are even with the 9.5 bugfixes branches (test 1, 3, 4 and 5).
If it not instantaneous then we have worse performances, it may be a few extra tenth at best and in a few rare cases a possible x2 execution time (the `COUNT` request would be almost as long as the main query in those cases like test 2).

* Remove 'no_search' option

The search engine used this option internally to know if we had a search with no criteria and thus would be able to use the LIMIT query (as described in "Current search engine" section above).
Since we now use `LIMIT` all the time this option is no longer needed and all associated code as been removed.

* Remove 'all' criteria

As we discussed the 'all' criteria is outdated and bugged so this PR is a good opportunity to remove it.

* Fix duplicate aliases

It seems we had a few case of duplicate aliases in the `SELECT` clause of our search requests.
This is not a problem in most simple requests as MySQL allow duplicates in the results.

It is however a problem with wrapped queries like `SELECT COUNT(*) FROM ( {wrapped query} )`, duplicated alias are not allowed in the wrapped query and will cause a SQL error.

There was two kind of duplicated aliases in the code base:
One of them was on Ticket/Problem/Changes and could be fixed in the searchoption by removing an extra field.
The other was on a special case of a meta criteria join on the current itemtype (for example searching for documents on documents).
This case required a fix in the search engine by adding a "META_" prefix to meta criteria aliases in the `SELECT` and `HAVING` clauses.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !22511

### DIFF
--- a/inc/search.class.php
+++ b/inc/search.class.php
@@ -866,7 +866,7 @@ class Search {
                      $GROUPBY.
                      $HAVING;
 
-                  if (empty(trim($GROUPBY)) && empty(trim($HAVING))) {
+                  if (empty(trim($HAVING))) {
                      // No HAVING clause: we can keep a simple query
                      $count_tmpquery = self::buildSimpleCountQuery(
                         $COUNT_FROM,
@@ -1026,7 +1026,7 @@ class Search {
                   $ORDER.
                   $LIMIT;
 
-         if (empty(trim($GROUPBY)) && empty(trim($HAVING))) {
+         if (empty(trim($HAVING))) {
             // No HAVING clause: we can keep a simple query
             $COUNT_QUERY = self::buildSimpleCountQuery($COUNT_FROM, $WHERE);
          } else {
--- a/inc/search.class.php
+++ b/inc/search.class.php
@@ -1373,6 +1373,7 @@ class Search {
          $data['data']['totalcount'] = 0;
          $result_num = $DBread->query($data['sql']['count']);
          $data['data']['totalcount'] += $DBread->result($result_num, 0, 0);
+         $data['data']['count'] = intval(min($data['data']['totalcount'], $data['search']['list_limit']));
 
          if ($onlycount) {
             //we just want to coutn results; no need to continue process
@@ -1483,10 +1484,6 @@ class Search {
          // Don't risk overflowing the memory limit by loading the entire
          // resultset, use a generator
          $data['data']['rows'] = self::yieldResults($data, $DBread, $result);
-
-         // We have two variables for the same value here since the LIMIT changes.
-         // Should maybe be merged into one var but it may impact existing code.
-         $data['data']['count'] = $data['data']['totalcount'];
       } else {
          echo $DBread->error();
       }
--- a/inc/search.class.php
+++ b/inc/search.class.php
@@ -1588,6 +1588,10 @@ class Search {
             );
          }
 
+         // Add items in item list
+         $current_type = (isset($newrow['TYPE']) ? $newrow['TYPE'] : $data['itemtype']);
+         Session::addToNavigateListItems($current_type, $newrow["id"]);
+
          yield $i => $newrow;
          $i++;
       }
@@ -1676,12 +1680,6 @@ class Search {
          'may_be_deleted'      => $item instanceof CommonDBTM && $item->maybeDeleted(),
          'may_be_located'      => $item instanceof CommonDBTM && $item->maybeLocated(),
       ]);
-
-      // Add items in item list
-      foreach ($data['data']['rows'] as $row) {
-         $current_type = (isset($row['TYPE']) ? $row['TYPE'] : $data['itemtype']);
-         Session::addToNavigateListItems($current_type, $row["id"]);
-      }
 
       // Clean previous selection
       $_SESSION['glpimassiveactionselected'] = [];
--- a/inc/search.class.php
+++ b/inc/search.class.php
@@ -1473,109 +1473,123 @@ class Search {
          }
 
          // Get rows
-
-         $i = $data['data']['begin'];
          $data['data']['warning']
             = "For compatibility keep raw data  (ITEM_X, META_X) at the top for the moment. Will be drop in next version";
 
          $data['data']['rows']  = [];
-         $data['data']['items'] = [];
 
          self::$output_type = $data['display_type'];
 
-         while (($i < $data['data']['totalcount']) && ($i <= $data['data']['end'])) {
-            $row = $DBread->fetchAssoc($result);
-            $newrow        = [];
-            $newrow['raw'] = $row;
+         // Don't risk overflowing the memory limit by loading the entire
+         // resultset, use a generator
+         $data['data']['rows'] = self::yieldResults($data, $DBread, $result);
 
-            // Parse datas
-            foreach ($newrow['raw'] as $key => $val) {
-               if (preg_match('/ITEM(_(\w[^\d]+))?_(\d+)(_(.+))?/', $key, $matches)) {
-                  $j = $matches[3];
-                  if (isset($matches[2]) && !empty($matches[2])) {
-                     $j = $matches[2] . '_' . $matches[3];
+         // We have two variables for the same value here since the LIMIT changes.
+         // Should maybe be merged into one var but it may impact existing code.
+         $data['data']['count'] = $data['data']['totalcount'];
+      } else {
+         echo $DBread->error();
+      }
+   }
+
+   /**
+    * Generator used for fetching and parsing an entire resultset with no impact
+    * on used memory
+    *
+    * @param array $data
+    * @param DBmysql $DBread
+    * @param mysqli_result $result
+    *
+    * @return Generator
+    */
+   private static function yieldResults(
+      array $data,
+      DBmysql $DBread,
+      mysqli_result $result
+   ): Generator {
+      $i = $data['data']['begin'];
+
+      // Iterate over the resultset
+      while ($row = $DBread->fetchAssoc($result)) {
+         $newrow = [
+            'raw' => $row
+         ];
+
+         // Parse raw data
+         foreach ($newrow['raw'] as $key => $val) {
+            if (preg_match('/ITEM(_(\w[^\d]+))?_(\d+)(_(.+))?/', $key, $matches)) {
+               $j = $matches[3];
+               if (isset($matches[2]) && !empty($matches[2])) {
+                  $j = $matches[2] . '_' . $matches[3];
+               }
+               $fieldname = 'name';
+               if (isset($matches[5])) {
+                  $fieldname = $matches[5];
+               }
+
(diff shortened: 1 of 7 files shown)


AdrienClairembault:
Tests are failing due to a duplicate key in the COUNT query for ticket, changes and problems: 
image

This is weird as the SELECT query have the same duplicate key even in 9.5/bugfixes but it doesn't trigger an error : 
image


Probably an error in some CommonITILSearchOptions... will try to look for it tomorrow.

orthagh:
Important point, I liked the previous solution as we did it in a single query and i was pretty sure it wouldn't changed significantly the query execution time.
Double query means double execution time and GLPI can be alredy slows at some pages with some complex criteria.
And adding more time can be catastrophic

AdrienClairembault:
Should be all good for reviews now, I've updated the initial description.

inc/search.class.php
```diff
@@ -522,6 +501,31 @@ static function prepareDatasForSearch($itemtype, array $params, array $forcedisp
...
    }
 
+   /**
+    * Parse the $criteria array to return a searchoption_id => value array
+    *
+    * @param array $criteria
+    *
+    * @return array
+    */
+   public static function getConditionSearchOptionsIds(array $criteria): array {
```
cedric-anne, line 511:
```suggestion
   private static function getConditionSearchOptionsIds(array $criteria): array {
```

inc/search.class.php
```diff
@@ -522,6 +501,31 @@ static function prepareDatasForSearch($itemtype, array $params, array $forcedisp
...
+      $condition_search_options_ids = [];
+      foreach ($criteria as $crit) {
+         if (isset($crit['criteria'])) {
+            // Recursion, parse child criteria
+            $child_criteria = self::getConditionSearchOptionsIds($crit['criteria']);
+            foreach (array_keys($child_criteria) as $child_crit) {
+               $condition_search_options_ids[$child_crit] = $child_crit['value'] ?? true;
```
cedric-anne, line 518: If `$child_crit` is a key of `$child_criteria`, then it cannot be an array, and `$child_crit['value']` is invalid.

inc/search.class.php
```diff
@@ -1001,9 +1072,16 @@ static function constructCriteriaSQL($criteria = [], $data = [], $searchopt = []
                   continue;
                }
 
-               $new_having = self::addHaving($LINK, $NOT, $itemtype,
-                                             $criterion['field'], $criterion['searchtype'],
-                                             $criterion['value']);
+               $is_meta =
```
cedric-anne, line 1075:
```suggestion
```

inc/search.class.php
```diff
@@ -1101,14 +1179,18 @@ static function constructCriteriaSQL($criteria = [], $data = [], $searchopt = []
     * @param  string &$FROM                TODO: should be a class property (output parameter)
     * @param  array  &$already_link_tables TODO: should be a class property (output parameter)
     * @param  array  &$data                TODO: should be a class property (output parameter)
+    * @param  bool   &$count_query         Are we building a could query ?
```
cedric-anne, line 1182:
```suggestion
    * @param  bool   $count_query          Are we building a could query ?
```

inc/search.class.php
```diff
@@ -7923,4 +8001,42 @@ public static function joinDropdownTranslations($alias, $table, $itemtype, $fiel
...
+
+   /**
+    * Build a simple count query
+    *
+    * @param string $FROM
+    * @param string $WHERE
+    *
+    * @return string
+    */
+   public static function buildSimpleCountQuery(
```
cedric-anne, line 8013:
```suggestion
   private static function buildSimpleCountQuery(
```

inc/search.class.php
```diff
@@ -7923,4 +8001,42 @@ public static function joinDropdownTranslations($alias, $table, $itemtype, $fiel
...
+    *
+    * @param string $SELECT
+    * @param string $FROM
+    * @param string $WHERE
+    * @param string $GROUPBY
+    * @param string $HAVING
+    *
+    * @return string
+    */
+   public static function buildWrappedCountQuery(
```
cedric-anne, line 8031:
```suggestion
   private static function buildWrappedCountQuery(
```

AdrienClairembault:
Will this fix be part of 9.5.6 ?

cedric-anne (approved):
It seems OK, but I do not really have enough data to do real tests.

AdrienClairembault:
Regarding the latest (6984fa1) commit, it's related to another memory mismanagement that was found in the internal ref.
It turns out we are also loading **all** search results into a PHP array (so taking the results out of a memory efficient mysql resultset and putting it all in a standard PHP array row by row).

This isn't an issue when you are browsing a small sized result (e.g. a standard 20 entries result page).
However if you want to export all search results to CSV then you may reach the memory limit very easily if you have a lot of items.

This commit fix it by using a PHP generator.

The downside is that the "parsing" process of the results set can now only be used to build `$data['data']['rows']`. 
Before the changes we would also build `$data['data']['currentuser']` and `$data['data']['items']`.

-> I'm not sure what was in `$data['data']['currentuser']` but it seems to be unused data as the `'currentuser'` string isn't used anywhere else in GLPI.
-> `$data['data']['items']` seemed to contain the ids of every items returned by the searchs.
Removing it doesn't break the unit tests and I can't find any code that would depend on it so it may be unused too.
Note that the `items` key is used in a lot of place in GLPI so it's hard to check only the ones related to search results...

inc/search.class.php
```diff
@@ -1227,21 +1314,9 @@ static function constructData(array &$data, $onlycount = false) {
...
-            } else {
-               foreach ($data['sql']['count'] as $sqlcount) {
-                  $result_num = $DBread->query($sqlcount);
-                  $data['data']['totalcount'] += $DBread->result($result_num, 0, 0);
-               }
-            }
-         }
+         $result_num = $DBread->query($data['sql']['count']);
+         $data['data']['totalcount'] += $DBread->result($result_num, 0, 0);
+         $data['data']['count'] = intval(min($data['data']['totalcount'], $data['search']['list_limit']));
```
cedric-anne, line 1319: Why there is a `intval` here ? If `$data['data']['totalcount']` or `$data['search']['list_limit']` may be something else than an integer, it should be clearly explained in a comment. Also, I guess it would be prefferable to apply the `intval` transformation prior to the `min` call.

cedric-anne:
IMHO, we should not introduce such changes on 9.5 branch. Indeed, we are introducing many changes on a critical part of GLPI to fix a behaviour that is present from a while and, I guess, is only a problem on some edge cases.

trasher:
> IMHO, we should not introduce such changes on 9.5 branch. Indeed, we are introducing many changes on a critical part of GLPI to fix a behaviour that is present from a while and, I guess, is only a problem on some edge cases.

:+1: this should not tagret 9.5 branch

inc/search.class.php
AdrienClairembault: Some tests were failing due to a string value in `$data['data']['count']`.
Min may return a string if the compared value are numbers stored in string: 
image

I've moved intval to only target `$data['search']['list_limit']` since it is the offender.
`$data['data']['totalcount']` is defined just above so it should be safe to assume it'll always be a integer.

AdrienClairembault:
> 
> 
> IMHO, we should not introduce such changes on 9.5 branch. Indeed, we are introducing many changes on a critical part of GLPI to fix a behaviour that is present from a while and, I guess, is only a problem on some edge cases.

Like @orthagh suggested in our weekly meeting, we can keep this PR open in 9.5 for a few weeks while we wait for the code to be "production tested" on the internal ref then moves and merge it to master later.

cedric-anne (approved):

AdrienClairembault:
The 4 latest commit reduce the memory used when exporting 140k computers to CSV from **1.4GB** to ~65MB.

65MB is still significant as there is yet another bottleneck happening simply because we load a lot of data from the database straight into PHP memory.
Getting rid of this bottleneck would require to use [unbuffered queries](https://www.php.net/manual/en/mysqlinfo.concepts.buffering.php) which would imply much bigger changes to the codebase so I don't see it happening in the near future.

PR #9687, opened 2021-10-10
Add cartidge/consumable restock target value

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

Add a stock target value for consumables and cartridges.
The alarm threshold is the lowest amount desired to have in stock before ordering more but previously there was no way to track how much we want in stock. Ordering just one or two at a time to go back over the threshold could mean a lot of small orders instead of fewer larger ones. There could be a lot of consumables in inventory and the stock target may not always be known by the person doing the ordering. Previously, only the remaining stock was known in the notification email and not the alarm threshold.

Example:
I have 10' HDMI cables in stock. I want to keep 20 in stock but don't want to order another batch until I only have 5 left.

The stock target of 20 is a "soft" upper limit.

I added the `stock_target` and `to_order` tags to the alarm notifications to know the target stock level and the amount to order right from the notification. If the stock target is 0 (default), the amount to order will be the amount needed to go one over the alarm threshold.


templates/generic_show_form.html.twig
```diff
@@ -789,6 +789,17 @@
                      ) }}
                   {% endif %}
 
+                  {% if item.isField('stock_target') %}
```
orthagh, line 792: i think this addition should be in a specific twig template like done in `templates/pages/assets/monitor.html.twig`

orthagh (commented):
globally ok with the proposed changes.

orthagh (approved):

trasher (approved):

PR #9692, opened 2021-10-11
Kanban Filters

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

Replaces #9273

Kanban search now supports tags like `title:"My title"` or `type:project` to filter the shown items. You can still search without any tags and it falls back to searching on the title only (could be expanded to search on more shown items). The plain search works in addition to the tags, you can use both at the same time like `type:project My project`.

Added tags:

- title - The card title
- type - The itemtype of the item the card represents
- milestone - If a project task is a milestone or not (Supports numeric or boolean values)
- content - The plain-text content/description of the item
- team - Search among team members of any type
- user - Search all User team members
- group - Search all Group team members
- supplier - Search all Supplier team members
- contact - Search all Contact team members

You can prefix a tag with `-` to make it an exclusion. For example `-type:Project` will show all items that are NOT Projects.

### DIFF
--- a/css/includes/components/_kanban.scss
+++ b/css/includes/components/_kanban.scss
@@ -439,3 +439,21 @@
 .flex-break {
    width: 100%;
 }
+
+.popover.kanban-filter-popover {
+   background-color: inherit;
+   color: inherit;
+   min-width: 200px;
+
+   .popover-arrow {
+      display: none;
+   }
+
+   .popover-body {
+      color: inherit;
+
+      ul {
+         list-style: none;
+      }
+   }
+}
--- a/css/includes/components/_kanban.scss
+++ b/css/includes/components/_kanban.scss
@@ -443,15 +443,22 @@
 .popover.kanban-filter-popover {
    background-color: inherit;
    color: inherit;
+   max-width: unset;
 
    .popover-arrow {
-      display: none;
+      top: calc(-0.5rem - 1px); // Fix for using non-standard placement
+      transform: unset !important; // Fix for using non-standard placement
+      &::before  {// Fix for using non-standard placement
+         top: 0;
+         border-width: 0 (1rem * .5) 0.5rem (1rem * .5);
+         border-bottom-color: $popover-arrow-outer-color;
+      }
    }
 
    .popover-body {
       width: 400px;
-      overflow: auto;
       color: inherit;
+      overflow: auto;
 
       ul {
          list-style: none;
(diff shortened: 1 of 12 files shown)


cedric-anne (approved):

trasher (approved):

orthagh:
I tried the branch just now and cannot get a full working solution (my tests are done on latest firefox):

- the hint panel content overflows (the 400px set in kanban.scss is to large for popover parent)
- the panel position is strange (cannot screenshot as the panel disappears when i launch the screenshot feature), it's currently centered, i think left alignment could be better. Maybe much better, the popover should also visually refers to the input with an arrow (see first example in https://getbootstrap.com/docs/4.0/components/popovers)
- the colors is strange also in light palette (the used background seems to be an header), there is background color set to inherit in kanban.scss
- in ticket global kanban, any try to filter on title (with or without keyword) empty all lists

cconard96:
The noted issues should be fixed now. By default, Bootstrap only supports centering the popover so I had to use the popperConfig option and then make a CSS workaround to get the arrow to show properly.

orthagh:
Sorry for the delay of review.

Last small issues with the pr i think :

    the arrow is too much on the left an the radius of the corner is do a empty space (i still can't screenshot)
    in example, we may set quotes, no ? like title:"The title of item"

After that, we'll be good i think.

cconard96:
>     * the arrow is too much on the left an the radius of the corner is do a empty space (i still can't screenshot)
Do you want it centered or left at the left but remove the border radius on that corner?

>     * in example, we may set quotes, no ? like title:"The title of item"
Yes. In fact they are required if there are spaces or colons `:`  to ensure they are parsed correctly.

orthagh:
> Do you want it centered or left at the left but remove the border radius on that corner?

I don't have any preference, choose the option requiring the less css on our side.

orthagh:
> Yes. In fact they are required if there are spaces or colons : to ensure they are parsed correctly.

I'm not sure i was very clear, the example provided in the popover doesn't contains theses quotes and can mislead users

orthagh:
Apart this last point, pr is ready to be merged

orthagh (approved):

trasher:
Tests are failing :/

cconard96:
> Tests are failing :/

I fixed the JS tests but now there are unrelated PHP functional test failures...

trasher:
Yes, that one happens from time to time :/ Not a blocker to merge current PR

PR #9757, opened 2021-10-22
fix card palettes cache using random id

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

issue : card using a gradient palette doesn't have their color applied correctly after a refresh
cause :  `Widget::getCssGradientPalette()` uses a generated css prefixed by a parent id.
This id was generated by a rand.
So the cache worked but return a css targeting nothing.
I change the id of impacted card for something more predictable.

At the moment, i just changed cards using the mentionned function but i can extends to all cards if you prefer.

### DIFF
--- a/inc/dashboard/widget.class.php
+++ b/inc/dashboard/widget.class.php
@@ -535,6 +535,8 @@ HTML;
       $dark_bg_color = Toolbox::getFgColor($p['color'], 80);
       $dark_fg_color = Toolbox::getFgColor($p['color'], 40);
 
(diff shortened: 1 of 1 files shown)


orthagh:
After retesting this changes, it seems ok finally.
So ready for review

cconard96 (approved):

trasher (approved):

cedric-anne (approved):

PR #9760, opened 2021-10-23
Re-add ability to add items on ticket creation

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

Re-implemented the previous code that let you select one or more items during ticket creation. Currently, you have to add all of the items one at a time after the ticket is created.

### DIFF
--- a/ajax/dropdownTrackingDeviceType.php
+++ b/ajax/dropdownTrackingDeviceType.php
@@ -102,10 +102,10 @@ if ($isValidItemtype) {
    );
 
    // Auto update summary of active or just solved tickets
-   echo "<span id='item_ticket_selection_information$rand' class='ms-1'></span>";
+   echo "<span id='item_ticket_selection_information{$_POST["myname"]}_$rand' class='ms-1'></span>";
    Ajax::updateItemOnSelectEvent(
       $field_id,
-      "item_ticket_selection_information$rand",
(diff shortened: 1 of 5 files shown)


cconard96:
>     1. If I add more than 5 items (using the `all_items_dropdown`), I only see 5 of them, and I do not see the `Display all items` link.

Seems OK for me. Shows up after 5+ items are already added and not pending to be added.
>     2. During selection of items in the `all_items_dropdown`, a `0` counter appears, then dissapear when item is added. I do not understand meaning of this counter.

This is the same behavior as 9.5 and not added by me. The counter is to show how many tickets are in progress or recently solved for the currently selected item. It is a good indicator so the user knows there is a chance they are entering a duplicate ticket or there is a repeat issue.

cedric-anne:
1) When more than 5 elements are linked to the ticket, the counter is incorrect in the title. Indeed, it will never show a value greater than 5.
image

2) Following counter should be displayed next to the used dropdown.
image

cedric-anne (approved):

trasher (approved):
Please fix lint

orthagh (approved):

PR #9762, opened 2021-10-24
Fix potential empty value

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

### DIFF
--- a/inc/impact.class.php
+++ b/inc/impact.class.php
@@ -475,7 +475,7 @@ class Impact extends CommonGLPI {
             $("#impact-list-settings").click(function() {
                glpi_html_dialog({
                   title: __("Settings"),
-                  body: '.$setting_dialog.',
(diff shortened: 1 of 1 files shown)


cconard96:
Actually, it seems I only saw this error because of a bad `impactcontexts_id` value. I'm not sure how that happened...

In a normal case, the settings button isn't shown without update rights or a valid context. The only issue is that the empty value malforms the JS even if the code would never be called. The JS property is JSON encoded HTML, not a plain HTML string so I put an empty object as the default. This is only to make the script valid and should never be run.

AdrienClairembault (approved):

cedric-anne (approved):

orthagh (approved):

Issue #7144, opened 2020-04-18
Multi-click on submit button returns csrf error page


AKorezin:
**Describe the bug**

Multi-clicking on submit button on any form submit button sends you to csrf error page with text: `The action you have requested is not allowed.`
Can be triggered by:
- fast double-click;
- multiple clicks if server is slow.

**Page(s) URL**

Any page with form to submit.

**To reproduce**

1. Go to `Create a ticket` page.
2. Fill any data in form to make it valid
3. Fast double-click on submit button at the bottom.
4. Wait until redirect.
5. See csrf error page.

**Expected behavior**

I don't want to see an error page, because my ticket was created successfully.

**Additional context**

It's reproducible in https://demo.glpi-project.org/
I've patched `session.class.php` to display more logs to debug this problem.
Redis/memcached sessions can create race condition if locking is disabled.

AKorezin:
`9.5.6` - reproduced.
https://demo.glpi-project.org site don't work.

tomolimo:
Hello @AKorezin 
To prevent this issue, you may use the Mask plugin
Thank you
Regards,
Tomolimo

AKorezin:
Bump
I think this feature should be in main branch.
Even there's the [mask plugin](https://github.com/tomolimo/mask).

PR #9908, opened 2021-11-22
Debounce submit and regular buttons

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #7144

Add ability to specify that a button should be debounced. For submit buttons, this means preventing them from triggering another form submit after clicking the first time. For other buttons, it means disabling them for a certain amount of time.

This does not affect any existing code except the buttons in the `components/form/buttons.html.twig` template because the debounce is only enabled by default when using a new macro in `components/form/fields_macros.html.twig`. When using the macro, the debounce functionality is on by default but can be disabled.

The original plan was to disable submit buttons for a certain amount of time but it broke form submission completely and I don't know of any situation when the actual form submission would fail so it was changed to just disabling the default behavior.

Replaces part of the functionality of the mask plugin.


js/common.js
```diff
@@ -1372,3 +1372,25 @@ function flashIconButton(button, button_classes, icon_classes, duration) {
...
    }, duration);
 }
+
+/**
+ * 'Debounces' the click event of a button.
+ * For submit buttons, this will prevent the form from being submitted multiple times.
+ * For other buttons, this will disable them for a specified period of time based on the data-debounce-timeout attribute (defaults to 1 second)
+ * @param btn
+ */
+function debounceButton(btn) {
```
cedric-anne, line 1382: 1. IMHO, same behaviour should be applied to both submit and base buttons.
2. `debounce-timeout` may lead to unexpected results. For the developper, I guess it would be preferable to have a `undebounceButton` function (or any other better name), that could be used to restore the default button state. Indeed if the `data-debounce` is manually set by developper, I guess he will be able to call this function on the most appropriate moment (on ajax success, ajax failure, on error display, ...).
3. Maybe adding a `<span class="spinner-border spinner-border-sm" role="status" aria-hidden="true"></span>` could enhance the UX.

templates/components/form/fields_macros.html.twig
```diff
@@ -639,6 +639,18 @@
...
 {% endmacro %}
 
+{% macro submit(name, content, value, options = {}) %}
+   {% set options = {
+      'rand': random(),
+      'class': 'btn btn-primary me-2',
+      'debounce': true,
+      'debounce_timeout': 1000
+   }|merge(options) %}
+   {% set on_click = options.debounce ? '' : 'return true;' %}
```
cedric-anne, line 649:
```suggestion
```

js/common.js
cconard96: If the developer is going to handle manually re-enabling the button, it doesn't make much since to use this function since all it does for non-submit buttons is disable them. They should just add a click listener that disables the button and then re-enable it when needed.

If the timeout isn't desired, I guess this function should only be for submit buttons where we expect the page to reload or change so it never needs re-enabled.

cedric-anne: It may make sence if this function does more than just disabling the button, for instance displaying a loader or any other visual hint.

js/common.js
```diff
@@ -1374,16 +1374,53 @@ function flashIconButton(button, button_classes, icon_classes, duration) {
...
+ * If debounce-timeout is not specified, the button must be manually undebounced with the `undebounceButton` function.
+ * @param btn
+ */
+function debounceButton(btn) {
+   btn.data('original-content', btn.html());
+   btn.html(`<span class="spinner-border spinner-border-sm" role="status" aria-hidden="true"></span>`);
+   if (btn.attr('type') === 'submit') {
+      btn.click((e) => {e.preventDefault();});
+   } else {
+      btn.prop('disabled', true);
```
cedric-anne, line 1404: I still think that behaviour should be the same for both `submit` and `button`. Indeed, `submit` could also be used in an AJAX context (to handle submission of for using `Enter` key).
Thus, I did not tested it, but I am pretty sure that using the `Enter` key would submit the form enven if the button has been debounced.

Finally, we may have mutiple submit or regular buttons on a form (each having a different value). Maybe this function should debounce all buttons of the form, to prevent execution of another action related to the same form.

cconard96: The ability to trigger a form submission when pressing enter or ctrl-enter in form fields added by us in `common.js` just triggers a click on the `add` or `update` submit button and only if there is only one of them present. Therefore, debouncing the submit button still prevents the form submission that way.

cconard96: I'm willing to just not worry about regular buttons for this PR just to get it merged. The entire point was to prevent the confusing `The action you have requested is not allowed.` error page you get when clicking a `submit` button more than once. Having some kind of functionality to handle regular buttons was just extra.

The submit buttons cannot be disabled because that prevents even the first form submission. Regular buttons cannot just have a click handler `e => e.preventDefault()` added because that doesn't prevent anything for some reason. Being able to re-enable the submit button doesn't make sense because it should only need to be clicked once and then the page reloads. Clicking a regular button more than once makes sense because in general, they shouldn't cause a page reload. Having some kind of timeout for regular buttons would have been nice for cases where you are making AJAX calls and don't them to stack or happen so frequently.

cedric-anne: > The ability to trigger a form submission when pressing enter

This one is a native HTML behaviour.

https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement/submit_event says:
> The submit event fires when the user clicks a submit button (<button> or <input type="submit">) or presses Enter while editing a field (e.g. <input type="text">) in a form. The event is not sent to the form when calling the form.submit() method directly.

But you are right, we should focus on submit buttons only for now.

cconard96:
Logic for debouncing regular buttons was removed.
Since it is just focusing on form submissions now, a new flag attribute was added on many traditional/non-AJAX forms `data-submit-once` that prevents submitting it again.

When a submit event is fired on a form with `data-submit-once`, it looks for the triggering button (or a best possible alternative) and shows a spinner in it to indicate that action is in progress. All submit buttons in the form are effectively disabled. A `data-submitted` attribute is added to the form to handle other cases where the form could be submitted beyond the disabled buttons.
If the form gets submitted again with the `data-submitted` attribute present, the event is cancelled.

templates/components/form/buttons.html.twig
```diff
@@ -42,45 +44,51 @@
       <div class="card-body mx-n2 mb-4  border-top">
          {% if withtemplate|length > 0 or item.isNewID(id) %}
             {% if id <= 0 or withtemplate == 2 %}
-               <button class="btn btn-primary me-2" type="submit" name="add" value="1">
+               {% set content %}
```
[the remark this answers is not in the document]
cconard96: > `content` may be too generic. As we almost always pass variables to included templates, you may unvoluntarly override an existing `content` variable.
> Maybe `button_content` would be more secure.

Actually, the submit button macro isn't needed anymore. Everything is completely handed during the form submit now and doesn't need any attributes on the submit button.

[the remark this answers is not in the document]
cedric-anne: Maybe you could revert this part then.

orthagh (commented):
Not sure if it's the change done in #10023, but when i click on trash a computer, the loader animation is replacing the save (primary) button

cconard96:
It's this PR. When you click one submit button, all the submit buttons have
that currently. I can change it to just disable the ones that weren't
clicked.

cedric-anne:
I squashed and rebased to fix conflicts.

js/common.js
```diff
@@ -1381,3 +1381,42 @@ function uniqid(prefix = "", more_entropy = false) {
...
+
+/**
+ * Prevent submitting the form again
+ *
+ * A spinner is shown in the triggering submit button and all others are disabled.
+ * @param form The form to block submits
+ * @param {SubmitEvent} e The submit event
+ */
+function blockFormSubmit(form, e) {
+   let submitter = $(e.submitter);
```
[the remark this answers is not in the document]
cedric-anne: Already fixed by last commit.

js/common.js
```diff
@@ -1381,3 +1381,42 @@ function uniqid(prefix = "", more_entropy = false) {
...
+   if (!submitter.is('button')) {
+      submitter = submitter.find('button[name="add"]:first, button[name="update"]:first');
+      // If no submit button was found, use the first submit button
+      if (submitter.length === 0) {
+         submitter = form.find('button[type="submit"]:first');
+      }
+   }
+
+   if (submitter.length > 0 && submitter.is('button')) {
+      submitter.html(`<span class="spinner-border spinner-border-sm" role="status" aria-hidden="true"></span>`);
```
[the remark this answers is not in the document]
orthagh: I don't know, i like the current version

orthagh (approved):
Ok for me

cedric-anne (approved):

PR #9948, opened 2021-11-26
more subtle style for goto menu

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

image

### DIFF
--- a/inc/html.class.php
+++ b/inc/html.class.php
@@ -6336,7 +6336,7 @@ JAVASCRIPT;
                      <div class="modal-content">
                         <div class="modal-header">
                            <h5 class="modal-title">
-                              <i class="fas fa-arrow-right me-2"></i>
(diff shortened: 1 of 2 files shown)


trasher (approved):

AdrienClairembault (approved):
Looks much better :+1:

Not super important but maybe the background is a bit too light ?

For comparison, mattermost use a color scheme very similar to ours and the background is darker :
image

cedric-anne (approved):

orthagh:
> Not super important but maybe the background is a bit too light ?

I'm currently relying on secondary color
changing that is a more complex change as i need to do also dark theme.
Maybe later

PR #9962, opened 2021-11-29
Better Kanban Filter UX

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

Done:
- Added a better visualization of the different tags entered
- Added ability to select the tag and term suggestions from the popover (keyboard support too)
- Separated the UI part of the SearchTokenizer into a SearchInput class and left just the tokenization part in the original file
- The Babel config file was renamed as part of work to get a Tabler script to import in tests. It may not have been read at all before, but it is now required for scripts in "node_modules" to be transformed to modules.

To do:
- [x] Make the tag part bold (title, milestone, etc)
- [x] Support regex terms

Basic preview of the current state:
Peek 2021-11-29 04-22


orthagh:
Some suggestions:

- use `.badge .bg-secondary` classes for tags (note you need to add `text-transform: inherit; font-weight: inherit` to avoid the forced changes of tabler)
- typing in search bar should filter the tags list suggestion panel, and a selection (by mouse or keyboard should remove the searched term)
- the suggestion list is very spaced vertically, addding `padding: 0.3rem 0.5rem;` for `.list-group_item` should be better
- with the new icons set, use `ti ti-x` for close icon
- with dev-tools opened, i have some placement issues
  image
probably not due to this pr, but in responsive mode, search input overlap viewport width. We may set project dropdown and search input on two rows, we maybe set the container (kanban-toolbar) as `.flex-column .flex-md-row`

cconard96:
>     * use `.badge .bg-secondary` classes for tags (note you need to add `text-transform: inherit; font-weight: inherit` to avoid the forced changes of tabler)

Done

>     * typing in search bar should filter the tags list suggestion panel, and a selection (by mouse or keyboard should remove the searched term)

Added filtering to the tag and autocomplete values when typing. Not sure what the second point refers to.

>     * the suggestion list is very spaced vertically, addding `padding: 0.3rem 0.5rem;` for `.list-group_item` should be better
> 
>     * with the new icons set, use `ti ti-x` for close icon

Done

>     * with dev-tools opened, i have some placement issues
>       image
>       probably not due to this pr, but in responsive mode, search input overlap viewport width. We may set project dropdown and search input on two rows, we maybe set the container (kanban-toolbar) as `.flex-column .flex-md-row`

The odd placement looks like the expected, default behavior for the Bootstrap popover. It cannot fit the contents under the search bar so it moved it to the left which allowed more of it to be shown. The arrow looks wrong because of the overrides I had to make in CSS to center it.
I'll see if there is a clean way to make the list smaller based on screen size and add a scroll.

orthagh:
> Added filtering to the tag and autocomplete values when typing. Not sure what the second point refers to.

ex for second term
image
I get this after typing part of the entry i want, and then pressing [enter]

> We may set project dropdown and search input on two rows, we maybe set the container (kanban-toolbar) as `.flex-column .flex-md-row`

For this part, it's my fault, but i may have desribe it as a separate point and with a picture like this:
image
Set this flex display as column under medium screen size ( `.flex-column .flex-md-row`)

Some suggestions for the current design of the suggestions list

- add btm-sm and ms-1 (when there's multiple button) to additional button
- i wanted to suggest to identity these buttons with ti or fa icons but i cannot found any good ones, maybe regex should be `#` and exclude maybe more name "not" (so `!` char)
- use `text-muted fst-italic` for helper text (ìtalic` doesn't exist)
- add `shadow` to the popover
- type sublist entries should be prefixed by their respective icons 
- the close button in badge could be spaced a little (add `ms-1`)

In summary, something like that:
image

orthagh:
something missing in the last commits ?

image

cconard96:
> something missing in the last commits ?
> 
> image
> 
> i got this after selecting a filter

Not missing, just a bug but I don't really know what caused it. I just pushed a commit that fixes it and forces the popover to refresh instead of showing the old auto-complete info.

orthagh:
Last batch of suggestions, i think, we're almost good:

- there is a similar js error with regex 
- i still get the behavior of my text remaining in search box when i want to only search for a specific filter
- `me-3` spacing between badges is very large, i think `me-1` should be enough
- icon in type dropdown is good, a recall in the badge could be better
- do we still can add raw text for filtering ? if yes, the suggest panel still display empty but i don't know how we can do better
   - maybe press [enter] key after typing an unknow tag transform it into a generic text one (surrounded by quotes and with a badge also). 
   - can be editable so
   - we can continue to add normal badges after a quoted text (we can't now)
   - What do you think ?

cconard96:
>     * icon in type dropdown is good, a recall in the badge could be better

This could be a complex addition to handle it. It would be easy to copy the icon into the badge when selecting it from the list, but I'm not sure a good, clean way to automatically show the icon when you manually type one of the supported itemtype and still have the Kanban and Search Input code separate.

>     * do we still can add raw text for filtering ? if yes, the suggest panel still display empty but i don't know how we can do better
>       
>       * maybe press [enter] key after typing an unknow tag transform it into a generic text one (surrounded by quotes and with a badge also).
>       * can be editable so
>       * we can continue to add normal badges after a quoted text (we can't now)
>       * What do you think ?

Raw/un-tagged search is possible still but not instant. You have to hit enter or click outside the search box for it to refresh the search. It is a bit weird though with undefined behaviors when mixed with tagged searches. The plain text stays in the placeholder input inside the search box. When you type a tagged search in the placeholder and press enter, the tagged search is moved out and into a new badge while the un-tagged text remains (or should). Right now it looks like having plain text in the placeholder input breaks the popover suggestions because it doesn't know where the plain text ends and where the desired tagged search starts.

orthagh:
Ok did a batch of tests right now, and it seems pretty good

last thing i found : i tried some regexes but "starting by" keyword seems not working:
image
if i validate the tag, all card disapears

cconard96:
Fixed. The title text has a lot of whitespace in it when pulled by jQuery. I trimmed the title text and also added a trim directly in the regex matcher.

orthagh (approved):

trasher (approved):

PR #9964, opened 2021-11-29
Fix project clone

AdrienClairembault:
When cloning a project, the "Project Team" data was not copied correctly:

image

1 -> My "original" team with one group (id =1) and one user (id = 3).
2 -> The incorrect data over multiple clone tests, you can see the project id was used instead of the correct user/group ids.
3 -> One more clone test after fix -> OK.

Seem like the `override_input` variable used wasn't reinitialized for each itemtype so some data was overlapping when it shouldn't.
Here I guess `items_id` was set from one of the previous clone relation (`Document_Item` or `ProjectTask`).

I've also added a new test to check that projects are cloned correctly.
Before fix:
image

After fix:
image

|  Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !22986

### DIFF
--- a/inc/features/clonable.class.php
+++ b/inc/features/clonable.class.php
@@ -56,6 +56,8 @@ trait Clonable {
 
       $clone_relations = $this->getCloneRelations();
       foreach ($clone_relations as $classname) {
+         $override_input = [];
+
          if (!is_a($classname, CommonDBConnexity::class, true)) {
             Toolbox::logWarning(
                sprintf(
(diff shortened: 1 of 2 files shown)


AdrienClairembault:
PR was not targeting the correct branch, fixed.

AdrienClairembault:
Force pushed to force the test to run, seems like they were stuck on "waiting for report" after branch change

orthagh (approved):

trasher (approved):
:+1:

cedric-anne (approved):

PR #10012, opened 2021-12-03
Central counter links fix

felipebart10:
When clicking on the counters next to "Problems tasks to do", the link would send user to "Tickets" section instead of "Problem" section. I noted this when I was adding "Changes" information on central screen.

I tested it and now the link is sending user to correct place. I did with a variable because in my instance I'm adding "Changes" info on central too. I would send a pull request for it too, but I've seen it was already implemented.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | don't know what that means

### DIFF
--- a/inc/commonitiltask.class.php
+++ b/inc/commonitiltask.class.php
@@ -1498,8 +1498,8 @@ abstract class CommonITILTask extends CommonDBTM implements CalDAVCompatibleItem
                } else if ($itemtype == "ProblemTask") {
                   $title = __("Problem tasks to do");
                }
-               $itemtype_url = strtolower(str_replace("Task","",$itemtype));
-               echo "<a href=\"".$CFG_GLPI["root_doc"]."/front/".$itemtype_url.".php?".
+               $linked_itemtype = str_replace("Task","",$itemtype);
+               echo "<a href=\"".$linked_itemtype::getSearchURL()."?".
                       Toolbox::append_params($options, '&amp;')."\">".
                       Html::makeTitle($title, $displayed_row_count, $total_row_count)."</a>";
                break;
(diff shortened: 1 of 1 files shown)


inc/commonitiltask.class.php
```diff
@@ -1498,7 +1498,8 @@ static function showCentralList($start, $status = 'todo', $showgrouptickets = tr
                } else if ($itemtype == "ProblemTask") {
                   $title = __("Problem tasks to do");
                }
-               echo "<a href=\"".$CFG_GLPI["root_doc"]."/front/ticket.php?".
+               $itemtype_url = strtolower(str_replace("Task","",$itemtype));
+               echo "<a href=\"".$CFG_GLPI["root_doc"]."/front/".$itemtype_url.".php?".
```
trasher, line 1502: Indeed, the URL is not good. Maybe better to do something like:
```suggestion
               $linked_itemtype = str_replace("Task","",$itemtype);
               echo "<a href=\"".$linked_itemtype::getSearchURL()."?".
```

felipebart10: Thanks, i'm not familiar with all glpi classes methods.

trasher (approved):

cedric-anne (approved):

inc/commonitiltask.class.php
```diff
@@ -1498,7 +1498,8 @@ static function showCentralList($start, $status = 'todo', $showgrouptickets = tr
                } else if ($itemtype == "ProblemTask") {
                   $title = __("Problem tasks to do");
                }
-               echo "<a href=\"".$CFG_GLPI["root_doc"]."/front/ticket.php?".
+               $linked_itemtype = str_replace("Task","",$itemtype);
```
trasher, line 1501:
```suggestion
               $linked_itemtype = str_replace("Task", "", $itemtype);
```

orthagh (approved):

PR #10098, opened 2021-12-13
Move item name to the main header

AdrienClairembault:
As discussed this afternoon, a possible UI change.

Before:

image

After:

image

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |


src/CommonGLPI.php
```diff
@@ -1096,10 +1077,39 @@ class='btn btn-sm btn-icon btn-ghost-secondary'
...
             echo "</div>"; // .navigationheader
         }
     }
 
+    /**
+     * Compute the name to be used in the main header of this item
+     *
+     * @return string
+     */
+    public function getHeaderName(): string {
```
trasher, line 1091:
```suggestion
    public function getHeaderName(): string
    {
```

trasher:
Seems better to me

src/CommonGLPI.php
```diff
@@ -1096,10 +1077,40 @@ class='btn btn-sm btn-icon btn-ghost-secondary'
...
+            && $this->isEntityAssign()
+        ) {
+            $entname = Dropdown::getDropdownName("glpi_entities", $this->fields["entities_id"]);
+            if ($this->isRecursive()) {
+                $entname = sprintf(__('%1$s + %2$s'), $entname, __('Child entities'));
+            }
+            $name = sprintf(__('%1$s (%2$s)'), $name, $entname);
+        }
+
+        return $name;
```
cedric-anne, line 1112: If item name contains `>`, `<` or `&` chars, they will be displayed as html entities. This method should return an unsanitized value, and twig will handle the escaping itself.

```suggestion
        return Sanitizer::unsanitize($name);
```

May require an additionnal use statement.

cedric-anne (approved):

trasher (approved):

src/CommonGLPI.php
```diff
@@ -1096,10 +1078,40 @@ class='btn btn-sm btn-icon btn-ghost-secondary'
...
+            isset($this->fields["entities_id"])
+            && Session::isMultiEntitiesMode()
+            && $this->isEntityAssign()
+        ) {
+            $entname = Dropdown::getDropdownName("glpi_entities", $this->fields["entities_id"]);
+            if ($this->isRecursive()) {
+                $entname = sprintf(__('%1$s + %2$s'), $entname, __('Child entities'));
+            }
+            $name = sprintf(__('%1$s (%2$s)'), $name, $entname);
+        }
```
orthagh, line 1111: I think the whole part about entity is not needed.
The header where the title is inserted already have a badge for displaying the entity

templates/components/form/header.html.twig
```diff
@@ -86,7 +87,7 @@
                   </div>
                {% endif %}
                <span>
-               {{ nametype }}
+               {{ nametype }} - {{ friendlyname }}
```
orthagh, line 90: I'm currently asking myself if the nametype is really necessary.
We have the information from the tab name

orthagh (approved):

templates/components/form/header.html.twig
AdrienClairembault: Not a big deal but if I had to choose I'd keep the nametype.
It makes sense to me that the most important header should contain the full information (type + name).

orthagh (approved):

AdrienClairembault:
Test failure seems to be unrelated ?
image

This specific test run without issues on my PC.

trasher:
> Test failure seems to be unrelated ?

Indeed, I see that one on several recent PRs... No idea what's wrong right now

PR #10100, opened 2021-12-13
SocketModel should use same namespace as Socket

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

Multiple places in the code expects an asset model class to be the fully qualified class of the asset followed by `Model`. For sockets, the `Socket` is in the Glpi namespace while the `SocketModel` is in the root so it cannot be automatically discovered.

### DIFF
--- a/front/socketmodel.form.php
+++ b/front/socketmodel.form.php
@@ -31,6 +31,8 @@
  * ---------------------------------------------------------------------
  */
 
+use Glpi\SocketModel;
+
 include('../inc/includes.php');
 
 $dropdown = new SocketModel();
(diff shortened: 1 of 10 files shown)


trasher:
Tests are failing :(

trasher (approved):
Assuming it is correct that `Socket` is part of `Glpi` namespace, sounds good to me.

cedric-anne (approved):

PR #10114, opened 2021-12-14
Improve browser tab names

AdrienClairembault:
Another UI proposal.

Two changes:

1\) Move "GLPI" to the end of the tab name.
Our tabs names are currently like this: "GLPI - XXXXX".
I've noticed that most website do the opposite and put the must important information first, for example :

image

The main advantage is that the information on the left is always shown while the information on the right side may be hidden if you have too many tabs and they start to shrink.
It makes sense to keep the most important information always visible.

2\) Remove ids and use objects name.
All commondbtm items tab are displayed as "GLPI - {itemtype plural} - {item id}".
This is not very helpful for the user as ids are more meant for database than for humans.
If you have multiple computer tabs open for example, it's not easy to remember which is which.
Also, using the plural here seems inaccurate since we are viewing a single item.

Before changes:

image

After changes:

image

This was only done on computers for now, I'll keep this PR as a draft and wait for validation before editing all ".form.php" files .

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |

### DIFF
--- a/front/agent.form.php
+++ b/front/agent.form.php
@@ -120,11 +120,9 @@ if (isset($_POST["add"])) {
     );
     Html::back();
 } else {//print agent information
-    Html::header(Agent::getTypeName(Session::getPluralNumber()), $_SERVER['PHP_SELF'], "admin", "glpi\inventory\inventory", "agent");
-   //show agent form to add
-    $agent->display([
-        'id'           => $_GET["id"],
-        'withtemplate' => $_GET["withtemplate"]
+    $menus = ["admin", "glpi\inventory\inventory", "agent"];
+    Agent::displayFullPageForItem((int) $_GET['id'], $menus, [
+        'withtemplate' => $_GET["withtemplate"],
+        'formoptions'  => "data-track-changes=true",
     ]);
-    Html::footer();
 }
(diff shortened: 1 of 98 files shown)


AdrienClairembault:
> Small feedback, it is good to use names for assets but I think it shouldn't be changed for Projects or ITIL Objects since their IDs are far more important than a shortened title/name.
> 
> I don't know how many tickets I've seen with near identical subjects.

I agree, ticket "names" should be just their id of "#id".

cconard96:
Actually I'm not sure. I guess it is situational. Ticket subjects would be OK in some cases where not many tabs are open or are different enough.

AdrienClairembault:
New item tab:

image

AdrienClairembault:
Up, as said in the description I'd like a first validation before modifying all relevant front files.

Also, looking back on this I wonder if putting the item name before the type would be even better context wise ?

cedric-anne:
> Up, as said in the description I'd like a first validation before modifying all relevant front files.
Seems technically OK for me.

> Also, looking back on this I wonder if putting the item name before the type would be even better context wise ?
Ticket ID may be really important for many people, and the name could be really long (and not always meaningful), so, maybe following patterns should be used:
- existing item: `{$itemtype_name} ({$id}) - {$name} - GLPI` (e.g. `Ticket (29564) - My computer does many bips and do not boot - GLPI`
- new item: `{$itemtype_name} - New item - GLPI`

orthagh (approved):

cedric-anne (approved):

PR #10115, opened 2021-12-14
Fix guess  IP from Agent

stonebuzz:
Actually, GLPI try to get all addresses from computer to request a remote inventory / status

from a standard inventory GLPI return three addresses (or more depending of inventory) (```NetworkPortLocal``` are excluded)
```https://192.168.1.24:62354/status``` (by ipv4)
```https://2a01:e0a:87c:4ea0:f178:597:e709:8aea:62354/status``` (by ipv6)
```https://Desktop:62354/status``` (by name)

As discuss with @g-bougard 

When agent contact GLPI save into ```glpi_agents``` table two new informations

- IP ```$_SERVER['REMOTE_ADDR']``` use by agent to contact GLPI (store as binary with ```inet_pton```)
- Protocol : ```http``` by default , ```https``` if plugin ```ssl``` is active agent side 

When user request ```/status```, ```/now``` (or other from plugins)
GLPI use this IP and make only one call.

| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #number

### DIFF
--- a/src/Agent.php
+++ b/src/Agent.php
@@ -166,18 +166,18 @@ class Agent extends CommonDBTM
         return $ong;
     }
 
-    static function getSpecificValueToDisplay($field, $values, array $options = []) 
+    public static function getSpecificValueToDisplay($field, $values, array $options = [])
     {
         if (!is_array($values)) {
-           $values = [$field => $values];
+            $values = [$field => $values];
         }
         switch ($field) {
-           case 'ip_binary':
-              if ($ipadress = inet_ntop($values[$field])) {
-                 return $ipadress;
-              } else {
-                 return __('IP address not properly formatted');
-              }
+            case 'ip_binary':
+                if ($ipadress = inet_ntop($values[$field])) {
+                    return $ipadress;
+                } else {
+                    return __('IP address not properly formatted');
+                }
         }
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Review by trasher — APPROVED** [review]

**Comment by trasher** [comment]
please run `vendor/bin/phpcbf install/migrations/update_9.5.x_to_10.0.0/native_inventory.php src/Agent.php`

File: src/Agent.php [hunk 1]
```diff
@@ -348,6 +370,24 @@ public function handleAgent($metadata)
             $input['tag'] = $metadata['tag'];
         }
 
+        if (isset($_SERVER['REMOTE_ADDR'])) {
+            if ($ip_binary = inet_pton($_SERVER['REMOTE_ADDR'])) {
```
trasher, line 374: As discussed, we cannot rely on that; it will be wrong in czse of uploading an XML file, or if there are proxies between agent and glpi.

trasher (changes requested):

g-bougard:
Would it be possible to resurrect this PR but updating guessAddress adding the know agent ip and not changing requestAgent ?

PR #10163, opened 2021-12-17
Browse items by category

Rom1-B:
Browse items by category, as already exists for kb items

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !22497

image

image

### DIFF
--- a/ajax/treebrowse.php
+++ b/ajax/treebrowse.php
@@ -61,6 +61,7 @@ switch ($_REQUEST['action']) {
             }
         }
 
+        $params['treecriteria'] = count($params['criteria']);
         $params['criteria'][] = [
             'link'   => "AND",
             'field'  => $field,
--- a/ajax/treebrowse.php
+++ b/ajax/treebrowse.php
@@ -1,0 +1,73 @@
+<?php
+
+/**
+ * ---------------------------------------------------------------------
+ * GLPI - Gestionnaire Libre de Parc Informatique
+ * Copyright (C) 2015-2022 Teclib' and contributors.
+ *
+ * http://glpi-project.org
+ *
+ * based on GLPI - Gestionnaire Libre de Parc Informatique
+ * Copyright (C) 2003-2014 by the INDEPNET Development Team.
+ *
+ * ---------------------------------------------------------------------
+ *
+ * LICENSE
+ *
+ * This file is part of GLPI.
+ *
+ * GLPI is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * GLPI is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with GLPI. If not, see <http://www.gnu.org/licenses/>.
+ * ---------------------------------------------------------------------
+ */
+
+include('../inc/includes.php');
+header("Content-Type: text/html; charset=UTF-8");
+Html::header_nocache();
+
+switch ($_REQUEST['action']) {
+    case "getItemslist":
+        $params = [
+            'start'              => $_REQUEST['start'],
+            'is_deleted'         => $_REQUEST['is_deleted'],
+            'browse'             => 1,
+            'as_map'             => 0,
+            'showmassiveactions' => false,
+            'criteria'           => $_REQUEST['criteria'],
+        ];
+
+        $itemtype = $_REQUEST['itemtype'];
+        $cat_table = getTableNameForForeignKeyField($_REQUEST['cat_field']);
+        $item = new $itemtype();
+        $so = $item->rawSearchOptions();
+
+        $field = 0;
+        foreach ($so as $value) {
+            if (isset($value['field'])) {
+                if (($value['field'] == 'name' || $value['field'] == 'completename') &&  $value['table'] == $cat_table) {
+                    $field = $value['id'];
+                }
+            }
+        }
+
+        $params['criteria'][] = [
+            'link'   => "AND",
+            'field'  => $field,
+            'searchtype'   => "equals",
+            'value'  => ($_REQUEST['cat_id'] > 0) ? $_REQUEST['cat_id'] : 0,
+        ];
+        Search::showList($itemtype, $params);
+        break;
+}
+http_response_code(400);
+return;
--- a/ajax/treebrowse.php
+++ b/ajax/treebrowse.php
@@ -47,14 +47,15 @@ switch ($_REQUEST['action']) {
         ];
 
         $itemtype = $_REQUEST['itemtype'];
-        $cat_table = getTableNameForForeignKeyField($_REQUEST['cat_field']);
+        $category_itemtype = $itemtype::getCategoryItemType($itemtype);
+        $category_table = $category_itemtype::getTable();
         $item = new $itemtype();
         $so = $item->rawSearchOptions();
 
         $field = 0;
         foreach ($so as $value) {
             if (isset($value['field'])) {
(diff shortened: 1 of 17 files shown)


ajax/document.php
```diff
@@ -0,0 +1,55 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with GLPI. If not, see <http://www.gnu.org/licenses/>.
+ * ---------------------------------------------------------------------
+ */
+
+include('../inc/includes.php');
```
cconard96, line 34:
```suggestion
include('../inc/includes.php');
header("Content-Type: text/html; charset=UTF-8");
```

ajax/document.php
```diff
@@ -0,0 +1,55 @@
...
+Html::header_nocache();
+
+switch ($_REQUEST['action']) {
+    case "getItemslist":
+        header("Content-Type: application/json; charset=UTF-8");
```
cconard96, line 39: Search::showList seems to only output HTML content, not JSON
```suggestion
```

ajax/document.php
```diff
@@ -0,0 +1,55 @@
...
+        ];
+        $params['criteria'][] = [
+            'link'   => "AND",
+            'field'  => "7",
+            'searchtype'   => "equals",
+            'value'  => $_REQUEST['cat_id']
+        ];
+        Search::showList('Document', $params);
+        break;
+}
```
cconard96, line 55:
```suggestion
}
http_response_code(400);
return;

```

stonebuzz:
do not display toggles when treeview mode is enabled

image

Rom1-B:
It should be ok now

src/Search.php
```diff
@@ -1728,6 +1731,7 @@ public static function displayData(array $data)
          ]),
          'may_be_deleted'      => $item instanceof CommonDBTM && $item->maybeDeleted(),
          'may_be_located'      => $item instanceof CommonDBTM && $item->maybeLocated(),
+         'may_be_browsed'      => $item instanceof CommonDBTM && $item->maybeBrowsed(),
```
cedric-anne, line 1734: Maybe here you could just check `Toolbox::hasTrait($item)`

cedric-anne (dismissed):
For me, this PR is almost ready to be merged.

The tree display should be improved. On my browser, the left border seems to be missing
image

Also, I still have some edge cases that may require a fix.

1) The selected category is not stored, so when the page is refresh (manually, or when user changes displayed columns), the previously selected category is lost and the items without any category are displayed.

2) When adding a new criterion, or when switching to "trashbin" view, page is refreshed, and selected category is lost. Even if it was stored, it may be empty when using new criteria, so it may not be present in the tree.

src/CommonITILObject.php
```diff
@@ -46,6 +46,7 @@ abstract class CommonITILObject extends CommonDBTM
     use \Glpi\Features\Timeline;
     use \Glpi\Features\Kanban;
     use Glpi\Features\Teamwork;
+    use Glpi\Features\TreeBrowse;
```
cedric-anne, line 49: It could also be added to Softwares, I guess.

src/Features/TreeBrowse.php
```diff
@@ -0,0 +1,313 @@
...
+                    if ($cat_item instanceof CommonTreeDropdown) {
+                        return $category['id'] == $element[$cat_fk];
+                    }
+                }
+            );
+
+            if (empty($children) && 0 == $category['items_count']) {
+                unset($categories[$index]);
+            }
+        }
```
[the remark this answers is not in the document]
Rom1-B: I don't know, it's a shame to display empty categories (especially if there are many).
???

cedric-anne:
We do a functionnal review with @orthagh and here are our comments.

1) This feature should be remove on tickets for now, we may reactivate it on a future GLPI version, but with some enhancements. No need to clean the code, just remove usage of `Glpi\Features\TreeBrowse` in `CommonITILObject`.

2) When the selected category contains no elements, it should still be displayed and selected in the tree, so user will not be "lost" when changing criteria result in having no elements in current category.

3) When the browse view is displayed, the javascript logic behind the search view seems not correctly handled. Indeed, changing criteria triggers a full page refresh, changing the number of items displayed is not working, ...

4) The tree seems to be build using a limited search results, so the number of items in category tree is not correct. For instance, if there is 10 categories with 1 item in each, they will be all displayed when at least 10 rows/page are displayed, but the tree will be limited to 5 categories when 5 rows/page are displayed.

5) When pagination is used, a new criteria `Characteristics - Heading > is > ...` is added to search criteria. So once page is reload, tree contains only the current category and is not browsable anymore.

6) On browse view, a scrollbar is always added, even when there is only a few items displayed. Removing the `height: 100%` on `#tree_browse` style seems to fix this, but it may have side effects on KB view.

src/Features/TreeBrowse.php
```diff
@@ -162,6 +162,7 @@ public static function getTreeCategoryList(string $itemtype, array $params): arr
 
         $cat_itemtype = static::getCategoryItemType($itemtype);
         $cat_item     = new $cat_itemtype();
+        $params['list_limit'] = 99999;
```
cedric-anne, line 165: IMHO, this is really a bad idea.

I did not think about it in my previous reviews, but displaying items counter on the category tree may make this feature  unusable on instances that have many items. Indeed, we already have perfomances issues on search view (see #9495), and here we will, I guess, use twice the already used memory to compute this tree.

Also, if we finally found a solution for #9495 issue, we will never be able to find a one for this tree computation, as we will always need to fetch all matching items.

So, I really think we should remove counters from this tree. We could eventually filter categories by doing a simple SQL query that fetched count of items for each category, without any additionnal condition.

ajax/treebrowse.php
```diff
@@ -61,6 +61,7 @@
             }
         }
 
+        $params['treecriteria'] = count($params['criteria']);
```
cedric-anne, line 64: I had to read the code multiple time to understand what it the purpose of this.

Maybe you should add a key `'virtual' => true` to the criterai added below, and, before generating `$globallinkto`, removing all criteria having `'virtual' => true`.

This would make this logic more generic and may help us to add more "virtual" criteria.

cconard96:
It looks like when switching to the tree browse view, it replaces the `search-display-data` element and doesn't use the `display_data.html.twig` template so the JS isn't initialized for fluid search.
I think a new view would have to be created, possibly extending `js/modules/Search/Table.js` to handle it.

I can work on fluid search compatibility if you want.

cedric-anne:
@Rom1-B Can you check if https://github.com/glpi-project/glpi/pull/10512/commits/817c07dbd0db762c945895da9de57b92e613d5ec fixes latests issues and integrate it in this PR ?

Rom1-B:
> @Rom1-B Can you check if [817c07d](https://github.com/glpi-project/glpi/commit/817c07dbd0db762c945895da9de57b92e613d5ec) fixes latests issues and integrate it in this PR ?

It looks better
Thanks @cconard96

cedric-anne:
This is still not OK.

When you use any of the search actions (change a criteria, use the pagination, ...), the category filter is not applied anymore.

cconard96:
I thought so so I made my PR a draft. It will need some changes on the JS side to inject the search criteria.

cedric-anne:
@cconard96 Do not hesitate to push your fixes here.

cconard96:
I'm not sure if I can. I've tried for other PRs in the past but I don't have access.

I thought only repo maintainers could push to other people's PRs.

Rom1-B:
It is not yet OK. The criteria are not applied immediately, you have to apply twice.
I must be missing something.

cconard96:
> I thought so so I made my PR a draft. It will need some changes on the JS side to inject the search criteria.

With the category tree filtering results after the regular search is applied, I'm not sure of the correct fix.

cedric-anne:
As seen internally, I merge this to be able to integrate it in 10.0.0-rc. Latest bug have to be fixed in another PR soon.

PR #10164, opened 2021-12-17
fix notice with timezone dropdown

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

This occurs when timezone are available on the system but are not setup in GLPI (in config > general)

### DIFF
--- a/src/Config.php
+++ b/src/Config.php
@@ -1380,7 +1380,7 @@ class Config extends CommonDBTM
                 'timezone',
                 $timezones,
                 [
-                'value'                 => $data["timezone"],
+                'value'                 => $data["timezone"] ?? "",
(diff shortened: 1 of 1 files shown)


cedric-anne (commented):
Maybe this issue is due to the lack of `timezone` in `$default_prefs` in `install/empty_data.php` file. It should be fixed there.
Also, a `$migration->addConfig(['timezone' => null]);` should be added in `install/migrations/update_9.5.x_to_10.0.0/configs.php`.

trasher (approved):

cconard96 (approved):

Issue #10172, opened 2021-12-19
Can not add profile on Knowledge Base article


99net:
### Code of Conduct

- [ ] I agree to follow this project's Code of Conduct

### Is there an existing issue for this?

- [ ] I have searched the existing issues

### Version

10

### Bug description

Can not add profile on Knowledge Base article as a super-admin.

![image](https://user-images.githubusercontent.com/43380069/146675097-926067f3-a7e1-42ff-b2a6-b881e4789b0c.png)

99net:
Thank You.

PR #10173, opened 2021-12-19
Fix KB profile visibility dropdown

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #10172

The removal of the `right` property in #9898 broke the Profile dropdown in the KB visibility tab. I tested the category dropdown from the other commit and it still worked with the `right` property restored. I'm not quite sure why it was removed.

### DIFF
--- a/src/KnowbaseItem.php
+++ b/src/KnowbaseItem.php
@@ -2206,6 +2206,7 @@ class KnowbaseItem extends CommonDBVisible implements ExtraVisibilityCriteria
     protected function getShowVisibilityDropdownParams()
     {
         $params = parent::getShowVisibilityDropdownParams();
+        $params['right'] = ($this->getField('is_faq') ? 'faq' : 'knowbase');
         $params['allusers'] = 1;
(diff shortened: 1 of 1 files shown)


trasher (approved):

src/KnowbaseItem.php
```diff
@@ -2206,6 +2206,7 @@ public function getAnswer()
     protected function getShowVisibilityDropdownParams()
     {
         $params = parent::getShowVisibilityDropdownParams();
+        $params['right'] = ($this->getField('is_faq') ? 'faq' : 'knowbase');
```
Rom1-B, line 2209: `$params['right']` was deleted because it seemed to be ignored if `$params['allusers'] = 1`

orthagh (approved):

Rom1-B (approved):

PR #10197, opened 2021-12-21
Update/inventory support

g-bougard:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |

First this PR add netdiscovery and netinventory tasks hooks, so we can modify glpiinventory plugin to handle this tasks providing required information during glpi-agent contact.
Secondly, a typo was hiding error handling in request addError() API as the HTTP error code wasn't set so we always got a 200 http code, even on error on glpi-agent side.
This revealed the line 118 at src/Inventory/Request.php should be removed to not fail on contact as glpi-agent is providing all the tasks it supports for information.
Finally, the `Method not supported` handling should return 405 as error code and it does now the typo is fixed.

### DIFF
--- a/src/Inventory/Request.php
+++ b/src/Inventory/Request.php
@@ -114,6 +114,12 @@ class Request extends AbstractRequest
             case self::INVENT_TASK:
                 return $this->handleInventoryTask();
             break;
+            case self::NETDISCOVERY_TASK:
+                return $this->handleNetDiscoveryTask();
+            break;
+            case self::NETINV_TASK:
+                return $this->handleNetInventoryTask();
+            break;
             default:
                 $this->addError("Task '$task' is not supported.", 400);
                 return [];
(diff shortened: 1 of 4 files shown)


tests/web/Glpi/Inventory/Request.php
```diff
@@ -60,37 +60,46 @@ public function beforeTestMethod($method)
     */
     private function checkResponse(GuzzleHttp\Psr7\Response $res, $reply)
     {
-        $this->integer($res->getStatusCode())->isIdenticalTo(200);
```
trasher, line 63: Why remove this one? It's correct as far as I can see

src/Inventory/Request.php
```diff
@@ -114,8 +114,13 @@ protected function handleTask($task): array
             case self::INVENT_TASK:
                 return $this->handleInventoryTask();
             break;
+            case self::NETDISCOVERY_TASK:
```
trasher, line 117: I do not agree, if there is no plugin that support those tasks; GLPI will just answer "nothing"

g-bougard: This is not true. handleTask() is just called during agent CONTACT requests to handle tasks hooks. GLPI will answer as usual and the contact answer won't be updated with any handled tasks related information. The inventory hook is actually only used by the database plugin to notify it knows some database credentials.

tests/web/Glpi/Inventory/Request.php
g-bougard: I removed this one as the test should be done by each test as the return code is now different in each cases. The expected value is now tested on lines 83, 102 and 122.

g-bougard:
Will be superseeded by #10207 & #10225

PR #10198, opened 2021-12-22
Replace not supported grep option

trasher:
Fix nightly builds


orthagh (approved):

cconard96 (changes requested):
The class name in `tests/units/Crontask.php` has to be `CronTask` not `Crontask`.
I assume atoum is very picky about test class names matching source class names exactly.
No idea why it didn't break before.

trasher:
see #10257

PR #10214, opened 2021-12-23
Improve plugin support for Kanban

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

Add missing support for plugins to hook into the new Kanban filtering feature.

Two new PHP hooks were added:

- kanban_item_metadata: Used to inject metadata values into the displayed cards that will be used later for filtering
- kanban_filters: Add new filters for different item types

The JS-side event `kanban:filter` was changed to pass the kanban element selector in addition to the filters. This is not strictly needed for core Kanbans since they all use `#kanban` but this isn't a restriction for Kanban's added elsewhere.

### DIFF
--- a/src/CommonITILObject.php
+++ b/src/CommonITILObject.php
@@ -8226,12 +8226,6 @@ abstract class CommonITILObject extends CommonDBTM
             'create_card_limited_columns'    => []
         ];
 
-        $plugin_filters = [];
-        Plugin::doHook(Hooks::KANBAN_FILTERS, [
-            'itemtype' => static::getType(),
-            'filters' => &$plugin_filters
-        ]);
-
         TemplateRenderer::getInstance()->display('components/kanban/kanban.html.twig', [
             'kanban_id'                   => 'kanban',
             'rights'                      => $rights,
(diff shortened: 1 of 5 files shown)


src/CommonITILObject.php
```diff
@@ -8214,54 +8218,60 @@ public static function showKanban($ID)
...
-         'create_card_limited_columns'    => []
+            'create_item'                    => self::canCreate(),
+            'delete_item'                    => self::canDelete(),
+            'create_column'                  => false,
+            'modify_view'                    => true,
+            'order_card'                     => true,
+            'create_card_limited_columns'    => []
         ];
 
+        $plugin_filters = [];
```
orthagh, line 8229: is this part redundant with getKanbanPluginFilters method, no ?

orthagh (approved):

cedric-anne (approved):

trasher (approved):

PR #10220, opened 2021-12-24
language dropdown twig macro

btry:
Add a twig macro to build a languages dropdown. 

Required to convert Formcreator to Twig.

image

| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #number


orthagh:
As discussed together before christmas days, i close this (without merge).
I'll prefer not adding stufff and stuff to macro file especially when glpi doesn't use it itself.
You can use a `do call` + `field()` to do a custom dropdown (we alredy do this for non generic calls).

PR #10250, opened 2022-01-02
Fix bad actor format for helpdesk ticket form

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

When using the helpdesk form to create tickets (simplified interface) and then doing something to trigger a page reload like changing the category, you would get the following error:
`Warning: Invalid argument supplied for foreach() in C:\xampp\htdocs\glpi\src\CommonITILObject.php on line 162`

The `_actors` parameter is a JSON string but is treated like an array.
When using the standard interface, this is an array as expected.

### DIFF
--- a/front/tracking.injector.php
+++ b/front/tracking.injector.php
@@ -56,6 +56,13 @@ if (isset($_POST["_type"]) && ($_POST["_type"] == "Helpdesk")) {
     Html::header(__('Simplified interface'), '', $_SESSION["glpiname"], "helpdesk", "tracking");
 }
 
+if (isset($_POST['_actors']) && is_string($_POST['_actors'])) {
+    try {
+        $_POST['_actors'] = json_decode(stripslashes($_POST['_actors']), true, 512, JSON_THROW_ON_ERROR);
+    } catch (\JsonException $e) {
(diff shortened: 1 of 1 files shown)


front/tracking.injector.php
```diff
@@ -56,6 +56,13 @@
     Html::header(__('Simplified interface'), '', $_SESSION["glpiname"], "helpdesk", "tracking");
 }
 
+if (isset($_POST['_actors']) && is_string($_POST['_actors'])) {
+    try {
+        $_POST['_actors'] = json_decode(stripslashes($_POST['_actors']), true, 512, JSON_THROW_ON_ERROR);
```
orthagh, line 61:
```suggestion
        $_POST['_actors'] = json_decode($_UPOST['_actors'], true, 512, JSON_THROW_ON_ERROR);
```

orthagh (changes requested):

orthagh (approved):

cedric-anne (approved):

PR #10273, opened 2022-01-06
Use configuration fro plugin if any, also display it

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #number

### DIFF
--- a/src/Agent.php
+++ b/src/Agent.php
@@ -264,21 +264,30 @@ class Agent extends CommonDBTM
             echo "<tr class='tab_bg_1'>";
             echo "<td>" . __('Network discovery threads') . "</td>";
             echo "<td>";
+            $general = __('General setup');
+            if (isset($CFG_GLPI['threads_networkdiscovery'])) {
+                $general .= sprintf(' (%1$s)', $CFG_GLPI['threads_networkdiscovery']);
+            }
             Dropdown::showNumber(
                 'threads_networkdiscovery',
                 [
-                'value' => $this->fields['threads_networkdiscovery'],
-                'toadd' => [__('General setup')]
+                    'value' => $this->fields['threads_networkdiscovery'],
+                    'toadd' => [$general],
+                    'min' => 1
                 ]
             );
             echo "</td>";
             echo "<td>" . __('Network discovery timeout') . "</td>";
             echo "<td>";
+            $general = __('General setup');
+            if (isset($CFG_GLPI['timeout_networkdiscovery'])) {
+                $general .= sprintf(' (%1$s)', $CFG_GLPI['timeout_networkdiscovery']);
+            }
             Dropdown::showNumber(
                 'timeout_networkdiscovery',
                 [
-                'value' => $this->fields['timeout_networkdiscovery'],
-                'toadd' => [__('General setup')]
+                    'value' => $this->fields['timeout_networkdiscovery'],
+                    'toadd' => [$general]
                 ]
             );
             echo "</td>";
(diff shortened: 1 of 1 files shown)


orthagh (approved):

src/Agent.php
```diff
@@ -264,21 +264,30 @@ public function showForm($id, array $options = [])
...
+            if (isset($CFG_GLPI['threads_networkdiscovery'])) {
+                $general = sprintf('%1$s (%2$s)', $general, $CFG_GLPI['threads_networkdiscovery']);
+            }
             Dropdown::showNumber(
                 'threads_networkdiscovery',
                 [
-                'value' => $this->fields['threads_networkdiscovery'],
-                'toadd' => [__('General setup')]
+                    'value' => $this->fields['threads_networkdiscovery'],
+                    'toadd' => [$general],
```
g-bougard, line 275: General should be bind to 0.
```suggestion
                    'toadd' => [0 => $general],
```

src/Agent.php
```diff
@@ -264,21 +264,30 @@ public function showForm($id, array $options = [])
...
+            if (isset($CFG_GLPI['timeout_networkdiscovery'])) {
+                $general = sprintf('%1$s (%2$s)', $general, $CFG_GLPI['timeout_networkdiscovery']);
+            }
             Dropdown::showNumber(
                 'timeout_networkdiscovery',
                 [
-                'value' => $this->fields['timeout_networkdiscovery'],
-                'toadd' => [__('General setup')]
+                    'value' => $this->fields['timeout_networkdiscovery'],
+                    'toadd' => [$general]
```
g-bougard, line 290: General should be bind to 0 and min set to 1.
```suggestion
                    'toadd' => [0 => $general],
                    'min' => 1
```

src/Agent.php
```diff
@@ -287,21 +296,30 @@ public function showForm($id, array $options = [])
...
+            if (isset($CFG_GLPI['threads_networkinventory'])) {
+                $general = sprintf('%1$s (%2$s)', $general, $CFG_GLPI['threads_networkinventory']);
+            }
             Dropdown::showNumber(
                 'threads_networkinventory',
                 [
-                'value' => $this->fields['threads_networkinventory'],
-                'toadd' => [__('General setup')]
+                    'value' => $this->fields['threads_networkinventory'],
+                    'toadd' => [$general],
```
g-bougard, line 307: General should be bind to 0.
```suggestion
                    'toadd' => [0 => $general],
```

src/Agent.php
```diff
@@ -287,21 +296,30 @@ public function showForm($id, array $options = [])
...
+            if (isset($CFG_GLPI['timeout_networkinventory'])) {
+                $general = sprintf('%1$s (%2$s)', $general, $CFG_GLPI['timeout_networkinventory']);
+            }
             Dropdown::showNumber(
                 'timeout_networkinventory',
                 [
-                'value' => $this->fields['timeout_networkinventory'],
-                'toadd' => [__('General setup')]
+                    'value' => $this->fields['timeout_networkinventory'],
+                    'toadd' => [-1 => $general]
```
g-bougard, line 322: General should be bind to 0. and min set to 1.
```suggestion
                    'toadd' => [0 => $general],
                    'min' => 1
```

src/Agent.php
```diff
@@ -388,6 +406,21 @@ public function prepareInputs(array $input)
...
+        global $CFG_GLPI;
+
+        if (isset($CFG_GLPI['threads_networkdiscovery']) && !isset($input['threads_networkdiscovery'])) {
+            $input['threads_networkdiscovery'] = 0;
+        }
+        if (isset($CFG_GLPI['threads_networkinventory']) && !isset($input['threads_networkinventory'])) {
+            $input['threads_networkinventory'] = 0;
+        }
+        if (isset($CFG_GLPI['timeout_networkdiscovery']) && !isset($input['timeout_networkdiscovery'])) {
+            $input['timeout_networkdiscovery'] = -1;
```
g-bougard, line 418: Default should be 0 which means use general setup.
```suggestion
            $input['timeout_networkdiscovery'] = 0;
```

src/Agent.php
```diff
@@ -388,6 +406,21 @@ public function prepareInputs(array $input)
...
+        }
+        if (isset($CFG_GLPI['timeout_networkinventory']) && !isset($input['timeout_networkinventory'])) {
+            $input['timeout_networkinventory'] = -1;
```
g-bougard, line 421: Default should be 0 which means use general setup.
```suggestion
            $input['timeout_networkinventory'] = 0;
```

g-bougard (changes requested):
With all the suggestions I propose here, everything seems to work as expected.
In particular, `0` should always be use to represent "General setup" even for timeout.
I can confirm you `0` is not support on agent side and it will always be replaced by a convenient not null default integer.

src/Agent.php
trasher: This is not strictly needed, it will be the default

src/Agent.php
trasher: For timeouts, "0" means disabled, so we have to keep it as is, and therefore keep `-1` for general configuration values

g-bougard (approved):

cedric-anne (approved):

PR #10275, opened 2022-01-06
Do not show entity in header if item doesn't support it

AdrienClairembault:
Avoid this issue where an item doesn't support entities:

image

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |

### DIFF
--- a/templates/components/form/header.html.twig
+++ b/templates/components/form/header.html.twig
@@ -102,7 +102,9 @@
                {% endif %}
             </h3>
 
+            {% set single_actions_ms_auto = true %}
             {% if item.isEntityAssign() and is_multi_entities_mode() and item is not instanceof('Entity') %}
+               {% set single_actions_ms_auto = false %}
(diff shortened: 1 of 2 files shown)


trasher (approved):

orthagh (approved):

cedric-anne (commented):
Looks like the single actions are only displayed if this condition is true.

`{{ include('components/form/single-action.html.twig') }}` should be moved outside this `if` block, I guess.

AdrienClairembault:
> Looks like the single actions are only displayed if this condition is true.
> 
> `{{ include('components/form/single-action.html.twig') }}` should be moved outside this `if` block, I guess.

Done but I also add to make sure the action dropdown get the 'ms-auto' class to be aligned to the right ONLY if the entity details are hidden.

Not sure if there is a better way to check for this.

cedric-anne (approved):

trasher (approved):

PR #10277, opened 2022-01-06
Flatpickr theme support

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

Currently, the default dark flatpickr theme is only used if the current palette is `darker` and doesn't account for any of the other dark themes.

To resolve this, I changed it to include the default dark flatpickr theme all the time (may have a cleaner look than light), and then overrode all colors in a new includes so that they would match the current palette colors.

Samples (Auror, Premium Red, Dark Auror, Midnight):
Selection_034
Selection_035
Selection_036
Selection_037

### DIFF
--- a/css/includes/components/_flatpickr.scss
+++ b/css/includes/components/_flatpickr.scss
@@ -5,6 +5,7 @@
    $picker-text-color: $dark;
    $picker-bg: $card-bg;
 }
+
 $picker-color-accent: rgba($link-color, 0.5) !default;
 $date-hover-bg: rgba($picker-color-accent, 0.4);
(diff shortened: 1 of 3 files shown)


orthagh (approved):

cedric-anne (approved):

orthagh:
I come back to this issue.
I validated without testing live on the previous review.

I noticed on my glpi this monday the new theme, it's globally ok, but i think we could add a regular shadow (see https://getbootstrap.com/docs/5.0/utilities/shadows/) to the dropdown panel

PR #10280, opened 2022-01-06
Add emailField twig macro

AdrienClairembault:
Add support for [\<input type="email"\>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/email).

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |

### DIFF
--- a/templates/components/form/fields_macros.html.twig
+++ b/templates/components/form/fields_macros.html.twig
@@ -91,6 +91,7 @@
              name="{{ name }}" value="{{ value|verbatim_value }}"
              {{ options.readonly ? 'readonly' : '' }}
              {{ options.disabled ? 'disabled' : '' }}
+             {{ options.multiple ? 'multiple' : '' }} {# Only for emailField #}
              {{ options.required ? 'required' : '' }} />
    {% endset %}
 
--- a/templates/components/form/fields_macros.html.twig
+++ b/templates/components/form/fields_macros.html.twig
@@ -345,6 +345,11 @@
    {{ _self.field(name, field, label, options) }}
 {% endmacro %}
 
+{% macro emailField(name, value, label = '', options = {}) %}
+   {% set options = {'type': 'email'}|merge(options) %}
+   {{ _self.textField(name, value, label, options) }}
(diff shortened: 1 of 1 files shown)


orthagh (approved):

trasher (approved):

AdrienClairembault:
Added support for the [multiple attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/email#multiple).

templates/components/form/fields_macros.html.twig
```diff
@@ -345,6 +346,11 @@
    {{ _self.field(name, field, label, options) }}
 {% endmacro %}
 
+{% macro emailField(name, value, label = '', options = {}) %}
+   {% set options = {'type': 'email'}|merge(options) %}
```
trasher, line 350:
```suggestion
   {% set options = {'type': 'email'}|merge(options) %}
```
should not multiple be handled here only aswell?

AdrienClairembault: I don't get what you mean, do you have a code example ?

trasher, line 350: maybe sthing like:
```suggestion
   {% set options = {'type': 'email', 'multiple': options.multiple ?? ''}|merge(options) %}
```

AdrienClairembault: Unless I'm missing something, the other multiple ref can't be removed: it is what actually write the 'multiple' option into the html input properties.

trasher: I really do not know... @orthagh?

orthagh: after checking, the initial proposal is indeed possibly the best

trasher: OK, it's good for me then

cedric-anne (approved):

PR #10293, opened 2022-01-08
Restore actors after incomplete form reload

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

While looking at a different issue, I noticed that in some cases the actors would disappear before I submit the final form.
To recreate:
1. Have category a mandatory field
2. Create ticket with actors and everything required except category
3. Try adding ticket and get error saying category is required and see actors are missing
This doesn't seem to happen when adding other actors or changing fields that cause the form to reload.

### DIFF
--- a/src/Change.php
+++ b/src/Change.php
@@ -1116,7 +1116,7 @@ class Change extends CommonITILObject
              'backoutplancontent'         => '',
              'checklistcontent'           => '',
              'items_id'                   => 0,
-            '_actors'                     => [],
+             '_actors'                     => [],
         ];
     }
 
--- a/src/Change.php
+++ b/src/Change.php
@@ -1077,45 +1077,46 @@ class Change extends CommonITILObject
     {
         $default_use_notif = Entity::getUsedConfig('is_notif_enable_default', $_SESSION['glpiactive_entity'], '', 1);
         return [
-         '_users_id_requester'        => Session::getLoginUserID(),
-         '_users_id_requester_notif'  => [
-            'use_notification'  => $default_use_notif,
-            'alternative_email' => ''
-         ],
-         '_groups_id_requester'       => 0,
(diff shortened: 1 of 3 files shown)


trasher (approved):

orthagh (approved):

cedric-anne (approved):

src/Change.php
```diff
@@ -1077,45 +1077,46 @@ public static function getDefaultValues($entity = 0)
...
+             '_add_validation'            => 0,
+             'users_id_validate'          => [],
+             '_tasktemplates_id'          => [],
+             'controlistcontent'          => '',
+             'impactcontent'              => '',
+             'rolloutplancontent'         => '',
+             'backoutplancontent'         => '',
+             'checklistcontent'           => '',
+             'items_id'                   => 0,
+            '_actors'                     => [],
```
trasher, line 1119:
```suggestion
             '_actors'                     => [],
```

src/Problem.php
```diff
@@ -1721,40 +1721,41 @@ public static function getDefaultValues($entity = 0)
...
+             'content'                    => '',
+             'name'                       => '',
+             'entities_id'                => $_SESSION['glpiactive_entity'],
+             'itilcategories_id'          => 0,
+             'actiontime'                 => 0,
+             '_add_validation'            => 0,
+             'users_id_validate'          => [],
+             '_tasktemplates_id'          => [],
+             'items_id'                   => 0,
+            '_actors'                     => [],
```
trasher, line 1758:
```suggestion
             '_actors'                     => [],
```

trasher (approved):

PR #10319, opened 2022-01-13
timeline_actions hook must load hooks of plugins

btry:
This fix ensure that hook.php from the plugin is loaded.

Fixes a feature of Formcreator.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #number

### DIFF
--- a/src/CommonITILObject.php
+++ b/src/CommonITILObject.php
@@ -6552,21 +6552,12 @@ abstract class CommonITILObject extends CommonDBTM
 
         $legacy_actions = '';
 
-        if (isset($PLUGIN_HOOKS[Hooks::TIMELINE_ACTIONS])) {
-            foreach ($PLUGIN_HOOKS[Hooks::TIMELINE_ACTIONS] as $plugin => $callback) {
-                if (!Plugin::isPluginActive($plugin)) {
-                    continue;
-                }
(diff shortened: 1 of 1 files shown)


cconard96 (approved):
Tested changes with Screenshot plugin. No breaks there.

cedric-anne (approved):

cedric-anne (changes requested):
IMHO, `Plugin::doHook(Hooks::TIMELINE_ACTIONS, [... ]);` should be called once, not for each plugin that register the hook. Inddeed, the `doHook` method itself will loop on each plugin.

orthagh (approved):

src/CommonITILObject.php
```diff
@@ -6552,21 +6552,12 @@ public function getLegacyTimelineActionsHTML(): string
 
         $legacy_actions = '';
 
-        if (isset($PLUGIN_HOOKS[Hooks::TIMELINE_ACTIONS])) {
```
btry, line 6555: if and foreach are already in the doHook method.

cconard96 (approved):

trasher (approved):

PR #10350, opened 2022-01-17
Cleanup inventory temp files

g-bougard:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

Add API to clean up temporary file created during XML conversion.
Use it to cleanup netdiscovery and netinventory tasks handling XML messages.
Use it when import is failing.

### DIFF
--- a/src/Inventory/Inventory.php
+++ b/src/Inventory/Inventory.php
@@ -416,6 +416,21 @@ class Inventory
     }
 
    /**
+    * Clenup temporary file
+    *
+    * @return void
+    */
+    public function cleanTempFile()
+    {
+        $ext = (Request::XML_MODE === $this->inventory_format ? 'xml' : 'json');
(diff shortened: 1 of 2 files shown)


trasher:
> Use it to cleanup netdiscovery and netinventory tasks handling XML messages.

As I said yesterday, those files should never be stored; this is not relevant.

trasher:
Also, this should already been handled in the existing `finally` (see @handleInventoryFile@).

I do not know exactly where the problem is, but current PR sounds wrong

g-bougard:
Actually, these files are stored as the data is transmitted to `setData()`, and the temporary file can only be removed in `handleInventoryFile()`, but in some case the code flow don't reach `handleInventoryFile()`.
If you think the PR is wrong, just close it and I'll open an issue to explain how to reproduce the case. That's not a big deal.

PR #10366, opened 2022-01-19
Suppress update notifications during ticket purge

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #9792

During ticket purge, the cleanup of actor links (like requesters, assigned to, etc) and ticket links both raise update notification events for the ticket. These events were checking for the presence of a `_disablenotif` property to skip sending the update notification, but at least in this case, `_no_notif` is used instead.

### DIFF
--- a/inc/commondbconnexity.class.php
+++ b/inc/commondbconnexity.class.php
@@ -121,7 +121,7 @@ abstract class CommonDBConnexity extends CommonDBTM {
       if ($criteria !== null) {
          $input = [
             '_no_history'     => true,
-            '_no_notif'       => true
+            '_disablenotif'       => true
          ];
 
          $iterator = $DB->request($criteria);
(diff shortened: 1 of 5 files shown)


trasher:
Should not we use `_disablenotif`only? Tho `no_notif` parameter is currently not used at all as far as I can see.

cedric-anne (commented):
On master branch, the `_no_notif` input is set only once, and is never used. All other occurences are commented. I suggest to replace usage of `_no_notif` by `_disablenotif`.

cedric-anne (approved):

orthagh (approved):

trasher (approved):

PR #10373, opened 2022-01-19
fix glpi_close_all_dialogs js function

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

Don't know why, but modal.hide() (or dispose()) doesn't anymore.

### DIFF
--- a/js/glpi_dialog.js
+++ b/js/glpi_dialog.js
@@ -306,7 +306,7 @@ var glpi_confirm = function({
  * Remove from dom all opened glpi dialog
  */
 var glpi_close_all_dialogs = function() {
-   $('.modal.show, .modal-backdrop.show').remove();
+   $('.modal.show').modal('hide').remove();
 };
 
 var toast_id = 0;
--- a/js/glpi_dialog.js
+++ b/js/glpi_dialog.js
@@ -303,13 +303,10 @@ var glpi_confirm = function({
 
 
 /**
- * Close and remove from dom all opened glpi dialog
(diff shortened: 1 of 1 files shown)


js/glpi_dialog.js
```diff
@@ -303,13 +303,10 @@ var glpi_confirm = function({
...
 /**
- * Close and remove from dom all opened glpi dialog
+ * Remove from dom all opened glpi dialog
  */
 var glpi_close_all_dialogs = function() {
-   document.querySelectorAll('.modal').forEach(function(modalElem) {
-      const myModal = new bootstrap.Modal(modalElem);
-      myModal.hide();
-   });
+   $('.modal.show, .modal-backdrop.show').remove();
```
cconard96, line 309:
```suggestion
   $('.modal).modal('hide');
```
This seems to work as an alternative. It isn't destructive so modals that are built once can still be re-shown and it will still trigger the `hide.bs.modal` event.

orthagh: indeed, i'll check again #10286 because i originaly found issue with modal before this pr

orthagh: strange, i cannot trigger the original issue

cedric-anne (approved):

cconard96 (approved):

trasher (approved):

Issue #10299, opened 2022-01-10
[Beta 10] Ticket status cannot be changed directly from NEW to Closed or Solved 


husisusi:
### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

GLPI 10 Beta

### Bug description

Ticket status cannot be changed directly from NEW to Closed or Solved.
No error, the status is reset to New after clicking on Save.

cconard96:
It seems intentional. There is a specific check in the code for updating that ignores the status change when going from New to anything else when the Ticket, Change, or Problem are not assigned. This code has existed for at least 9 years, so it doesn't seem to be a change with GLPI 10.

husisusi:
Hum! but in GLPI version 9.5.6 i can do it.
![grafik](https://user-images.githubusercontent.com/8606735/148853448-b3468a01-f979-4704-98f3-552e0b19c62c.png)

orthagh:
Please check the lifecycle tab of your current profile on both instances.
Maybe the 10 instance forbid this change:

![image](https://user-images.githubusercontent.com/418844/149530922-de745815-f446-43d3-837f-65e414d99c40.png)

orthagh:
On my current instance, i have no problem to close a new ticket

cconard96:
> Indeed, the status update doesn't work with this state, i'll check. @cconard96 do you mind point the historic code you found ? We may drop it, but i want to check some history

https://github.com/glpi-project/glpi/blob/79f9719d7995cc6bd38661cbc03fa0601824b236/src/CommonITILObject.php#L1532-L1553

husisusi:
Thank you @cconard96.
Yes that's it. It works after I changed line 1551 to (only to test):
`             $this->fields['status'] = $this->fields['status']; `

cconard96:
Fixed in #10386

PR #10386, opened 2022-01-20
Fix Ticket status change

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes (change in behavior)
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #10299

Before GLPI 10 it seemed like you could change a ticket's status from New directly to Solved or Closed without having anyone assigned to it. The behavior that blocked it was not called at least for tickets, but it is now.

I don't understand the purpose of this forced behavior and don't think it is wanted for Tickets, Changes, or Problems.
If it is wanted, I think it would be best to have this be a configurable behavior somewhere else.

### DIFF
--- a/src/CommonITILObject.php
+++ b/src/CommonITILObject.php
@@ -1531,27 +1531,6 @@ abstract class CommonITILObject extends CommonDBTM
         }
 
         if (isset($this->input["status"])) {
-            if (
-                ($this->input["status"] != self::WAITING)
-                && ($this->countSuppliers(CommonITILActor::ASSIGN) == 0)
-                && ($this->countUsers(CommonITILActor::ASSIGN) == 0)
-                && ($this->countGroups(CommonITILActor::ASSIGN) == 0)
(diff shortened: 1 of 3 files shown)


orthagh:
Did you checked also when status is Processing (assigned) on ticket (but maybe also on other ITIL items) and after removing actors, you return to Incoming status ?

cconard96:
> Did you checked also when status is Processing (assigned) on ticket (but maybe also on other ITIL items) and after removing actors, you return to Incoming status ?

Covered by https://github.com/glpi-project/glpi/blob/86487b3012de4533bd56e20e7324d4a4f4a61060/tests/functionnal/Ticket.php#L1643?

orthagh (approved):

trasher (approved):

cedric-anne (approved):

Issue #9287, opened 2021-07-15
[9.5.6] Active Directory location sync creates locations in root entity instead of user's entity


CireSacud26:
Hi GLPI Team !

I'm facing what it seems to be au bug :

**Bug Description**

In AD User object, physicaldeliveryofficename attribute is set to give a location of the user like [Site x] > [Floor y] > [Room z], according of the existing locations set in a given entity in GLPI. But user sync creates a duplicate location in root entity and don't take the existing one in the user's entity.

**To reproduce**
It would be hard work to explain all steps to reproduce this issue... I'll try with this :
1- In _Home >Setup > Dropdowns > Locations_, a location is set as **[Site x] > [Floor y] > [Room z]** in **[root Entity] / [Entity1]**
2- Considering a fully functionnal AD or LDAP user sync to GLPI. Users are successfully imported in **[root Entity] / [Entity1]**
3- In _Home >Setup > Authentication > LDAP directories > Users_, **physicaldeliveryofficename** attribute is binded with "**location**" field
4- In AD user's object, the **physicaldeliveryofficename** attribute is set as **[Site x] > [Floor y] > [Room z]**
5- Forcing user sync from AD in GLPI : A new location **[Site x] > [Floor y] > [Room z]** is created in **[root entity]** instead of using the existing one **[root Entity] / [Entity1] > [Site x] > [Floor y] > [Room z]**

**Expected behavior**

The AD attribute should link the existing location of the user's entity.

Regards,
Eric

CireSacud26:
Hi !

I realized just now I forgot to post my system info. Sorry... Here is it : 
[code]   GLPI 9.5.5 ( => G:) Installation mode: TARBALL
--

[...]

PR #10403, opened 2022-01-22
Preserve location during user AD sync

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #9287

### DIFF
--- a/inc/user.class.php
+++ b/inc/user.class.php
@@ -1650,7 +1650,7 @@ class User extends CommonDBTM {
                      // use import to build the location tree
                      $this->fields[$k] = Dropdown::import('Location',
                                                           ['completename' => $val,
-                                                           'entities_id'  => 0,
+                                                           'entities_id'  => $import ? 0 : $this->fields['entities_id'],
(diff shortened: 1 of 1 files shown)


trasher (approved):

orthagh (approved):

cconard96:
Not sure how this would work without extensive changes.

By the time `User::getFromLDAP` is called, the user may or may not represent a User stored in the DB. Both cases are seen even when $import is false so the User should already be imported into the DB. There isn't even an `id` field sometimes.

If you try using `User::getFromDBByDn` to try and load the fields, this may or may not work because in some cases the `users_dn` field is missing or null(?).

Even so, authorization assignment rules are run after the user's data like the location and category are imported so they should be transferred but I don't see anywhere where the rules cause a transfer. They only seem to directly set the `entities_id` property.

This issue seems go beyond the initial scope of this PR so I'm closing it.

PR #10406, opened 2022-01-22
Fix infocom inheritance for assets from template

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #9666

If you have the option "Enable the financial and administrative information by default" enabled, the infocom data from asset templates is not cloned to the created assets because a blank infocom item is created first.

I added a check to skip the automatic creation if it is during a clone operation (creating asset from template).


trasher:
Maybe this would need a unit test

orthagh (approved):

trasher:
I'm not sure the result is correct. If you activate the option, and clone anything, you'll end with no infocom created, right? (juste like the option ha not been enabled)

cconard96:
> I'm not sure the result is correct. If you activate the option, and clone anything, you'll end with no infocom created, right? (juste like the option ha not been enabled)

No. That is handled somewhere else. Here, it just creates a a blank Infocom (enable but don't fill).
I tested to be sure. We don't want anything to be made here ever if we are going to get a clone from a template or a clone source item because the cloning process tries creating a new Infocom with the cloned data.

cedric-anne (approved):

cedric-anne (changes requested):
You have a test failure:
```
=> tests\units\Computer::testCloneWithAutoCreateInfocom():
In file /var/glpi/tests/functionnal/Computer.php on line 544, array() failed: array(3) is not identical to array(3)
-Expected
+Actual
@@ -7 +7 @@
-  string(6) "800.00"
+  string(8) "800.0000"
```

trasher:
could you please rebase?

cedric-anne (approved):

trasher (approved):

orthagh (approved):

Issue #10041, opened 2021-12-08
[GLPI10] Simplified interface - Icon savesearches


tsmr:
### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

Master

### Bug description

On simplified interface - the Icon for load savesearches doesn't exist like 9.5.X

![image](https://user-images.githubusercontent.com/13178158/145201080-d42bb34b-b606-40fe-8c70-e6baed29eb2a.png)

### Page URL

http://localhost/glpi/front/helpdesk.public.php

PR #10409, opened 2022-01-23
Add context links for Tickets in simplified interface

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #10041

Previously `Html::helpHeader` didn't support any context links (buttons link add, search, lists, etc).
I added some basic support for them in the context of Tickets.

As a side-effect, the menu template mistakenly thought that the Tickets menu item had children when it only had those context links. I added an extra check in the template to see if there were any real sub-items.

### DIFF
--- a/front/helpdesk.faq.php
+++ b/front/helpdesk.faq.php
@@ -41,7 +41,7 @@ if (isset($_GET["redirect"])) {
 Session::checkFaqAccess();
 
 if (Session::getLoginUserID()) {
-    Html::helpHeader(__('FAQ'), $_SERVER['PHP_SELF'], $_SESSION["glpiname"]);
+    Html::helpHeader(__('FAQ'));
 } else {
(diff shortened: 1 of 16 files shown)


cconard96:
Last commit fixes invalid usage of parameters for helpHeader function. They were invalid before with the $url parameter being unused for over 11 years (didn't look back further than that) and some having a third parameter of the current user's name which hasn't existed at all at least in the last 11 years.

Leaving these invalid parameters could lead to undesired behavior and confusion if code gets copied to new files or referenced.

trasher (approved):

orthagh (approved):

cedric-anne (approved):

Issue #10184, opened 2021-12-20
[GLPI 10] No icon for plugins on simplified interface


tsmr:
### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

Master

### Bug description

There is no icon hook for plugins on simplified interface

![image](https://user-images.githubusercontent.com/13178158/146787382-3c7f9674-8551-4a64-a1fe-c013ef28a08e.png)

I can do a PR with a new hook "helpdesk_menu_entry_icon" ? 

```
diff --git a/src/Html.php b/src/Html.php
index 96c7c346c9..f0882d4a68 100644
--- a/src/Html.php
+++ b/src/Html.php
@@ -1885,6 +1886,9 @@ HTML;
                     if (is_string($PLUGIN_HOOKS["helpdesk_menu_entry"][$plugin])) {
                         $link = $PLUGIN_HOOKS["helpdesk_menu_entry"][$plugin];
                     }
+                    if (isset($PLUGIN_HOOKS["helpdesk_menu_entry_icon"][$plugin])) {
+                      $infos['icon'] = $PLUGIN_HOOKS["helpdesk_menu_entry_icon"][$plugin];
+                    }
                     $infos['page'] = $link;
                     $infos['title'] = $infos['name'];
                     $menu['plugins']['content'][$plugin] = $infos;

```

### Page URL

http://localhost/glpi/front/helpdesk.public.php

orthagh:
Yes you can submit the pr.
Don't forget to do also a pr for the empty plugin and in docdev

tsmr:
Thanks for your job curtis

PR #10412, opened 2022-01-23
Add helpdesk_menu_entry_icon hook

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #10184

### DIFF
--- a/src/Plugin/Hooks.php
+++ b/src/Plugin/Hooks.php
@@ -209,12 +209,4 @@ class Hooks
             self::TIMELINE_ACTIONS,
         ];
     }
-
-    public static function getHelpdeskMenuHooks(): array
-    {
-        return [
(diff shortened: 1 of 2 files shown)


orthagh (approved):

src/Plugin/Hooks.php
```diff
@@ -205,4 +209,12 @@ public static function getItemHooks(): array
             self::TIMELINE_ACTIONS,
         ];
     }
+
+    public static function getHelpdeskMenuHooks(): array
```
orthagh, line 213: I'm not sur this method is required as you don't use it

orthagh (approved):

Issue #10043, opened 2021-12-08
[GLPI10] Simplified interface - Can see childs entities


tsmr:
### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

Master

### Bug description

In simplified interface, even with an no recursive profile, you'll see childs entities

![image](https://user-images.githubusercontent.com/13178158/145201376-bde97fad-5071-4997-99e3-021ca3fec897.png)

### Page URL

http://localhost/glpi/front/helpdesk.public.php

cconard96:
I cannot recreate the issue.
I removed the recursive profile from the default post-only user and added the "Self Service" profile without recursion. User could only see the root entity and none of the children.
I tried adding a second profile (Observer) that was assigned recursively and I could see children entities only when the Observer profile was active.

tsmr:
Hello Curtis

I checked on my side. In fact it is when you change profile that this happens.

You connect in self-service: Display OK
You change your profile to super-admin: OK
You change your profile again to Self-Service: KO: The list of entities displays all the entities.

I'm thinking a cache problem, right?

tsmr:
Hello. Sorry but your patch doesn't resolve the issue

tsmr:
Curtis in fact in ajax/entitytreesons.php if i comment : 

```
if (array_key_exists($subckey, $all_entitiestree)) {
    echo json_encode($all_entitiestree[$subckey]);
    exit;
}
```

It's good

tsmr:
So it was a cache problem too i think

PR #10418, opened 2022-01-23
Fix entity switcher when using a non-recursive profile

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #10043 

Entity switcher was showing child entities even when using a profile assigned non-recursively because no checks for `is_recursive` were done.

### DIFF
--- a/ajax/entitytreesons.php
+++ b/ajax/entitytreesons.php
@@ -59,7 +59,7 @@ $entitiestree = [];
 foreach ($_SESSION['glpiactiveprofile']['entities'] as $default_entity) {
     $default_entity_id = $default_entity['id'];
 
-    $entitytree  = getTreeForItem('glpi_entities', $default_entity_id);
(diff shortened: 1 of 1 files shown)


orthagh (approved):

trasher (approved):

cedric-anne (approved):

cedric-anne (changes requested):
See comment in #10043. Maybe cache key must be adapted.

trasher:
Cache issue is not related to the initial problem this PR tries to address, am I wrong?

cconard96:
> Cache issue is not related to the initial problem this PR tries to address, am I wrong?

I don't understand what the issue with the cache would be and I cannot seem to find any further issues on my end.

cedric-anne (approved):

PR #10424, opened 2022-01-23
Fix undefined cache_key error on printer page counters

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

### DIFF
--- a/src/Dashboard/Widget.php
+++ b/src/Dashboard/Widget.php
@@ -522,6 +522,7 @@ HTML;
             'rand'         => mt_rand(),
         ];
         $p = array_merge($default, $params);
+        $p['cache_key'] = $p['cache_key'] ?? $p['rand'];
         $default_entry = [
             'url'    => '',
             'icon'   => '',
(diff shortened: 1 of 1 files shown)


trasher (approved):

orthagh (commented):
I'm not sure about this one. 
I need to check in the morning, but we shouldn't miss this entry here.
We can indeed avoid warning but this params should have been passed by  the parent card

orthagh:
After a quick investigation, cache_key doesn't exist a card creation.
I checked the code of Widget, we may fix this by adding this key  to default options and remove it from supercharging methods.

Like, adding the default option to `pie`, `getBarsGraph`, `getLinesGraph` and removing the existing options from `multipleLines` and `multipleAreas`

cedric-anne (approved):

trasher (approved):

PR #10446, opened 2022-01-26
fix button alignement in legacy timeline actions

btry:
Seen with Formcreator, which adds a timeline action

### before 
image

### intermediate (without change in Formcreator)

image

### after (with change in Formcreator)

To have a consistent height of the button I replaced the HTML I tag with a BUTTON tag. I think this is acceptable to switch to GLPI 10,

image

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #number

### DIFF
--- a/css/includes/components/itilobject/_timeline.scss
+++ b/css/includes/components/itilobject/_timeline.scss
@@ -531,11 +531,12 @@ ul.legacy-timeline-actions {
     list-style: none;
     display: inline-flex;
     margin-bottom: 0;
+    position: relative;
+    top: -3px;
 
     li {
         display: inline;
         margin-left: 5px;
-        border: $input-border-width solid $input-border-color;
         border-radius: 5px;
         padding: 5px;
 


css/includes/components/itilobject/_timeline.scss
```diff
@@ -531,11 +531,12 @@ ul.legacy-timeline-actions {
     list-style: none;
     display: inline-flex;
     margin-bottom: 0;
+    position: relative;
+    top: -3px;
 
     li {
         display: inline;
         margin-left: 5px;
-        border: $input-border-width solid $input-border-color;
```
orthagh, line 538: can you explain this part ?

btry: If I keep the current CSS on GLPI master and use a button in Formcreator (see the linked PR), I get this :

image

I think the border should be on the button itself, instead of the parent LI tag.

orthagh: ok

AdrienClairembault:
As a side note, did you check what the "Cancel my Ticket" button was supposed to do ?
To make sure it isn't doing the same as the new "thrashcan" button (and avoid having two buttons for the same action).

cconard96 (approved):

trasher (approved):

btry:
@AdrienClairembault 

 As a side note, did you check what the "Cancel my Ticket" button was supposed to do ?

The button "cancel my ticket" belongs to Formcreator. It allows the requester to close his own ticket when the ticket is still new and does not has any followup. Useful when the requester solved his problem himself.

cedric-anne (approved):

orthagh (approved):

PR #10451, opened 2022-01-27
make title sticky in itilobject

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

Follow #10114 

image

image

Last concern, the title, when scrollin is on the top, is repeated twice but i don't have any idea atm

I tested with reverse order for timeline, thinking it will be strange, but in fact, i think it display ok.

image

### DIFF
--- a/css/includes/components/itilobject/_timeline.scss
+++ b/css/includes/components/itilobject/_timeline.scss
@@ -29,6 +29,36 @@
  * ---------------------------------------------------------------------
  */
 
+.itil-title {
+    position: sticky;
+    top: -0.5rem;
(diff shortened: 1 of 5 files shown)


trasher (approved):

AdrienClairembault (approved):

cconard96 (approved):

orthagh:
I pushed a simpler solution (similar to what we had previously)

image

trasher (approved):

cedric-anne (approved):

PR #10452, opened 2022-01-27
Replace user-facing master/slave DB usage with main/replica

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

Using master/slave wording has bad connotations in English. I propose replacing the wording with main/replica.

Also, there was already multiple places where main/replica was used in the UI in addition to the other wording so this will also unify it more.

I left the config file name as "config_db_slave" for now.
For the status checker, I left the old properties in place to avoid causing issues with monitoring solutions and added the new properties (Not sure how bad it would be to remove the old ones without deprecation). The plain-text output was left as-is since it is already deprecated.

### DIFF
--- a/src/System/Status/StatusChecker.php
+++ b/src/System/Status/StatusChecker.php
@@ -173,7 +173,7 @@ final class StatusChecker
                     'servers' => []
                 ]
             ];
-           // Check slave server connection
+           // Check replica SQL server connection
             if (DBConnection::isDBSlaveActive()) {
                 $DBslave = DBConnection::getDBSlaveConf();
                 if (is_array($DBslave->dbhost)) {
@@ -220,6 +220,10 @@ final class StatusChecker
             }
         }
 
+        // Set new properties. Master and slave are deprecated given their implications in English.
+        $status['main'] = $status['master'];
+        $status['replicas'] = $status['slaves'];
+
         return $status;
     }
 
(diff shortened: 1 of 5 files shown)


src/System/Status/StatusChecker.php
```diff
@@ -560,7 +564,7 @@ private static function getPlaintextOutput(array $status): string
                 $output .= "GLPI_DBSLAVE_{$num}_{$slave_info['status']}\n";
             }
         } else {
-            $output .= "No slave DB\n";
+            $output .= "No slave DB\n"; // Leave as "slave" since plain text is already deprecated
```
trasher, line 567: Then, maybe just:
```suggestion
            $output .= "No DB replica\n";
```

src/System/Status/StatusChecker.php
```diff
@@ -220,6 +220,10 @@ public static function getDBStatus($public_only = true): array
             }
         }
 
+        // Set new properties. Master and slave are deprecated given their implications in English.
+        $status['main'] = $status['master'];
```
trasher, line 224: I wonder if this would not be preferable to really rename all occurrences, and potentially keep a master/slave here for backward compatibility.

src/System/Status/StatusChecker.php
cconard96: Changing the text would be a BC break if people are still monitoring using the plain-text output.
Not sure how much people actually use it currently and would be affected.

orthagh (approved):

src/System/Status/StatusChecker.php
trasher: ok, yes

trasher (approved):

Issue #10203, opened 2021-12-22
[GPLPI 10.0]+[AgentGLPI] : Uncaught Exception TypeError: Return value of Agent::requestAgent() must be an instance of GuzzleHttp\Psr7\Response


steph461:
### Code of Conduct

- [ ] I agree to follow this project's Code of Conduct

### Is there an existing issue for this?

- [ ] I have searched the existing issues

### Version

10.0

### Bug description

I go back the first time the computer with the agent glpi see version on the photo.

But if I want to force an inventory or see the status of the agent from GLPI 10.0. The arrows turn loop.

The error is generated in the log

![glpi10+agentglpi](https://user-images.githubusercontent.com/68054991/147110219-2d539dce-3d70-4c28-a12a-ba9b2f4dfba4.png)
p

### Relevant log output

```shell
[2021-12-22 15:35:39] glpiphplog.CRITICAL:   *** Uncaught Exception TypeError: Return value of Agent::requestAgent() must be an instance of GuzzleHttp\Psr7\Response, null returned in /var/www/html/glpi10beta/src/Agent.php at line 584
  Backtrace :
  src/Agent.php:608                                  Agent->requestAgent()
  ajax/agent.php:54                                  Agent->requestInventory()
  
[2021-12-22 15:35:42] glpiphplog.CRITICAL:   *** Uncaught Exception TypeError: Return value of Agent::requestAgent() must be an instance of GuzzleHttp\Psr7\Response, null returned in /var/www/html/glpi10beta/src/Agent.php at line 584
  Backtrace :
  src/Agent.php:595                                  Agent->requestAgent()
  ajax/agent.php:58                                  Agent->requestStatus()
```

### Steps To reproduce

1. Go to Computer
2. Go click inventory

steph461:
If i go http://localhost:62354/ It is OK
Contact Computer ==> GLPI ok
Contact GLPI ===> Computer Ko

Contact Computer ==> GLPI ok Look in image
![image](https://user-images.githubusercontent.com/68054991/147113668-196baa4e-9b7a-426f-b818-b07ef05f0557.png)

steph461:
<details><summary>Informations sur le système, l'installation et la configuration</summary><pre>GLPI 10.0.0-dev (/glpi10beta => /var/www/html/glpi10beta)
Installation mode: TARBALL
Current language:fr_FR

</details><details><summary>Server</summary><pre> 
[...]

PR #10454, opened 2022-01-27
Update GuzzleHttp client exception handling when requesting agents

g-bougard:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #10203

Avoid an exception when a request to an agent fails and return "Unknown" in that case.
```
[2022-01-27 11:36:59] glpiphplog.CRITICAL:   *** Uncaught Exception TypeError: Return value of Agent::requestAgent() must be an instance of GuzzleHttp\Psr7\Response, null returned in /srv/glpi/src/Agent.php at line 619
  Backtrace :
  src/Agent.php:630                                  Agent->requestAgent()
  ajax/agent.php:58                                  Agent->requestStatus()
```

### DIFF
--- a/src/Agent.php
+++ b/src/Agent.php
@@ -611,11 +611,17 @@ class Agent extends CommonDBTM
                 $response = $httpClient->request('GET', $endpoint, []);
                 self::$found_adress = $adress;
                 break;
-            } catch (\GuzzleHttp\Exception\ConnectException $e) {
+            } catch (Exception $e) {
                 //many adresses will be incorrect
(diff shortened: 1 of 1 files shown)


orthagh (approved):

cedric-anne (approved):

g-bougard:
I just realized it should fix #10203

PR #10460, opened 2022-01-28
Fix decimal dropdown

AdrienClairembault:
Dropdowns with a decimal step are not displayed correctly since b7457cbbba5d063a828502c099c2cff66068bc15.

image

After fix:

image

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !23325

### DIFF
--- a/inc/dropdown.class.php
+++ b/inc/dropdown.class.php
@@ -1512,15 +1512,13 @@ class Dropdown {
     *
     * @since 0.84
     *
-    * @param integer  $value    numeric value
-    * @param string   $unit     unit (maybe year, month, day, hour, % for standard management)
-    * @param int|null $decimals number of decimal, null to not handle decimals at all
-    *
-    * @return string
-    */
-   static function getValueWithUnit($value, $unit, ?int $decimals = 0): string {
+    * @param integer $value numeric value
+    * @param string $unit unit (maybe year, month, day, hour, % for standard management)
+    * @param integer $decimals number of decimal
+    **/
+   static function getValueWithUnit($value, $unit, $decimals = 0) {
 
-      $formatted_number = !is_null($decimals) && is_numeric($value)
+      $formatted_number = is_numeric($value)
          ? Html::formatNumber($value, false, $decimals)
          : $value;
 
--- a/inc/dropdown.class.php
+++ b/inc/dropdown.class.php
@@ -1512,10 +1512,10 @@ class Dropdown {
     *
     * @since 0.84
     *
-    * @param integer $value numeric value
-    * @param string $unit unit (maybe year, month, day, hour, % for standard management)
+    * @param integer $value    numeric value
+    * @param string  $unit     unit (maybe year, month, day, hour, % for standard management)
     * @param integer $decimals number of decimal
-    **/
+   **/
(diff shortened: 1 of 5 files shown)


trasher (approved):

orthagh (approved):

inc/dropdown.class.php
```diff
@@ -1478,7 +1478,8 @@ static function showNumber($myname, $options = []) {
 
       $field_id = Html::cleanId("dropdown_".$myname.$p['rand']);
       if (!isset($p['toadd'][$p['value']])) {
-         $valuename = self::getValueWithUnit($p['value'], $p['unit']);
+         $decimals = Toolbox::isFloat($p['step']) ? null : 0;
```
cedric-anne, line 1481: I do not agree with this. The number should be formatted even if it is a float. Indeed, the decimal separator is not always a `.`.

Another solution could be to extract the number of decimals from the step value, like this:
```suggestion
         $decimals = strlen(preg_replace('/\d*\./', '', floatval($p['step'])));
```

AdrienClairembault: Done, but I've keep "0" if the number is not a float, to avoid adding unnecessary `,0000` parts to integers values.

AdrienClairembault: image

inc/toolbox.class.php
```diff
@@ -3599,4 +3599,50 @@ public static function getAvailablesTabs(string $itemtype, $id = null): array {
...
+
+      return (floatval($value) - intval($value)) > 0;
+   }
+
+   /**
+    * Get the number of decimals for a given value
+    *
+    * @param mixed $value A possible float
+    *
+    * @return bool
```
cedric-anne, line 3629:
```suggestion
    * @return int
```

inc/dropdown.class.php
```diff
@@ -1511,13 +1512,15 @@ static function showNumber($myname, $options = []) {
...
-    * @param integer $decimals number of decimal
-   **/
-   static function getValueWithUnit($value, $unit, $decimals = 0) {
+    * @param integer  $value    numeric value
+    * @param string   $unit     unit (maybe year, month, day, hour, % for standard management)
+    * @param int|null $decimals number of decimal, null to not handle decimals at all
+    *
+    * @return string
+    */
+   static function getValueWithUnit($value, $unit, ?int $decimals = 0): string {
```
cedric-anne, line 1521:
```suggestion
    * @param integer $value numeric value
    * @param string $unit unit (maybe year, month, day, hour, % for standard management)
    * @param integer $decimals number of decimal
    **/
   static function getValueWithUnit($value, $unit, $decimals = 0) {
```

inc/dropdown.class.php
```diff
@@ -1511,13 +1512,15 @@ static function showNumber($myname, $options = []) {
...
 
-      $formatted_number = is_numeric($value)
+      $formatted_number = !is_null($decimals) && is_numeric($value)
```
cedric-anne, line 1523:
```suggestion
      $formatted_number = is_numeric($value)
```

PR #10465, opened 2022-01-28
permits to disable display of initials in timeline

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

Again, following a meeting about glpi 10 with Nicolas, he says me one of its colleagues initials is a bad idea to display (in France context)  in the timeline.

So here is a config to disable display of initials in timeline.
When set to no, user avatar will display a generic picture (the same used for anonymous heldesk):

image

### DIFF
--- a/install/empty_data.php
+++ b/install/empty_data.php
@@ -2178,6 +2178,7 @@ $tables['glpi_entities'] = [
         'suppliers_as_private'                 => 0,
         'enable_custom_css'                    => 0,
         'anonymize_support_agents'             => 0,
+        'display_users_initials'               => 1,
         'contracts_strategy_default'           => 0,
         'transfers_strategy'                   => 0,
     ],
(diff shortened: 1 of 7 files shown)


cconard96:
Does the Kanban need changed too? For users the default badge shows initials.

orthagh:
same indeed.
Could you create a pr (starting from this branch) and do the same in kanban

cconard96 (approved):

cconard96:
Would it only need done for ITILObjects or global?

I'll start it tonight after my other job.

orthagh:
Globally. The issue is some initials can be bad in some context, like SS for example, but in France PD is also bad (it means faggot).
So the initial feature should be deactivatable everywhere.

cedric-anne (approved):

PR #10486, opened 2022-02-01
Allow custom Kanban URLs

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

Tasklists plugin has custom URLs for the different contexts it uses in its Kanban. Currently, the URLs in the switcher are hard coded to the item's form URL with the Kanban tab. Same with the global Kanban for an itemtype.
This adds methods to the Kanban trait which can be overridden by different itemtypes to specify different URLs.


cedric-anne (approved):

trasher (approved):
Please fix cs ;)

PR #10599, opened 2022-02-10
Fix OLA progress bar

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #8179 (maybe)

### DIFF
--- a/src/Search.php
+++ b/src/Search.php
@@ -6559,15 +6559,19 @@ JAVASCRIPT;
                         switch ($table . '.' . $field) {
                             case "glpi_tickets.time_to_resolve":
                                 $slaField = 'slas_id_ttr';
+                                $sla_class = 'SLA';
                                 break;
                             case "glpi_tickets.time_to_own":
                                 $slaField = 'slas_id_tto';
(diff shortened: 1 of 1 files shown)


trasher (approved):

carlosbayerl:
Could you tell me which file and in which folder it is recorded? So that I can fix it.

cconard96:
> Could you tell me which file and in which folder it is recorded? So that I can fix it.

The files changed tab shows each file that and changes (only src/Search.php here), the and the lines that were changed. For 9.5.X, the file would be inc/Search.class.php but the lines would be different depending on the version.

carlosbayerl:
@cconard96 Hello.
The version I use is 9.5.6. and the toto file is different. I replaced the entire file, it still didn't work.  Can you help me?

carlosbayerl:
I inserted the new lines in the file. And it still didn't work. Then I copied the code presented on the forum and it didn't work either.

cedric-anne (approved):

Issue #10594, opened 2022-02-09
TTR calculation for SLAs defined in days and Default Calendar


VladoTTX:
### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

9.5.6

### Bug description

There is bug in accounting of Pending time when this is met:
1. TTR is defined in days
2. TTR is using calendar without working days for some weekdays and/or holidays (ex. Default calendar)

Function getActiveTimeBetween expects calendar to be 24/7 without holidays see:
https://github.com/glpi-project/glpi/blob/c2a3f2c05e7eee62739dd5fd9e00dca648ef4731/inc/calendar.class.php#L320

But SLA end date calculation computeEndDate counts with holidays (and working hours):
https://github.com/glpi-project/glpi/blob/c2a3f2c05e7eee62739dd5fd9e00dca648ef4731/inc/calendar.class.php#L507

TTR calculation with waiting works like this, first getActiveTimeBetween is called to compute sla_waiting_duration and this is passed to computeEndDate afterwards.
This will lead to issues when ticket is pending over the weekend for example and it will resume after. Too much delay will be accounted.

### Steps To reproduce

1. Create SLA to have TTR 1 day and attach it to default calendar
2. Create ticket attach the created SLA
3. Change status to Pending status
4. Leave it pending over the weekend
5. Resume ticket to working status Processing (Assigned)

Also weekend will be accounted as active time so the new TTR time will be +3 days not just one.

### Your GLPI setup information

Default installation of 9.5.6 without plugins

### Anything else?

I tried to think about fix but it does not seem to be so straightforward, if we would like to account correctly for reminders of working time (coming from working to pending status multiple times).

cconard96:
I closed the PR for my fix because I found it wasn't correct. It wasn't easy to determine the intention of the legacy code and was very complex.

PR #10600, opened 2022-02-10
Fix calendar active time when working in days

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | yes (time calculation change)
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #10594

I'm not quite sure this is the correct fix yet. The original logic had been unchanged for years.

To me it looks like it should return a timestamp in days (n * DAY_TIMESTAMP) for working time, but it doesn't. The behavior currently just ignores all non-working time and returns ($end - $start).

In the context of an SLA (with non 24/7 calendar):
Ticket created 10 minutes before end of day with TTR of 1 day (having Days in SLA sets work_in_days to true).

The current behavior gives technicians until 10 minutes before closing the next day regardless of working hours. If the next day is a holiday, it looks like it would give technicians 10 minutes to meet the SLA.

### DIFF
--- a/src/Calendar.php
+++ b/src/Calendar.php
@@ -285,50 +285,45 @@ class Calendar extends CommonDropdown
 
         $activetime = 0;
 
-        $cache_duration = $this->getDurationsCache();
-
-        for ($actualtime = $timestart; $actualtime <= $timerealend; $actualtime += DAY_TIMESTAMP) {
-            $actualdate = date('Y-m-d', $actualtime);
-
-            if (!$this->isHoliday($actualdate)) {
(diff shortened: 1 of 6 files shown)


cconard96:
Commit c4e21340a14ffe3f124d665a719ec0294ec4e2e7 seems like it would be the correct fix. If that is the case, I added tests for the LevelAgreement active time and improved the documentation of the `work_in_days` parameter in `Calendar::getActiveTimeBetween` so its behavior is more obvious.

Issue #10686, opened 2022-02-17
[GLPI 10] Display Items menu without rights


tsmr:
### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

Master

### Bug description

Profile self-service : 
No rights for inventory
![image](https://user-images.githubusercontent.com/13178158/154437648-bd1cfa40-d0a5-4adc-9a96-4c331b3bee64.png)

In a existant ticket, the menu still appears : 
![image](https://user-images.githubusercontent.com/13178158/154437892-8777a051-cfa2-4011-aed2-d4f2129b9bd1.png)

### Page URL

http://localhost/glpi/front/ticket.form.php?id=X

ZalemCitizen:
Hi,
Same error here, and for "Technician" field too.

cconard96:
> Hi, Same error here, and for "Technician" field too.

Can you open a new issue for the other field/issue?

PR #10695, opened 2022-02-17
Hide item add button in tickets when no permission to add

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #10686

### DIFF
--- a/src/Item_Ticket.php
+++ b/src/Item_Ticket.php
@@ -205,7 +205,8 @@ class Item_Ticket extends CommonItilObject_Item
             return false;
         }
 
-        $canedit = ($ticket->can($params['id'], UPDATE)
+        $can_add_items = $_SESSION["glpiactiveprofile"]["helpdesk_hardware"] & pow(2, Ticket::HELPDESK_MY_HARDWARE) || $_SESSION["glpiactiveprofile"]["helpdesk_hardware"] & pow(2, Ticket::HELPDESK_ALL_HARDWARE);
+        $canedit = ($can_add_items && $ticket->can($params['id'], UPDATE)
                   && $params['_canupdate']);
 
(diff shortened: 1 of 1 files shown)


cedric-anne (approved):

src/Item_Ticket.php
```diff
@@ -205,7 +205,8 @@ public static function itemAddForm(Ticket $ticket, $options = [])
             return false;
         }
 
-        $canedit = ($ticket->can($params['id'], UPDATE)
+        $can_add_items = $_SESSION["glpiactiveprofile"]["helpdesk_hardware"] & pow(2, Ticket::HELPDESK_MY_HARDWARE) || $_SESSION["glpiactiveprofile"]["helpdesk_hardware"] & pow(2, Ticket::HELPDESK_ALL_HARDWARE);
```
orthagh, line 208:
```suggestion
        $can_add_items = Session::haveRightsOr('helpdesk_hardware', [Ticket::HELPDESK_MY_HARDWARE, Ticket::HELPDESK_ALL_HARDWARE]);
```

no ?

orthagh (changes requested):

src/Item_Ticket.php
cconard96: helpdesk_hardware isn't a regular right and this code was used other places so I thought it was required. I tested with no options selected for this right and then all options. These two cases worked. I then chose to see only All items and nothing was showing in the ticket with this proposed change. With the original fix, the one dropdown was showing as expected.

orthagh (approved):

PR #10707, opened 2022-02-18
Fix clone name uniqueness

AdrienClairembault:
Cloning an entity fail silently because entity names must be unique.

To fix this, I've improved the cloning system to make sure it always set unique name to new items.
For example, if you clone an item named "Test", the clone will be named "Copy of Test (1)".
If "Copy of Test (1)" already exist then we move on to "Copy of Test (2)" and so on.

This also makes it easier to spot cloned items (and renamed them manually as needed):

image

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !23470

### DIFF
--- a/inc/commondbtm.class.php
+++ b/inc/commondbtm.class.php
@@ -1403,15 +1403,13 @@ class CommonDBTM extends CommonGLPI {
       string $current_name,
       ?int $copy_index = null
    ): string {
-      $base_name = sprintf(__("Copy of %s"), $current_name);
-
-      if (is_null($copy_index)) {
-         return $base_name;
+      // No index specified or first copy
+      if (is_null($copy_index) || $copy_index == 1) {
+         return sprintf(__("%s (copy)"), $current_name);
       }
 
-      // Note, we can't use sprintf(__("Copy of %s (%d)"), $current_name, $copy_index) here
-      // We need to rely on the same translation to garantee uniqueness
-      return $base_name . " ($copy_index)";
+      // 1+ copy, add index
+      return sprintf(__("%s (copy %d)"), $current_name, $copy_index);
    }
 
    /**
@@ -1433,14 +1431,14 @@ class CommonDBTM extends CommonGLPI {
       // Force uniqueness for the name field
       if (isset($input[static::getNameField()])) {
          $current_name = $input[static::getNameField()];
+         $copy_name = "";
 
-         // Find all exisiting copy names
-         $search = $this->computeCloneName($current_name) . "%";
+         // Find all exisiting names starting like our current item
          $data = $DB->request([
             'SELECT' => [static::getNameField()],
             'FROM'   => $this->getTable(),
             'WHERE'  => [
-               static::getNameField() => ['LIKE', $search]
+               static::getNameField() => ['LIKE', "$current_name%"]
             ]
          ]);
          $names = [];
@@ -1450,7 +1448,8 @@ class CommonDBTM extends CommonGLPI {
             $names[] = $row[static::getNameField()];
          }
 
-         // The goal is to set the copy name as "Copy of {name} ({i})"
+         // The goal is to set the copy name as "{name} (copy {i})" unless it's
+         // the first copy: in this case just "{name} (copy)" is acceptable
          $copy_index = 0;
 
          // Increment the index until we find an available name
--- a/inc/commondbtm.class.php
+++ b/inc/commondbtm.class.php
@@ -168,18 +168,18 @@ class CommonDBTM extends CommonGLPI {
    static $undisclosedFields = [];
 
    /**
-    * Constructor
-   **/
-   function __construct () {
-   }
-
-   /**
     * Cache used to keep track of the last clone index
     * Usefull when cloning an item multiple time to avoid iterating on the sql
     * results multiple time while looking for an available unique name
     * @var null|int
     */
-   public static ?int $last_clone_index = null;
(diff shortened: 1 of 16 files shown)


inc/commondbtm.class.php
```diff
@@ -1391,19 +1391,78 @@ function prepareInputForAdd($input) {
...
+   ): string {
+      $base_name = sprintf(__("Copy of %s"), $current_name);
+
+      if (is_null($copy_index)) {
+         return $base_name;
+      }
+
+      // Note, we can't use sprintf(__("Copy of %s (%d)"), $current_name, $copy_index) here
+      // We need to rely on the same translation to garantee uniqueness
+      return $base_name . " ($copy_index)";
```
cedric-anne, line 1414: IMHO, using a prefix is not the best option. Indeed, in many cases, items are ordered by name, so, if prefix is used, an item an its clone may be distant from each other in a list.
If a suffix is used instead of a prefix (e.g. `My item (copy)`), an item will be followed by its clones in an alphabetically ordered list.

AdrienClairembault: Indeed, I agree with this.  
I only went for the "Copy My item" approach because it was already used for some items (rules).  

If we are allowed to change this format then how about one of the following:
1) My item (1)
2) My item - copy (1)
3) My item (copy 1)

I like 1 or 2, not sure about 3.

cedric-anne: Solution 1 is the one used by Chrome (at least on my Ubuntu system) when downloading a file multiple times.

Solution 3 is the one used by the default Ubuntu file browser:
image
If I remeber right, it is also the system used by Windows file browser.

Problem with solution 1 is that it may be ambiguous with the format used to append ID on items. If you see `Some computer (3)` in GLPI UI, will it referes to the `Some computer` which ID is `3`, or will it refers to the third clone of this item ?

AdrienClairembault: As an additional info, Windows files explorer use the first solution but start the index at 2:
image

Understood for the confusion with the ID part, it's indeed not ideal.

If we follow Ubuntu implementation of the third solution we end up with:
- My item
- My item (copy)
- My item (2nd copy) 
- My item (3rd copy) 
- My item (4th copy) 

Would this be OK ?
No issues with translation for the dynamic 2nd/3rd/4th part ?

cedric-anne: I guess `(copy)`, `(copy 2)`, `(copy 3)`, ... would be easier to handle as it would not require to handle translation of each number.

AdrienClairembault: Ok for me.
Is this good for you too @orthagh @trasher ?

inc/commondbtm.class.php
```diff
@@ -1391,19 +1391,77 @@ function prepareInputForAdd($input) {
...
+         $copy_name = "";
+
+         // Find all exisiting names starting like our current item
+         $data = $DB->request([
+            'SELECT' => [static::getNameField()],
+            'FROM'   => $this->getTable(),
+            'WHERE'  => [
+               static::getNameField() => ['LIKE', "$current_name%"]
+            ]
+         ]);
```
cedric-anne, line 1443: In some languages, the base name may be at a different position, so assuming that it is always placed at the beginning of the computed clone name is not correct.
Also, the `computeCloneName` method can be overridden and produce something that may not exactly contains the base name (i.e. cloning `test-----user` will produce `test-user-copy`).

So I guess you have no choice than using a `while (countElementsInTable($table, [$namefield => $copy_name]) > 0)` condition.

cedric-anne (changes requested):

inc/commondbtm.class.php
AdrienClairembault: Yeah I've tried to avoid doing this but I guess it's not possible to avoid it.

I'll look into having some kind of cache, because with this solution if you clone an item `n` times you will have at least (if all copy names are available) `n * (n +1) /2` queries (because each copy will start over it's search from the first 'copy index').
This means if you clone an item 60 times you will run 1830 queries.

I suppose it's not a very common usage but It might be better to think ahead and avoid crashing the database if someone actually tries it.

orthagh (approved):
Apart removing the remaining code commented : OK for me

AdrienClairembault:
> Apart removing the remaining code commented : OK for me

Indeed this was a mistake, forgot to delete it. Done now

trasher (approved):

inc/commondbtm.class.php
```diff
@@ -167,13 +167,20 @@ class CommonDBTM extends CommonGLPI {
     */
    static $undisclosedFields = [];
 
+   /**
+    * Cache used to keep track of the last clone index
+    * Usefull when cloning an item multiple time to avoid iterating on the sql
+    * results multiple time while looking for an available unique name
+    * @var null|int
+    */
+   public static $last_clone_index = null;
```
cedric-anne, line 176: You are using this field on a non static context. I think it is preferable to not define it as static.
```suggestion
   public $last_clone_index = null;
```

cedric-anne (changes requested):

inc/commondbtm.class.php
AdrienClairembault: Done

cedric-anne (approved):

trasher (approved):

Issue #10592, opened 2022-02-09
[GLPI 10.0.0-rc1] Display bug in small | split screen


atomiksteak77:
### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

10.0RC1

### Bug description

Bonjour,

Lors de l'utilisation en "petit" écran , la barre "home" ne passe pas sous le menu deroulant:

Hello,

When using the "small" screen, the "home" bar does not go under the scrolling menu:

Exemple 1
![image](https://user-images.githubusercontent.com/99338821/153214044-09b44364-7369-4ea5-b0af-22007596135d.png)

Exemple 2
![image](https://user-images.githubusercontent.com/99338821/153214186-957ad039-15bf-4ca8-be27-16c53c3d659d.png)

### Steps To reproduce

1.Ouvrir GLPI en petite fenetre 
2.Aller dans le menu setup / dropdown ( Par exemple)
3.Descendre la page avec le menu deroulant ouvert .

trasher:
Please write in english only.

PR #10767, opened 2022-02-23
Fix navigation menu z-index

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #10592

Set z-index higher so that it can show over other elements when in "mobile" mode.

### DIFF
--- a/css/includes/components/_global-menu.scss
+++ b/css/includes/components/_global-menu.scss
@@ -37,6 +37,7 @@
 .topbar {
     background-color: $mainmenu-bg;
     color: $mainmenu-fg;
+    z-index: 2000;
 
     .btn-ghost-secondary {
         color: rgba($color: $mainmenu-fg, $alpha: 80%);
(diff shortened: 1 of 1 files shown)


css/includes/components/_global-menu.scss
```diff
@@ -37,6 +37,7 @@
 .topbar {
     background-color: $mainmenu-bg;
     color: $mainmenu-fg;
+    z-index: 2000;
```
orthagh, line 40: You should use bootstrap variable for z-indexes : https://getbootstrap.com/docs/5.0/layout/z-index/

Test to be sure, but i think we need fixed or sticky var

orthagh (changes requested):

orthagh (approved):

cedric-anne (approved):

Issue #10663, opened 2022-02-15
[GLPI 10.0-RC1] Mini Dashboard no longer shows up in tickets


steph461:
### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

10.0-RC1

### Bug description

In my preferences I no longer see the mini dashboard menu. On my production version GLPI 9.5.7, the mini dashboard displays fine

9.5.7 the menu mini tableau de bord ok
![image](https://user-images.githubusercontent.com/68054991/152989669-98c21ae3-be9c-4e05-95e1-0028b03887c6.png)

10.0 RC1 mini tableau de bord doesnt exist
![image](https://user-images.githubusercontent.com/68054991/152989814-d3bef9d4-c2f1-4d62-a3d1-2ae1e5389c26.png)

Dashboard tickets : 
9.5.7
![image](https://user-images.githubusercontent.com/68054991/152990262-b9da14e3-7d46-4661-823f-b9ae7817880e.png)

10.0 RC1 nothing
![image](https://user-images.githubusercontent.com/68054991/152990319-b26ab927-68dd-47e4-b69d-3422315302b2.png)

### Relevant log output

```shell

```

### Page URL

/front/ticket.php

### Steps To reproduce
1. Ticket shows

_No response_

cconard96:
Fixed via #10771

PR #10771, opened 2022-02-23
Rework dashboard dropdown

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #10663

Core issue this tries to fix is how the Dashboard dropdowns in the user preferences work. Currently there are two contexts for dashboards (core and mini_core) where everything is thrown into the "core" context except the one default mini dashboard for Tickets. I'm not sure if there is a way to build a custom mini dashboard or not.
The dropdown asking what you want your mini Ticket dashboard to be didn't specify the context and the underlying code didn't support specifying it either.

The Grid class used a static class property to define the context but it was being set in a constructor. This didn't make sense to me. I removed the static property completely (the constructor parameter wasn't needed anywhere either).

I added a `context` parameter to `Dashboard::getAll`. Previously, this function wouldn't retrieve all dashboards. It would instead load everything in "core" context because that was what the static property was set to and never changed. The new behavior is that it will always load all dashboards into its cache, and then if a specific context is passed, it will filter those out for the result.

The only other use of the static context property is passing its value into a JS string for a context property. Since the function this is in is an instance method, we can use the context from the dashboard that it represents instead.

There remains an issue with the loading of the mini dashboard but this is a separate issue and not related to this as far as I can tell. I'll make another PR for that. I wanted to keep the scope for this one as small as possible.

### DIFF
--- a/src/Dashboard/Dashboard.php
+++ b/src/Dashboard/Dashboard.php
@@ -363,56 +363,51 @@ class Dashboard extends \CommonDBTM
      *
      * @param bool   $force don't check dashboard are already loaded and force their load
      * @param bool   $check_rights use to remove rights checking (use in embed)
-     * @param string $context only dashboard for given context
+     * @param ?string $context only dashboard for given context
      *
      * @return array dasboards
      */
-    public static function getAll(
-        bool $force = false,
-        bool $check_rights = true,
-        string $context = 'core'
-    ): array {
+    public static function getAll(bool $force = false, bool $check_rights = true, ?string $context = 'core'): array {
         global $DB;
 
-        if (!$force && count(self::$all_dashboards) > 0) {
-            return self::$all_dashboards;
-        }
+        if ($force || count(self::$all_dashboards) == 0) {
+            self::$all_dashboards = [];
 
-       // empty previous data
-        self::$all_dashboards = [];
+            $dashboards = iterator_to_array($DB->request(self::getTable()));
+            $items      = iterator_to_array($DB->request(Item::getTable()));
+            $rights     = iterator_to_array($DB->request(Right::getTable()));
 
-        $dashboard_criteria = [];
-        if (strlen($context)) {
-            $dashboard_criteria['context'] = $context;
-        }
+            foreach ($dashboards as $dashboard) {
+                $key = $dashboard['key'];
+                $id  = $dashboard['id'];
 
-        $dashboards = iterator_to_array($DB->request(self::getTable(), ['WHERE' => $dashboard_criteria]));
-        $items      = iterator_to_array($DB->request(Item::getTable()));
-        $rights     = iterator_to_array($DB->request(Right::getTable()));
+                $d_rights = array_filter($rights, static function ($right_line) use ($id) {
+                    return $right_line['dashboards_dashboards_id'] == $id;
+                });
+                $dashboardItem = new self($key);
+                if ($check_rights && !$dashboardItem->canViewCurrent()) {
+                    continue;
+                }
+                $dashboard['rights'] = self::convertRights($d_rights);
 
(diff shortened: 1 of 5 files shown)


cedric-anne (approved):

src/Dashboard/Grid.php
```diff
@@ -54,22 +54,15 @@ class Grid
...
-    public static $context            = '';
     public static $all_dashboards     = [];
 
-    public function __construct(
-        string $dashboard_key = "central",
-        int $grid_cols = 26,
-        int $grid_rows = 24,
-        string $context = "core"
-    ) {
+    public function __construct(string $dashboard_key = "central", int $grid_cols = 26, int $grid_rows = 24) {
```
trasher, line 59:
```suggestion
    public function __construct(string $dashboard_key = "central", int $grid_cols = 26, int $grid_rows = 24)
    {
```

src/Dashboard/Dashboard.php
```diff
@@ -363,56 +363,51 @@ public function cloneCurrent(): array
...
+     * @param ?string $context only dashboard for given context
      *
      * @return array dasboards
      */
-    public static function getAll(
-        bool $force = false,
-        bool $check_rights = true,
-        string $context = 'core'
-    ): array {
+    public static function getAll(bool $force = false, bool $check_rights = true, ?string $context = 'core'): array {
```
trasher, line 370:
```suggestion
    public static function getAll(bool $force = false, bool $check_rights = true, ?string $context = 'core'): array
    {
```

trasher (approved):

orthagh:
There is, i think, a small regression with the proposed changed but not directly in glpi.
The grid had a `$context` variable to let developers use the whole grid system without being polluted by dashboards from the core context.
The dashboards dropdown displayed on top of the grid load all dashboards relative to it's grid context.

It's difficult to say if it still work (i guess not) after your changes, because i'm not aware of existing implementations.

I'm agree with most of the proposed changes,  but not with removing the context relative to a grid.
I think we shoul keep this part, and pass it into `$params` of `Grid::dropdownDashboard`  called in `Grid::show`
With this, a plugin will still be able to display a Dashboard grid with only its own dashboards listed in the top left dropdown.

orthagh:
Thanks for the update.
Could you rebase the pr to ease live tests (currently i'm soft blocked by rc1/rc2 update) ?

cconard96:
> Thanks for the update. Could you rebase the pr to ease live tests (currently i'm soft blocked by rc1/rc2 update) ?

Done

src/Dashboard/Dashboard.php
```diff
@@ -363,56 +363,52 @@ public function cloneCurrent(): array
...
+
+            $dashboards = iterator_to_array($DB->request(self::getTable()));
+            $items      = iterator_to_array($DB->request(Item::getTable()));
+            $rights     = iterator_to_array($DB->request(Right::getTable()));
+
+            foreach ($dashboards as $dashboard) {
+                $key = $dashboard['key'];
+                $id  = $dashboard['id'];
+
+                $d_rights = array_filter($rights, static function ($right_line) use ($id) {
```
orthagh, line 385: What's the purpose of static anonymous function here (and below) ?

cconard96: Same as old code but the use of the "static" keyword on anonymous functions is preferred if you don't need access to properties or methods of the object instance. It is probably just a minor performance improvement.

orthagh: ok

orthagh (approved):

cedric-anne (approved):

PR #10777, opened 2022-02-24
feat(Ruleticket): activates the major priority actions of the rules

stonebuzz:
activates the "major" priority actions of the rules

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !23501

### DIFF
--- a/src/RuleAction.php
+++ b/src/RuleAction.php
@@ -585,7 +585,8 @@ class RuleAction extends CommonDBChild
 
                         case "dropdown_priority":
                             if ($_POST["action_type"] != 'compute') {
-                                $param['name']  = 'value';
+                                $param['name']      = 'value';
+                                $param['withmajor'] = true;
                                 Ticket::dropdownPriority($param);
                             }
                             $display = true;


cedric-anne (approved):

orthagh (changes requested):
As seen together, this pr should be refused and an explanation should be given to the initial requester.

The major priority is a special case which is designed to be managed manually only.
The two other dropdowns implicated in priority matrix doesn't have this major level, the maximum is "Very high".
If we remove the manual side only of "Major" level, there is no need for it to exists at all, a "very high" level is sufficient to achieve ordering and prioritization.

The purpose of this mechanism is, regardless of all automations or matrix setup in config, you have a manual level to prioritize above everything else when things go very bad in your IT service.

In a normal usage of GLPI, you should never use this level, and the max level of your priority should be "Very high" (also for VIP requests).

PR #10820, opened 2022-03-01
Add rule action to skip remaining rules

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

Add new global rule action for `Stop rules processing` which will tell the engine to skip all other rules in the collection.
Currently, some rule collections are set to stop after the first match but the default and the case for most is to run all matching rules. With this new option along with the existing ability to change the rule priority/order, this will give the user more control over which rules run and when they run.

This does not stop other collections from running for the same event.

### DIFF
--- a/src/Rule.php
+++ b/src/Rule.php
@@ -2874,7 +2874,12 @@ class Rule extends CommonDBTM
 
     public function getActions()
     {
-        return [];
+        return [
+            '_stop_rules_processing' => [
+                'name' => __('Stop rules processing'),
+                'type' => 'yesonly',
+            ]
+        ];
     }
 
 
--- a/src/Rule.php
+++ b/src/Rule.php
@@ -2876,7 +2876,7 @@ class Rule extends CommonDBTM
     {
         return [
             '_stop_rules_processing' => [
-                'name' => __('Stop rules processing'),
+                'name' => __('Skip remaining rules'),
                 'type' => 'yesonly',
             ]
         ];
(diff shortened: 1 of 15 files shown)


orthagh (commented):
Globally ok with this proposal, it could be usefull.
Two things:

- the wording for "Stop rules processing" may be interpreted by canceling rules changes, we may add a small precision with `(and apply changes)` or something like that (suggestions welcome)
- i don't understand the changes relative to `parent::getActions`, can you explain ?

cconard96:
>     * the wording for "Stop rules processing" may be interpreted by canceling rules changes, we may add a small precision with `(and apply changes)` or something like that (suggestions welcome)

Would "Skip remaining rules" work better?

>     * i don't understand the changes relative to `parent::getActions`, can you explain ?
Instead of adding this new action to every rule collection that allows multiple rules to run, it is declared in the parent class and inherited.

orthagh:
> Would "Skip remaining rules" work better?

yes completly

> it is declared in the parent class and inherited.

Indeed, i miss that in my first reading (it's the first change ...)

orthagh (approved):
Ok for me (after rewording)

trasher (approved):

cedric-anne (approved):

PR #10829, opened 2022-03-02
fix(dashboard): prevent error from php 8

stonebuzz:
Error mey occurs from PHP >= 8.0

image

The ```call_user_func_array()``` function will use they array keys as parameter names, and this can be a breaking-change in certain situations, where the parameter array is an associative array with non-numeric keys. 

see https://php.watch/versions/8.0/named-parameters#named-params-call_user_func_array

This PR fix this

| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #23519

### DIFF
--- a/src/Dashboard/Grid.php
+++ b/src/Dashboard/Grid.php
@@ -930,7 +930,7 @@ HTML;
             if (isset($card_options['args']['apply_filters'])) {
                 $provider_args['params']['apply_filters'] = $card_options['args']['apply_filters'];
             }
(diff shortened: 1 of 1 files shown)


orthagh (approved):
This is indeed very strange we didn't see much more issue due to change in php8

cedric-anne (approved):

trasher (approved):

Issue #10292, opened 2022-01-08
Actors duplicated in tickets if I sublit incomplete form, fix it, and submit again


btry:
### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

10

### Bug description

If a user submits a new ticket with a mandatory field empty, then fills the mandatory field and submits again, the requester and the assigned users appear twice in the created ticket.

### Page URL

/front/ticket.form.php

### Steps To reproduce

1. use the default ticket template with title mandator in your entity
2. open the ticket creation page
3. filll any description, but leave the title empty
4. submit the ticket; the browser tells the title is required
5. fill the title
6. submit the form
7. open the created ticket : see the actors appears twice

![image](https://user-images.githubusercontent.com/14139801/148637116-9447c7f6-8833-44c8-99d0-54d0a8c05941.png)

cconard96:
I've tried to recreate this issue a few times over the last week or so but haven't been able to.
Is the issue still present for you in the latest GLPI 10 state?

btry:
Hi

It seems thngs changed since my report. I remember that I found the issue by creating a ticket without a title or a content, despite they were both mandatory with Default template. It is no longer possible to submit such form as the browser is now aware if the mandatory property on both fields.

I'm trying to reproduce with an other mandatory field.

btry:
I did a try with mandatory observer, then mandatory category.

In both cases, when the submission is rejected, I go back to an empty ticket; all inputs seems lost. There is maybe a regression compared  to GLPI 9.5, and it is no longer possible to reproduce this issue.

cconard96:
All related PRs have been merged. The issues listed here should all be resolved now.

PR #10846, opened 2022-03-03
Fix restore of saved input

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

Related to #10292.
Since #10114, a new form option `loaded` was added during the loading of a tab and this caused the saved input of a form to be overridden with this one value before it could be restored.

### DIFF
--- a/src/CommonGLPI.php
+++ b/src/CommonGLPI.php
@@ -1124,6 +1124,7 @@ JAVASCRIPT;
     {
         // Item might already be loaded, skip load and rights checks
         $item_loaded = $options['loaded'] ?? false;
+        unset($options['loaded']);
         if (!$item_loaded) {
             if (
                 isset($options['id'])


trasher (approved):

orthagh (approved):

AdrienClairembault:
`CommonDBTM::restoreInput()` has a single parameter containing the default values (which are in fact more like "fallback" values):

image

In most places, we use the current item fields as the default values (make sense, if we can't load the saved input then we fallback to the current fields).
Example:

image

However, in `CommonDBTM::initForm()`, we indeed use `$options` instead:

image


I'm not sure that using `$options` here makes sense.
Indeed, why use an option array as the fallback values ?
Unless I'm missing something, it will never contains any relevant values.

Sending a new option to the display method shouldn't break an unrelated feature.
Also, shouldn't others existing options like `'withtemplate` cause the same issue ?
As far as I understand, it doesn't seem unique to the new `loaded` option

cconard96:
In this case, `restoreInput` never returns the defaults with or without the fix. The issue is with saveInput which is done in `CommonGLPI::showTabsContent` and only for new ITILObjects. In that case, I'm pretty sure that the `withtemplate` option never worked with ITILObjects (I've tried before to use it to manually select a ticket template) and doesn't get used regularly as I guess all ITILObjects have a template.

cedric-anne (commented):
I tried to understand why it prevents input restoration, but was not able to find what is the problem.

If adding this option is an issue, maybe `loaded` should be defined as a second argument, instead of being set before method call and unset inside method call.

AdrienClairembault:
I still have doubts that the `loaded` option impact this (or maybe there something else that need to be fixed too ?).

I've applied your changes, logged off and on to clear my session, and tried the following steps:

1) Go to ticket creation
2) Add any requester
3) Write a random description
4) Submit

I always end up with two requested: the selected requester and my current user: 

image

image

I do not have any active ticket rules and this option is disabled in my user prefs:

image

If I skip the second step (add requester) then I am still set as the requester:
image

Am I missing something or is there another issue ? Do you reproduce the same behavior ?

AdrienClairembault:
I think I found the issue.

Check out this code from `Commonitilobject::prepareInputForAdd` :

image

If you create a ticket on GLPI 9.5 from the ticket form, you never reach this code because `_users_id_requester` exist even if there is no requester:

image

On GLPI 10, there is no `_users_id_requester` key, even if you specify a requester.

image

This mean we validate this condition every time, which trigger setting the current user as the requester.

AdrienClairembault:
As a side note, I'm also able to reduce it on the 10.0.0-beta tag which was before https://github.com/glpi-project/glpi/pull/10114.

I think it's an issue that originated with the rewrite of UI (maybe the `_users_id_requester` input doesn't exist anymore but some code still depends on it ?) and has probably been here for a long time, it's weird no one notified it before.

orthagh:
Regarding the part found by @AdrienClairembault : https://github.com/glpi-project/glpi/blob/10.0.0-rc2/src%2FCommonITILObject.php#L1955-L1962
I'm not sure what it try to achieve.

`autoimport` key is set to 1 on mailcollector and anomymous helpdesk form and it aims in others part to do some process differently when we are collecting an email i think.

In the current cited code, I don't know why you would set current user as requester when a parameter exists for that (as mentioned above). Seems to be a relic of the past (very long ago), hidden until today with the removal of `_users_id_requester` from post.

Except this piece of code, all other usages of `_auto_import` seems legit.


So my conclusion is we may want to remove the autofill of requester in commonitilobject.

cconard96:
I'm having trouble understanding how the last few comments are related to this PR.

This addresses an issue reported in a comment on the linked issue where all data is lost if you forget a mandatory field. The original issue reported was indeed that there were duplicated actors but this seemed to be resolved.

orthagh:
So let's resume as we have a misunderstanding since the befinning.
We have two issues talked here:

- autofilling of user_id_requester : fixed by #10941
- loss of user inputs (like title/content) when the form is submitted with a mandatory empty category

This last is indeed since #10114 (c707f0be5a3bb18387434d2266639a4f86853526 - after a git bisect).
I don't know yet if `loaded` is the good way, but let's not mix subject. The first one should adressed in #10941 only.
let's concentrate here on the restore input issue

orthagh:
And so for the original issue (override of the session saved input), i agree with @cconard96 
The addition of `loaded` key triggers this piece of code : https://github.com/glpi-project/glpi/blame/master/src/CommonGLPI.php#L857-L874

Before #10114, `$options` was empty when passing in this place, and now as it contains only the `loaded`, the function `saveInput` is called and previous inputs erased before being restored.

That's said, the `saveInput` here seems not legits. It's already called in `CommonDBTM::add` (before `prepareInputForAdd`).
This was been added by #2682.
I'm in favor to remove this call (and so not touching `loaded` key) and check we don't add regressions of issues #2031 and #2553

orthagh:
Ok after more investigation, the content of ticket after reload (change of request type for example), is passed within GET and so is truncated indeed.
I don't find a more efficient way than indeed calling saveInput.

orthagh (approved):

orthagh:
So let's merge, as `loaded` is not used after

AdrienClairembault (approved):

PR #10867, opened 2022-03-05
Fix item name created from template

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

Creating an asset from a template always added "(copy)" to the name even though I was specifying the name. Automatic generation of names should probably not happen if a name is specified in `override_input`.

### DIFF
--- a/src/Features/Clonable.php
+++ b/src/Features/Clonable.php
@@ -200,8 +200,11 @@ trait Clonable
         }
         $input = $new_item->cleanCloneInput($input);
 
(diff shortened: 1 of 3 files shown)


trasher (approved):

trasher:
I do not really understand why Software tests are broken :/

orthagh (approved):

cedric-anne (approved):

PR #10874, opened 2022-03-06
Allow assigning without UPDATE right

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #10364

Fix for #10364 done for GLPI 10. I think the only issue in 9.5 was the UPDATE right check in the front file. For GLPI 10, there were other issues due to using `_actors` instead of individual fields for each actor type (_itil_assign`).

### DIFF
--- a/front/ticket.form.php
+++ b/front/ticket.form.php
@@ -76,9 +76,7 @@ if (isset($_POST["add"])) {
     }
     Html::back();
 } else if (isset($_POST['update'])) {
-    if (!$track->canAssign()) {
-        $track->check($_POST['id'], UPDATE);
-    }
+    $track->check($_POST['id'], UPDATE);
     $track->update($_POST);
 
     if (isset($_POST['kb_linked_id'])) {
--- a/front/ticket.form.php
+++ b/front/ticket.form.php
@@ -76,7 +76,9 @@ if (isset($_POST["add"])) {
     }
     Html::back();
 } else if (isset($_POST['update'])) {
-    $track->check($_POST['id'], UPDATE);
+    if (!$track::canUpdate()) {
+        Html::displayRightError();
+    }
     $track->update($_POST);
 
     if (isset($_POST['kb_linked_id'])) {
--- a/front/ticket.form.php
+++ b/front/ticket.form.php
@@ -76,7 +76,9 @@ if (isset($_POST["add"])) {
     }
     Html::back();
 } else if (isset($_POST['update'])) {
-    $track->check($_POST['id'], UPDATE);
+    if (!$track->canAssign()) {
+        $track->check($_POST['id'], UPDATE);
+    }
     $track->update($_POST);
 
     if (isset($_POST['kb_linked_id'])) {
--- a/src/Ticket.php
+++ b/src/Ticket.php
@@ -537,6 +537,7 @@ class Ticket extends CommonITILObject
         if (
             !Session::haveRight(self::$rightname, UPDATE)
             && !$this->ownItem()
+            && !$this->canAssign()
         ) {
            //we always return false, as ownItem() = true is managed by below self::canUpdate
             return false;
--- a/src/Ticket.php
+++ b/src/Ticket.php
@@ -536,8 +536,7 @@ class Ticket extends CommonITILObject
        // if we don't have global UPDATE right, maybe we can own the current ticket
         if (
             !Session::haveRight(self::$rightname, UPDATE)
-            && !$this->ownItem()
-            && !$this->canAssign()
+            && (!$this->ownItem() || !$this->canAssign())
         ) {
            //we always return false, as ownItem() = true is managed by below self::canUpdate
             return false;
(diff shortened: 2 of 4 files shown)


front/ticket.form.php
```diff
@@ -76,7 +76,9 @@
     }
     Html::back();
 } else if (isset($_POST['update'])) {
-    $track->check($_POST['id'], UPDATE);
+    if (!$track->canAssign()) {
```
orthagh, line 79: I'm not sure about adding a specifity in the controller part.
I would prefer a change in canUpdatexxx methods i think.

More globally, this is very strange that 9.5.x is impacted as assign right is already tested as i remember (maybe not as you complete the test method).
Where is more strange, there is a case for this in Ticket::canUpdate 
image

orthagh: maybe it's just adding a line `!$this->canAssign()` here : https://github.com/glpi-project/glpi/blob/master/src%2FTicket.php#L539

cconard96: I made those changes and there are 3 test failures currently (only checked Ticket tests). I'll apply the improved test error messages from #10824 and look into why it fails some more.

cconard96: There is one test that fails with this method that makes it look like having the right to assign also gives global update permission for tickets. The access denied screen when trying to just assign users has returned.

glpi123:
Hello cconard96],
Thank you for your response.
I have GLPI 9.5.5, and i made the proposed changes only in file the front/ticket.form.php, and it worked. I did some global functional test, all seems in order.
Please, does orthagh's comment have any impact on the global functioning?

Best regards

cedric-anne (approved):

cconard96:
Current PR is not OK. Front file still throws the access denied error when trying to assign users and the UPDATE/ASSIGN rights seem broke based on the failed test.

src/Ticket.php
```diff
@@ -536,7 +536,7 @@ public function canUpdateItem()
        // if we don't have global UPDATE right, maybe we can own the current ticket
         if (
             !Session::haveRight(self::$rightname, UPDATE)
-            && !$this->ownItem()
+            && (!$this->ownItem() || !$this->canAssign())
```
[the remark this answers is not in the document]
orthagh: Problem is, this removes the read-only attributes on ticket fields

orthagh:
We have the same problem with priority update right (at least on master, i didn't checked on 9.5).

The suggested change in canUpdateItem is not good as it removes the read-only attribute on fields when you have only own or change_priority right.

I'm currently trying to find another way, but maybe, the only way is indeed a change on controller part.

cconard96:
Tested again but still get access denied error when trying to assign a user to a ticket.

Profile has the following permissions:
- Purge
- See all tickets
- See assigned
- Assign
- Beeing in charge
- Change the priority

src/Ticket.php
```diff
@@ -536,7 +536,7 @@ public function canUpdateItem()
        // if we don't have global UPDATE right, maybe we can own the current ticket
         if (
             !Session::haveRight(self::$rightname, UPDATE)
-            && !$this->ownItem()
+            && (!$this->ownItem())
```
cedric-anne, line 539:
```suggestion
            && !$this->ownItem()
```

orthagh:
@cconard96 did you pull locally the changes ?
I checked on another computer to be sure, and i can save the new assignment (and also priority).

cconard96:
I thought it pulled all changes but I tested again this morning and it looks OK.
I cannot approve my own PR, but the changes are OK for me.

orthagh (approved):

cconard96:
I think both of the other actor PRs try to accomplish the same thing but do it in a different way. I'm not sure which is the better solution.

cedric-anne (approved):

trasher (approved):

glpi123:
Hello cconard96,
I tried to operate the proposed modifications in GLPI 9.5.5, the second modification in the file src/CommonITILObject.php is not found : 

  // parse posted actors
        foreach ($this->input['_actors'] as $actortype_str => $actors) {
            $actortype = constant("CommonITILActor::" . strtoupper($actortype_str));
            if ($actortype === CommonITILActor::ASSIGN && !$this->canAssign()) {
                continue;
            }
            if ($actortype !== CommonITILActor::ASSIGN && !$this->canUpdateItem()) {
                continue;
            }
            $existings = $this->getActorsForType($actortype);
            $added   = [];

Please let me know where I should add the change in GLPI 9.5.5

Best regards.

front/ticket.form.php
```diff
@@ -76,7 +76,9 @@
     }
     Html::back();
 } else if (isset($_POST['update'])) {
-    $track->check($_POST['id'], UPDATE);
```
VladoTTX, line 79: Is it possible still call the check method and achieve the same thing?

PR #10875, opened 2022-03-06
Fix truncated double SQL warning

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #10536 (Issue in comment)

### DIFF
--- a/src/Dropdown.php
+++ b/src/Dropdown.php
@@ -440,7 +440,9 @@ class Dropdown
         $name    = "";
         $comment = "";
 
-        if ($id) {
+        $id_set = $id !== null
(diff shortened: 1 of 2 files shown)


trasher (approved):

cedric-anne (commented):
Having something else than an id here is not normal. A `trigger_error(sprintf('Invalid id "%s".', $id), E_USER_WARNING);`, as the source of this wrong value has still to be fixed.

cedric-anne (approved):

trasher (approved):

PR #10941, opened 2022-03-10
remove autoset of _user_id_request without config, see #10846

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

Fix the autofilling of `_users_id_requester`without regarding the user preferences

### DIFF
--- a/tests/functionnal/Glpi/ContentTemplates/Parameters/ChangeParameters.php
+++ b/tests/functionnal/Glpi/ContentTemplates/Parameters/ChangeParameters.php
@@ -85,19 +85,7 @@ class ChangeParameters extends AbstractParameters
                 'completename' => 'category_testGetValues',
             ],
             'requesters' => [
-                'users'  => [
-                    [
-                        'id'       => getItemByTypeName("User", TU_USER, true),
-                        'login'     => TU_USER,
-                        'fullname' => TU_USER,
-                        'email'    => "_test_user@glpi.com",
-                        'phone'    => null,
-                        'phone2'   => null,
-                        'mobile'   => null,
-                        'firstname'  => null,
-                        'realname'   => null,
-                        'used_items' => [],
-                    ],
+                'users'  => [],
                 ],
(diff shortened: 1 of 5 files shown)


trasher (approved):

cedric-anne (approved):

tests/functionnal/Glpi/ContentTemplates/Parameters/ChangeParameters.php
```diff
@@ -85,20 +85,7 @@ public function testGetValues(): void
...
-                        'fullname' => TU_USER,
-                        'email'    => "_test_user@glpi.com",
-                        'phone'    => null,
-                        'phone2'   => null,
-                        'mobile'   => null,
-                        'firstname'  => null,
-                        'realname'   => null,
-                        'used_items' => [],
-                    ],
-                ],
```
AdrienClairembault, line 101: Maybe it would be better to properly set the requester in the code above (`$this->createItem('Change', [...`) rather than removing him from the expected data ?

AdrienClairembault (changes requested):
Did you also test for other failures related to the removal of the `_user_id_request` field from the main ticket form ?
I can confirm that there is at least one failure regarding setting a requester from a ticket template but I'm afraid there may be more.

tests/functionnal/Glpi/ContentTemplates/Parameters/ChangeParameters.php
orthagh: not yet.
But you're right, i need to check also this behavior, you mentioned it yesterday.

orthagh: Please check #10950 for this issue.
If you agree the current pr can be merged.

AdrienClairembault (approved):

PR #10943, opened 2022-03-10
Improve timeline item form layouts for non-xxl screens

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #10931

The timeline item forms (such as tasks) can feel very cramped unless the browser window is very large. I some cases, the fields are too small to interact with. This is also an issue if, like me, you have the browser's console open often and it is docked on the left or right.

The following changes were made to alleviate these issues:
- Less padding around timeline item forms on non-xxl screens
- Adjusted timeline layout so the left side would have more space on non-xxl screens
- Adjusted col breakpoints in timeline item forms to match xxl and not md screens and added a second break point to slightly reduce the size of the `content` field before falling back on having the extra fields show under the content.

I'm still looking at the behavior of the right-side panel. The expand/collapse isn't working right and I notice some odd resize issues before my changes.

### DIFF
--- a/templates/components/itilobject/answer.html.twig
+++ b/templates/components/itilobject/answer.html.twig
@@ -29,7 +29,7 @@
  # ---------------------------------------------------------------------
  #}
 
-<div id="new-itilobject-form" class="ms-2">
+<div id="new-itilobject-form" class="ms-auto ps-3">
    {% for timeline_itemtype in timeline_itemtypes %}
       {% set show_kb_sol = load_kb_sol > 0 and timeline_itemtype.type == 'ITILSolution' %}
       <div class="timeline-item mb-3  {{ timeline_itemtype.type }} collapse {{ show_kb_sol ? 'show' : '' }}"
--- a/templates/components/itilobject/answer.html.twig
+++ b/templates/components/itilobject/answer.html.twig
@@ -29,7 +29,7 @@
  # ---------------------------------------------------------------------
  #}
 
-<div id="new-itilobject-form" class="ms-auto">
+<div id="new-itilobject-form" class="ms-2">
    {% for timeline_itemtype in timeline_itemtypes %}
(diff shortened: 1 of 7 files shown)


orthagh:
Can you add a screenshot ?

cconard96:
Using the same screen size comparing `master` to the changes in this PR.
Before:
Screenshot from 2022-03-10 09-52-34


Current:
Screenshot from 2022-03-10 09-53-27

It still isn't great, but this is the smallest the fields get in this PR so is a worst-case scenario. There may be other tweaks that can be done to improve it more.

cconard96:
> it's better but not great indeed. under (and equals) xl screens maybe we can wrap the right column. And maybe hide it under a toggle for cleaner display.

Since there are already controls for collapsing and expanding the panel, maybe some JS can collapse it automatically when the timeline loads if the screen is too small.
For the timeline forms themselves, the right column does wrap for screens smaller than `xl`. Since the breakpoints are based on the screen size and not parent element, these breakpoints start to mean less when you have a ton of nested elements and have a lot of elements horizontally as we do here. It would almost have to wrap and show the fields vertically under the content for anything less than `xxl`.

orthagh:
> the right column does wrap for screens smaller than xl

Indeed, tested live, it's ok with these breakpoints (xxl spacing, xl wrap)


Here is a suggestion for gaining 30-40 px more.
image


One last thing not related to this pr (i have the issue on current master), we currently have an horizontal scrollbar on the timeline.
Don't know why (yet)

orthagh:
i tried and in fact, the last issue is more than the margin-start.
when the `#new-itilobject-form` take all the width, we need at least -3 spacing
but when it doesn't, it require the ms-auto to right aligned.

I suggest so: `ms-auto ps-3`

orthagh (approved):

cedric-anne (approved):

PR #9761, opened 2021-10-23
Manage teams from Kanban

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

PR #10950, opened 2022-03-11
permits to reach actors loading from template when opening new ticket…

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | https://github.com/glpi-project/glpi/pull/10941#pullrequestreview-905977936

Issue is within this part : https://github.com/glpi-project/glpi/blob/c243e976801fa4689a8c33d7c67ce22088043d80/src/CommonITILObject.php#L164-L192

To note, @AdrienClairembault talked about `_users_id_requester` but all actor types and itemtypes (observer/assign and group/supplier) are also impacted.

At first load when opening a ticket form, `$params['_actors']`is an empty array and after a page reload (when selecting a category for example), it contains sub arrays, one for each actors types.
Also, we exactly see after the `isset` test we parse each requester type.
**If we enter in this check, we will always return the empty `$actor` array.**

So checking for   `$params['_actors']` lenght (wich is 0 in the case we currently parsing) permits to reach the part where actors are loaded from template : https://github.com/glpi-project/glpi/blob/c243e976801fa4689a8c33d7c67ce22088043d80/src/CommonITILObject.php#L195-L241


trasher (approved):

cedric-anne (approved):

cconard96 (approved):

AdrienClairembault (changes requested):
Does not seems to work for me.

Configuration:
image

Ticket form seems OK:
image

But after submitting the form, the values are gone:
image

Also the header says one actor when there is zero.
There is indeed a value inside `glpi_groups_tickets` but without a valid `groups_id`:
image

AdrienClairembault:
Also, another issue with the form.

As you've seen in my previous comment my template add a requester and an assigned group.
If I add any other requester/observer/technician then the preset assigned group is lost.

Before adding another requester/observer/technician:

image

After: 

image

trasher (approved):

orthagh:
Wait for merge, the subject is clearly complex, i need to a global check on all forms
and a bunch of unit tests could be welcome

cconard96:
> The case may happens as we have 3 different keys to add actors:
> 
>     * former one `_itemtype_id_actortype` (like `_user_id_requester`)
> 
>     * additional actors from rules (this one already de-duplicates actors)
> 
>     * the new way from twig rework push an `_actors`  key

> I'm not sure this is the best way and if we want to keep the code in `prepareInputForAdd` but for avoiding future pain (finding how an actor is added is very painful as we have many process who can this), i though avoiding duplication on actor add process may be useful.

There is also now the idea of "Teams" from a Kanban work (#9761) that somewhat unifies actor management for ITIL Objects, Projects and Project Tasks :cry: . Luckily this is only used in the context of the Kanban currently.

I don't really have an opinion on this or alternatives only because it is very complex/redundant currently and I don't entirely know every different method's usage and behavior.

cedric-anne:
> Th only "issue", with removing the code for default actor in the prepare function , could be tickets created by api, they will not benefit anymore the glpiset_default_tech feature.

`set_default_tech` configuration label is `Pre-select me as a technician when creating a ticket`. It would seems logical to me that   behaviour of this configuration would be to preselect current user in actors dropdown in the ticket creation form.
So, based on this, IMHO:
 - it should have an effect only if the field is not marked as hidden and/or readonly in the corresponding template;
 - it should not have any effect on API calls.


> Another think, like described above, [addAdditionalActors](https://github.com/glpi-project/glpi/blob/10.0.0-rc2/src/CommonITILObject.php#L2534) deduplicates its own entry with existing actors since the recent #10826, we may remove the changes done in `CommonItilObject`

Anything that can reduce the amount of code related to ticket actors is welcome.

trasher (approved):

cedric-anne (approved):

AdrienClairembault (approved):

cconard96 (approved):

PR #10957, opened 2022-03-11
 add device and item_device BigNumber card in dashboard 

delcroip:
replace  add device and item_device BigNumber card in dashboard #10946 

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #number

### DIFF
--- a/src/Dashboard/Grid.php
+++ b/src/Dashboard/Grid.php
@@ -1139,7 +1139,7 @@ HTML;
             }
 
             foreach ($CFG_GLPI['itemdevices'] as $itemtype) {
-                $fk_itemtype = $itemtype::$itemtype_2;
+                $fk_itemtype = $itemtype::getDeviceType();
                 $label = sprintf(
                     __("Number of %s by type"),
                     $itemtype::getTypeName(Session::getPluralNumber()),


cconard96 (changes requested):
The `plugins/remove.txt` file should not be removed. It is here to ensure the folder exists since git doesn't store empty folders.

delcroip:
> The `plugins/remove.txt` file should not be removed. It is here to ensure the folder exists since git doesn't store empty folders.

I have to double check why it is till removed because I added it back already

delcroip:
I didn't managed the trad yet 
`__("Number type of %s")`
`__("%s by type")` 

because I don't want to mess up the pot file

src/Dashboard/Provider.php
```diff
@@ -803,7 +803,9 @@ public static function nbItemByFk(
         if ($fk_item instanceof CommonTreeDropdown) {
             $name = 'completename';
         }
-
+        if (!($item instanceof CommonDevice)) {
```
cconard96, line 807: CommonDevice is in the global namespace while Provider is not, so this reference isn't valid.
You can add `use CommonDevice;` to the top of the file with the other use statements to fix the reference.

cconard96: Nevermind, I see you just changed that.

cconard96:
> I didn't managed the trad yet `__("Number type of %s")` `__("%s by type")`
> 
> because I don't want to mess up the pot file

No need to change the pot file in this PR.

src/Dashboard/Provider.php
```diff
@@ -803,7 +804,9 @@ public static function nbItemByFk(
         if ($fk_item instanceof CommonTreeDropdown) {
             $name = 'completename';
         }
-
+        if (is_subclass_of($fk_item, 'CommonDevice')) {
```
cedric-anne, line 807:
```suggestion
        if ($fk_item instanceof CommonDevice) {
```

cedric-anne:
@delcroip Could you please squash all your commits into only one and revert useless changes ?

src/Dashboard/Grid.php
```diff
@@ -1138,6 +1142,56 @@ public function getAllDasboardCards($force = false): array
                 }
             }
 
+            foreach ($CFG_GLPI['itemdevices'] as $itemtype) {
+                if (property_exists($itemtype, 'itemtype_2')) {
```
cedric-anne, line 1146: Instead of using `itemtype_2` property, you should rely on `$itemtype::getDeviceForeignKey()`.

Also, I guess an `Item_Devices` implementation has always a related device class.

delcroip: indeed it works without the check

src/Dashboard/Provider.php
delcroip: will be updated

src/Dashboard/Grid.php
```diff
@@ -1385,7 +1438,6 @@ public function getAllDasboardCards($force = false): array
         return $cards;
     }
```
cedric-anne, line 1440:
```suggestion


```

src/Dashboard/Grid.php
```diff
@@ -57,8 +57,12 @@ class Grid
...
     public static $all_dashboards     = [];
 
-    public function __construct(string $dashboard_key = "central", int $grid_cols = 26, int $grid_rows = 24, string $context = 'core')
-    {
+    public function __construct(
+        string $dashboard_key = "central",
+        int $grid_cols = 26,
+        int $grid_rows = 24,
+        string $context = 'core'
+    ) {
```
cedric-anne, line 65:
```suggestion
    public function __construct(string $dashboard_key = "central", int $grid_cols = 26, int $grid_rows = 24, string $context = 'core')
    {
```

src/Dashboard/Grid.php
```diff
@@ -1138,6 +1138,55 @@ public function getAllDasboardCards($force = false): array
                 }
             }
 
+            foreach ($CFG_GLPI['itemdevices'] as $itemtype) {
+                $fk_itemtype = $itemtype::$itemtype_2;
```
cedric-anne, line 1142:
```suggestion
                $fk_itemtype = $itemtype::getDeviceType();
```

cedric-anne:
image

On left, the number of `Item_DeviceProcessor` and, on right, the number of `DeviceProcessor`.
This is due to the fact that `Item_DeviceProcessor::getTypeName()` returns the same value as `DeviceProcessor::getTypeName()`:
```php
    public static function getTypeName($nb = 0)
    {
        $device_type = static::getDeviceType();
        return $device_type::getTypeName($nb);
    }
```

@orthagh @trasher Should we change the device items typename ?

orthagh:
Yes we can, @AdrienClairembault added recently some missing getTypeName for item relation class
Most are name `__("##Parent name## item", ""##Parent name## items", nb)`
The same can be done on `Item_Devices` sub classes.
In fact we already do this for UI (not a generic call btw) : https://github.com/glpi-project/glpi/blob/10.0.0-rc2/src/CommonDevice.php#L128

So adding a `getTypeName` method in `Item_Devices` calling the one of "Parent device" with a sprintf should do the work in a generic way

cedric-anne:
> 

Done in #10976.

cedric-anne (approved):

delcroip:
Then I Think the title box names must be updated 
 something like: 
Number of %s (processor)
Number of %s (processor item)
Number of %s item per %s  (processor item, processor)

orthagh (approved):

trasher (approved):

PR #11006, opened 2022-03-17
Group approval requests

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

Allow requesting approval from a group where a member of that group can then answer the request on the behalf of the group.
This is different than the previous option for "Group" approvals where it was just a filter for a list of users.

Previous "Group" option is know labelled "Group User(s)" while the new option has the old label "Group".
`users_id_validate` field no longer represents who was asked for an approval and is only used to represent who answered the request. `itemtype_target` (User or Group) and `items_id_target` are now used instead to know who was asked for approval.

On the timeline, approval request answers that were asked of a group have a special header that say `Created n time ago by
'xxx' for the group 'yyyy'` to indicate the answer was done by a user on the group's behalf.


src/CommonITILValidation.php
```diff
@@ -1360,96 +1444,6 @@ public function getRights($interface = 'central')
...
-     *  - id                      : ID of object > 0 Update, < 0 New
-     *  - entity                  : ID of entity
-     *  - right                   : validation rights
-     *  - groups_id               : ID of group validator
-     *  - users_id_validate       : ID of user validator
-     *  - applyto
-     *
-     * @return void Output is printed
-     **/
-    public static function dropdownValidator(array $options = [])
```
cedric-anne, line 1378: This one is used in `templates/components/itilobject/fields/global_validation.html.twig`. Thus, method should just be deprecated, to not result in a BC break.

src/CommonITILValidation.php
```diff
@@ -168,22 +168,71 @@ public function canDeleteItem()
...
+     * @return bool
+     */
+    public function isCurrentUserValidationTarget(bool $include_groups = true): bool
+    {
+        $is_target = false;
+        //TODO Remove BC-fix code
+        if (!isset($this->fields['itemtype_target'])) {
+            $this->fields['itemtype_target'] = 'User';
+            $this->fields['items_id_target'] = $this->fields['users_id_validate'];
+        }
```
cedric-anne, line 199: Is this still needed ?

cconard96: The function is public so I can't say how it is used outside the core. The note is for a future version when adding a breaking change is OK.

src/CommonITILValidation.php
cedric-anne: I removed the deprecation state and update the validator types. It is actually used in ticket creation form:
image

cedric-anne (approved):

cedric-anne (approved):

PR #11103, opened 2022-03-24
Fix missing users_id value for KB item

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

The hidden field used for the `users_id` value was missing. It looks like it was accidentally removed when KB categories were reworked because the hidden field was placed between the Category header and the Category field.

### DIFF
--- a/src/KnowbaseItem.php
+++ b/src/KnowbaseItem.php
@@ -817,7 +817,7 @@ class KnowbaseItem extends CommonDBVisible implements ExtraVisibilityCriteria
         echo "<tr class='tab_bg_1'>";
         echo "<td colspan=2></td>";
         echo "<td>";
-        echo "<input type='hidden' name='users_id' value=\"".Session::getLoginUserID()."\">";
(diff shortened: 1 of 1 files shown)


trasher (approved):

src/KnowbaseItem.php
```diff
@@ -817,6 +817,7 @@ public function showForm($ID, array $options = [])
         echo "<tr class='tab_bg_1'>";
         echo "<td colspan=2></td>";
         echo "<td>";
+        echo "<input type='hidden' name='users_id' value=\"".Session::getLoginUserID()."\">";
```
trasher, line 820:
```suggestion
        echo "<input type='hidden' name='users_id' value=\"" . Session::getLoginUserID() . "\">";
```

orthagh (approved):

PR #11118, opened 2022-03-28
Command line inventory

trasher:
Usage:

```
php front/inventory.php < /path/to/inventory.xml
```

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

### DIFF
--- a/front/inventory.php
+++ b/front/inventory.php
@@ -34,7 +34,7 @@
 use Glpi\Inventory\Request;
 
 if (!defined('GLPI_ROOT')) {
-    include('../inc/includes.php');
+    include(__DIR__ . '/../inc/includes.php');
 }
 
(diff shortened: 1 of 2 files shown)


cedric-anne (approved):

front/inventory.php
```diff
@@ -81,6 +90,9 @@
     $redirect_url = $refused->handleInventoryRequest($inventory_request);
     Html::redirect($redirect_url);
 } else {
+    if (isCommandLine()) {
```
g-bougard, line 93: I think you should put this check before `if (isset($_GET['refused'])) {` as this later check as no sens while using commandline import.

PR #11133, opened 2022-03-29
don't check right when updating CommonDBRelation on inventory process

orthagh:
ex update of OS

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | https://forum.glpi-project.org/viewtopic.php?pid=490556

- update of OS done here https://github.com/glpi-project/glpi/blob/ab8b4df675c0d1b188e5669de484e32f611f1318/src/Inventory/Asset/OperatingSystem.php#L136
- `CommonRelation::prepareInputForUpdate` called with a check on rights : https://github.com/glpi-project/glpi/blob/ab8b4df675c0d1b188e5669de484e32f611f1318/src/CommonDBRelation.php#L766-L770
- as we don't have any rights when doing inventories, false is returned, OS is not updated

Maybe we can up more, by adding return true to `CommonDBTM::can` and `Session::haveRight` when `Session::isIventory` return true also, but it seems too much

### DIFF
--- a/src/CommonDBConnexity.php
+++ b/src/CommonDBConnexity.php
@@ -299,7 +299,7 @@ abstract class CommonDBConnexity extends CommonDBTM
             }
         }
 
-        if ($have_to_check) {
+        if ($have_to_check && !Session::isInventory()) {
             $new_item = clone $this;
 
            // Solution 1 : If we cannot create the new item or delete the old item,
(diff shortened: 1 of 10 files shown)


trasher (approved):

trasher (approved):

orthagh:
OK for the last additions

PR #11200, opened 2022-04-06
Add separate right for impersonation

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

Adds a new User right for impersonation.
The new permission only allows users whose current profile is higher or equally privileged as every profile of the other user. For example, a technician with this right could impersonation self-service users but not admins. This is mainly to avoid privilege escalation.

If user's have the update permission for the config right, then they can impersonate anyone which is the previous behavior.

### DIFF
--- a/install/migrations/update_10.0.x_to_10.1.0/impersonate_right.php
+++ b/install/migrations/update_10.0.x_to_10.1.0/impersonate_right.php
@@ -1,0 +1,41 @@
+<?php
+
+/**
+ * ---------------------------------------------------------------------
+ *
+ * GLPI - Gestionnaire Libre de Parc Informatique
+ *
+ * http://glpi-project.org
+ *
+ * @copyright 2015-2022 Teclib' and contributors.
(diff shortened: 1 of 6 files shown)


cedric-anne (approved):

install/migrations/update_10.0.x_to_10.1.0/impersonate_right.php
```diff
@@ -0,0 +1,43 @@
...
+if (!defined('GLPI_ROOT')) {
+    die("Sorry. You can't access this file directly");
+}
+
+/**
+ * @var DB $DB
+ * @var Migration $migration
+ */
+
+$migration->addRight('user', [User::IMPERSONATE]);
```
orthagh, line 43:
```suggestion
$migration->addRight('user', [User::IMPERSONATE], ['config' => UPDATE]);
```

trasher (approved):

PR #11224, opened 2022-04-08
feat(users): creation

Rom1-B:
Allow to create a user from a ticket in creation:
- simplified data (name, first name, email, phone)
- entity = ticket entity
- default profile

Allow to create a user from an email sent to the receiver:
- login & name = email
- entity = ticket entity
- default profile

For both, send a password initialization email to the new user.

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | gitlab>erp-orders>425


trasher:
This won't be included in GLPI 10, therefore, you cannpot use 9.5.X-to-10.0 migration files. Another PR already exists that provide the new files for reference.

src/User.php
```diff
@@ -6468,4 +6577,44 @@ public static function getUserByForgottenPasswordToken(string $token): ?User
...
+
+        if (count($iterator) > 0) {
+            return false;
+        }
+
+        $user = new self();
+        $user->add([
+            'name'           => $email,
+            'realname'       => $email,
+            'password'       => Toolbox::getRandomString(16),
```
cedric-anne, line 6608: You should probably move the randomization of the password in the `prepareInputForAdd` method. This would permit to use this feature from API.

src/User.php
```diff
@@ -4990,6 +5036,10 @@ public static function getOrImportByEmail($email = '')
                     }
                 }
             }
+            if ($createuserfromemail) {
```
cedric-anne, line 5039: IMHO, this logic may produce unexpected result when `is_users_auto_add` is set to false. Indeed, user will not be imported from LDAP, but will be created anyway, and maybe with a login that differs from the LDAP value. I do not really know how it will be handled after that in LDAP sync process, but I guess it could result in having 2 users with the same email, the first created from mail collector and the second created from LDAP sync.

front/initpassword.php
```diff
@@ -40,6 +40,11 @@
         ['itemtype' => 'User', 'event' => 'passwordinit', 'is_active' => 1]
     )
 ) {
+    Session::addMessageAfterRedirect(
```
cedric-anne, line 43: This will not display the message, as user is not redirected.

src/User.php
```diff
@@ -4943,7 +4991,7 @@ public function showItems($tech)
      *
      * @return integer ID of user, 0 if not found nor imported
      */
-    public static function getOrImportByEmail($email = '')
+    public static function getOrImportByEmail($email = '', $createuserfromemail = false)
```
cedric-anne, line 4994: I am still not confident with conflicts between `$CFG_GLPI['is_users_auto_add']` and `create_user_from_email` collector configuration.

On previous implementation, a user that exists in LDAP would have been created without using LDAP informations, and I guess it would have lead to unexpected sync issues.

On current implementation, `create_user_from_email` collector configuration will not result in user creation if `$CFG_GLPI["is_users_auto_add"]` is set to false.

So, IMHO, if `create_user_from_email` is set to true, we should use LDAP information if they are available, regardless of `$CFG_GLPI["is_users_auto_add"]`  value, but I would like a confirmation from either @orthagh or @flegastelois . This could be done by using LDAP import using `if ($CFG_GLPI["is_users_auto_add"] || $createuserfromemail)` condition.

cedric-anne: We discuss about it with @orthagh , and we choose to use your first proposal.
It may generate duplicated users, so we decided to add a tooltip to informa administrators about this.

Changes done on 375398a561fe7b82d02b28fa58c6f262d254e4a6 .

cedric-anne (approved):

trasher (approved):

templates/components/user/create_user.html.twig
```diff
@@ -0,0 +1,43 @@
...
+ # MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ # GNU General Public License for more details.
+ #
+ # You should have received a copy of the GNU General Public License
+ # along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ #
+ # ---------------------------------------------------------------------
+ #}
+
+<button type="button"
```
orthagh, line 34: Current design is not ideal i think but i don't have ideas about it currently.

Some suggestions despite that:
- move to `ti ti-user-plus` icon
- a single button for all actors could be better no ?
- do we have any way to disable the display of the button

orthagh: maybe something like that if there is no conflict event with accordion

image

cedric-anne: > maybe something like that if there is no conflict event with accordion

:+1: I prefer this design.

Rom1-B: Button moved, but I can't manage not to trigger the folding/unfolding of the menu.

cedric-anne: > I can't manage not to trigger the folding/unfolding of the menu.

I fixed it.

Rom1-B: > I fixed it.

Thank you

orthagh (approved):

PR #11267, opened 2022-04-11
Mini dashboard fixes

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

Fixed issue where the second row of the mini dashboard was non-functional and only worked to fit the display of cards two rows tall.
There is also an issue with dragging cards to the left but I haven't determined the cause. The dragging seems to be handled by jQuery UI in GridStack but it also seems to work properly for full size dashboards so I can't tell if the issue in on our end or not. In some cases when moving a card to an empty space on the left there are areas that don't get highlighted as if they are not valid drop areas.

### DIFF
--- a/js/dashboard.js
+++ b/js/dashboard.js
@@ -94,7 +94,7 @@ var Dashboard = {
 
         Dashboard.grid = GridStack.init({
             column: options.cols,
-            maxRow: (options.rows + 1), // +1 for a hidden item at bottom (to fix height)
+            maxRow: options.rows,
             margin : this.cell_margin,
             float: true, // widget can be placed anywhere on the grid, not only on top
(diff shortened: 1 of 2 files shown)


cconard96:
Nevermind, second row is broken again...
Cannot resize or move a card into the second row again.

dxdemetriou:
> Nevermind, second row is broken again... Cannot resize or move a card into the second row again.
It seems that's only updated GLPI from 9.x to 10.x are affected.

I tried those fixes and I could resize & move everything on GLPI 10.0.2 upgraded from 9.5.8

On GLPI 10.0.0 fresh install updated to 10.0.2 it worked without any change

PR #11356, opened 2022-04-22
feat(ticket): entity notes

Rom1-B:
Display the entity's notes in a ticket

image

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | gitlab>erp-orders>427


src/Notepad.php
```diff
@@ -416,4 +416,46 @@ public static function showForItem(CommonDBTM $item, $withtemplate = 0)
         }
         return true;
     }
+
+    public static function showNotes(CommonDBTM $item)
```
cedric-anne, line 420: I guess it could be possible to replace this whole method by a template file that could be called like this:
```
{{ include('components/notepad/notes.html.twig', {'item': item}, with_context = false) }}
```

templates/components/notepad/notes.html.twig
```diff
@@ -0,0 +1,70 @@
...
+                        user.fields['realname'],
+                        user.fields['firstname']
+                    ]) %}
+                {% endset %}
+                <span class="avatar avatar-sm rounded" style="{{ style }}" title="{{ user_name }}">
+                    {% if thumbnail_url is empty %}
+                        {{ user.getUserInitials() }}
+                    {% endif %}
+                </span>
+            </a>
```
cedric-anne, line 62: I guess it could be replaced by
```{{ include('components/user/picture.html.twig', {'users_id': note['users_id_lastupdater']}, with_context = false) }}```

File: src/Ticket.php [hunk 3]
```diff
@@ -4415,6 +4415,7 @@ public function showForm($ID, array $options = [])
             'canassigntome'      => $canassigntome,
             'load_kb_sol'        => $options['load_kb_sol'] ?? 0,
             'userentities'       => $userentities,
+            'notepad'            => new Notepad(),
```
cedric-anne, line 4418:
```suggestion
```

templates/components/notepad/notes.html.twig
```diff
@@ -0,0 +1,53 @@
...
+ #
+ # You should have received a copy of the GNU General Public License
+ # along with GLPI. If not, see <http://www.gnu.org/licenses/>.
+ # ---------------------------------------------------------------------
+ #}
+
+{% set rand = random() %}
+{% set entity = get_item('Entity', item.fields['entities_id']) %}
+
+{% set notes = notepad.getAllForItem(entity) %}
```
cedric-anne, line 35: No need to pass an instance of item to call a static method.
```suggestion
{% set notes = call('Notepad::getAllForItem', [entity]) %}
```

cedric-anne (changes requested):

cedric-anne (commented):
1. When there is no notes on entity, having an empty accordion item it disturbing. I propose to add following text: `There is no notes on this entity.`.

2. Rendering can be improved. You reused the notes styles, but IMHO, the notes background style does not go with ticket form styles.

3. Rendering is different on existing ticket (see image below). Indeed, there is no spacing between avatar and text.
image

4. Only users with `READNOTE` right on entity should be able to see these notes.

5. I do not know what kind of information people are pushing into entity notes, but I am not sure that every existing entities notes could be safely published on ticket form. Also, they may contains information that are not relevant in ticket form context. Maybe we should add a configuration option for this.

orthagh:
1. I agree, no notes -> no accordion item. It also hide the feature for those who don't need it

2. cleary proposed UI is dated: i suggest card and card footer to display contextual information like user, ex:
image
see: https://preview.tabler.io/tasks.html

5. agree, a flag above entity notes to publish them into ticket could be appreciated

orthagh (changes requested):
Last thing missing i think is to not display the accordion item when no notes is available

orthagh (approved):

trasher (approved):

cedric-anne (changes requested):

cedric-anne (commented):
The visibility switch UX may be enhanced.
1. I first though that it was usable on existing notes, without having to enter "edit" mode. At least the "pointer" mouse cursor, and maybe the tooltip too, should be removed when it is disabled.
2. The contrast is really low, and, for someone like me, it is really uncomfortable.
image

PR #11389, opened 2022-04-26
display tab spinner at a better position

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #11387

### DIFF
--- a/src/Ajax.php
+++ b/src/Ajax.php
@@ -362,7 +362,7 @@ JAVASCRIPT;
             if ($(target).html() && !force_reload) {
                 return;
             }
-            $(target).html('<i class=\"fas fa-3x fa-spinner fa-pulse position-absolute top-50 start-50\"></i>');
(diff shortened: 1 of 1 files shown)


trasher (approved):

derry43:
The fix below did not really place the spinner where I would have expected.
```
            $(target).html('<i class=\"fas fa-3x fa-spinner fa-pulse position-absolute m-5 start-50\"></i>');
```

The above really positioned the spinner towards the top right of the page rather than the middle of the viewport. I think the following actually positions the spinner correctly in the middle of the viewport (and it stays there, as expected, even if you scroll the page).

```
            $(target).html('<i class=\"fas fa-3x fa-spinner fa-pulse position-fixed top-50 start-50\"></i>');
```

PR #11415, opened 2022-04-28
avoid notices in ticket's stats tab

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

Avoid:

```
[2022-04-28 13:56:56] glpiphplog.NOTICE:   *** PHP Deprecated function (8192): strtotime(): Passing null to parameter #1 ($datetime) of type string is deprecated in .../src/Ticket.php at line 6285
```

### DIFF
--- a/src/Ticket.php
+++ b/src/Ticket.php
@@ -6280,12 +6280,12 @@ JAVASCRIPT;
         if ($date_takeintoaccount == $date_creation) {
             $date_takeintoaccount  = 0;
         }
-        $internal_time_to_own     = strtotime($this->fields['internal_time_to_own']);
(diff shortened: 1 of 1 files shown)


trasher (approved):

cedric-anne (commented):
Maybe logic could be simplified.

Here fields values are converted into timestamps, then in `showDatesTimelineGraph()` timestamps are converted back to strings, then these stings are formatted in twig template.

Maybe we could directly pass the field value into the `date` key, and fille this key in `showDatesTimelineGraph()` only if it is not already set.

With this, the template would just have to check `{% if data['date'] is not empty %}`

PR #11437, opened 2022-04-30
Fix password reset form

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #11426

The password reset form seems unique in that there is a user session but no active profile or entity. Parts of the simple header used are reliant on at least an active entity when they weren't before in GLPI 9.5 and older. This PR tries detecting when the user is logged in like this and show a similar interface as before (`Html::simpleHeader`).

For example when needing to reset the password:
Screenshot from 2022-04-29 21-09-04

Reference to previous `Html::simpleHeader` implementation:
https://github.com/glpi-project/glpi/blame/9.5/bugfixes/inc/html.class.php#L1844

### DIFF
--- a/templates/layout/parts/page_header.html.twig
+++ b/templates/layout/parts/page_header.html.twig
@@ -29,6 +29,8 @@
  # ---------------------------------------------------------------------
  #}
 
+{% set anonymous = session('glpiactiveprofile') is null %}
+
 {% set is_vertical = user_pref('page_layout') == 'vertical' %}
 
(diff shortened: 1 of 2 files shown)


cconard96:
As a note, the username is actually pre-filled in the password reset form. It just doesn't show up in the screenshot because of issues with Chrome's internal stylesheet and auto-filled text. The color is changed to black so it doesn't show when using the midnight palette.

trasher (approved):

PR #11480, opened 2022-05-04
feat(DCbreadcrumb): add capability to display location (and use it on Cable)

stonebuzz:
Add capability to display location from ```DCBreadcrumb```

(and use it from ```Cable``` object)

Before : 
image

After : 
image

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #number


orthagh:
Did you test with very long location name ?
if not, maybe some text-overflow classes should be added

stonebuzz:
> text-overflow

(without change for the moment)

image

stonebuzz:
> Did you test with very long location name ? if not, maybe some text-overflow classes should be added

Maybe better with ```text-truncate```

image

btry:
Hi

Interesting feature. After viewing the rendering, I got an idea. The breadcrumb consumes 2 lines very fast, without the need of long strings. Then why not show the locations in the 2nd line directly, horizontally aligned with the matching breadcrumb item ?

Something like that : 
```preformatted 
General > Datacenter > Server room > Rack
          Marseille >  Marseille >   Marseille

```

cconard96:
Realistically, how often is a rack going to have a different location than its server room or a server room with a different location than its datacenter?

I think some optimization can be done with that. If a breadcrumb item has the same location as the one before it, there shouldn't be a need to display the location again. If a breadcrumb item in in a child location of the item before it, only the difference in the complete name needs shown. If a breadcrumb item has no location, the badge can be hidden too so it appears that the location is inherited.
If there is some weird case where the location of two breadcrumb items are not in the same hierarchy, then the full location can be shown I guess.

I'm not sure how complicated that would be to implement though.

stonebuzz:
> Hi
> 
> Interesting feature. After viewing the rendering, I got an idea. The breadcrumb consumes 2 lines very fast, without the need of long strings. Then why not show the locations in the 2nd line directly, horizontally aligned with the matching breadcrumb item ?
> 
> Something like that :
> 
> ```
> General > Datacenter > Server room > Rack
>           Marseille >  Marseille >   Marseille
> ```
> 
> or
> 
> ```
> General > Datacenter > Server room > Rack
>           Marseille  > Marseille   > Marseille
> ```

new version 

image

orthagh:
How the breadcrumb behaves when there is no location for an item ?

stonebuzz:
> Realistically, how often is a rack going to have a different location than its server room or a server room with a different location than its datacenter?
> 
> I think some optimization can be done with that. If a breadcrumb item has the same location as the one before it, there shouldn't be a need to display the location again. If a breadcrumb item in in a child location of the item before it, only the difference in the complete name needs shown. If a breadcrumb item has no location, the badge can be hidden too so it appears that the location is inherited. If there is some weird case where the location of two breadcrumb items are not in the same hierarchy, then the full location can be shown I guess.
> 
> I'm not sure how complicated that would be to implement though.

@cconard96 

Some examples:

GLPI location tree

image

Datacenter -> ```A```
DCRoom -> ```A > A1```
Rack -> ```B > B1```

Breadcrumb

image



Datacenter -> ```A```
DCRoom -> ```A```
Rack -> ```A1```

Breadcrumb 

image

stonebuzz:
> How the breadcrumb behaves when there is no location for an item ?

image

stonebuzz:
> It may be good to prepend child locations with "> " to differentiate them from completely separate locations.

Like this

image

src/Features/DCBreadcrumb.php
```diff
@@ -171,10 +262,32 @@ private function getItemEnclosurePosition($itemtype, $items_id)
...
 
+    /**
+     * get item location from Enclosure
+     *
+     * @param string  $itemtype Item type
+     * @param integer $items_id Item ID
+     *
+     * @return string
+     */
+    private function getItemLocation($itemtype, $items_id)
```
orthagh, line 277: a suggestion here, this method seems not related to DCBreadcrumb class and doesn't require any `$this`.
I suggest to move it to Location class as private method.
Like `Location::getForItem(CommonDBTM $item)`  (avoid itemtype, items_id parameter when you the item initialy)

templates/layout/parts/dcbreadcrumbs.html.twig
```diff
@@ -0,0 +1,56 @@
...
+         <div class="col-auto p-1">
+            >
+         </div>
+      {% else %}
+         <div class="col-auto pt-1 p-0">
+            {{ breadcrumb_item['position']|raw }}
+         </div>
+      {% endif %}
+   {% endfor %}
+</div>
```
orthagh, line 56:
```suggestion
</div>

```

src/Features/DCBreadcrumb.php
```diff
@@ -111,11 +142,71 @@ public function getDcBreadcrumb()
...
+        $breadcrumb = array_reverse($breadcrumb);
+
+        if ($with_location) {
+            $breadcrumb = $this->cleanLocationInforamtions($breadcrumb);
+        }
+
+        return $breadcrumb;
+    }
+
+    public function cleanLocationInforamtions(&$breadcrumb)
```
orthagh, line 168: a dockblock could help

src/Location.php
```diff
@@ -399,6 +399,19 @@ public static function displayTabContentForItem(CommonGLPI $item, $tabnum = 1, $
...
     }
 
+    /**
+     * get item location
+     *
+     * @param CommonDBTM  $item
+     *
+     * @return CommonDBTM
+     **/
+    public static function getItemLocation(CommonDBTM $item)
```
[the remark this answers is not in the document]
stonebuzz: done

src/Location.php
```diff
@@ -399,6 +399,23 @@ public static function displayTabContentForItem(CommonGLPI $item, $tabnum = 1, $
         return true;
     }
 
+    /**
+     * get item location
+     *
+     * @param CommonDBTM  $item
+     *
+     * @return Location if succeed else false
```
orthagh, line 407:
```suggestion
     * @return Location|false
```

orthagh (approved):

trasher:
Please fix tests ;)

stonebuzz:
Ex (from Computer side ) with small location name

image

With long location name

image

templates/layout/parts/dcbreadcrumbs.html.twig
```diff
@@ -0,0 +1,56 @@
...
+ #}
+<div class="row">
+   {% for breadcrumb_item in breadcrumbs %}
+
+      <div class="col-auto p-1">
+         {{ breadcrumb_item['link']|raw }}
+
+         {% if breadcrumb_item['location']['name'] is defined %}
+            <br>
+            <span class='p-0 float-left badge bg-blue-lt d-inline-block text-truncate' data-toggle='tooltip' title='{{ breadcrumb_item['location']['name'] }}' style='max-width: 100px;'>
```
orthagh, line 41:
```suggestion
            <span class='p-0 float-left badge bg-blue-lt d-inline-block text-truncate' data-bs-toggle='tooltip' title='{{ breadcrumb_item['location']['name'] }}' style='max-width: 100px;'>
```

orthagh (approved):

src/Features/DCBreadcrumb.php
```diff
@@ -111,11 +141,85 @@ public function getDcBreadcrumb()
...
+     * Clean Location information foreach item breadcrumb
+     * If Current item Location = Parent item Location              -> do not display Location
+     * If Current item Location is child of Parent item Location    -> display only Location name instead of completename
+     * Else Current item Location != Parent item Location           -> display Location completename
+     *
+     * @param array $breadcrumb
+     *
+     * @return array cleanded $breadcrumb
+     */
+    public function cleanLocationInformations(&$breadcrumb)
```
cedric-anne, line 177: 1. Your return the updated array, so no need to use a reference.
2. Method should be private as it is used only here.
```suggestion
    private function cleanLocationInformations(array $breadcrumb)
```

cedric-anne (changes requested):
Functionnaly, it seems OK, but technically, it is not OK for me. Indeed, you changed existing public methods signature, and, even if they does not seems to be used in plugins, it is preferable to avoid such changes.

I will adapt the code and fix tests.

cedric-anne (approved):

orthagh (approved):

PR #11484, opened 2022-05-04
remove gantt feature from core

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

Related to #11475, dhtmlx/gantt is not compatible anymore with GLPI license.
Moving to another library requires too much work.
Just separate the feature into a dedicated plugin do the trick and may help also maintenance of both projects

Here is the plugin : https://github.com/pluginsGLPI/gantt

Todo:

- [x] insert a link into projects context for global gantt
- [x] `project.show_on_global_gantt` field


cconard96:
The minimum version for the plugin could be changed to 10.0.1 to avoid issues with users on 10.0.0 trying to install and use it while the feature is still integrated.

trasher (approved):

orthagh:
To note, 
Project|ProjectTask::getDataToDisplayOnGantt seems to be leftovers of previous gantt implementation.
I didn't find any call to these methods.
If someone can confirm

cconard96:
`show_on_global_gantt` has no affect on the Kanban.
The Kanban feature shows everything on the global project Kanban with one exception; Tasks and child projects of a project that has a status that is considered finished are hidden to avoid a lot of clutter.

I think the flag can be removed from the core entirely. If the behavior of the Kanban needs to change, it could be done in a later version.

orthagh:
the idea (and i wanted to ask you) is to make this field more generic for both views.
If we remove it, i need to manage it in the plugin and i'm not happy about it (to have DB management for just one field).

The question i wanted to ask, is how much is difficult to get kanban filtered by this field (and can i do it myself in a reasonable time)

cconard96:
It should be as simple as adding the `show_on_global_view` to the criteria here:
https://github.com/glpi-project/glpi/blob/ef727f996f7ad5dcbe903d1c9d3381d84fde349e/src/Project.php#L2083-L2097

This would leave the hidden filter in place of hiding child projects and tasks of finished projects.

I'm just not sure how to handle this with a bugfix version as it would change how the global Kanban is displayed for users especially since the field was set to no by default.

cconard96:
I guess in the form display the field can be conditionally shown based on `Plugin::isActivated('gantt')` and remove reference to the Kanban for now. This is the only case I found where the core checks the existence of a specific plugin for anything, but it would avoid adding more breaking changes than needed for a bugfix version.

cconard96 (approved):

PR #11500, opened 2022-05-05
fix(export PDF): html entities

Rom1-B:
In a PDF export, the HTML entities were shown instead of the special characters.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !23945


trasher:
We're going to change PDF library; see #11488. Note that I've not reproduced the original issue with TCPDF.

PR #11504, opened 2022-05-05
Fix strtotime deprecation messages in Ticket stats timeline

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

In PHP 8.1, passing null to `strtotime` is deprecated. If a date field is not set, it is null by default. The solution is to use `?? ''` as an invalid date string will return false still. The only change is that the argument must be a string.

### DIFF
--- a/src/Ticket.php
+++ b/src/Ticket.php
@@ -6275,17 +6275,17 @@ JAVASCRIPT;
     public function showStatsDates()
     {
         $now                      = time();
-        $date_creation            = strtotime($this->fields['date']);
+        $date_creation            = strtotime($this->fields['date'] ?? '');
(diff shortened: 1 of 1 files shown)


trasher (approved):

orthagh:
duplicate of #11415. I didn't have time to continue for now. You can take over if you want.
@cedric-anne is right when he says that we do multiple conversion for nothing.

trasher:
Current PR fix issues with php 8.1; rewriting may be done in another one.

Issue #11492, opened 2022-05-05
GLPI 10.0.0 - the options bar displaying problem


WebGregGit:
### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

10.0.0

### Bug description

Top bar with options (Actions button and to the right of it) and the bottom bar (rows/page, showing x to y of z) are pinned if the page scrolls vertically. If there are more columns and you scroll horizontally - these options are also scrolled.

### Page URL

front/computer.php

### Steps To reproduce

1. Add a lot of columns so that they do not fit on the screen.
2. Scroll horizontally the page to see the last columns that are currently outside the visible area.
3. The options bar will also scroll.

### Anything else?

![Schowek02](https://user-images.githubusercontent.com/68922721/166892836-6458ad77-cf33-4426-96ac-6d4be608bd7d.jpg)

Additionally, the bottom bar is aligned to the width of the columns. By adding more columns you make the data "Showing x to y of z" go further to the right, until it overlaps the navigation buttons:

![Schowek05](https://user-images.githubusercontent.com/68922721/166893492-17d6ab52-eb7b-43ba-b49e-a9e6e56b7757.jpg)

cconard96:
Can you test #11506?

PR #11506, opened 2022-05-06
Fix search table horizontal scroll

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #11492

Instead of the table scrolling horizontally when there are more columns than can be displayed at once, the parent card element was scrolling instead.

### DIFF
--- a/css/includes/pages/_search.scss
+++ b/css/includes/pages/_search.scss
@@ -90,7 +90,6 @@
         .search-card {
             @include media-breakpoint-up(sm) {
                 min-width: 100%;
-                width: min-content;
 
                 .search-header {
                     border-bottom: 0;
--- a/templates/components/search/table.html.twig
+++ b/templates/components/search/table.html.twig
@@ -31,7 +31,7 @@
 
 {% set searchform_id = data['searchform_id']|default('search_' ~ rand) %}
 
-<div class="table-responsive-md">
+<div class="table-responsive">
    <table class="search-results table card-table table-hover {{ data['search']['is_deleted'] ? "table-danger deleted-results" : "table-striped" }}"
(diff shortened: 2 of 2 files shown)


trasher (approved):

orthagh (approved):

cedric-anne (approved):

orthagh:
@cconard96 The pr seems to add an empty space inside the thead

image

cconard96:
> @cconard96 The pr seems to add an empty space inside the thead
> 
> image

There are conflicts with a responsive table and the sticky table headers. I think this was noticeable now because I changed the table-responsive class used.

Sticky elements cannot work properly if any parent element has any overflow value other than visible. Even if the header x position remains the same and only sticks vertically, you cannot have overflow-x set which is done when the table is responsive.
https://github.com/w3c/csswg-drafts/issues/865
I've been looking at workarounds for a while now.

orthagh:
ok i think i get it

if we something like 

.search-container > table > tr > th directly, the sticky part seems to be ok.
I think as we have like 4 or 5 intermediate blocks (div, form, etc), it failed to apply correctly.

i don't how to reorganize correctly the form, the container and the table.
but i'm sure we can remove the table-responsive (and move the properties to search-container) at least

orthagh:
ok better : just removing the .table-responsive (and move the table just after the card header) do the trick
We still have the responsive done by `.search-container`

The issue seems to be we have two place where overflow-x applied

cconard96:
Horizontal scrolling doesn't seem to work again after that.

Issue #11405, opened 2022-04-27
10.0 : GLPI+Agent  lots of files *.xml files are generated in /var/www/html/glpi/files/_inventories.


steph461:
### Code of Conduct

- [ ] I agree to follow this project's Code of Conduct

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

10.0

### Bug description

When I launch my glpi agent, temp *.xml files are generated in /var/www/html/glpi/files/_inventories.

I had lots of *.xml files (4000 files)

Is this wanted ?

every night I do a cron rm *.xml --force

exemple : 
![image](https://user-images.githubusercontent.com/68054991/165507177-08271ed7-8b77-45b9-885d-e7afc350df2d.png)

### Relevant log output

```shell
no log
```

### Your GLPI setup information

1. start agent with task

steph461:
in the file  example :  same details on all files

<?xml version="1.0" encoding="UTF-8"?>
<REQUEST>
  <CONTENT>
    <AGENT>
      <AGENTVERSION>1.2</AGENTVERSION>
      <START>1</START>
    </AGENT>
    <MODULEVERSION>5.1</MODULEVERSION>
    <PROCESSNUMBER>235948</PROCESSNUMBER>
  </CONTENT>
  <DEVICEID>xxxxxxxxxxxxxxxxxxxxxxxxxxxxxx-2022-02-16-12-47-12</DEVICEID>
  <QUERY>SNMPQUERY</QUERY>
</REQUEST>

trasher:
Duplicates #10356

g-bougard:
Hi @steph461 
you can give #11518 a try. It should really fix this issue or just enable the cleantmp automatic action.

steph461:
its ok , i see the delete file automatically

Issue #10356, opened 2022-01-18
Temporary files not removed from `files/_inventories`


g-bougard:
### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

10-beta

### Bug description

I found 2 cases where a file is kept in `files/_inventories` but it shouldn't be.

1. First case is when a json is not respecting JSON inventory spec.
2. The second case is related to netdiscovery and netinventory tasks support with **GlpiInventory** plugin. These tasks are generating state message for the plugin that are finally kept in temporary file.

### Steps To reproduce

1. For the first case, here is sample you can try to inject in server with `glpi-injector` even if no related asset is known:
```
{
   "action": "inventory",
   "content": {
      "bios": {
         "bmanufacturer": "Apple Inc.",
         "bversion": "IM112.88Z.0057.B00.1005031455",
         "mmanufacturer": "Apple Inc.",
         "mmodel": "Mac-F2238AC8",
         "msn": "C020XXXXXXXXCJN1M",
         "smanufacturer": "Apple Inc.",
         "smodel": "iMac11,2",
         "ssn": "WXXXXMWXXXX"
      },
      "hardware": {
         "chassis_type": "All in One",
         "dns": "127.0.0.53",
         "name": "xxxx",
         "uuid": "ba56962a-xxxx-xxxx-xxxx-a58fb09d46cb",
         "vmsystem": "Physical"
      },
      "remote_mgmt": [
         {
            "id": 1043952882,
            "type": "teamviewer"
         }
      ],
      "versionclient": "GLPI-Inventory_v1.0-1"
   },
   "deviceid": "xxxxxxxx-2022-01-18-09-03-56",
   "itemtype": "Computer",
   "partial": true
}
```
This file was generated by a still fixed bug in GLPI-Agent 1.0. But the problem is a temporary file with that content will remain in `files/_inventories`.

2. For the second case, `netdiscovery` and `netinventory` tasks generates job state messages for the plugin like the following:
```
<?xml version="1.0" encoding="UTF-8" ?>
<REQUEST>
  <CONTENT>
    <AGENT>
      <START>1</START>
    </AGENT>
    <MODULEVERSION>4.4</MODULEVERSION>
    <PROCESSNUMBER>500</PROCESSNUMBER>
  </CONTENT>
  <DEVICEID>xxxxxxxx-2021-11-30-09-57-34</DEVICEID>
[...]

PR #11518, opened 2022-05-06
Delete temporary inventory file via destructor

g-bougard:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #10356,#11405

This implements the deletion of temporary inventory file using Inventory object destructor.

I tested this patch with the case reported in #10356 and it worked as expected.


g-bougard:
Replaced by #11550

PR #11523, opened 2022-05-09
make search header/footer components sticking horinzontally

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

Follow #11521 

Last thing i need (try) to fix, is `.page-infos` div scrolling due to the `.justify-content-between` class of its parent.
It's annonying, but not blocking i think, i contained it at 250px left/right

@cconard96 Do you see any remaining bugs  ?

### DIFF
--- a/css/includes/pages/_search.scss
+++ b/css/includes/pages/_search.scss
@@ -100,6 +100,11 @@
                     top: -1px;
                     background-color: inherit;
                     z-index: #{$zindex-sticky + 1};
+
+                    .search-controls {
+                        position: sticky;
+                        left: 22px;
+                    }
                 }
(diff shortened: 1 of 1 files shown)


cconard96:
I does seem to work properly with these new CSS rules apart from the last issue you noted.

cconard96 (approved):

orthagh:
Cannot do better.
So pr is ready for merge

trasher (approved):

cconard96 (approved):

PR #11524, opened 2022-05-09
feat(dashboard label): set min font-size

stonebuzz:
Set min ```font-size``` for ```big-number``` card

Useful for single  ```big-number``` cards

Before 
image

After
image

| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #number

### DIFF
--- a/js/dashboard.js
+++ b/js/dashboard.js
@@ -791,7 +791,7 @@ var Dashboard = {
 
         parent_item
             .find('.big-number')
-            .find('.label').fitText(text_offset - 0.2);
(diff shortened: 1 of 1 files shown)


orthagh (approved):
I agree with the change, having a minimum size is good.
But font size on dashboard cards are an old topic with at least 3 or 4 pr before

So check all the possiblee cases please

AdrienClairembault (approved):

PR #11538, opened 2022-05-10
Add global locks

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

### DIFF
--- a/src/Inventory/Inventory.php
+++ b/src/Inventory/Inventory.php
@@ -493,7 +493,9 @@ class Inventory
                 'icon'  => Lockedfield::getIcon(),
                 'title' => Lockedfield::getTypeName(Session::getPluralNumber()),
                 'page'  => Lockedfield::getSearchURL(false),
-                'links' => $links
+                'links' => [
+                    "<i class=\"ti ti-plus\" title=\"" . __('Add global lock') . "\"></i><span class='d-none d-xxl-block'>" . __('Add global lock') . "</span>" => Lockedfield::getFormURL(false)
+                ] + $links
             ];
         }
 
--- a/src/Lockedfield.php
+++ b/src/Lockedfield.php
@@ -33,6 +33,8 @@
  * ---------------------------------------------------------------------
  */
 
+use Glpi\Application\View\TemplateRenderer;
+
 /**
  *  Locked fields for inventory
  **/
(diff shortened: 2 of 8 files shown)


src/Inventory/Inventory.php
```diff
@@ -479,7 +479,9 @@ public static function getMenuContent()
                 'icon'  => Lockedfield::getIcon(),
                 'title' => Lockedfield::getTypeName(Session::getPluralNumber()),
                 'page'  => Lockedfield::getSearchURL(false),
-                'links' => $links
+                'links' => [
+                    "<i class=\"ti ti-plus\" title=\"" . __('Add global lock') . "\"></i><span class='d-none d-xxl-block'>" . __('Add global lock') . "</span>" => Lockedfield::getFormURL()
```
orthagh, line 483:
```suggestion
                    "<i class=\"ti ti-plus\" title=\"" . __('Add global lock') . "\"></i><span class='d-none d-xxl-block'>" . __('Add global lock') . "</span>" => Lockedfield::getFormURL(false)
```

src/Lockedfield.php
```diff
@@ -221,6 +237,101 @@ public function getForbiddenStandardMassiveAction()
...
+            $input['is_global'] = 1;
+        }
+
+        return $input;
+    }
+
+    public function showForm($ID, array $options = [])
+    {
+        $this->initForm($ID, $options);
+        TemplateRenderer::getInstance()->display('pages/admin/inventory/lockedfield.html.twig', [
```
orthagh, line 269:
```suggestion
         unset($this->fields['is_global']);
         TemplateRenderer::getInstance()->display('pages/admin/inventory/lockedfield.html.twig', [
```

This to avoid the `is_global` field display (is_global is normally used by Monitor/Peripheral/Printer to have a generic item), as it make no sense in this context

orthagh (changes requested):
A quick suggestion as you are working on the lock field management.
On an asset, when a lock icon is displayed, maybe we can set it as a link for ease removal, no ?

trasher:
As discussed, it ould be quite difficult to add removal links the way lock icons are currently rendered.

orthagh (approved):

orthagh (approved):

PR #11540, opened 2022-05-10
avoid to set width on responsive, min-width is better

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

Follow #11521

Avoid empty space when not enough columns to fill 

image

### DIFF
--- a/css/includes/pages/_search.scss
+++ b/css/includes/pages/_search.scss
@@ -42,13 +42,13 @@
 }
 
 .search_page {
-    @include media-breakpoint-up(sm) {
+    @include media-breakpoint-up(lg) {
         display: flex;
         align-items: stretch;
     }
 
     .search-container {
-        @include media-breakpoint-up(sm) {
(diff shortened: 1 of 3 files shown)


trasher (approved):

AdrienClairembault (changes requested):
It seems there is still some issue with the sticky header:

image

Does not seem to be a layout specific issues, tested on both with same results

AdrienClairembault (approved):

orthagh:
@cconard96 your answer crossed my last fix.
Do you mind do your tests again ?

cconard96:
> @cconard96 your answer crossed my last fix. Do you mind do your tests again ?

I saw the new commit just after my comment. After testing again, it does seem to be working as expected for all screen sizes now.

cconard96 (approved):

PR #11563, opened 2022-05-11
clear timer when select a dropdown in logs filter

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

In history tab of an asset, after enabling logs filtering, if we select a value in the "Field" column, the filter action triggered twice:
- at selection
- after the timer (normally dedicated to text inputs)

### DIFF
--- a/js/log_filters.js
+++ b/js/log_filters.js
@@ -49,6 +49,8 @@ $(function() {
         }
     };
 
+    var delay_timer = null;
+
     var bindFilterChange = function () {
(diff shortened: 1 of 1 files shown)


trasher (approved):

g-bougard (approved):
Tested, thx.

cedric-anne (approved):

PR #11564, opened 2022-05-11
Add cartridge level search options

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !23397

Add search options for toner percentages which are stored in `glpi_printers_cartridgeinfos` and gathered by inventory agents.

This required an addition to the search engine since the data is not stored in a structured schema (just property => value pairs) and the use of additionalfields (which isn't great because it seems like it will grab all properties for the printer for every search option) would would cause extra values to display. I had to add an `aggregate` property to search options and change how the specific value display gets handled so that multiple values can be combined to one displayed value.

### DIFF
--- a/src/Printer_CartridgeInfo.php
+++ b/src/Printer_CartridgeInfo.php
@@ -261,6 +261,39 @@ class Printer_CartridgeInfo extends CommonDBChild
         return parent::getSpecificValueToSelect($field, $name, $values, $options);
     }
 
+    private static function getProgressColorsForColor($color)
+    {
+        $fg_transparency_hex = '80';
+        switch ($color) {
+            case 'cyan':
+            case 'light_cyan':
+                return [
+                    'fg' => '#00ffff'.$fg_transparency_hex,
+                    'text' => 'var(--tblr-body-color)'
+                ];
+            case 'magenta':
+            case 'light_magenta':
+                return [
+                    'fg' => '#ff00ff'.$fg_transparency_hex,
+                    'text' => 'var(--tblr-body-color)'
+                ];
+            case 'yellow':
+                return [
+                    'fg' => '#ffff00'.$fg_transparency_hex,
+                    'text' => 'var(--tblr-body-color)'
+                ];
+            case 'black':
+            case 'grey':
+            case 'darkgrey':
+                return [
+                    'fg' => '#303030'.$fg_transparency_hex,
+                    'text' => 'var(--tblr-body-color)'
+                ];
+        }
+
+        return null;
+    }
+
     public static function getSpecificValueToDisplay($field, $values, array $options = [])
     {
         $printer = new Printer();
@@ -299,7 +332,8 @@ class Printer_CartridgeInfo extends CommonDBChild
                         'percent' => $percent_remaining,
                         'message' => $percent_remaining.'%',
                         'display' => false,
-                        'create' => true
+                        'create' => true,
+                        'colors' => self::getProgressColorsForColor($color)
                     ]);
                 }
             }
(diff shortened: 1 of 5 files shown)


cconard96:
Screenshot from 2022-05-11 09-33-16

cconard96:
With colored progress bars and without placeholder NA values:

Selection_072

Selection_073

It may be better to review and merge this as-is, without the search capability, and then work on trying to make these fields searchable in another PR.

cconard96:
For reference, the search engine adds a `property` and `value` property to the select for every one of these virtual search options. The resulting values for each are from every single property/value pair for that printer. I don't fully understand the format (partly why making these fields searchable will be painful if possible at all).

Sample `property` column result:
```
tonerblackmax$#$13$$##$$tonerblackused$#$14$$##$$tonerblackremaining$#$15$$##$$tonercyanmax$#$16$$##$$tonercyanused$#$17$$##$$tonermagentamax$#$18$$##$$tonermagentaused$#$19$$##$$tonermagentaremaining$#$20$$##$$toneryellowmax$#$21$$##$$toneryellowused$#$22
```

Sample `value` column result:
```
1000$#$13$$##$$700$#$14$$##$$300$#$15$$##$$1000$#$16$$##$$654$#$17$$##$$1000$#$18$$##$$545$#$19$$##$$455$#$20$$##$$980$#$21$$##$$258$#$22
```

I don't think this odd result can be avoided given the table schema without a large rework/update to the search engine.

orthagh:
> It may be better to review and merge this as-is, without the search capability, and then work on trying to make these fields searchable in another PR.

I agree

orthagh (approved):

trasher (approved):

src/Printer_CartridgeInfo.php
```diff
@@ -115,4 +115,147 @@ public function showForPrinter(Printer $printer)
...
+    }
+
+    public static function getSpecificValueToDisplay($field, $values, array $options = [])
+    {
+        $color_aliases = [
+            'grey'      => 'gray',
+            'darkgrey'  => 'darkgray'
+        ];
+        $printer = new Printer();
+        if (strpos($field, '_virtual_toner') === 0) {
```
cedric-anne, line 205: You can use the PHP 8 function `str_starts_with`. Polyfill lib is loaded for PHP < 8.0.
```suggestion
        if (str_starts_with($field, '_virtual_toner')) {
```

src/Search.php
```diff
@@ -7281,56 +7281,91 @@ public static function giveItem(
...
+                        $field,
+                        $tmpdata,
+                        [
+                            'html'      => true,
+                            'searchopt' => $so,
+                            'raw_data'  => $data
+                        ]
+                    );
+
+                    $append_specific($specific, $tmpdata, $out);
```
cedric-anne, line 7341: I do not get why call this method, as it cannot do more than appending `$specific`. Thus, adding the `else if (isset($field_data['values']))` block would not be required if appending is done directly here.
```php
            if (!empty($specific)) {
                $out .= $specific;
            } else if (isset($field_data['values'])) {
                // Aggregate values; No special handling
                return;
            }
```

cconard96: If no specific value is returned from `getSpecificValueToDisplay`, there was fallback logic here that doesn't apply for aggregated search options. Confusingly, if this code is reached, it doesn't necessarily mean that the search option datatype was set to `specific`; only that it wasn't one of the other types. In fact, I don't see the `specific` datatype referenced at all in the search engine, but that isn't really important here.

The alternative is to check if the return from `getSpecificValueToDisplay` is empty before calling the function in the `if ($aggregate) {` block instead of having that logic inside the function itself.

cedric-anne (approved):

PR #11583, opened 2022-05-13
Fix null entity in self service form

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #10570 

The template that shows actor fields uses the entity from the item (or empty item is case of a new ticket). When using the self service form to create a ticket, the entity information is not set in the item fields (It wasn't in old versions either). This shouldn't have mattered since the entity is instead passed into the self service template in the `params` variable and then into the `main` actor template as `entities_id`. However, since the `main` actor template and child `field` template only checked the item fields, it was using a null value.

The solution is to set `entities_id` to the value in the item fields only if the `entities_id` variable isn't already set. Then in that template and `field`, use the `entities_id` variable instead of using the item field directly.

### DIFF
--- a/src/CommonITILObject.php
+++ b/src/CommonITILObject.php
@@ -408,6 +408,7 @@ abstract class CommonITILObject extends CommonDBTM
             'timeline_itemtypes'      => $this->getTimelineItemtypes(),
             'legacy_timeline_actions' => $this->getLegacyTimelineActionsHTML(),
             'params'                  => $options,
+            'entities_id'             => $ID ? $this->fields['entities_id'] : $options['entities_id'],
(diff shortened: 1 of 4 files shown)


trasher (approved):

cedric-anne (approved):

orthagh (approved):

soc10004:
@cconard96  @cedric-anne Thanks for the release, i apply this changes merged on my production version who is not Version 10 and i work... The error is solved. Great job

PR #11669, opened 2022-05-24
Improved theme support

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

Adds a Theme class to represent a theme/palette including the key, display name, if it is a dark theme or not, and if it was included by GLPI itself.

Core themes are statically defined in the new class. Any theme in the palette folder that is not defined as a core theme is considered custom. With custom themes, it tries to guess if it is a dark theme based on the file name. Custom themes were not technically documented/supported before, but they did work properly so I felt that some support to prevent breaking functionality was needed.

With this, it is now possible to know if the current theme is a dark theme or not. On the PHP side, there was no real way to know this without checking for specific palette names or parsing the stylesheet. On the client side, a hack was used to check the "--is-dark" CSS variable. It also was not possible for CSS stylesheets from plugins to have any logic based on if the current theme was dark or not as the SCSS variable `$is-dark` is not available to them.

Now, the current palette/theme is added as a data attribute on the "html" element along with a data attribute indicating if it is dark or not. JS code can now check the new data attribute instead of the root CSS, and plugin stylesheets can use the attribute in a selector.

In the palette selector, there are now cleaner display names for the core palettes thanks to being able to define it through the new Theme class. For custom palettes, the name is automatically generated similar to how it was before, but it replaces "_" with a space now.


src/Theme.php
```diff
@@ -0,0 +1,205 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ * ---------------------------------------------------------------------
+ */
+
+namespace Glpi;
```
cedric-anne, line 36: It is quite difficult to predict to what extent we will use namespaces in the future, but ideally all classes should be in a namespace. To avoid that everything is found in the namespace `Glpi`, it will then be necessary to create a tree structure. I propose here to use `Glpi\UI`, where we can put everything that corresponds to purely UI logic.

src/Theme.php
```diff
@@ -0,0 +1,205 @@
...
+        return null;
+    }
+
+    public function __construct(string $key, string $name, bool $is_dark, bool $is_custom)
+    {
+        $this->key = $key;
+        $this->name = $name;
+        $this->is_dark = $is_dark;
+        $this->is_custom = $is_custom;
+    }
```
cedric-anne, line 175: Constructor should probably moved to be the first method declared. Not sure it is a convention, but it is always declared at this place in GLPI classes, and also in most of projects I know.

src/Theme.php
```diff
@@ -0,0 +1,205 @@
...
+            $core_themes = self::getCoreThemes();
+            $core_keys = [];
+            foreach ($core_themes as $core_theme) {
+                $core_keys[] = $core_theme->getKey();
+            }
+
+            foreach ($file_matches as $file_name => $file) {
+                if (!in_array($file_name, $core_keys, true)) {
+                    // Guess dark mode based on the filename ("dark" with - or _ before or after)
+                    $pattern = '/dark[-_]|[-_]dark/i';
```
cedric-anne, line 120: Using `\b` would handle more cases.
```suggestion
                    $pattern = '/\bdark\b/i';
```

src/Theme.php
```diff
@@ -0,0 +1,205 @@
...
+            ];
+            foreach ($patterns as $pattern) {
+                foreach (glob(self::CORE_THEME_ROOT . $pattern) as $file) {
+                    $file_name = pathinfo($file, PATHINFO_FILENAME);
+                    if (str_starts_with($file_name, '_')) {
+                        continue;
+                    }
+                    $file_matches[$file_name] = $file;
+                }
+            }
```
cedric-anne, line 109: If we want to officially support custom themes, I think with you use one of the following solutions:
1. Adding a hook that would permit to any plugin to register one or more themes (and so the plugin would be itself able to declare the `$is_dark` value).
2. Using a new `GLPI_THEMES_DIR` constant (with default value to `_files/themes`), and parse this directory.

I personally prefer the first solution.

templates/layout/parts/head.html.twig
```diff
@@ -32,7 +32,8 @@
  #}
 
 <!DOCTYPE html>
-<html lang="{{ lang }}" {% if high_contrast %}data-high-contrast="1"{% endif %}>
+<html lang="{{ lang }}" {% if high_contrast %}data-high-contrast="1"{% endif %}
+      data-glpi-theme="{{ theme.key }}" {% if theme.isDarkTheme %}data-glpi-theme-dark="1"{% endif %}>
```
cedric-anne, line 36:
```suggestion
      data-glpi-theme="{{ theme.key }}" data-glpi-theme-dark="{{ theme.isDarkTheme() ? '1' : '0' }}">
```

cedric-anne (changes requested):

src/Theme.php
cconard96: For the first solution, I don't know how approachable it would be for users. IT staff may not necessarily know PHP or be familiar with creating a plugin or want to maintain one (even small changes). I would be nice to have it available as an option just in case someone wants to maintain some kind of theme pack, but I don't think it is good as the only solution.

The second solution more closely matches what it seems like users are already doing which is creating their own CSS files and dumping them in the `css/palettes` folder. At least with them being in `_files/themes` they won't conflict and are clearly user-defined.

With the new UI, creating new themes can be very simple with just copying an existing one and changing color values as needed.

cedric-anne: First solution is not so complicated. It would not require much than following code.

```php
<?php
function plugin_init_mycompany() {
    global $PLUGIN_HOOKS;
    $PLUGIN_HOOKS['custom_theme']['mycompany'] = [
        // key pattern: `plugin:${plugin_name}/${css_basename}`
        // will search for (s)css file in `css` directory of plugin:
        // - try to load `mycompany/css/mytheme.scss`
        // - fallback to `mycompany/css/mytheme.css` 
        return new \Glpi\UI\Theme('plugin:mycompany/mytheme', 'My company theme', true, true);
    ];
}

function plugin_version_mycompany() {
   return [
      'name'           => 'My company plugin',
      'version'        => '1.0.0',
   ];
}
```

And documenting it would not be more complicated than documenting the other solution.

@orthagh @trasher Which solution would you prefer ?

orthagh: i agree with @cconard96, php code, even minimal, can be a step too much complex for some users.
I think the `GLPI_THEMES_DIR`|`_files/themes` is a requirements.

Both solutions could be implemented but to be sure, what's the advantage to declare it by a plugin hook ? (fixing the is_dark issue only ?)

stonebuzz:
@cconard96 
With this PR, will it be possible to use CSS variables from the entity configuration ?
image

cconard96:
> @cconard96 With this PR, will it be possible to use CSS variables from the entity configuration ? image

No. The customization CSS is regular CSS and not SCSS so it cannot use Sass style variables at all. They can however use any regular CSS variables declared by the main styles or palettes.

src/Theme.php
cconard96: The plugin hook could be useful if their file name doesn't match what is expected for the light/dark check or if they want a different display name for the palette than what gets computed from the file name.
Although, I think both are edge cases.

It could also be used to group similar palettes together and/or distribute them via the existing plugin/marketplace system.

cedric-anne: > i agree with @cconard96, php code, even minimal, can be a step too much complex for some users.

For users that have no technical background, it is still possible to use the entity configuration to override the CSS.

> I think the `GLPI_THEMES_DIR`|`_files/themes` is a requirement.

Indeed, if we want to legacy behaviour (i.e. custom themes files that have been put in `css` directory), then we should probably trigger a warning to ensure that administrator will move them in the `GLPI_THEMES_DIR`.

> Both solutions could be implemented but to be sure, what's the advantage to declare it by a plugin hook ?

Main advantage is that it would be possible to develop theme packs as plugins that could be available through marketplace. It would then be possible to use the GLPI<>plugin compatibility logic to ensure that theme pack is downloaded in the correct version.

Another minimal advantage is to be able to have a translated name, and also, as you said, to have no ambiguity about the light/dark mode.

src/UI/Theme.php
```diff
@@ -0,0 +1,205 @@
...
+
+            $core_themes = self::getCoreThemes();
+            $core_keys = [];
+            foreach ($core_themes as $core_theme) {
+                $core_keys[] = $core_theme->getKey();
+            }
+
+            foreach ($file_matches as $file_name => $file) {
+                if (!in_array($file_name, $core_keys, true)) {
+                    // Guess dark mode based on the filename ("dark" with - or _ before or after)
```
orthagh, line 127: Currently discussing with @cedric-anne about this pr.
Just a quick note about this part.
It could be replaced by a detection of `$is-dark: true;` in the content of the file instead of trying to guess from the filename

orthagh:
Summary of our discussion with @cedric-anne 

- we add the GLPI_THEMES_DIR constant (default on `files/_themes`) for let users adding custom theme files
- if we find a custom theme in the base dir, we manage it but with a deprecation warning.
- we try to autodetect custom/dark themes based on their content (and not on their name)
- we abandon the idea of plugin for the moment (if needed for us, let discuss again on this)

cedric-anne:
@cconard96 Could you please rebase branch to solve conflicts ?

orthagh (commented):
Seems ok on my side
Just a suggestion.
@cedric-anne shows me `vfsStream` Wednesday and i think it could be applicable also for your tests (for checkkling custom and/or deprecated themes)

See the tests files in my [pr logs viewer](https://github.com/glpi-project/glpi/pull/12636/files#diff-307daff16bf18cd25b3c6c848348035d6fc8ddec594c643d4842701d1662eef6R46)

cedric-anne (commented):
I think the new class should be splitted in 2 parts.

1. `Theme`, that corresponds to a specific theme.
2. `ThemeManager`, that would contains all methods to compute available themes list and method to get current theme.

Also, there should be no `static` method, except maybe a `ThemeManager::getInstance()` that would be usefull until we will have a DI component.

Building classes with no static method could help to build better tests, and will make introduction of a DI container easier (when we will decide to introduce such a component).

.composer-require-checker.config.json
```diff
@@ -35,6 +35,7 @@
     "GLPI_NETWORK_SERVICES", "GLPI_PICTURE_DIR", "GLPI_PLUGIN_DOC_DIR", "GLPI_RSS_DIR", "GLPI_SESSION_DIR",
     "GLPI_TELEMETRY_URI", "GLPI_TMP_DIR", "GLPI_UPLOAD_DIR", "GLPI_USE_CSRF_CHECK", "GLPI_USE_IDOR_CHECK",
     "GLPI_USER_AGENT_EXTRA_COMMENTS", "GLPI_VAR_DIR", "GLPI_CENTRAL_WARNINGS", "GLPI_SERVERSIDE_URL_ALLOWLIST",
+    "GLPI_THEMES_DIR",
```
cedric-anne, line 38: This has to be define now in `stubs/glpi_constants.php`.

cedric-anne (changes requested):
1) Now that we handle a custom directory, the CSS inclusion path should be adapted (see `$_SESSION['glpipalette']` usages).

`Theme` must probably have a new `path` property that indicates the path of the them file.

2) While reading this PR, I figured out that the code could probably be reduced drastically by using the same logic to build both base and custom themes and by removing the `$is_custom` gflag that is actually useless.
We could indeed have a `getThemesFromDirectory()` method that would build theme key/name/is_dark based on filename/filecontent.
There would be only 2 small differences with your proposal:
 - we could not anymore translate/beautify palettes names,
 - we could not anymore add a deprecation warning for custom themes located in `css` directory.

I think that these 2 small differences are not valuable enough compared to the requirement of maintaining a hardcoded list, but maintaining that list should not be too difficult either.

@orthagh What is your opinion on this second point ?

cedric-anne (approved):

trasher (approved):

orthagh (approved):

Issue #11672, opened 2022-05-24
[GLPI 10 Bugfixes] No opened tickets for requester


tsmr:
### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

[GLPI 10 Bugfixes]

### Bug description

On previous versions, you can see on ticket creation the number of opened tickets for requester

![image](https://user-images.githubusercontent.com/13178158/169981986-fc74d512-c548-4bd0-8b0d-4ddc7d0012b9.png)

PR #11673, opened 2022-05-24
[GLPI 10 Bugfixes] Fix Requester Count Display

tsmr:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #11672


orthagh (changes requested):
I agree with the miss of the previous feature.
But maybe you can factorize the call of assign/requester information in a single ajax file and just change the title/parameter/etc.

Having a new file for requester doing the same for assign is little too much

templates/components/itilobject/actors/field.html.twig
```diff
@@ -140,27 +148,37 @@
          var option_element = $(`<span class="actor_entry" data-itemtype="${itemtype}" data-items-id="${items_id}" data-actortype="${actorytype}" data-text="${text}">${indent}${icon}${option_text}${actions}</span>`);
 
          // manage ticket information (number of assigned ticket for an actor)
-         if (is_selection && itemtype != "Email" && "{{ actortype }}" == "assign") {
-            var already_assign_element = $(
-               `<span class="assign_infos ms-1" title="{{ __('Number of tickets already assigned') }}"
+         if (is_selection && itemtype != "Email") {
+            if ("{{ actortype }}" == "assign") {
+               var already_element = $(
```
trasher, line 153: Variable name is strange.
```suggestion
               var existing_element = $(
```

trasher:
And also, you have CS fails to fix. Appart of that, sounds correct to me that me.

trasher (changes requested):
See previous comments

trasher (approved):

orthagh (changes requested):
Last thing, rename ticketinformation.php to actorinformation.php

Functional review ok

orthagh (approved):

PR #11815, opened 2022-06-05
Fix add column button tag and role

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

The button used to add/choose columns in the Kanban was using a custom element tag because of a typo. This didn't seem to break anything for me in Chrome even though it did not have a matching closing tag. Also, the role attribute was misnamed. On button elements, the role attribute is not needed. At one point, I think this button was a different type of element so it would have been needed before for accessibility reasons.


trasher (approved):

orthagh (commented):
> On button elements, the role attribute is not needed

Are you sure, AFAIK default role is `submit`, set role=button may avoid unwanted behavior.

cedric-anne (approved):

PR #11818, opened 2022-06-05
Move crontask UI to twig templates

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

Move automatic action (crontask) UI to twig templates.

Main form:
Selection_078

Statistics tab (just a table; nothing special):
Selection_077

### DIFF
--- a/templates/pages/setup/crontask/crontask.html.twig
+++ b/templates/pages/setup/crontask/crontask.html.twig
@@ -1,35 +1,35 @@
 {#
-# ---------------------------------------------------------------------
-#
-# GLPI - Gestionnaire Libre de Parc Informatique
-#
-# http://glpi-project.org
-#
-# @copyright 2015-2022 Teclib' and contributors.
-# @copyright 2003-2014 by the INDEPNET Development Team.
-# @licence   https://www.gnu.org/licenses/gpl-3.0.html
-#
-# ---------------------------------------------------------------------
-#
-# LICENSE
-#
-# This file is part of GLPI.
-#
-# This program is free software: you can redistribute it and/or modify
-# it under the terms of the GNU General Public License as published by
-# the Free Software Foundation, either version 3 of the License, or
-# (at your option) any later version.
-#
-# This program is distributed in the hope that it will be useful,
-# but WITHOUT ANY WARRANTY; without even the implied warranty of
-# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-# GNU General Public License for more details.
-#
-# You should have received a copy of the GNU General Public License
-# along with this program.  If not, see <https://www.gnu.org/licenses/>.
-#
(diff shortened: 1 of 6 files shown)


templates/pages/setup/crontask/statistics.html.twig
```diff
@@ -0,0 +1,98 @@
...
+ #
+ # You should have received a copy of the GNU General Public License
+ # along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ #
+ # ---------------------------------------------------------------------
+ #}
+
+<div class="d-flex justify-content-center">
+   <div class="col-sm-8 col-md-6 col-xlg-4 col-12 mx-2 mw-50">
+      <table class="table table-striped">
```
orthagh, line 36: Instead of table layout, it prefer to use some .row + .col
To have some style, you can either use ` .list-group` or much simple `divide-y` on the row div (i don't like the current stripped design, and prefer hover style like available in bootstrap lists group, but if you prefer stripped one, ok for me).

I'm not 100% sure by the way.
html tables should be reserved for tabuled data.
In the particular case we have two cols, and i cannot choose myself what design is better, a list or a table

templates/pages/setup/crontask/statistics.html.twig
```diff
@@ -0,0 +1,98 @@
...
+         <thead>
+            <tr><th colspan="2">{{ __('Statistics') }}</th></tr>
+         </thead>
+         <tbody>
+            <tr>
+               <td>{{ __('Run count') }}</td>
+               <td class="text-end">
+                  {{ _n('%s start', '%s starts', stats.runs.starts)|format(stats.runs.starts) }}
+                  <br>
```
orthagh, line 45: avoid `<br>` if possible and do these design with `ul.list-unstyled > li`

cconard96:
Using rows, cols, and list groups:
Selection_087

templates/pages/setup/crontask/crontask.html.twig
```diff
@@ -0,0 +1,177 @@
...
+                              no_label: true
+                           }) }}
+                           <span class="mx-2 text-nowrap">-></span>
+                           {{ fields.dropdownNumberField('hourmax', item.fields['hourmax'], null, {
+                              min: 0,
+                              max: 24,
+                              no_label: true
+                           }) }}
+                        </div>
+                     {% endset %}
```
orthagh, line 99: This part renders packed too much with the field on the right.
I think we should remove col-12 attribute to avoid overextend of the fields

orthagh:
> Using rows, cols, and list groups:

Seems great now

orthagh (approved):

cedric-anne (approved):

PR #11841, opened 2022-06-07
Fix multiple issues with Inventory code and PHPDoc

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

Fix multiple potential bugs found with the PHPStan static analysis tool and a review of the PHPDoc.

Summary of changes:
- The delete method of CommonDBTM should be called with a boolean for the force parameter even if it historically used an integer. The PHPDoc for the method properly typehints it as a boolean and internally, the method can handle it since it only checks the truthiness of the value.
- Fixed cases where the type in a PHPDoc block was not specific enough (InventoryAsset used but MainAsset methods used and always expected)
- Avoiding cases where variables may not be declared yet
- Missing method/property existence checks where their presence is not enforced by an abstract class, interface or trait.
- Remove passing arguments to a method that has none
- Fix missing return
- Fix unneeded return (unreachable or void result)
- Add missing PHPDoc
- Add property declarations to the untyped, dynamic properties on the Conf class based on the information I found from the array of default values.
- Avoid cases where a variable may be null but that case is not handled properly

One issue that was noticed by PHPStan that I also could not figure out how it would work was in `Glpi\Inventory\NetworkPort::handleMacConnection` where the `connection_ports` is used in multiple for loops but is always empty. A comment in that method makes it seem like it would be set by rules, but I wasn't able to see how that would be done. Either it is a bug, or there may need to be something added to hint to the static analysis tool that this property is actually set. I noted these cases with a FIXME comment.

This PR was not a complete fix. There are still a lot of detections at the different levels of analysis but these are the more obvious issues and ones that were easiest to fix right now.

### DIFF
--- a/src/Inventory/Asset/NetworkPort.php
+++ b/src/Inventory/Asset/NetworkPort.php
@@ -280,6 +280,7 @@ class NetworkPort extends InventoryAsset
         }
 
         $found_macs = [];
+        //FIXME Static analysis says this property is always an empty array. I don't see it get set even within the rule(s)
         foreach ($this->connection_ports as $ids) {
             $found_macs += $ids;
         }
@@ -289,6 +290,7 @@ class NetworkPort extends InventoryAsset
         }
 
        // Try detect phone + computer on this port
+        //FIXME Static analysis says this property is always an empty array. I don't see it get set even within the rule(s)
         if (isset($this->connection_ports['Phone']) && count($found_macs) == 2) {
             trigger_error('Phone/Computer MAC linked', E_USER_WARNING);
             return;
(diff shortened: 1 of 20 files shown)


cedric-anne (approved):

src/Inventory/Asset/NetworkPort.php
```diff
@@ -280,6 +280,7 @@ private function handleMacConnection(\stdClass $port, int $netports_id)
         }
 
         $found_macs = [];
+        //FIXME Static analysis says this property is always an empty array. I don't see it get set even within the rule(s)
```
trasher, line 283: As far as I can see; shit is set in the same file:
```php
$this->connection_ports[$itemtype][$ports_id] = $ports_id;
```

But I'm pretty sure this part may not work; I never had any data that suits those cases.

trasher (approved):

Issue #11843, opened 2022-06-08
[GLPI 10 Bugfixes] php-errors on update


tsmr:
### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

[GLPI 10 Bugfixes]

### Bug description

When update glpi with console : 

```
[2022-06-08 09:21:24] glpiphplog.NOTICE:   *** PHP Notice (8): Undefined index: glpiis_ids_visible in /var/www/html/glpi/src/CommonDBTM.php at line 3643
  Backtrace :
  rc/Socket.php:568                                  CommonDBTM->getNameID()
  rc/CommonDBTM.php:1298                             Glpi\Socket->post_addItem()
  ...migrations/update_9.5.x_to_10.0.0/cable.php:223 CommonDBTM->add()
  nstall/migrations/update_9.5.x_to_10.0.0.php:59    require()
  rc/Update.php:260                                  update95xto1000()
  rc/Console/Database/UpdateCommand.php:191          Update->doUpdates()
  endor/symfony/console/Command/Command.php:298      Glpi\Console\Database\UpdateCommand->execute()
  endor/symfony/console/Application.php:1015         Symfony\Component\Console\Command\Command->run()
  rc/Console/Application.php:256                     Symfony\Component\Console\Application->doRunCommand()
  endor/symfony/console/Application.php:299          Glpi\Console\Application->doRunCommand()
  endor/symfony/console/Application.php:171          Symfony\Component\Console\Application->doRun()
  /usr/bin/glpi-console:102                          Symfony\Component\Console\Application->run()
```

PR #11852, opened 2022-06-08
Always use ID in socket location history

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #11843

During the 10.0 migration, sockets are added using the standard `add` method. In the `post_addItem`, a history entry is added to the related location using the socket's `getNameID` method. This uses a user preference to determine if the ID should be shown in the name or not but there is no session during CLI migration.

In similar use cases, the ID is always forced so that the logged data is consistent and this should bypass the need for a user session.

### DIFF
--- a/src/CommonTreeDropdown.php
+++ b/src/CommonTreeDropdown.php
@@ -376,7 +376,7 @@ abstract class CommonTreeDropdown extends CommonDropdown
             $changes = [
                 0,
                 '',
-                addslashes($this->getNameID()),
+                addslashes($this->getNameID(['forceid' => true])),
             ];
             Log::history(
                 $parent,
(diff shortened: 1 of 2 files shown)


trasher (approved):

orthagh (approved):

cedric-anne (commented):
Same change should be added to 
- `Socket::post_deleteFromDB()`
- Calls to `Log::history()` made in `CommonTreeDropdown`.

cedric-anne (approved):

PR #11895, opened 2022-06-14
feat(socket): mutualize view from location and permit to link location to socket

stonebuzz:
Mutualize socket view between ```assets``` and ```Location```

image

Allow to link ```Location``` to ```Socket```  (to represent socket outlets)

image

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #number

### DIFF
--- a/inc/define.php
+++ b/inc/define.php
@@ -331,7 +331,7 @@ $CFG_GLPI['device_types']                 = ['DeviceMotherboard', 'DeviceFirmwar
 
 
 $CFG_GLPI["socket_types"]                  = ['Computer','NetworkEquipment',
-    'Peripheral','Phone','Printer', 'PassiveDCEquipment'
+    'Peripheral','Phone','Printer', 'PassiveDCEquipment', 'Location'
 ];
 
 $CFG_GLPI['itemdevices'] = [];
--- a/src/Socket.php
+++ b/src/Socket.php
@@ -812,8 +812,6 @@ class Socket extends CommonDBChild
      **/
     public static function showForLocation($item)
     {
-        Toolbox::deprecated();
-
         global $DB;
 
         $ID       = $item->getField('id');
--- a/src/Socket.php
+++ b/src/Socket.php
@@ -49,6 +49,7 @@ use Location;
 use Log;
 use NetworkPort;
 use Session;
+use Toolbox;
 
 /// Socket class
 class Socket extends CommonDBChild
@@ -648,7 +649,6 @@ class Socket extends CommonDBChild
 
     public static function displayTabContentForItem(CommonGLPI $item, $tabnum = 1, $withtemplate = 0)
     {
-        global $CFG_GLPI;
         self::showListForItem($item);
         return true;
     }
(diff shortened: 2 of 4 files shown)


src/Socket.php
```diff
@@ -800,217 +800,6 @@ public static function showListForItem($item)
...
 
 
-    /**
-     * Print the HTML array of the Socket associated to a Location
-     *
-     * @param $item Location
-     *
-     * @return void
-     **/
-    public static function showForLocation($item)
```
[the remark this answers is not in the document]
stonebuzz: done

stonebuzz:
@rbrannath  can you test this PR ?

KongNan:
I'm sorry, although I was able to install your build successfully, I (and my translation engine) don't recognize the French in your picture and I can't make any more judgments. However, from your description, you want to "represent" the outlet, which is great, and I hope that this representation appropriately reflects "an outlet located in a room". Thank you!

src/Socket.php
```diff
@@ -824,6 +829,13 @@ public static function showForLocation($item)
         $number = countElementsInTable('glpi_sockets', ['locations_id' => $ID ]);
 
         if ($canedit) {
+            echo '<table class="tab_cadre_fixe">';
+            echo '<tbody>';
+            echo '<tr class="tab_bg_1">';
+            echo '<th colspan="2" class="">';
+            echo __('Add socket');
+            echo '</th></tr></tbody><table>';
+
```
cedric-anne, line 838: I guess you made a mistake here. You added new features to a deprecated method.

rbrannath:
@stonebuzz, this is working as expected.

inc/define.php
```diff
@@ -331,7 +331,7 @@
 
 
 $CFG_GLPI["socket_types"]                  = ['Computer','NetworkEquipment',
-    'Peripheral','Phone','Printer', 'PassiveDCEquipment'
+    'Peripheral','Phone','Printer', 'PassiveDCEquipment', 'Location'
```
cedric-anne, line 334: I am not confident with addition of `Location` here, in the middle of an asset types list. I am not sure that it will not induce unwanted side effects.

src/Socket.php
stonebuzz: fixed

rbrannath:
@trasher also this PR

trasher:
Same answer. I'm already asked to do a review, please stop ping me.

stonebuzz:
Hi @KongNan 
can you do a functional review of current PR

stonebuzz:
@KongNan 

Note that we have doubts about the addition of locations in the ```socket_type``` list.

A ```Location``` is not an ```asset```.

Best regards

trasher:
> Note that we have doubts about the addition of locations in the `socket_type` list.

Little precision: this means current PR may never be merged. Fact is what you want to achieve is probably not possible with only a few changes like done here. There is a better solution we've been talking about in the team; but that would require much more time to be done. 
Maybe should you consider asking a specific development, [contact pro services](https://portal.glpi-network.com/marketplace/formcreator/front/formdisplay.php?id=15) in that case.

KongNan:
> Hi @KongNan can you do a functional review of current PR?


I'm very sorry, because my English is not good and I don't know French, so I feel helpless in this matter (as I said before: I (and my translation engine) don't recognize the French in your picture and I can't make any more judgments.)

Can you request @rbrannath to do it? He has great insight on the subject, as described in https://github.com/glpi-project/glpi/issues/10720#issuecomment-1077588981, but unfortunately that issue is closed. If possible, I hope you can address this scenario in your future considerations: https://github.com/glpi-project/glpi/issues/10720#issuecomment-1183446568 . If I continue to speak, it will probably only be to my superficial understanding to the detriment of actually solving the problem. So again, sorry, We need  rbrannath .

Regards

KongNan:
> you can download a build for this PR (just install it as an official release)

For completeness, I actually found https://github.com/stonebuzz/glpi/actions/runs/3005615443 at https://github.com/stonebuzz/glpi/actions and installed it.

Then used Google Lens to translate your screenshot from French to English and found Lieu for Location.

So I went into DropDown -> Locations, added the socket and took a screenshot comparing it as follows, just FYI.


image

image

rbrannath:
> @KongNan
> 
> Note that we have doubts about the addition of locations in the `socket_type` list.
> 
> A `Location` is not an `asset`.
> 
> Best regards
@stonebuzz 
Agree, `Location` in `socket_type` list appeals to be a good idea at the first glance, but considering there is already a field `Location` it might be confusing for some users (also your hint that is isn't an asset, is true as well).

Wouldn't it be the best and easiest approach, to simply add an field "is_outlet" to the socket and then disable/delete itemtype / asset connections to this socket?

Right now, I could live with the existing field `Location` and just leaving the itemtype / asset fields blank. In the `Location` form, there is already a functionality to add several sockets, but if no itemtype is selected, the socket is broken (because no itemtype).

rbrannath:
> > Note that we have doubts about the addition of locations in the `socket_type` list.
> 
> Little precision: this means current PR may never be merged. Fact is what you want to achieve is probably not possible with only a few changes like done here. There is a better solution we've been talking about in the team; but that would require much more time to be done. Maybe should you consider asking a specific development, [contact pro services](https://portal.glpi-network.com/marketplace/formcreator/front/formdisplay.php?id=15) in that case.

@trasher, would you mind to share the approach you brought up in your team? I don't know, if this PR is the right place for this, though.

cedric-anne:
@rbrannath 

A way to handle this would be to create a new `NetworkOutlet` asset type, that would clearly represent a network outlet. I personnaly think this is the best solution, as it would permit to handle it like any other asset. This is similar to what we have done with `PassiveDCEquipment` to handle patch panels in datacenters.

@trasher @orthagh Would you consider this solution as acceptable ?

orthagh:
yes

cedric-anne:
I close this PR as an alternative solution has to be proposed. See my previous comment.

PR #11910, opened 2022-06-15
Fix saving cards with transparency

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

Saving a dashboard card with transparency breaks the entire dashboard. The client sends the value as a rgba string instead of hex like the server expects and this is never checked. Errors thrown by the PHPColor library are not caught and break the entire dashboard.
This PR adds client-side conversion to ensure it sends the correct format.

PR is draft as I am looking to improve the reliability of the dashboard if there are issues with individual cards.

### DIFF
--- a/js/dashboard.js
+++ b/js/dashboard.js
@@ -500,6 +500,14 @@ var Dashboard = {
 
         var edit_item = "old_id" in form_data && form_data.old_id.length > 0;
 
+        // Force color to hex format
+        if (form_data.color.startsWith('rgb')) {
+            const rgb = form_data.color.match(/^rgba?[\s+]?\([\s+]?(\d+)[\s+]?,[\s+]?(\d+)[\s+]?,[\s+]?(\d+)[\s+]?/i);
+            form_data.color = (rgb && rgb.length === 4) ? "#" +
+                ("0" + parseInt(rgb[1],10).toString(16)).slice(-2) +
+                ("0" + parseInt(rgb[2],10).toString(16)).slice(-2) +
+                ("0" + parseInt(rgb[3],10).toString(16)).slice(-2) : '';
+        }
         // prepare options
         form_data.card_options.color        = form_data.color || null;
         form_data.card_options.widgettype   = form_data.widgettype || null;
--- a/js/dashboard.js
+++ b/js/dashboard.js
@@ -500,14 +500,6 @@ var Dashboard = {
 
         var edit_item = "old_id" in form_data && form_data.old_id.length > 0;
 
-        // Force color to hex format
-        if (form_data.color.startsWith('rgb')) {
-            const rgb = form_data.color.match(/^rgba?\s*\(\s*(\d+)\s*,\s*(\d+)\s*,\s*(\d+)\s*/i);
-            form_data.color = (rgb && rgb.length === 4) ? "#" +
-                ("0" + parseInt(rgb[1],10).toString(16)).slice(-2) +
-                ("0" + parseInt(rgb[2],10).toString(16)).slice(-2) +
-                ("0" + parseInt(rgb[3],10).toString(16)).slice(-2) : '';
-        }
         // prepare options
         form_data.card_options.color        = form_data.color || null;
         form_data.card_options.widgettype   = form_data.widgettype || null;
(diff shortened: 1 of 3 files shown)


orthagh (approved):

trasher (approved):

js/dashboard.js
```diff
@@ -500,6 +500,14 @@ var Dashboard = {
 
         var edit_item = "old_id" in form_data && form_data.old_id.length > 0;
 
+        // Force color to hex format
+        if (form_data.color.startsWith('rgb')) {
+            const rgb = form_data.color.match(/^rgba?[\s+]?\([\s+]?(\d+)[\s+]?,[\s+]?(\d+)[\s+]?,[\s+]?(\d+)[\s+]?/i);
```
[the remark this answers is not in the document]
cconard96: I'll check later, but this was copy/pasted from the tag plugin's JS code so of there is an issue, it needs fixed there too.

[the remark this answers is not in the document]
cconard96: The provided alternate pattern does not seem to work. It does not match any provided rgba strings. It only matches rgb strings.

- rgba(10, 10,10,4) - No match
- rgba(1,2,3,4) - No match
- rgb(10, 10,10) - Match
- rgb(1,2,3) - Match

js/dashboard.js
```diff
@@ -500,6 +500,14 @@ var Dashboard = {
 
         var edit_item = "old_id" in form_data && form_data.old_id.length > 0;
 
+        // Force color to hex format
+        if (form_data.color.startsWith('rgb')) {
```
cedric-anne, line 504: Using this, you will loose the `alpha` value. I do not know why `spectrum` is using the `rgba` form (we use the `preferredFormat: "hex"` option), but it is possible to use the `#rrggbbaa` form to preserve the alpha value and respect the `hex` format (see https://caniuse.com/css-rrggbbaa).

Also, maybe the best place to do this is when value of field is changed, as we probably want to keep hex format globally.

Spectrum has a `color.toHex8String()` method, but it return `#aarrggbb` instead of `#rrggbbaa`, so, the fix could be:
```diff
diff --git a/templates/components/form/fields_macros.html.twig b/templates/components/form/fields_macros.html.twig
index f6ca1d0391..d606c6995e 100644
--- a/templates/components/form/fields_macros.html.twig
+++ b/templates/components/form/fields_macros.html.twig
@@ -369,6 +369,13 @@
             showInput: true,
             preferredFormat: "hex",
             type: "text",
+            change: function(color) {
+                if (color.getAlpha() !== 1) {
+                    let hex = color.toHexString();
+                    hex += ("0" + Math.round(parseFloat(color.getAlpha()) * 255).toString(16)).slice(-2);
+                    this.value = hex;
+                }
+            }
         });
       });
       </script>
```

It seems to work as expected.

cconard96: That does fix the issue with the alpha channel data not being sent.
However, there is still an issue with the server-side PHPColors library's handling of colors with transparency.
`Toolbox::getFgColor` removes the alpha part if it is present before passing it to `Color`. For some reason, it also has a `inherit_if_transparent` parameter that is never used but I think that is a fairly important thing to have. I do not know why it is not implemented.

The use of the library in the Dashboard code does not do anything with the alpha channel.
In the case of `Widget::showSearchList`, it uses the library to determine if the picked background color is light or not and then uses that to get the foreground color with `Toolbox::getFgColor` as an offset. It is not possible to rely on the library to know the rendered color of the background when transparency is involved since we have no good way to know the color of the elements behind the card from the server.

With the midnight theme and a card background set to "FFFFFF00" (white transparent), it may try setting the foreground color to black which would make it invisible against the black background of the dashboard.

Unless you only use colors with a neutral luminosity and low transparency (or none), I do not see how the current implementation would allow a dashboard to show well on both light and dark themes. If it is possible to move the logic of picking colors to the client, it may be easier to accomplish since we can get the current background color for the dashboard.

cedric-anne: A solution would be to pass a `allow_transparency` boolean to the for macro, and to use it to both define the  `show_alpha` option of spectrum and define whether the alpha information should be kept in returned hex code.

cedric-anne: As seen with @orthagh , a better handling of transparency in dashboards could be done in another PR.

Issue #11935, opened 2022-06-19
10.0.1 - No naming consistency for lock fields


WebGregGit:
### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

10.0.1

### Bug description

No naming consistency between field names and global lock fields list. For example: "Technician in charge of the hardware" and "users_id_tech"? Or "**Inventory number**" and "**Computer - otherserial**".
The addition of "_id" is also rather redundant.
Here is no "Alternate username" field.

### Steps To reproduce

1. Open the page with computer e.g.: **_front/computer.form.php?id=1_**
2. Here is the "**Inventory number**" field.
3. If you want to lock it, you have to select on the page _**front/lockedfield.form.php**_: "**Computer - otherserial**".

### Anything else?

front/lockedfield.php - I have on list "Computer - states_id". What is it actually blocking? I thought the "Status" field on the front/computer.form.php tab, but no. And now I don't know if the definition of this field lock does not work, or if the names do not match too.

Issue #11934, opened 2022-06-19
10.0.1 - After adding a global lock, the field is still selectable


WebGregGit:
### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

10.0.1

### Bug description

After adding a global lock, the field is still selectable. This makes management difficult because you have to go back to the site front/lockedfield.php, remember list and than you can add next global lock.

A small note, by the way - it would be easier with a link (add button?) to **_front/lockedfield.form.php_** from _**front/computer.form.php?id=x**_ - Locks tab.

### Relevant log output

```shell
In sql-errors.log I found:

[2022-06-16 22:42:17] glpisqllog.ERROR: DBmysql::query() in /var/www/html/glpi/src/DBmysql.php line 370
*** MySQL query error:
SQL: INSERT INTO glpi_lockedfields (itemtype, items_id, field, is_global, date_creation, date_mod) VALUES ('Computer', '0', 'otherserial', '1', '2022-06-16 22:>
Error: Duplicate entry 'Computer-0-otherserial' for key 'glpi_lockedfields.unicity'
Backtrace :
src/DBmysql.php:1312
src/CommonDBTM.php:705 DBmysql->insert()
src/CommonDBTM.php:1297 CommonDBTM->addToDB()
front/lockedfield.form.php:51 CommonDBTM->add()
{"user":"6@glpi"}

This error is related to problem 1. It appears when you try to add a previously added field again.
```

### Steps To reproduce

1. Open the page **_front/lockedfield.form.php_**
2. For example, add "Computer - name"
3. "Computer - name" - still on the list

### Your GLPI setup information

GLPI 10.0.1 ( => /var/www/html/glpi)
Installation mode: TARBALL
Current language:en_GB

Operating system: Linux glpi 5.4.0-117-generic #132-Ubuntu SMP Thu Jun 2 00:39:06 UTC 2022 x86_64
PHP 7.4.3 apache2handler (Core, FFI, PDO, Phar, Reflection, SPL, SimpleXML, Zend OPcache, apache2handler, apc, apcu, bz2,
	calendar, ctype, curl, date, dom, exif, fileinfo, filter, ftp, gd, gettext, hash, iconv, imap, intl, json, ldap, libxml,
	mbstring, mysqli, mysqlnd, openssl, pcre, pdo_mysql, posix, readline, session, shmop, snmp, soap, sockets, sodium, standard,
[...]

PR #11963, opened 2022-06-21
Fix/global locks

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #11934 #11935

### DIFF
--- a/src/Lockedfield.php
+++ b/src/Lockedfield.php
@@ -292,6 +292,7 @@ class Lockedfield extends CommonDBTM
         global $CFG_GLPI, $DB;
 
         $iterator = $DB->request([
+            'SELECT' => ['itemtype', 'field'],
             'FROM'   => $this->getTable(),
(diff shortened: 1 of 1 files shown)


orthagh (approved):

src/Lockedfield.php
```diff
@@ -291,6 +291,16 @@ public function getFieldsToLock(): array
     {
         global $CFG_GLPI, $DB;
 
+        $iterator = $DB->request([
+            'FROM'   => $this->getTable(),
```
cconard96, line 295:
```suggestion
            'SELECT' => ['itemtype', 'field'],
            'FROM'   => $this->getTable(),
```

cconard96: Selecting only the needed fields here has better performance

cconard96 (approved):

PR #11966, opened 2022-06-21
Keep inventory agent updated for any item type

g-bougard:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a

This PR is another solution to the problem #11948 fixes. I figured out, #11948 if fixing the case agent object property for an inventory is missing tag as it is not fully updates. But it fixes the problem just before running rules engine.

Here, when starting the inventory import, we keep the agent updated even if the agent type is not matching. In that case, we also need to forget the items_id which is the agent linked item id. So it is not preloaded as item object property at [src/Inventory/Inventory.php#L308](https://github.com/glpi-project/glpi/blob/fbf1c2f137c1a698896e86b6ee0d5d89ccbdc4c0/src/Inventory/Inventory.php#L308).

### DIFF
--- a/src/Agent.php
+++ b/src/Agent.php
@@ -298,7 +298,8 @@ class Agent extends CommonDBTM
             $input['tag'] = $metadata['tag'];
         }
 
-        if ($deviceid === 'foo' || !in_array($metadata['itemtype'], $CFG_GLPI['agent_types'])) {
+        $has_expected_agent_type = in_array($metadata['itemtype'], $CFG_GLPI['agent_types']);
+        if ($deviceid === 'foo' || (!$has_expected_agent_type && !$aid)) {
             $input += [
                 'items_id' => 0,
                 'id' => 0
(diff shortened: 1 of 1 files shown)


g-bougard:
I forgot to talk about an important point, with this PR we will have the tag during rules processing for network discovery AND network inventory files.

cedric-anne (approved):

orthagh:
This pr replace #11948, right ?

We should close one of these.

trasher:
I guess this would need tests

g-bougard:
> This pr replace #11948, right ?
> 
> We should close one of these.

As I said in @stonebuzz PR, #11948 will only permit to use expected tag for netinventory file. It will work if he removes the test on is_netdiscovery property. #11948 also instantiate another Agent object where this could be avoided by using this one.


> I guess this would need tests
I was also thinking about that. In which file would you like to see them ?

trasher:
> I was also thinking about that. In which file would you like to see them ?

No idea, this does not really matters anyways

stonebuzz:
Ok for me, the agent tag is correctly passed to the rules (original issue)

stonebuzz (approved):

orthagh (approved):

Issue #11493, opened 2022-05-05
10.0.0 - no object name on additional tabs


WebGregGit:
### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

10.0.0

### Bug description

Object name is displayed only for the first tab. I will refer to computers, but it also applies to other objects.

### Page URL

front/computer.php

### Steps To reproduce

1. Go to Assets > Computers > open any
2. Computer tab - at the top is the name of the computer.
3. Change the tab.
4. The computer name is no longer displayed.

### Anything else?

Why is this annoying? For example, select the tickets tab. At the top, there are navigation buttons that allow you to switch between computers (it's a pity that they are not pinned and disappear when you scroll the page vertically). If you switch the computer - you don't know which computer you are viewing the tickets tab. You need to go back to the computer tab, check the name and go back to the tickets tab. 

As a result, the functionality optimizing the browsing activities of specific data in a selected group of computers has been lost.

And maybe one more detail. The navigation buttons in the object's directory are at the top, and as I mentioned - scrolling along with the page. However, directly on front/computer.form.php the navigation buttons for changing pages are at the bottom and are pinned. This is just a loose comment - maybe it is worth standardizing it? It would be more convenient if the possibility of changing the object was also at the bottom and pinned.

WebGregGit:
As I can see, the work is still in progress.

WebGregGit:
Solved. Thank you.

PR #11968, opened 2022-06-21
Move main asset header back to the nav header

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #11493

Visually moved the item type and name header back to the navigation header so that is would be visible in each tab of an asset form.

Selection_079

~Still working on the functionality of the massive action dropdown and child entity checkbox as they are no longer located within the main form and are therefore missing some required inputs. Some more testing with modals needs done too. Modals of at least dropdown itemtypes seem to work properly with showing the header inside the form rather than the navigation header.~

### DIFF
--- a/src/CommonGLPI.php
+++ b/src/CommonGLPI.php
@@ -1013,21 +1013,39 @@ class CommonGLPI implements CommonGLPIInterface
             // First set of header pagination actions, displayed on the left side of the page
             echo "<div>";
 
-            if ($first >= 0) {
-                echo "<a href='$cleantarget?id=$first$extraparamhtml'
-                     class='btn btn-sm btn-icon btn-ghost-secondary' title=\"" . __s('First') . "\"
-                     data-bs-toggle='tooltip' data-bs-placement='bottom'>
-                     <i class='fa-lg ti ti-chevrons-left'></i>
-                  </a>";
+            if (!$glpilisttitle) {
+                $glpilisttitle = __s('List');
             }
+            $list = "<a href='$glpilisturl' title=\"$glpilisttitle\"
+                  class='btn btn-sm btn-icon btn-ghost-secondary'
+                  data-bs-toggle='tooltip' data-bs-placement='bottom'>
+                  <i class='far fa-lg fa-list-alt'></i>
+               </a>";
+            $list_shown = false;
 
+            if ($first < 0) {
+                // Show list icon before the placeholder space for the first pagination button
+                echo $list;
+                $list_shown = true;
+            }
+            echo "<a href='$cleantarget?id=$first$extraparamhtml'
+                 class='btn btn-sm btn-icon btn-ghost-secondary " . ($first >= 0 ? '' : 'bs-invisible') . "' title=\"" . __s('First') . "\"
+                 data-bs-toggle='tooltip' data-bs-placement='bottom'>
+                 <i class='fa-lg ti ti-chevrons-left'></i>
+              </a>";
+
+            if (!$list_shown && $prev < 0) {
+                // Show list icon before the placeholder space for the "prev" pagination button
+                echo $list;
+                $list_shown = true;
+            }
+            echo "<a href='$cleantarget?id=$prev$extraparamhtml'
+                 id='previouspage'
+                 class='btn btn-sm btn-icon btn-ghost-secondary " . ($prev >= 0 ? '' : 'bs-invisible') . "' title=\"" . __s('Previous') . "\"
+                 data-bs-toggle='tooltip' data-bs-placement='bottom'>
+                 <i class='fa-lg ti ti-chevron-left'></i>
+              </a>";
             if ($prev >= 0) {
-                echo "<a href='$cleantarget?id=$prev$extraparamhtml'
-                     id='previouspage'
-                     class='btn btn-sm btn-icon btn-ghost-secondary' title=\"" . __s('Previous') . "\"
-                     data-bs-toggle='tooltip' data-bs-placement='bottom'>
-                     <i class='fa-lg ti ti-chevron-left'></i>
-                  </a>";
                 $js = '$("body").keydown(function(e) {
                        if ($("input, textarea").is(":focus") === false) {
                           if(e.keyCode == 37 && e.ctrlKey) {
@@ -1038,14 +1056,10 @@ class CommonGLPI implements CommonGLPIInterface
                 echo Html::scriptBlock($js);
             }
 
-            if (!$glpilisttitle) {
-                $glpilisttitle = __s('List');
+            // If both first and prev buttons shown, the list should be added now
+            if (!$list_shown) {
+                echo $list;
             }
-            echo "<a href='$glpilisturl' title=\"$glpilisttitle\"
-                  class='btn btn-sm btn-icon btn-ghost-secondary'
-                  data-bs-toggle='tooltip' data-bs-placement='bottom'>
-                  <i class='far fa-lg fa-list-alt'></i>
-               </a>";
 
             echo "</div>";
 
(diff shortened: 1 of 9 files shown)


orthagh:
i have an issue with the current pr:

image

You probably missed it with the usage of the midnight theme

AdrienClairembault:
I don't think it's gonna be easy to get something "clean" UI wise for this change.
Maybe it would be simpler to add the header to every tabs ? 

The "header" could become a fixed part of the page and the tabs would only reload the space under it.

cconard96:
I did notice the background color difference.
If that is the only part that makes it not clean, it won't be difficult to use a CSS rule to remove the colored background from the header when it is in the navigation header.

AdrienClairembault:
I've just tested the branch and it seems to works great functionally, however I'm not really sold on the UI side.

@orthagh what are you thoughts ?

image

AdrienClairembault:
Up @orthagh

orthagh:
After a live discussion with @AdrienClairembault, we agreed to the following design:
image

cconard96:
Selection_098

The alignment of the asset name with the end of the tabs only works if the tab names are at their minimum width of 200px. If they stretch beyond that, it won't line up anymore.

If the new changes I made are OK, I'll rebase for 10.0.

cconard96:
Selection_100

Changed header align-self from baseline to end. I think it looks a bit better.

orthagh:
I have a strange rendering with the current iteration on both chrome and firefox:

image

Alignement seems to be broken

cconard96:
Odd. For me, it looks OK in Chrome in Firefox.

orthagh:
After testing at office (and on a ubuntu computer) i got the good display of the header.
So, i checked with @AdrienClairembault who uses Windows on its laptop and he have the bad displays on both firefox and edge.

So there something on windows breaking the flex vertical alignment.
I can't try to fix myself before thursday, if you can check @cconard96 ?

cconard96:
I also don't see that extra line in the header that you have even when I switch the palette back to Auror. I can't tell what that would be from.

husisusi:
> I have a strange rendering with the current iteration on both chrome and firefox:
> 
> Alignement seems to be broken

same here!
image

AdrienClairembault:
I've done some debugging on different OS :
* Firefox linux + PHP linux (ubuntu)
* Firefox windows+ PHP linux (wsl ubuntu)

On linux, I could see this css rule: 

image

Which didn't seem to exist on windows (or at least wasn't active).

I've tried to add some "!important" here :

image
... and it works.
Then I've removed the "!important" and it still works ???

So in the end it seems it was only some caching issue ... 
No clue why it only happens on windows (+ WSL backend) but now I'm stuck with the "correct" css and can't get back to the bugged version no matter how hard I try so I guess the code is correct.

Still found one issue tho, the item name isn't aligned correctly if you are on the first search result:

image

image

AdrienClairembault:
Any update on this ?
Also do we still want this change to make it into 10.0.3 ? It's getting close.

cconard96:
I added a commit that would keep the pagination controls, but just hide them (with visibility property so they still take up space) to help with the alignment. It won't always be perfect as the tab widths can change.

AdrienClairembault:
Its a clever solution to keep the same width but it pushes the list icon alone in the middle:

image

It can be fixed easily by moving the list icon at the start of the line if we detect no "left side" pagination:
You can do something like that:

```diff
diff --git a/src/CommonGLPI.php b/src/CommonGLPI.php
index 22d9e4f5b5..b2481e02b1 100644
--- a/src/CommonGLPI.php
+++ b/src/CommonGLPI.php
@@ -995,6 +995,21 @@ class CommonGLPI implements CommonGLPIInterface
             // First set of header pagination actions, displayed on the left side of the page
             echo "<div>";

+            if (!$glpilisttitle) {
+                $glpilisttitle = __s('List');
+            }
+
+            $list = "<a href='$glpilisturl' title=\"$glpilisttitle\"
+                class='btn btn-sm btn-icon btn-ghost-secondary'
+                data-bs-toggle='tooltip' data-bs-placement='bottom'>
+                <i class='far fa-lg fa-list-alt'></i>
+            </a>";
+
+            if ($first < 0) {
+                // No pagination, show the list icon first
+                echo $list;
+            }
+
             echo "<a href='$cleantarget?id=$first$extraparamhtml'
                  class='btn btn-sm btn-icon btn-ghost-secondary " . ($first >= 0 ? '' : 'bs-invisible') . "' title=\"" . __s('First') . "\"
                  data-bs-toggle='tooltip' data-bs-placement='bottom'>
@@ -1018,14 +1033,11 @@ class CommonGLPI implements CommonGLPIInterface
                 echo Html::scriptBlock($js);
             }

-            if (!$glpilisttitle) {
-                $glpilisttitle = __s('List');
+            if ($first >= 0) {
+                // Show list after pagination options
+                echo $list;
             }
-            echo "<a href='$glpilisturl' title=\"$glpilisttitle\"
-                  class='btn btn-sm btn-icon btn-ghost-secondary'
-                  data-bs-toggle='tooltip' data-bs-placement='bottom'>
-                  <i class='far fa-lg fa-list-alt'></i>
-               </a>";
+

             echo "</div>";
```

image

AdrienClairembault:
The rendering on small screens seems a bit chaotic compare to the previous versions.
GLPI 10:
image

These changes:
image

cconard96:
Last commit improves the alignment on small screens.
Selection_104

AdrienClairembault (approved):

templates/generic_show_form.html.twig
```diff
@@ -39,7 +39,7 @@
 {% endif %}
 
 <div class="asset {{ bg }}">
-   {{ include('components/form/header.html.twig') }}
+   {% include 'components/form/header.html.twig' with {'in_twig': true} %}
```
cedric-anne, line 42: Please use the `include` function, as recommanded in [Twig docs](https://twig.symfony.com/doc/3.x/tags/include.html).
```suggestion
   {{ include('components/form/header.html.twig', {'in_twig': true}) }}
```

src/CommonGLPI.php
```diff
@@ -1058,23 +1072,35 @@ class='btn btn-sm btn-icon btn-ghost-secondary'
...
+                    'params'        => $options,
+                    'in_navheader'  => true
+                ];
+                if (method_exists($this, 'getFormHeaderToolbar')) {
+                    $header_toolbar = $this->getFormHeaderToolbar();
+                    // Do not trust return type will always be string
+                    if (is_array($header_toolbar)) {
+                        $twig_params['header_toolbar'] = $header_toolbar;
+                    }
+                }
```
cedric-anne, line 1087: The proper way to ensure that method returns an array is probably to declare it in `CommonGLPI` with an empty implementation: `protected function getFormHeaderToolbar(): array { return []; }`.

```suggestion
                $twig_params['header_toolbar'] = $this->getFormHeaderToolbar();
```

templates/components/form/header.html.twig
```diff
@@ -55,108 +52,21 @@
...
-      {% set template_name = item.fields['template_name']|verbatim_value %}
-      {% if withtemplate == 2 and not item.isNewItem() %}
-         <input type="hidden" name="template_name" value="{{ template_name }}" />
-         {% set nametype = __('Created from the template %s')|format(template_name) %}
-      {% elseif withtemplate == 1 %}
-         <input type="hidden" name="is_template" value="1" />
-      {% elseif item.isNewItem() %}
-         {% set nametype = __('%1$s - %2$s')|format(__('New item'), nametype) %}
+      {% if no_header == false and (in_twig is defined or _get._in_modal|default(false)) %}
+         {% include 'components/form/header_content.html.twig' with {inside_main: true} %}
```
cedric-anne, line 67: Please use the `include` function, as recommanded in [Twig docs](https://twig.symfony.com/doc/3.x/tags/include.html).
```suggestion
         {{ include('components/form/header_content.html.twig', {'inside_main': true}) }}
```

orthagh:
@cconard96 we plan to include this in 10.0.3 if fixes could be done in the following day.
It needs so a rebase on 10.0/bf branch please.

Just after, we (at least @cedric-anne, @AdrienClairembault  and me) will do an extensive serie of functionnal tests to be sure we don't break anything juste before release.

cconard96:
Changes from review have been made and the PR was rebased on the 10.0 bugfixes branch.

templates/components/form/header.html.twig
```diff
@@ -55,108 +52,21 @@
...
    <div id="mainformtable">
-      {% set template_name = item.fields['template_name']|verbatim_value %}
-      {% if withtemplate == 2 and not item.isNewItem() %}
-         <input type="hidden" name="template_name" value="{{ template_name }}" />
-         {% set nametype = __('Created from the template %s')|format(template_name) %}
-      {% elseif withtemplate == 1 %}
-         <input type="hidden" name="is_template" value="1" />
-      {% elseif item.isNewItem() %}
-         {% set nametype = __('%1$s - %2$s')|format(__('New item'), nametype) %}
+      {% if no_header == false and ((in_twig is defined or _get._in_modal|default(false)) or (in_twig is not defined and item.isNewID(item.fields['id']))) %}
```
cedric-anne, line 66: Could you please move this in a `{% set include_header_content %}` variable declaration on line 57 and add some comments to explain all these conditions, especially usage of `in_twig` variable ?

cedric-anne:
There is still a bug. Following header is not displayed anymore in escalade config form:
image

Also, as a side effect, some items that already defined custom headers in their sub tabs will now have a weird presentation. In fields plugin:
image
In genericobject plugin:
image

cconard96:
Escalade is calling `showForm` directly from the front file and there is also no way I know of to tell that the config is a singleton item so it would never have a search list or navigation bar.

The objects plugin is also calling showForm from within another form.

These would all likely need fixed on the plugin side.

cedric-anne:
> Escalade is calling `showForm` directly from the front file and there is also no way I know of to tell that the config is a singleton item so it would never have a search list or navigation bar.
> 
> The objects plugin is also calling showForm from within another form.
> 
> These would all likely need fixed on the plugin side.

All plugins should then probably reviewed using current branch and fixed with something that is compatible also with GLPI 10.0.3. I do not know if other plugins are imacted, or if any itemtype in GLPI can be impacted too.

AdrienClairembault (approved):

trasher (approved):

PR #11978, opened 2022-06-22
add processes list

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

Following this morning webminar, i was asked for process list.
The fact is we already have the data in inventory files, we "just" need to store and display them.
Ans so this POC was very quickly done.

TODO:
- [x] tests
- [x] finish generic datatable

image


trasher:
With a quick test on an old inventory file; I got the following error:
```
  SQL: INSERT INTO `glpi_items_processes` (`cmd`, `cpuusage`, `pid`, `started`, `tty`, `user`, `virtualmemory`, `memusage`, `is_dynamic`, `items_id`, `itemtype`) VALUES ('/usr/bin/earlyoom -r 0 -m 4 -M 409600 --prefer ^Web Content$ --avoid ^(dnf|packagekitd|gnome-shell|gnome-session-c|gnome-session-b|lightdm|sddm|sddm-helper|gdm|gdm-wayland-ses|gdm-session-wor|gdm-x-session|Xorg|Xwayland|systemd|systemd-logind|dbus-daemon|dbus-broker|cinnamon|cinnamon-sessio|kwin_x11|kwin_wayland|plasmashell|ksmserver|plasma_session|startplasma-way|xfce4-session|mate-session|marco|lxqt-session|openbox)$', '0.0', '6143', '2020-12-21 08:03:00', '?', 'earlyoom', '2484', '0.0', '1', '1', 'Computer')
  Error: Data too long for column 'cmd' at row 1
```

src/Inventory/Conf.php
```diff
@@ -68,9 +68,10 @@ class Conf extends CommonGLPI
         'import_volume'                  => 1,
         'import_antivirus'               => 1,
         'import_registry'                => 1,
-        'import_process'                 => 1,
+        'import_process'                 => 0,
         'import_vm'                      => 1,
         'import_monitor_on_partial_sn'   => 0,
+        'import_process'                 => 1,
```
trasher, line 74:
```suggestion
```

src/Inventory/Asset/Process.php
```diff
@@ -0,0 +1,157 @@
...
+                    $itemProcess->update(Toolbox::addslashes_deep($input));
+                    unset($value[$key]);
+                    unset($db_itemProcess[$keydb]);
+                    break;
+                }
+            }
+        }
+
+        if ((!$this->main_asset || !$this->main_asset->isPartial()) && count($db_itemProcess) != 0) {
+           // Delete Item_Disk in DB
```
trasher, line 132:
```suggestion
           // Delete Item_Process in DB
```

tests/functionnal/Glpi/Inventory/Assets/Process.php
```diff
@@ -0,0 +1,144 @@
...
+
+        $result = $asset->prepare();
+        $this->object($result[0])->isEqualTo(json_decode($expected));
+    }
+
+    public function testHandle()
+    {
+        $computer = getItemByTypeName('Computer', '_test_pc01');
+
+       //first, check there are no volume linked to this computer
```
trasher, line 116:
```suggestion
       //first, check there are no process linked to this computer
```

trasher (approved):

trasher:
tests are failing because default configuration is to not import processes. You have to update (and reset) the configuration.

trasher (approved):

trasher:
Tested with deiffrent inventory files; all seems good

cconard96:
I updated my DB and then forced an inventory of my Windows Server VM, but I do not see the Processes tab.
There is nothing in the DB table.

cconard96:
The agent log says the Processes module is disabled (among a lot of others).
No processes are in the inventory file because of that I assume.
When I installed it, the only feature I didn't include was ESX remote inventory so I don't know where it would need enabled.
I toggled the import option on the server side and it didn't change anything.

cconard96:
I don't think it supports Windows.
In the related perl file the `isEnabled` method checks if `OSName` is not equal to `MSWin32` and if it can run `ps`.

cconard96:
Installed on a Linux VM and it works to get the info to the server.
The `code` HTML elements used in the command column have a color that is too dark to see in the midnight theme. The color comes from Bootstrap and doesn't seem to be handled by Tabler for dark mode support
Selection_086
.

orthagh:
> The code HTML elements used in the command column have a color that is too dark to see in the midnight theme. 

Fixed

> Also, the ID column seems to be the id column in the database instead of the process ID pid. Was that intended?

yes, more to have sorting capacity on table. The column itself is not this important

cconard96:
> yes, more to have sorting capacity on table. The column itself is not this important

Maybe the process ID should be included as well so it is more clear and to help differentiate between processes with similar commands. I don't know how this tab would be used though.

orthagh:
> Maybe the process ID should be included as well so it is more clear and to help differentiate between processes with similar commands.

I can move to processes id, as i said above, it doesn't matter.

> I don't know how this tab would be used though.

Not very usefull in my opinion, but on the last webinar about inventory, someone asked for this.
As the processes was already present in inventory file, i though it should be quick to add the feature.

But for some reasons, i took the opportunity to work around a generic datables twig template (this took a lot more time) to present long list of read-only data

trasher (approved):

cconard96 (approved):

PR #11983, opened 2022-06-23
fix(massive action): update location empty

Rom1-B:
In the ticket's massive actions, the list of locations was always empty.

image

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !24307

### DIFF
--- a/src/MassiveAction.php
+++ b/src/MassiveAction.php
@@ -926,7 +926,7 @@ class MassiveAction
 
     public static function showMassiveActionsSubForm(MassiveAction $ma)
     {
-        global $CFG_GLPI;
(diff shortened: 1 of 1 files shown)


trasher (approved):

orthagh (approved):

cedric-anne (commented):
Can you check if cases `Change` and `Problem` should be updated too ?

Rom1-B:
> Can you check if cases `Change` and `Problem` should be updated too ?

Already checked, I haven't identified any fields in the same case, so it doesn't seem necessary.

PR #11996, opened 2022-06-23
Add access error log

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

When an access-related issue occurs and causes the script execution to stop, a generic error is given to the user (good for security to not give exact details to them) but more information isn't available anywhere on the server for debugging.
This PR adds an "access-errors.log" file to log additional information including the user ID, request URI, and a sanitized stack trace (no arguments) and an optional message that can be changed based on the context.

### DIFF
--- a/src/Session.php
+++ b/src/Session.php
@@ -885,7 +885,8 @@ class Session
         if (Session::getCurrentInterface() != "central") {
            // Gestion timeout session
             self::redirectIfNotLoggedIn();
-            Html::displayRightError();
+            $user_id = self::getLoginUserID() ?? 'Anonymous';
+            Html::displayRightError("User ID: $user_id The current profile does not use the standard interface");
         }
     }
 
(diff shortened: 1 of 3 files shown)


src/Session.php
```diff
@@ -934,10 +934,37 @@ public static function checkLoginUser()
...
         }
     }
 
+    /**
+     * Get the name of the right.
+     *
+     * This only works for well-known rights.
+     * @param int $right The right
+     * @return string The right name
+     * @internal No backwards compatibility promise. Use in core only.
```
cconard96, line 947: Note that this function is marked as internal to indicate that it has no BC promise.
I imagine this can be moved into the PermissionManager class after that PR is merged for 10.1.

trasher (approved):

PR #11998, opened 2022-06-23
Itemtype dropdown: do not impact glpiUnsavedFormChanges when setting initial value

AdrienClairembault:
It seems that the initial value of this dropdown is set using js, which will trigger `glpiUnsavedFormChanges  = true`.
This will trigger the "unsaved changes" warning on page exit every time, which we do not want to happen as no data was actually modified by the user.

```php
Html::jsSetDropdownValue($field_id, $params['default_itemtype']);
```

Note that you may not be able to reproduce this issue on a standard GLPI installation, as we do not use this dropdown on any page with the "unsaved changes" warning enabled (as far as I know).

However, some plugins or specific patches (or potential future developments) may add this dropdown and cause the issue described above.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !24304

### DIFF
--- a/src/Dropdown.php
+++ b/src/Dropdown.php
@@ -1478,6 +1478,7 @@ class Dropdown
         $params['checkright']          = false;
         $params['toupdate']            = '';
         $params['display']             = true;
+        $params['track_changes']       = true;
 
         if (is_array($options) && count($options)) {
             foreach ($options as $key => $val) {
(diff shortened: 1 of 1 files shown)


cconard96:
This seems OK for me but a cleaner solution is to add a dropdown option for tracking changes set to true by default, but false for the itemtype dropdown used in `Dropdown::showSelectItemFromItemtypes` rather than saving and restoring the unsaved change flag.

Any input with `data-track-changes=""` inside a form that tracks changes would be ignored and not change the unsaved change flag when its value is changed.

src/Dropdown.php
```diff
@@ -2103,6 +2107,10 @@ public static function showFromArray($name, array $elements, $options = [])
                 $output .= " required='required'";
             }
 
+            if (!$param[$param['track_changes']]) {
```
cconard96:
```suggestion
            if (!$param['track_changes']) {
```

cconard96 (changes requested):

src/Dropdown.php
```diff
@@ -1543,19 +1544,20 @@ public static function showSelectItemFromItemtypes(array $options = [])
...
+        $params['entity_restrict']        = -1;
+        $params['onlyglobal']             = false;
+        $params['checkright']             = false;
+        $params['showItemSpecificity']    = '';
+        $params['emptylabel']             = self::EMPTY_VALUE;
+        $params['used']                   = [];
+        $params['ajax_page']              = $CFG_GLPI["root_doc"] . "/ajax/dropdownAllItems.php";
+        $params['display']                = true;
+        $params['rand']                   = mt_rand();
+        $params['itemtype_track_changes'] = true;
```
cconard96, line 1560: Should be false by default in this case since changing the itemtype dropdown on its own doesn't mean anything beyond changing which item dropdown values are shown.

AdrienClairembault:
Should be all good now

cconard96 (approved):

trasher (approved):

orthagh (approved):

PR #12008, opened 2022-06-24
Re-implement agent cleanup

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes (Fixed regression)
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

The agent cleanup options remain in the GLPI Inventory plugin, but they are not used by the plugin or by native inventory. The cleanup configuration was not re-added when the agent class was migrated.

### DIFF
--- a/src/Agent.php
+++ b/src/Agent.php
@@ -667,4 +667,63 @@ class Agent extends CommonDBTM
     {
         return "ti ti-robot";
     }
+
+    /**
+     * Cron task: clean and do other defined actions when agent not have been contacted
+     * the server since xx days
+     *
+     * @global object $DB
+     * @param object $task
+     * @return boolean true if successful, otherwise false
+     */
+    public static function cronCleanoldagents($task = null)
+    {
+        global $DB;
+
+        $config = \Config::getConfigurationValues('inventory');
+
+        $retention_time = $config['agents_old_days'] ?? 0;
+        if ($retention_time <= 0) {
+            return true;
+        }
+
+        $iterator = $DB->request([
+            'FROM' => self::getTable(),
+            'WHERE' => [
+                'last_contact' => ['<', new QueryExpression("date_add(now(), interval -" . $retention_time . " day)")]
+            ]
+        ]);
+
+        $cron_status = false;
+        if (count($iterator)) {
+            //agents_action 0=clean 1=change status
+            $action = (int)($config['agents_action'] ?? 0);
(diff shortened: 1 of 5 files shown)


orthagh (approved):

src/Agent.php
```diff
@@ -667,4 +667,63 @@ public static function getIcon()
...
+            if ($action === 0) {
+                //delete agents
+                $agent = new self();
+                foreach ($iterator as $data) {
+                    $agent->delete($data);
+                    $task->addVolume(1);
+                    $cron_status = true;
+                }
+            } else if (isset($config['agents_status'])) {
+                //change status of agents
```
g-bougard, line 710: About the comment, this is indeed the status of the linked asset which is changed.
The comment should be updated to avoid confusion.

src/Inventory/Conf.php
```diff
@@ -768,6 +793,59 @@ public function showConfigForm()
...
+        echo "<tr class='tab_bg_1'><td colspan=2></td>";
+        echo "<td>";
+        echo "<span id='blocaction1' style='display:none'>";
+        echo __('Change the status', 'glpiinventory');
+        echo "</span>";
+        echo "</td>";
+        echo "<td width='20%'>";
+        echo "<span id='blocaction2' style='display:none'>";
+        State::dropdown(['name'   => 'agents_status',
+            'value'  => $config['agents_status'] ?? -1,
```
g-bougard, line 842: Does this mean the value stored in database could be `-1` if `$config['agents_status']` is not defined or is this an UX syntax ? Imho, I expect to see `0` as default in db.

src/Agent.php
g-bougard: Even, the configuration name is not well named, but I'm not sure this is possible to change.

src/Inventory/Conf.php
cconard96: Either one should work. It just avoids a PHP notice when the config option is not set yet. Both will default to the empty option which doesn't make sense, but it was like this in FI.

In the client request to allow both actions, the empty choice label will become "No change".

src/Agent.php
g-bougard:
```suggestion
                //change status of agents linked assets
```

g-bougard (approved):

PR #12043, opened 2022-06-29
[GLPI 10 Bugfixes] Fix Rule SQL errors Import Asset

tsmr:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #10708


trasher:
Was using `state_types` because Fusion was using `state_types`... I have no idea if that breaks any feature or not - and this code part is certainly untested.

cedric-anne (approved):

PR #12045, opened 2022-06-29
Fix/priority matrix config

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

The commits should be kept at merge as they fix different issue.

- 131a6173cde7e2badcbb8146746add8b74007cfe fix a regression introduced by #10956 where the filter is applied in config (where we define the filter :) )
- 2d51ea08e2601b57bac2222456a31b608eba9052 avoid an sql error when log system truncate a json entry.
- ce84fd030b0dc82276e6b8612ffe7b8974aedc7f avoid removing priority entry corresponding to the existing value of the itil object (and selecting by default major priority)

### DIFF
--- a/src/CommonITILObject.php
+++ b/src/CommonITILObject.php
@@ -2730,7 +2730,11 @@ abstract class CommonITILObject extends CommonDBTM
                         if ($p['withmajor'] && $priority == 6) {
                             continue;
                         }
-                        unset($values[$priority]);
+
+                        // don't unset current value (to avoid selecting major priority on existing item)
+                        if ($priority != $p['value']) {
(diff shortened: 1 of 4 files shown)


flegastelois:
I think it's also necessary to correct the "Major" priority which is displayed in the ticket when the ticket priority value in the database is no longer available in the matrix.

trasher:
Maybe tests should be added?

cedric-anne (approved):

trasher (approved):

PR #12052, opened 2022-06-29
Fix global search

stonebuzz:
When first criteria is a global rule,

if we want to remove it 

image

all search form is removed

image

Remove useless 'dot' 

image

which remains when we add criterion (global or not)

image

| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #number

### DIFF
--- a/src/Search.php
+++ b/src/Search.php
@@ -3032,6 +3032,8 @@ JAVASCRIPT;
         }
         echo "</div>";
         echo "</div>";
+        echo "</div>";
+        echo "</div>";
     }
 
     /**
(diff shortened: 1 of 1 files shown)


orthagh:
It's better than the previous pr but the "..." seems to be still here
When i removed all criteria (not the meta ones):
image

orthagh:
my bad, i tested a different instance than the one where the pr was checkout.

orthagh (approved):

trasher (approved):

cedric-anne (approved):

PR #12060, opened 2022-06-30
Separate agent clean actions and add uninstall integration

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !23415

Separates the agent cleanup actions so that you have the option to cleanup the agent and change the status of the related item instead of choosing one action or the other.
This also adds integration with the uninstall plugin to allow applying an uninstall model to the related item during cleanup.

### DIFF
--- a/src/Agent.php
+++ b/src/Agent.php
@@ -37,6 +37,7 @@
 use Glpi\Application\ErrorHandler;
 use Glpi\Application\View\TemplateRenderer;
 use Glpi\Inventory\Conf;
+use Glpi\Plugin\Hooks;
 use Glpi\Toolbox\Sanitizer;
 use GuzzleHttp\Client as Guzzle_Client;
 use GuzzleHttp\Psr7\Response;
@@ -689,23 +690,51 @@ class Agent extends CommonDBTM
      */
     private static function cleanAgent(Agent $agent, array $config): bool
     {
-        // Actions that will need the associated item
-        static $item_actions = ['change_status', 'apply_uninstall_profile'];
+        global $PLUGIN_HOOKS;
+
         // Actions to apply to the agent or item
+        /** @phpstan-var array{item_action: bool, callback: callable(?Agent, array, ?CommonDBTM):bool}[] $actions_to_apply */
         $actions_to_apply = [];
+        $need_item = false;
         if (isset($config['stale_agents_status']) && $config['stale_agents_status']) {
-            $actions_to_apply[] = 'change_status';
+            $need_item = true;
+            $actions_to_apply[] = [
+                'item_action' => true,
+                'callback' => static function(?Agent $agent, array $config, ?CommonDBTM $item): bool {
+                    return $item !== null && $item->update([
+                        'id' => $item->fields['id'],
+                        'states_id' => $config['agents_status']
+                    ]);
+                }
+            ];
         }
-        if (isset($config['stale_agents_uninstall']) && $config['stale_agents_uninstall']) {
-            $actions_to_apply[] = 'apply_uninstall_profile';
+        $plugin_actions = $PLUGIN_HOOKS[Hooks::STALE_AGENT_CONFIG];
+        /**
+         * @var string $plugin
+         * @phpstan-var array{label: string, item_action: boolean, render_callback: callable, action_callback: callable}[] $actions
+         */
+        foreach ($plugin_actions as $plugin => $actions) {
+            if (is_array($actions) && \Plugin::isPluginActive($plugin)) {
+                foreach ($actions as $action) {
+                    if ($action['item_action']) {
+                        $need_item = true;
+                    }
+                    $actions_to_apply[] = $action['action_callback'];
+                }
+            }
         }
         if  (isset($config['stale_agents_clean']) && $config['stale_agents_clean']) {
-            $actions_to_apply[] = 'clean';
+            $actions_to_apply[] = [
+                'item_action' => false,
+                'callback' => static function(?Agent $agent, array $config, ?CommonDBTM $item): bool {
+                    return $agent !== null && $agent->delete(['id' => $agent->fields['id']]);
+                }
+            ];
         }
         $item = null;
 
         // If an action to apply needs the item
-        if (count(array_intersect($actions_to_apply, $item_actions)) > 0) {
+        if ($need_item) {
             $orphan = false;
             try {
                 $item = new $agent->fields['itemtype'];
@@ -725,6 +754,16 @@ class Agent extends CommonDBTM
         }
 
         // Run all actions, with the clean action running last
+        foreach ($actions_to_apply as $action) {
+            // If the action needs the item, and we don't have one, skip it
+            if ($action['item_action'] && $item === null) {
+                continue;
+            }
+            // Run the action
+            if (!$action['callback']($agent, $config, $item)) {
+                return false;
+            }
+        }
         if ($item !== null && in_array('change_status', $actions_to_apply, true)) {
             $result = $item->update([
                 'id' => $item->fields['id'],
--- a/src/Agent.php
+++ b/src/Agent.php
@@ -695,19 +695,20 @@ class Agent extends CommonDBTM
         /** @phpstan-var array{item_action: bool, callback: callable(?Agent, array, ?CommonDBTM):bool}[] $actions_to_apply */
         $actions_to_apply = [];
         $need_item = false;
-        if (isset($config['stale_agents_status']) && $config['stale_agents_status']) {
+        if (isset($config['stale_agents_status']) && $config['stale_agents_status'] >= 0) {
+            // 0 = none, -1 or other negative = no change
             $need_item = true;
             $actions_to_apply[] = [
                 'item_action' => true,
-                'callback' => static function(?Agent $agent, array $config, ?CommonDBTM $item): bool {
+                'callback' => static function (?Agent $agent, array $config, ?CommonDBTM $item): bool {
                     return $item !== null && $item->update([
                         'id' => $item->fields['id'],
-                        'states_id' => $config['agents_status']
+                        'states_id' => $config['stale_agents_status']
                     ]);
                 }
             ];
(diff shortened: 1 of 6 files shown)


cconard96:
I don't have a good solution for adding the integration on the plugin side.
A hook could be added in GLPI for handling stale agents since the agent may not be deleted and therefore the deletion hooks cannot be used, but I don't know a good way to keep the configuration centralized rather than having a separate option somewhere in the plugin just for agent cleanup.

trasher:
I agree, core should not have any plugin call references, this kind of thing would be a real nightmare.

cedric-anne:
@cconard96 Could you rebase your branch ? 10.0/bugfixes branch has be merged into main branch, so first commits of current PR could be removed now.

src/Agent.php
```diff
@@ -681,13 +681,99 @@ public static function getIcon()
...
+                $item = null;
+            }
+        }
+
+        // Run all actions, with the clean action running last
+        foreach ($actions_to_apply as $action) {
+            // If the action needs the item, and we don't have one, skip it
+            if ($action['item_action'] && $item === null) {
+                continue;
+            }
```
cedric-anne, line 761: For a plugin developper that want to handle an action that acts on the agent itself, and optionnaly on related item, if one exists, current behaviour will require to declare 2 callables in plugin (the first one that will always be called, even when item is null, and the second one that will be called when item is not null). Not sure we will be able to document it clearly.
Also, plugins will probably want to act on item in most cases (I do not see any relevant use case of a plugin that would act only on the agent).
And, finally, a plugin that requires the item but that is declaring `'item_action' => false` may still receive the item in its callback, depending on GLPI conf and on other plugins, and so may luckily work in certain conditions on not on others.

IMHO, we could consider that `$need_item` is true as long as a plugin hook is declared. Plugin would have to check `$item !== null` in their callback, as you did in yours.
```suggestion
```

src/Agent.php
```diff
@@ -681,13 +681,99 @@ public static function getIcon()
...
+            ];
+        }
+        $plugin_actions = $PLUGIN_HOOKS[Hooks::STALE_AGENT_CONFIG] ?? [];
+        /**
+         * @var string $plugin
+         * @phpstan-var array{label: string, item_action: boolean, render_callback: callable, action_callback: callable}[] $actions
+         */
+        foreach ($plugin_actions as $plugin => $actions) {
+            if (is_array($actions) && \Plugin::isPluginActive($plugin)) {
+                foreach ($actions as $action) {
```
cedric-anne, line 718: A check on hook declaration would probably be more secure.
```suggestion
                foreach ($actions as $action) {
                    if (!is_callable($action['action_callback'] ?? null)) {
                        trigger_error(
                            sprintf('Invalid plugin "%s" action callback for "%s" hook.', $plugin, Hooks::STALE_AGENT_CONFIG),
                            E_USER_WARNING
                        );
                        continue;
                    }
```

src/Inventory/Conf.php
```diff
@@ -822,52 +833,62 @@ public function showConfigForm()
...
+        foreach ($plugin_actions as $plugin => $actions) {
+            if (is_array($actions) && \Plugin::isPluginActive($plugin)) {
+                foreach ($actions as $action) {
+                    if (!$odd) {
+                        echo "<tr class='tab_bg_1'>";
+                    }
+                    $field = $action['render_callback']($config);
+                    if (!empty($field)) {
+                        echo "<td>";
+                        echo $action['label'];
```
cedric-anne, line 874:
```suggestion
                        echo $action['label'] ?? '';
```

tests/functionnal/Agent.php
```diff
@@ -321,4 +321,254 @@ public function testTagFromXML()
...
+        $DB->updateOrInsert(
+            \Config::getTable(),
+            [
+                'name' => 'stale_agents_clean',
+                'value' => $original_stale_clean
+            ],
+            [
+                'name' => 'stale_agents_clean'
+            ]
+        );
```
cedric-anne, line 421: As each test is made in a DB transaction that is rollback just after the test, I think this piece of code is useless.
```suggestion
```

tests/functionnal/Agent.php
```diff
@@ -321,4 +321,254 @@ public function testTagFromXML()
...
+
+        // Force run the cleanup cron task (direct call the function)
+        $crontask = new \CronTask();
+        // Should return true even if the agent has an invalid item
+        $this->boolean(\Agent::cronCleanoldagents($crontask))->isTrue();
+        // Verify that the agent has not been deleted
+        $this->boolean((bool) $agent->getFromDB($agents_id))->isTrue();
+        if ($items_id > 0) {
```
cedric-anne, line 518: This case does not exist in in your test data provider.

cedric-anne (changes requested):

src/Inventory/Conf.php
```diff
@@ -822,52 +833,62 @@ public function showConfigForm()
...
+         * @var string $plugin
+         * @phpstan-var array{label: string, item_action: boolean, render_callback: callable, action_callback: callable}[] $actions
+         */
+        foreach ($plugin_actions as $plugin => $actions) {
+            if (is_array($actions) && \Plugin::isPluginActive($plugin)) {
+                foreach ($actions as $action) {
+                    if (!$odd) {
+                        echo "<tr class='tab_bg_1'>";
+                    }
+                    $field = $action['render_callback']($config);
```
cedric-anne, line 871: I just take a look at https://github.com/pluginsGLPI/uninstall/pull/68 and, even if it is a good idea from UX point of view, it will lead to writing a configuration value in the `inventory` context instead of writing in the expected `plugin::uninstall` context.

There is already a configuration page for this plugin, so I would prefer to move this configuration field in plugin configuration context. The render callback could still be used to display the current configuration (read only mode), with a link to the plugin configuration form, or even a modal form to update the plugin configuration, accessible via a :memo: icon.

@orthagh @trasher, your opinion ?

src/Agent.php
cconard96: `item_action` was intended to be a boolean indicating if the plugin's action works on the agent's related item instead of or in addition to the agent itself.

src/Inventory/Conf.php
cconard96: I don't see this hook being used very often beyond this one case, but if having the options in the inventory context isn't desired, then the config field name can be used to also specify the context optionally or try to enforce it.

If the field name is not in the list of defaults and starts with "plugin_uninstall_" for example, then plugin:uninstall is the context and everything to the right is the name.
We can try to enforce this field naming by unsetting the value before it is saved if it contains "plugin" and is not in the list of defaults or just rejecting if it isn't in the defaults regardless of the name.

Having the option be read only in the agent cleanup actions config and linking to the plugin's own page for just a single field seems like bad UX. If the config option has to be shown in the uninstall plugin's own config form, then it should be in addition to the option in the inventory config.

src/Agent.php
cedric-anne: Indeed, but you may want to work on both agent and its related item, and in this case, you probably want to work on agent even if related item is not available anymore.

src/Inventory/Conf.php
cedric-anne: Saving the value in the right context could probably be handled in a `Hooks::PRE_ITEM_UPDATE` callback.

Displaying the additionnal field could probably be done using `Hooks::POST_ITEM_FORM` callback. In this case, it could simplify the `Hooks::STALE_AGENT_CONFIG` expected format, which would expect the `action_callback`.

cconard96: That hook doesn't account for different sections within the same form though. It would only look somewhat like it should because the stale agent config happens to be the last section currently.

orthagh: Maybe we can add a starting _  for this kind of field (like we do in other place) and add a method `Config::pre_updateInDB` to remove all fields starting with this char.
The plugin can save (or alter the context) itself the value with PRE_ITEM_UPDATE hook.
Maybe just this behavior (just the plugin intercept the call) is needed that said.

I quickly checked if there any existing data like this in my current DB:
image

IMO, none of these are valid one and are in the same scenario where we store unwanted data in the table.

cedric-anne: > Maybe we can add a starting _ for this kind of field (like we do in other place) and add a method `Config::pre_updateInDB` to remove all fields starting with this char.

This requires some modifications on multiple places, so if we want to do this, I think it could be better to do it in a separated PR, to make the review easier.

tests/functionnal/Agent.php
cconard96: If the items_id in the provider is null, a new item is created for the test and linked to the agent. If it is 0, it simulates an orphaned item.

orthagh (approved):

src/Agent.php
```diff
@@ -714,8 +714,15 @@ private static function cleanAgent(Agent $agent, array $config): bool
...
+                    if (!is_callable($action['action_callback'] ?? null)) {
+                        trigger_error(
+                            sprintf('Invalid plugin "%s" action callback for "%s" hook.', $plugin, Hooks::STALE_AGENT_CONFIG),
+                            E_USER_WARNING
+                        );
+                        continue;
+                    }
                     if ($action['item_action']) {
                         $need_item = true;
                     }
```
cedric-anne, line 728: We could probably consider that a plugin that acts on stale agents will require the corresponding item in most cases. Having a `item_action` param would require to write a documentation about it, and developpers does not always read documentation. The fact that plugins may receive a different payload depending on `stale_agents_status` would probably be incorrectly handled for developpers that does not understand the usage of `item_action`. So, IMHO, removing it may prevent issues on plugins.

```suggestion
```

cedric-anne (changes requested):

cedric-anne (approved):

Issue #12056, opened 2022-06-30
Can't assign group to a child entity directly from LDAP import


0xNath:
### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

10.0.2

### Bug description

I have 4 levels of entities.
let say that I set my user to the top entitie and I want to import a group to an under entitie :
<img width="832" alt="image" src="https://user-images.githubusercontent.com/67143274/176642344-f4ca3ded-6f1b-4ddd-bbf0-ecac5edccb48.png">

When I change the entities of destination of the groups and do the import, they go to the entity that my user has currently :
![image](https://user-images.githubusercontent.com/67143274/176649413-602ffc46-914a-4b6c-a980-888239e6e809.png)

### Page URL

https://localhost/front/ldap.group.import.php?next=servers?mode=0&action=show

### Steps To reproduce

1. setup GLPI
2. add an LDAP authentification method
3. go the the group pages
4. click on the "LDAP directory link button"
5. click on button "Import of new groups"
6. Select the previously authentification method in the list
7. Select two random groups, change the "destination entity" of one of the groups to a different entity that your user is set to
8. When going into the groups page, you should see that the two groups are under the same entity

### Your GLPI setup information

<details><summary>Informations sur le système, l'installation et la configuration</summary><pre>GLPI 10.0.2 ( => C:\xampp\htdocs\glpi)
Installation mode: TARBALL
Current language:fr_FR

</details><details><summary>Server</summary><pre> 
Operating system: Windows NT serverName 10.0 build 17763 (Windows Server 2019) AMD64
PHP 8.1.6 apache2handler (Core, PDO, Phar, Reflection, SPL, SimpleXML, Zend OPcache, apache2handler, bcmath, bz2, calendar,
	ctype, curl, date, dom, exif, fileinfo, filter, ftp, gd, gettext, hash, iconv, intl, json, ldap, libxml, mbstring, mysqli,
	mysqlnd, openssl, pcre, pdo_mysql, pdo_sqlite, readline, session, standard, tokenizer, xml, xmlreader, xmlwriter, zip, zlib)
Setup: max_execution_time="600" memory_limit="512M" post_max_size="40M" safe_mode="" session.save_handler="files"
	upload_max_filesize="40M" 
[...]

PR #12068, opened 2022-06-30
Fix LDAP group import options

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #12056

3 fixes:
1. The JS used to set the massive action inputs in the massive action modal used the `value` attribute of elements with a specific data attribute. This was not working for the select element to select the target entity. The `val` method needed to be used instead.
2. The recursive entity option was shown as a massive action checkbox which was not using the provided name and instead set it to the expected name as if it were the main checkbox used to select the items.
3. The recursive entity option was not passed in the massive actions. It seems like the `data-glpicore-ma-tags` has to be set to `common` otherwise, there is no effect.

### DIFF
--- a/src/AuthLDAP.php
+++ b/src/AuthLDAP.php
@@ -2298,13 +2298,13 @@ class AuthLDAP extends CommonDBTM
                               $dn_index,
                               ['massive_tags'  => 'select_item_child_entities',
                                   'name'          => "ldap_import_recursive[$dn_index]",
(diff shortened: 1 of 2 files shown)


cedric-anne (approved):

trasher (approved):

wadaki:
I'm looking for a fix on 9.x - because there is the same issue available. 
Thank you in advance.

PR #12143, opened 2022-07-05
Fix display of allowed statuses for new ITIL Objects

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !24027

See: https://github.com/glpi-project/glpi/issues/11858#issuecomment-1168514243

The `value_calculation` property may be an empty string which `??` would not have made `value` be used instead. The `?:` operator would be needed so that any time the `value_calculation` property is falsy (including empty strings), the `value` is used instead.

### DIFF
--- a/src/CommonITILObject.php
+++ b/src/CommonITILObject.php
@@ -3313,7 +3313,10 @@ abstract class CommonITILObject extends CommonDBTM
 
         switch ($p['showtype']) {
             case 'allowed':
-                $tab = static::getAllowedStatusArray($p['value_calculation'] ?: $p['value']);
+                $current = isset($p['value_calculation']) && $p['value_calculation'] !== ''
+                    ? $p['value_calculation']
+                    : $p['value'];
+                $tab = static::getAllowedStatusArray($current);
                 break;
 
             case 'search':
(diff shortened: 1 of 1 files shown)


trasher (approved):

src/CommonITILObject.php
```diff
@@ -3313,7 +3313,7 @@ public static function dropdownStatus(array $options = [])
 
         switch ($p['showtype']) {
             case 'allowed':
-                $tab = static::getAllowedStatusArray($p['value_calculation'] ?? $p['value']);
+                $tab = static::getAllowedStatusArray($p['value_calculation'] ?: $p['value']);
```
orthagh, line 3316: issue is, with elvis operator, you now can have a warning for undefined key if `value_calculation` is not set in `$p` array.

I'm not sure you can have a short syntax for this
You will need to have something like this:

```php 
isset($p['value_calculation']) && strlen($p['value_calculation']) > 0
   ? $p['value_calculation']
   : $p['value'];
```

cedric-anne (approved):

trasher (approved):

PR #12145, opened 2022-07-06
Ticketd: fix timeline scroll with middle click

AdrienClairembault:
When viewing a ticket, the right panel (tickets details) is scroll-able using the middle mouse button but the left side (ticket timeline) isn't.

image

This seems to be caused by an obsolete flex div:

image

This div contains only one element (the timeline container) so the `d-flex` and `flex-column-reverse` classes do not do anything.
Removing them fix the scrolling issues.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |

### DIFF
--- a/templates/components/itilobject/layout.html.twig
+++ b/templates/components/itilobject/layout.html.twig
@@ -61,9 +61,7 @@
    {% endif %}
 
    <div class="row d-flex flex-column alin-items-stretch itil-object">
-      {% set is_timeline_reversed = user_pref('timeline_order') == constant('CommonITILObject::TIMELINE_ORDER_REVERSE') %}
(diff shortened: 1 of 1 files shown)


orthagh (changes requested):
the `column-reverse` aims to have the scroll at the end without javascript.
We can't remove this feature

AdrienClairembault:
In this case, it seems to be an issue with firefox which doesn't support middle click scrolling on `column-reverse` content.
I don't see how we can fix this without remove the `column-reverse` part and adding some client side scroll using javascript.

We could also try reporting the issue to firefox (it doesn't seem to be a known issue ATM) and wait for a fix.

PR #12168, opened 2022-07-07
Add option to reset Kanban state

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

In some cases, the Kanban state may become corrupt or the user may want to reset it for some other reason and there is no way for it to be cleared except directly in the database.
If the user can modify the state, an option will show in the toolbar to reset the view now.
Resetting the view will reset the shown columns and remove custom card ordering.

### DIFF
--- a/ajax/kanban.php
+++ b/ajax/kanban.php
@@ -220,6 +220,11 @@ if (($_POST['action'] ?? null) === 'update') {
         'timestamp' => $_SESSION['glpi_currenttime']
     ];
     echo json_encode($response, JSON_FORCE_OBJECT);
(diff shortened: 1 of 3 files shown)


cconard96:
Even when the Kanban's AJAX endpoint dies with an error code, the toolbar is still displayed and this option is still available for user's with the right permissions. In most cases when there is an issue with a bad view, there should be a way to resolve the issue without going into the DB directly.

trasher (approved):

orthagh (approved):

PR #12176, opened 2022-07-08
fix(search): prevent error for ambigous column

stonebuzz:
Fix ambigous column when trying to display ```IP``` and ```MAC```

Before : 
image

After : 
image

| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #number

### DIFF
--- a/src/NetworkPort.php
+++ b/src/NetworkPort.php
@@ -1401,7 +1401,7 @@ class NetworkPort extends CommonDBChild
             'massiveaction'      => false,
             'joinparams'         => ['jointype' => 'itemtype_item',
                 'condition' => ['NOT' => [
-                    'instantiation_type'  => NetworkPortAggregate::getType()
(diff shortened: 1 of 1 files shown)


trasher (approved):

WebGregGit:
Hi @stonebuzz 

I have a slightly similar problem when I try to sort by IP address. But I don't know if that has anything to do with this fix. Can I also create a new issue for this?

image

stonebuzz:
> Hi @stonebuzz
> 
> I have a slightly similar problem when I try to sort by IP address. But I don't know if that has anything to do with this fix. Can I also create a new issue for this?
> 

Yes please, is not the same issue here the MySQL function ```inet_atom``` does not receive an IP.

Issue #12258, opened 2022-07-20
Search AND NOT for antivirus still not work


loyolajavi:
### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

9.5.7 with FusionInventory 9.5+4.2 

### Bug description

In Computer search when I filter by "not have some antivirus", if a computer have two or more antivirus, eg: Symantec and Kaspersky, that computer appears in the result, even though there is the filter to exclude:
Filter:
![image](https://user-images.githubusercontent.com/17304522/180032796-9c1cf517-76f2-4db2-b36c-076def96a939.png)

That computer have this antiviruses:
![image](https://user-images.githubusercontent.com/17304522/180030958-2996fe66-8f0d-440f-9363-76fb58020238.png)

Besides, the filter "IS NOT" doesn't show any antivirus to choose from:
![image](https://user-images.githubusercontent.com/17304522/180029241-94cf3abe-e901-4651-acdd-e76a906ba3d2.png)

### Steps To reproduce

1. Assets -> Computers
2. Create filter A or B: 
      A. Antiviruses Name not contain Symantec
      B. Antiviruses Name IS NOT Symantec

### Anything else?

I view the issue closed: https://github.com/glpi-project/glpi/issues/4894, but the bug persists in GLPI 9.5.7

cconard96:
Please test with the criteria:
NOT Antiviruses - Name is Symantec

Since you can have more than one antivirus, the negative search criteria does not work as you may expect. Using Is Not or Not Contains would actually match items if ANY of the antiviruses were not Symantec or if ANY of the antiviruses did not contain Symantec.

loyolajavi:
Hi Curtis,

Thanks for your response, I'm afraid the "Antiviruses - Name" "is" and "is not" criterias aren't working properly (at least in my installation).. do not show me any antivirus for selection. 
![image](https://user-images.githubusercontent.com/17304522/180314190-03460cbe-bd48-497d-b652-170e8719ef21.png)
Do you get antiviruses displayed when you select the "is"/"is not" filters? (I have 9.5.7 version)

cconard96:
Is/Is Not probably should not be options. Currently, antiviruses are only stored as sub-items of computers rather than individual items that get linked to one or more computers.

PR #12271, opened 2022-07-22
Allow only contains and not contains search for Antivirus name

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #12258

Currently, is/is not search types are broken with computer antiviruses because they do not have an entity field (SQL error) but the classes `isEntityAssign` returns true because their parent item type is assigned to entities.
Since antiviruses are only a child itemtype, I do not think it makes sense to have this search type at this time even if it had the `entities_id` field. There could be duplicate entries in the dropdown with different values where each selection only matched the antivirus of a single computer even if the names were exactly the same in other computers.

### DIFF
--- a/src/ComputerAntivirus.php
+++ b/src/ComputerAntivirus.php
@@ -153,7 +153,8 @@ class ComputerAntivirus extends CommonDBChild
             'datatype'           => 'dropdown',
             'joinparams'         => [
                 'jointype'           => 'child'
-            ]
+            ],
+            'searchtype'         => ['contains'],
         ];
 
         $tab[] = [


trasher (approved):

cedric-anne (commented):
#12275 fixes the SQL query. If it is validated, is/is not operator would be usable.

cconard96:
> #12275 fixes the SQL query. If it is validated, is/is not operator would be usable.

It may fix the SQL error, but it is not "usable".
Screenshot from 2022-07-22 05-54-15

As I said before, each antivirus only exists as a child item of a single computer rather than distinct items that are linked to one or more computers. There are duplicate options in the dropdown and their IDs only match the record for a single computer.

cedric-anne (approved):

PR #12288, opened 2022-07-25
Plugin: fix 'isInstalled' check

AdrienClairembault:
`Plugin:->isInstalled()` rely on `Plugin->isPluginLoaded()` which can't be trusted every page, for example every plugin are always loaded on `front/plugin.php` if their name are displayed: 

image

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |

### DIFF
--- a/src/Plugin.php
+++ b/src/Plugin.php
@@ -98,13 +98,6 @@ class Plugin extends CommonDBTM
     public static $rightname = 'config';
 
     /**
-     * Plugin init state.
-     *
-     * @var boolean
-     */
-    private static $plugins_init = false;
-
-    /**
      * Activated plugin list
      *
      * @var string[]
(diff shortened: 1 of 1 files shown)


cedric-anne (approved):

cedric-anne (approved):
As the removed check was made to prevent useless DB queries, I pushed a commit to optimize the way we check whether a plugin is active or not. Now, list of active plugins is computed before actually loading them, so calls to `Plugin::isPluginActive()` will not anymore do any DB query.

I may save a lot of DB queries when many plugins are active, as most of them are using this method on their init function.

AdrienClairembault:
👍

trasher (approved):

PR #12295, opened 2022-07-25
Workaround for config record ID requirement

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #12267

Replace #12270

### DIFF
--- a/front/config.form.php
+++ b/front/config.form.php
@@ -89,7 +89,6 @@ if (!empty($_POST['reset_translation_cache'])) {
     Html::redirect(Toolbox::getItemTypeFormURL('Config'));
 }
 
-$config_id = $_POST['id'];
-Config::displayFullPageForItem($config_id, ["config", "config"], [
+Config::displayFullPageForItem($_POST['id'], ["config", "config"], [
     'formoptions'  => "data-track-changes=true"
 ]);
--- a/front/config.form.php
+++ b/front/config.form.php
@@ -89,6 +89,7 @@ if (!empty($_POST['reset_translation_cache'])) {
     Html::redirect(Toolbox::getItemTypeFormURL('Config'));
 }
 
-Config::displayFullPageForItem(1, ["config", "config"], [
+$config_id = Config::getConfigIDForContext('core');
+Config::displayFullPageForItem($config_id, ["config", "config"], [
     'formoptions'  => "data-track-changes=true"
 ]);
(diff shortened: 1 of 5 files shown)


dsci-naw:
Hi, tested and worked fine for general config page! thanks

But in notification page didn't work /front/notificationmailingsetting.form.php

I think the new method is needed here too (found in search):
- https://github.com/glpi-project/glpi/blob/10.0/bugfixes/front/notificationajaxsetting.form.php#L51
- https://github.com/glpi-project/glpi/blob/10.0/bugfixes/front/notificationmailingsetting.form.php#L51

And also the new method here to save the notification settings:
- https://github.com/glpi-project/glpi/blob/10.0/bugfixes/src/NotificationSettingConfig.php#L52
- https://github.com/glpi-project/glpi/blob/10.0/bugfixes/src/NotificationSettingConfig.php#L69

cedric-anne (commented):
I do not like this kind of workarounds, but I dig a bit and handling it differently, like it was proposed in previours PR, would require lots of changes.
So, for me, this kind of change is OK.
Same kind of change should be added on notification settings.

cedric-anne (approved):

cconard96:
Based on @dsci-naw comment, ID 1 is hard coded other places for configs and saving doesn't work for them so this fix doesn't seem to be complete yet.

cconard96:
@dsci-naw Can you test again?

Some of the usages of ID 1 you mentioned seem to just be to handle the older databases which may not have the new configs table layout yet (useful during the update process).

dsci-naw:
All seems to work except general config "Save" that is not saving (notifications are ok anda saving now)

cconard96:
Are there any new errors in the GLPI logs (access-errors, php-errors, or sql-errors)?

dsci-naw:
unfortunately no, nothing

dsci-naw:
just one note, i think this can be used to reproduce the issue (not sure since don't know glpi database structure), change the id 1 to other id

> UPDATE glpi_configs SET id=1000 WHERE name = 'version';
> UPDATE glpi_logs SET items_id=1000 WHERE itemtype = 'Config';

and for rollback
> UPDATE glpi_configs SET id=1 WHERE name = 'version';
> UPDATE glpi_logs SET items_id=1 WHERE itemtype = 'Config';

dsci-naw:
with the latest change, now config options save. All seems to work fine.

trasher (approved):

cedric-anne (approved):

Issue #12297, opened 2022-07-26
Minor PHP error in logs when reloading marketplace


SI-AURAL:
### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

10.0.2

### Bug description

Minor bug with no impact

### Relevant log output

```shell
[2022-07-26 10:01:23] glpiphplog.CRITICAL:   *** Uncaught Exception TypeError: Glpi\Marketplace\View::getPaginationHtml(): Argument #2 ($total) must be of type int, string given, called in /srv/glpi/prod/ajax/marketplace.php on line 116 in /srv/glpi/prod/src/Marketplace/View.php at line 959
  Backtrace :
  ajax/marketplace.php:116                           Glpi\Marketplace\View::getPaginationHtml()
```

### Page URL

https://glpi.ad.auralyon.fr/front/marketplace.php

### Steps To reproduce

Access to marketplace then refresh plugin list

### Your GLPI setup information

<details><summary>Informations sur le système, l'installation et la configuration</summary><pre>GLPI 10.0.2 ( => /srv/glpi/prod)
Installation mode: TARBALL
Current language:fr_FR

</details><details><summary>Server</summary><pre> 
Operating system: Linux SRV-Parc 5.10.0-15-amd64 #1 SMP Debian 5.10.120-1 (2022-06-09) x86_64
PHP 8.1.7 fpm-fcgi (Core, FFI, PDO, Phar, Reflection, SPL, SimpleXML, Zend OPcache, apcu, bcmath, bz2, calendar, cgi-fcgi,
	ctype, curl, date, dom, exif, fileinfo, filter, ftp, gd, gettext, gmp, hash, iconv, imagick, imap, intl, json, ldap, libxml,
	mbstring, mcrypt, memcache, mysqli, mysqlnd, openssl, pcre, pdo_mysql, pdo_sqlite, posix, pspell, readline, session, shmop,
	snmp, soap, sockets, sodium, sqlite3, standard, sysvmsg, sysvsem, sysvshm, tidy, tokenizer, uploadprogress, xml, xmlreader,
	xmlrpc, xmlwriter, xsl, zip, zlib)
Setup: max_execution_time="600" memory_limit="128M" post_max_size="8M" safe_mode="" session.save_handler="files"
	upload_max_filesize="25M" 
Software: Apache ()
[...]

PR #12309, opened 2022-07-27
Enforce types of request params in Marketplace ajax endpoint

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #12297


cedric-anne (approved):
I added 2 commits.

1) Initial issue was due to call to `getPagination` action from the `installed` tab. This tab has no pagination, so I removed this call on this context, to save a useless query, and I removed fallback values in `getPagination` as these parameters are required to ensure displayed pagination is correct.

2) When using a string filter, pagination was not displayed, as total count of plugin was based on paginated list, instead of full list.

trasher (approved):

PR #12323, opened 2022-07-28
API: decode html

AdrienClairembault:
One issue with our API is that results are returned with encoded HTML:

image

image

It happens because we store HTML encoded data in the database (bad practice), but I don't see us changing that soon because of how much work would be involved.

What we can do now is decode the data before sending it in the API response.

Regarding the impact on existing code that use the API, I expect most code would already be decoding the content.
In this case, decoding a second time shouldn't break anything.

Is there some use case we should be worried about breaking with this change ?

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !24579

### DIFF
--- a/src/Api/API.php
+++ b/src/Api/API.php
@@ -1845,8 +1845,15 @@ abstract class API
                     if ($new_id === false) {
                         $failed++;
                     }
+
+                    $message = $this->getGlpiLastMessage();
+                    if (!$this->returnSanitizedContent()) {
+                        // Message may contains the created item name, which may
+                        // contains some encoded html
+                        $message = html_entity_decode($message);
+                    }
                     $current_res = ['id'      => $new_id,
-                        'message' => $this->getGlpiLastMessage()
+                        'message' => $message
                     ];
                 }
 
@@ -2330,7 +2337,7 @@ abstract class API
     /**
      * Get last message added in $_SESSION by Session::addMessageAfterRedirect
      *
-     * @return array  of messages
+     * @return string Last message
      */
     private function getGlpiLastMessage()
     {
--- a/src/Api/API.php
+++ b/src/Api/API.php
@@ -1028,7 +1028,12 @@ abstract class API
         }
 
         // Decode HTML
-        $fields = array_map(fn ($f) => is_string($f) ? html_entity_decode($f) : $f, $fields);
+        if (!$this->returnSanitizedContent()) {
+            $fields = array_map(
+                fn ($f) => is_string($f) ? html_entity_decode($f) : $f,
+                $fields
+            );
+        }
 
         return $fields;
     }
@@ -1300,7 +1305,12 @@ abstract class API
             }
 
             // Decode HTML
-            $fields = array_map(fn ($f) => is_string($f) ? html_entity_decode($f) : $f, $fields);
+            if (!$this->returnSanitizedContent()) {
+                $fields = array_map(
+                    fn ($f) => is_string($f) ? html_entity_decode($f) : $f,
+                    $fields
+                );
+            }
         }
        // Break reference
         unset($fields);
@@ -3361,5 +3371,15 @@ abstract class API
             "changeActiveEntities",
             "changeActiveProfile",
         ];
+    }
+
+    /**
+     * Will the API content be sanitized ?
+     *
+     * @return bool
+     */
+    public function returnSanitizedContent(): bool
+    {
+        return true;
     }
 }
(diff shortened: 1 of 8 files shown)


cconard96:
I don't agree with decoding every field in the response, at least not now given that it has always been encoded.

What if they were passing the encoded values as-is to another system that accepts encoded content only? Security-wise, I don't know how this could affect clients. On fields we expect to have rich-text, we sanitize the content to try an remove harmful content but how would a plain text field value like a name of "<script>alert()</script>" affect the client if it was sent as that instead of encoded? If the value was given as a response on its own later, then you could run into XSS through MIME sniffing (Client assumes the correct content type based on the content instead of the Content-Type header).

It sounds safer to put the responsibility on the API client/script to handle the encoded properties properly based on their needs.
I am not even sure it would be good to send decoded content in the new API (#12221) but if it was determined to be OK, that would be a better place to introduce the change where we could have more control over API versioning and handling individual response properties.

cedric-anne:
I agree, it may result in security issue in API clients.

To prevent introducing a BC-break in a non major version, we could add a `X-Glpi-Sanitized-Content` boolean header in 10.x version, having `true` as default value. If client pass `X-Glpi-Sanitized-Content: false`, in this case, we could return an unsanitized value.

cconard96:
Another possibility is a header that lets the client specify which fields they want to be unsanitized like:
`X-Glpi-Unsanitized-Fields: name, content`

cedric-anne:
> Another possibility is a header that lets the client specify which fields they want to be unsanitized like: `X-Glpi-Unsanitized-Fields: name, content`

In most cases, the escaping logic is global to an application. So having to declare all fields that should be returned in their original value (the unsanitized one) may be a pain for developper.

AdrienClairembault:
> I agree, it may result in security issue in API clients.
> 
> To prevent introducing a BC-break in a non major version, we could add a `X-Glpi-Sanitized-Content` boolean header in 10.x version, having `true` as default value. If client pass `X-Glpi-Sanitized-Content: false`, in this case, we could return an unsanitized value.

I like this option, I think I'll add it to the pull request

cedric-anne:
You should also add some tests using special chars to validate current behaviour. The fact that you were able to change encoding handling without having any failure in tests is a not very reassuring.

cedric-anne (approved):

trasher (approved):

cconard96:
I cannot test this at the moment. The API authenticates me but doesn't load my profiles or entities into the session. I do not think it is related to this PR, but I haven't found the cause.

AdrienClairembault:
I've added another commit to handle the API success message on item creation:

image

image

src/Api/API.php
```diff
@@ -1027,6 +1027,14 @@ protected function getItem($itemtype, $id, $params = [])
             );
         }
 
+        // Decode HTML
+        if (!$this->returnSanitizedContent()) {
+            $fields = array_map(
+                fn ($f) => is_string($f) ? html_entity_decode($f) : $f,
```
cedric-anne, line 1033: We should only revert sanitization process.
```suggestion
                fn ($f) => is_string($f) ? Sanitizer::decodeHtmlSpecialChars($f) : $f,
```

src/Api/API.php
```diff
@@ -1295,6 +1303,14 @@ protected function getItems($itemtype, $params = [], &$totalcount = 0)
                     ];
                 }
             }
+
+            // Decode HTML
+            if (!$this->returnSanitizedContent()) {
+                $fields = array_map(
+                    fn ($f) => is_string($f) ? html_entity_decode($f) : $f,
```
cedric-anne, line 1310:
```suggestion
                    fn ($f) => is_string($f) ? Sanitizer::decodeHtmlSpecialChars($f) : $f,
```

src/Api/API.php
```diff
@@ -1829,8 +1845,15 @@ protected function createItems($itemtype, $params = [])
                     if ($new_id === false) {
                         $failed++;
                     }
+
+                    $message = $this->getGlpiLastMessage();
+                    if (!$this->returnSanitizedContent()) {
+                        // Message may contains the created item name, which may
+                        // contains some encoded html
+                        $message = html_entity_decode($message);
```
cedric-anne, line 1853:
```suggestion
                        $message = Sanitizer::decodeHtmlSpecialChars($message);
```

cedric-anne (changes requested):

cedric-anne (approved):

orthagh (approved):

PR #12326, opened 2022-07-28
Solution type default should be 0 id.

miguelanruiz:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | N/A

When trying to add a solution, a null value is displayed so the form is submitted incorrectly by default:
image

Expected behavior shoul be defined with id as 0:
image

### DIFF
--- a/templates/components/itilobject/timeline/form_solution.html.twig
+++ b/templates/components/itilobject/timeline/form_solution.html.twig
@@ -168,7 +168,7 @@
                      {{ fields.dropdownField(
                         'SolutionType',
                         'solutiontypes_id',
-                        0,
+                        subitem.fields['solutiontypes_id'] ?? 0,
                         sol_type_lbl,
                         {
                            'full_width': true,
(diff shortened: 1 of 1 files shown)


templates/components/itilobject/timeline/form_solution.html.twig
```diff
@@ -168,7 +168,7 @@
                      {{ fields.dropdownField(
                         'SolutionType',
                         'solutiontypes_id',
-                        subitem.fields['solutiontypes_id'],
+                        0,
```
cconard96, line 171: This change would make it appear as if nothing is selected when editing a solution even if a choice was made and if the solution is saved, it would set unset the type.
The template field has the value set to 0 statically because it is not a field that gets stored in the database. It is only used while the form is open to change the other fields.

The easiest solution seems to be:
```suggestion
                        subitem.fields['solutiontypes_id'] ?? 0,
```

This would set the value to 0 if the current value is null or not set.

cconard96 (changes requested):

templates/components/itilobject/timeline/form_solution.html.twig
miguelanruiz: Great, is a good idea I didn't think about editing, thank you

cedric-anne (approved):

cconard96 (approved):

trasher (approved):

Issue #12336, opened 2022-07-29
[GLPI 10 bugfixes] Error display from Kanban with user surname with quotes


tsmr:
### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

[GLPI 10 Bugfixes]

### Bug description

User : 
Firstname : Xavier
Surname : D'ALES

Problem of display with mouse on Kanban : 

![image](https://user-images.githubusercontent.com/13178158/181779070-71aae284-a743-4807-b850-62fcca29db17.png)

PR #12361, opened 2022-08-01
Fix tooltip for team members with quote in name

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #12336

### DIFF
--- a/js/modules/Kanban/Kanban.js
+++ b/js/modules/Kanban/Kanban.js
@@ -1552,7 +1552,7 @@ class GLPIKanbanRights {
         /**
        * Generate a user image based on the user's initials.
        * @since 9.5.0
-       * @param {string} teammember The teammember array/object that represents the user.
+       * @param {{}} teammember The teammember array/object that represents the user.
        * @return {string} HTML image of the generated user badge.
        */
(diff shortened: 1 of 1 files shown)


js/modules/Kanban/Kanban.js
```diff
@@ -1552,7 +1552,7 @@ class GLPIKanbanRights {
         /**
        * Generate a user image based on the user's initials.
        * @since 9.5.0
-       * @param {string} teammember The teammember array/object that represents the user.
+       * @param {{}} teammember The teammember array/object that represents the user.
```
trasher, line 1555: Is that intended?

trasher (approved):

cedric-anne (approved):

js/modules/Kanban/Kanban.js
cconard96: Yes it was. The previous JSDoc type was a mistake and my IDE was giving an inspection error when trying to use the `replace` method on the `name` element.

PR #12363, opened 2022-08-02
Status dropdown: fix undefined values

AdrienClairembault:
When filtering the status dropdown, the values that didn't match the filter are shown are shown as "undefined" instead of being hidden from the results.

Before:

image

After:

image

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |

### DIFF
--- a/js/common.js
+++ b/js/common.js
@@ -891,7 +891,7 @@ var templateSelection = function (selection) {
 
 var templateItilStatus = function(option) {
     if (option === false) {
-        // option is false when element does not match searched terms
+        // Option is false when element does not match searched terms
         return null;
     }
     var status = option.id || 0;
@@ -946,6 +946,11 @@ var templateItilStatus = function(option) {
 };
 
 var templateValidation = function(option) {
+    if (option === false) {
+        // Option is false when element does not match searched terms
+        return null;
+    }
+
     var status = option.id || 0;
 
     var classes = "";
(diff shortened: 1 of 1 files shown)


js/common.js
```diff
@@ -892,6 +892,11 @@ var templateSelection = function (selection) {
 var templateItilStatus = function(option) {
```
cedric-anne, line 892: First argument passed to callback of function referenced in `templateSelection` is `false` when element do not match selection. Not sure if it is a bug in select2 or not.
```suggestion
var templateItilStatus = function(option) {
    if (option === false) {
        // option is false when element does not match searched terms
        return null;
    }
```

js/common.js
```diff
@@ -892,6 +892,11 @@ var templateSelection = function (selection) {
...
     var status = option.id || 0;
 
+    // Only show valid items
+    if (option.text == undefined) {
+        return null;
+    }
+
```
cedric-anne, line 899:
```suggestion
```

cedric-anne (changes requested):
All methods referenced in `templateSelection` option have to be fiexd. For instance, same bug appears when trying to filter priority.

AdrienClairembault:
Looking into the others functions

AdrienClairembault:
Done, I've also changed the priority rendering, I think the black dot wasn't displayed on purpose.

Before: 
image

After: 
image

This was it match the status dropdown that removes the icon for the "multi status" options.

cedric-anne (approved):

trasher (approved):

PR #12368, opened 2022-08-02
Allow any object in URL to a document

btry:
Related to #10945

Formcreator needs to create URL to documents with query parameters itemtype=xxxx&items_id=xxxx . (see https://github.com/pluginsGLPI/formcreator/pull/2907)

This PR avoids code duplication currently added in the above PR.

### DIFF
--- a/src/Toolbox.php
+++ b/src/Toolbox.php
@@ -2618,21 +2618,22 @@ class Toolbox
                         $document->getFromDB($id)
                         && strpos($document->fields['mime'], 'image/') !== false
                     ) {
-                        // append itil object reference in image link
-                        $itil_object = null;
-                        if ($item instanceof CommonITILObject) {
-                            $itil_object = $item;
-                        } else if (
-                            isset($item->input['_job'])
+                        // append object reference in image link
+                        $linked_object = null;
+                        if (
+                              !($item instanceof CommonITILObject)
+                              && isset($item->input['_job'])
                               && $item->input['_job'] instanceof CommonITILObject
                         ) {
-                            $itil_object = $item->input['_job'];
+                            $linked_object = $item->input['_job'];
+                        } else if ($item instanceof CommonDBTM) {
+                            $linked_object = $item;
                         }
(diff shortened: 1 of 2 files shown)


src/Toolbox.php
```diff
@@ -2618,21 +2618,21 @@ public static function convertTagToImage($content_text, CommonDBTM $item, $doc_d
...
-                        $itil_object = null;
-                        if ($item instanceof CommonITILObject) {
-                            $itil_object = $item;
+                        // append object reference in image link
+                        $linked_object = null;
+                        if ($item instanceof CommonDBTM) {
+                            $linked_object = $item;
                         } else if (
                             isset($item->input['_job'])
                               && $item->input['_job'] instanceof CommonITILObject
```
btry, line 2627: I don't know where _job key may come from. I leave it as is for now.

src/Toolbox.php
```diff
@@ -2618,21 +2618,21 @@ public static function convertTagToImage($content_text, CommonDBTM $item, $doc_d
...
-                        if ($item instanceof CommonITILObject) {
-                            $itil_object = $item;
-                        } else if (
-                            isset($item->input['_job'])
+                        // append object reference in image link
+                        $linked_object = null;
+                        if (!($item instanceof CommonITILObject)
+                              && isset($item->input['_job'])
                               && $item->input['_job'] instanceof CommonITILObject
                         ) {
```
btry, line 2626: From my understanding, this case matches ITILFollowup and ITILSolution

I think this is a inclear and risky expression. ITILFollowup and ITILSolution both inherit from CommonDBChild. I think we could make this better if we introduce a new (abstract) class as a parent of those itemtypes. As they show in a timeline, maybe a CommonITILTimeline class could be helpful (the classes have some identical code to deduplicate).

orthagh (approved):

trasher (approved):

cconard96 (approved):

cedric-anne (approved):

PR #12383, opened 2022-08-04
Console: fix build:compile_scss in production builds

AdrienClairembault:
The `tools` folder doesn't exist in GLPI official releases.
Therefore,  `tools/HEADER` can't be fetched by the `build:compile_scss` command.

We must either:
*  Include this file in the releases
*  Have a fallback to an empty string when it's missing (which is the change proposed by this PR)

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |

### DIFF
--- a/src/Console/Build/CompileScssCommand.php
+++ b/src/Console/Build/CompileScssCommand.php
@@ -175,10 +175,17 @@ class CompileScssCommand extends Command
      */
     private function getLicenceHeaderString(): string
     {
-       // Extract header lines
-        $lines = file(GLPI_ROOT . '/tools/HEADER');
+        // Extract header lines
+        $header_file = GLPI_ROOT . '/tools/HEADER';
 
-       // Add * prefix on all lines
+        if (!file_exists($header_file)) {
+            // Production build, there is no tools/HEADER file
+            return "";
+        }
+
+        $lines = file($header_file);
+
+        // Add * prefix on all lines
         $lines = array_map(
             function ($line) {
                 $line_prefix = ' * ';
@@ -187,7 +194,7 @@ class CompileScssCommand extends Command
             $lines
         );
 
-       // Surround by opening and closing lines
+        // Surround by opening and closing lines
         $lines = array_merge(["/**\n"], $lines, [" */\n\n"]);
 
         return implode($lines);
--- a/src/Console/Build/CompileScssCommand.php
+++ b/src/Console/Build/CompileScssCommand.php
@@ -119,8 +119,6 @@ class CompileScssCommand extends Command
             }
         }
 
-        $licence_header = $this->getLicenceHeaderString();
-
         foreach ($files as $file) {
             $output->writeln(
                 '<comment>' . sprintf('Processing "%s".', $file) . '</comment>',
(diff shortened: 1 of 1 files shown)


trasher:
I guess HEADER file should be included; GLPI license should be applied on built CSS.

cedric-anne:
I propose to move the licence header template file into root path (see https://github.com/glpi-project/tools/pull/57). This will permit to preserve it without having to modify the way dev files are cleaned.

```diff
diff --git a/tools/HEADER b/.licence-header
similarity index 100%
rename from tools/HEADER
rename to .licence-header
diff --git a/src/Console/Build/CompileScssCommand.php b/src/Console/Build/CompileScssCommand.php
index be46277df6..6ab2a37def 100644
--- a/src/Console/Build/CompileScssCommand.php
+++ b/src/Console/Build/CompileScssCommand.php
@@ -176,7 +176,7 @@ class CompileScssCommand extends Command
     private function getLicenceHeaderString(): string
     {
         // Extract header lines
-        $header_file = GLPI_ROOT . '/tools/HEADER';
+        $header_file = GLPI_ROOT . '/.licence-header';
 
         if (!file_exists($header_file)) {
             // Production build, there is no tools/HEADER file
diff --git a/tools/build_glpi.sh b/tools/build_glpi.sh
index 62fb4426f5..18e8f3c380 100755
--- a/tools/build_glpi.sh
+++ b/tools/build_glpi.sh
@@ -69,8 +69,8 @@ composer update nothing --ansi --no-interaction --ignore-platform-reqs --no-dev
 # Remove user generated files (i.e. cache and log from CLI commands ran during release)
 find $WORKING_DIR/files -depth -mindepth 2 ! -iname "remove.txt" -exec rm -rf {} \;
 
-# Remove hidden files and directory, except .htaccess files
-find $WORKING_DIR -depth \( -iname ".*" ! -iname ".htaccess" \) -exec rm -rf {} \;
+# Remove hidden files and directory, except .htaccess files and .licence-header file
+find $WORKING_DIR -depth \( -iname ".*" ! -iname ".htaccess" ! -iname ".licence-header" \) -exec rm -rf {} \;
 
 # Remove useless dev files and directories
 dev_nodes=(
 ```

cconard96:
How important is it to have the header added again?
In my testing for #12277, the compiler didn't strip out any comments from the source CSS files so there was a copy of the GLPI header before every section from our CSS files.

cedric-anne:
Your are right, comments are preserved, and licence headers are so present in compiled scss if they are present in base SCSS file. So the `getLicenceHeaderString()` method is useless and can even be removed.

AdrienClairembault:
I've removed it in the latest commit.
After doing a test run, I've seen that each compiled file contains the license multiple times (40+ times):

image


image

I suppose it relates to what @cconard96 said and the license is included each time a file is imported.

trasher (approved):

cedric-anne:
> Is this something to worry about ?

Probably in another PR. Maybe we could use `setOutputStyle(OutputStyle::COMPRESSED)` and use the `/*!` comment prefix in main scss files only, but not in scss files prefixes by a `_` (those that are not meant to be used directly).

cedric-anne (approved):

PR #12437, opened 2022-08-12
Ticket: fix reopen by requester group

AdrienClairembault:
Here `post-only` is part of the `Pro services` group and add a new followup to a pending ticket.

image
-> Ticket is still pending which is incorrect: the ticket must be "reopened" when a requester add a followup.

The issue comes from `needReopen` calling `isGroup` incorrectly (sending an array of groups id instead of a single group id).
These changes:
- Fix the incorrect method parameter sent in `needReopen`
- Enforce types in `isGroup` to prevent future issues
- Improve unit tests coverage

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |

### DIFF
--- a/src/CommonITILObject.php
+++ b/src/CommonITILObject.php
@@ -887,7 +887,8 @@ abstract class CommonITILObject extends CommonDBTM
      *
      * @return bool
      **/
-    public function isGroup(int $type, int $groups_id): bool
+    // FIXME add params typehint in GLPI 10.1
+    public function isGroup($type, $groups_id): bool
     {
         if (isset($this->groups[$type])) {
             foreach ($this->groups[$type] as $data) {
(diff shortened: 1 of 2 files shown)


src/CommonITILObject.php
```diff
@@ -870,14 +882,13 @@ public function isUser($type, $users_id)
...
-     * @param integer $type      type to search (see constants)
-     * @param integer $groups_id group ID
+     * @param int $type      type to search (see constants)
+     * @param int $groups_id group ID
      *
-     * @return boolean
+     * @return bool
      **/
-    public function isGroup($type, $groups_id)
+    public function isGroup(int $type, int $groups_id): bool
```
trasher, line 890: Not sure this won't cause issues; I think we should avoid this kind of changes in a bugfixes release.

AdrienClairembault: Indeed, let's be safe.
Done ->[8d6219a](https://github.com/glpi-project/glpi/pull/12437/commits/8d6219af72b27b54e19a37958c372121c10ef4b3)

trasher (approved):

orthagh (approved):

cedric-anne (approved):

PR #12440, opened 2022-08-12
Collasped sidebar tweaks

AdrienClairembault:
A few minor improvements to the collasped sidebar.

Before:
image

After: 
image

Side by side:
image    --- image

The issues that were bugging me with the sidebar were:

1\) Not enough top padding for the logo.

image

Now the padding match the padding used for the "non-collapsed" logo:
image

2\) The "go to menu" icon didn't match the size of the others icons.

image

It was even more noticeable if you swaped the icon for the same as the one below:

image

3\)  Not enough spacing between each menu:

It felt like the icons were accumulated at the top of the page.
They are now spaced a bit more evenly, matching the spacing of the "non-collapsed" menu.

4\) Can't read the active menu:

I don't think this one was intentional but it definitely needed to be fixed:

image

After fix:

image

I realize that some of these changes might be subjective, so let me know what you think :) 

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |

### DIFF
--- a/css/includes/components/_global-menu.scss
+++ b/css/includes/components/_global-menu.scss
@@ -65,6 +65,14 @@
     }
 }
 
+.navbar-vertical {
+    .trigger-fuzzy {
+        border: 0;
+        font-size: 1em;
+        font-weight: 400;
+    }
+}
+
 .sidebar {
     & ~ .navbar {
         z-index: $zindex-fixed;
@@ -149,11 +157,6 @@
     }
 
     body.navbar-collapsed {
-        .trigger-fuzzy {
-            border: 0;
-            font-size: 1em;
-        }
-
         #navbar-menu .nav-link {
             padding: 0.58rem 1.5rem !important;
         }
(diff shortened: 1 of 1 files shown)


trasher (approved):

orthagh (commented):
> The "go to menu" icon didn't match the size of the others icons.

It affects also the non collapsed layout. if you can push a fix for it.

> I realize that some of these changes might be subjective, so let me know what you think :)

seems good for me

AdrienClairembault:
> It affects also the non collapsed layout. if you can push a fix for it.

Done 6c95915.

Note that this icon difference is because its a button instead of an `<li>` element.
Maybe another PR could be done to convert it to an `<li>` element for the vertical layout.

orthagh:
as seen together and to be picky, fix of GLPI logo position could be done also on expanded state

orthagh (approved):
Go for merge, logo part will be fixed in another PR

PR #12448, opened 2022-08-17
fix(core): load 'inventory' context to CFG_GLPI

stonebuzz:
From inventory context, ```$CFG_GLPI['inventory_frequency']``` is always set to ```DEFAULT_FREQUENCY``` because

GLPI only load  legacy configuration with ```core``` context

Fix agent expiration always set to 24 hours

| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #12116

### DIFF
--- a/src/Config.php
+++ b/src/Config.php
@@ -3027,8 +3027,9 @@ HTML;
         } else {
            // multiple rows = 0.85+ config
             $values = [];
+            $allowed_context = ['core', 'inventory'];
             foreach ($iterator as $row) {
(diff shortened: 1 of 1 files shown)


g-bougard (approved):
Just tested and it works as expected.
Thx

trasher (approved):

PR #12454, opened 2022-08-18
ID in front for all ITIL Objects in Search results

VladoTTX:
Why not to have interface of Search results to be same for all ITIL Objects (Ticket, Changes, Problems..)

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | NA

### DIFF
--- a/src/Search.php
+++ b/src/Search.php
@@ -637,8 +637,8 @@ class Search
         $to_add_view = array_diff($p['sort'], $data['toview']);
         array_push($data['toview'], ...$to_add_view);
 
-       // Special case for Ticket : put ID in front
-        if ($itemtype == 'Ticket') {
+       // Special case for CommonITILObjects : put ID in front
+        if (is_a($itemtype,'CommonITILObject',true)) {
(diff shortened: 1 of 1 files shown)


src/Search.php
```diff
@@ -637,8 +637,8 @@ public static function prepareDatasForSearch($itemtype, array $params, array $fo
         $to_add_view = array_diff($p['sort'], $data['toview']);
         array_push($data['toview'], ...$to_add_view);
 
-       // Special case for Ticket : put ID in front
-        if ($itemtype == 'Ticket') {
+       // Special case for CommonITILObjects : put ID in front
+        if (is_a($itemtype,'CommonITILObject',true)) {
```
trasher, line 641:
```suggestion
        if (is_a($itemtype, CommonITILObject::class, true)) {
```

orthagh (approved):

trasher (approved):

PR #12459, opened 2022-08-22
fix(timeline):  the img display size in timeline is not correct.

stonebuzz:
When the user copies/pastes the image from richtext and changes its size (smaller)

on display, the size is reworked by GLPI to match the size of the text area -> perfect

But when GLPI generates the notification, it takes into account the ```width``` and ```height``` in the ```img``` DOM

This PR allow GLPI to handle image size.

**EDIT**
The real issue is -> " the display size in timeline is not correct."
Images are always risize by GLPI even if user change it

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !24145

### DIFF
--- a/src/RichText/RichText.php
+++ b/src/RichText/RichText.php
@@ -355,10 +355,10 @@ final class RichText
                     $gallery = self::imageGallery([
                         [
                             'src' => $docsrc,
-                            'w'   => $imgsize[0], //used by photoSwipe gallery
-                            'h'   => $imgsize[1], //used by photoSwipe gallery
-                            'user_w' => $width, //use from timeline
-                            'user_h' => $height,//use from timeline
+                            'w'   => $imgsize[0],
+                            'h'   => $imgsize[1],
+                            'thumbnail_w' => $width,
+                            'thumbnail_h' => $height,
                         ]
                     ]);
                     $content = str_replace($img_tag, $gallery, $content);
--- a/src/RichText/RichText.php
+++ b/src/RichText/RichText.php
@@ -453,7 +453,9 @@ final class RichText
             }
             $out .= "<figure itemprop='associatedMedia' itemscope itemtype='http://schema.org/ImageObject'>";
             $out .= "<a href='{$img['src']}' itemprop='contentUrl' data-index='0'>";
-            $out .= "<img src='{$img['thumbnail_src']}' itemprop='thumbnail' loading='lazy' width='{$img['user_w']}' height='{$img['user_h']}' >";
+            $width_attr = isset($img['thumbnail_w']) ? "width='{$img['thumbnail_w']}'" : "";
+            $height_attr = isset($img['thumbnail_h']) ? "height='{$img['thumbnail_h']}'" : "";
+            $out .= "<img src='{$img['thumbnail_src']}' itemprop='thumbnail' loading='lazy' {$width_attr} {$height_attr}>";
             $out .= "</a>";
             $out .= "</figure>";
         }
--- a/src/RichText/RichText.php
+++ b/src/RichText/RichText.php
@@ -339,12 +339,27 @@ final class RichText
             if ($document->getFromDB($docid)) {
                 $docpath = GLPI_DOC_DIR . '/' . $document->fields['filepath'];
                 if (Document::isImage($docpath)) {
+
+                    //find width / height define by user
+                    $width = null;
+                    if (preg_match("/width=[\"|'](\d+)(\.\d+)?[\"|']/", $img_tag, $wmatches)) {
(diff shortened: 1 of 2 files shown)


trasher (approved):

cedric-anne (changes requested):
I think resizing was wanted here, to ensure that image is sent using the size defined in rich text, to have a rendering similar in both browser and mail contexts.
There is probably something wrong in width/height handling.

stonebuzz:
Currently : 

With this image
image


- Paste image (GLPI resize it to match rich text size)
image

- Manually resize the image and save 
image

- GLPI display like this
image

- And like this when editing the follow-up
image

cedric-anne:
IMHO, the display size in timeline is not correct.

stonebuzz:
Now, ```img``` from ```timeline``` or ```richtext``` use ```user``` size if exist 
image
image

```PhotoSwipe``` gallery continues to use the real size of the image
image

```NotificationEventMailing``` continues to use ```user``` size if exist

**Review by trasher — APPROVED** [review]

File: src/RichText/RichText.php [hunk 1]
```diff
@@ -339,12 +339,27 @@ private static function replaceImagesByGallery(string $content): string
             if ($document->getFromDB($docid)) {
                 $docpath = GLPI_DOC_DIR . '/' . $document->fields['filepath'];
                 if (Document::isImage($docpath)) {
+
```
trasher, line 342:
```suggestion
```

cedric-anne (changes requested):
Initial problem is that the HTML tag may contain only `height` or `width` but not both. In that case, for the undefined dimension, `Document::getImage()` uses a default value for `mail` context. IMHO, if `height` or `width` is defined, but not both, we should compute the other dimension using same ratio, and used default values only if both are not defined.

src/RichText/RichText.php
```diff
@@ -339,12 +339,26 @@ private static function replaceImagesByGallery(string $content): string
...
                     $gallery = self::imageGallery([
                         [
                             'src' => $docsrc,
-                            'w'   => $imgsize[0],
-                            'h'   => $imgsize[1]
+                            'w'   => $imgsize[0], //used by photoSwipe gallery
+                            'h'   => $imgsize[1], //used by photoSwipe gallery
+                            'user_w' => $width, //use from timeline
+                            'user_h' => $height,//use from timeline
```
cedric-anne, line 361: `thumbnail` prefix will be easier to understand. No need to comment it.
```suggestion
                            'w'   => $imgsize[0],
                            'h'   => $imgsize[1],
                            'thumbnail_w' => $width,
                            'thumbnail_h' => $height,
```

src/RichText/RichText.php
```diff
@@ -439,7 +453,7 @@ private static function imageGallery(array $imgs, array $options = []): string
             }
             $out .= "<figure itemprop='associatedMedia' itemscope itemtype='http://schema.org/ImageObject'>";
             $out .= "<a href='{$img['src']}' itemprop='contentUrl' data-index='0'>";
-            $out .= "<img src='{$img['thumbnail_src']}' itemprop='thumbnail' loading='lazy'>";
+            $out .= "<img src='{$img['thumbnail_src']}' itemprop='thumbnail' loading='lazy' width='{$img['user_w']}' height='{$img['user_h']}' >";
```
cedric-anne, line 456: Values were not required before this PR, so missing values have to be handled.
```suggestion
            $width_attr = isset($img['thumbnail_w']) ? "width='{$img['thumbnail_w']}'" : "";
            $height_attr = isset($img['thumbnail_h']) ? "height='{$img['thumbnail_h']}'" : "";
            $out .= "<img src='{$img['thumbnail_src']}' itemprop='thumbnail' loading='lazy' {$width_attr} {$height_attr}>";
```

cedric-anne (changes requested):

cedric-anne:
> Initial problem is that the HTML tag may contain only `height` or `width` but not both. In that case, for the undefined dimension, `Document::getImage()` uses a default value for `mail` context. IMHO, if `height` or `width` is defined, but not both, we should compute the other dimension using same ratio, and used default values only if both are not defined.

I am not really sure to understand this piece of code. I guess it is quiet obsolete and could probably be simplified (it should not be a problem anymore to send images in real size in emails in 2022), but it is another topic.

So this PR is OK for me.

cedric-anne (approved):

trasher (approved):

orthagh (approved):

cconard96 (approved):

Issue #12472, opened 2022-08-23
10.0.3-dev: Network port metrics - no graph Input/Output 


WebGregGit:
### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

10.0.3-dev

### Bug description

1. Go to the Assets > Network devices
2. Open some switch.
3. Go to the 'Network' ports tab.
4. Choose an active port and enter it.
5. Go to the 'Network port metrics' tab.

![image](https://user-images.githubusercontent.com/68922721/186199518-a361a315-8739-4d35-a1b6-3a489bb84e08.png)

It looks like something should be displayed here, but there is no data.

### Page URL

front/networkport.form.php?id=

### Anything else?

I found this [https://github.com/glpi-project/glpi/pull/10987](https://github.com/glpi-project/glpi/pull/10987) but I have made these changes, so it's not relevant

cconard96:
Do you have any errors in your browser's console (F12 key > Console tab)?

WebGregGit:
Hi @cconard96 

Yes I do

![image](https://user-images.githubusercontent.com/68922721/186345722-279032d5-954a-4af2-a257-cb993fa8f6e3.png)

And after expanded:

<html>
<body>


  | (anonymous) | @ | VM233:5
-- | -- | -- | --
  | mightThrow | @ | jquery.js:3766
  | process | @ | jquery.js:3834
  | setTimeout (async) |   |  
  | jQuery.readyException | @ | jquery.js:4058
  | (anonymous) | @ | jquery.js:4078
  | mightThrow | @ | jquery.js:3766
  | process | @ | jquery.js:3834
  | setTimeout (async) |   |  
  | (anonymous) | @ | jquery.js:3872
  | fire | @ | jquery.js:3500
  | fireWith | @ | jquery.js:3630
  | fire | @ | jquery.js:3638
  | fire | @ | jquery.js:3500
  | fireWith | @ | jquery.js:3630
  | process | @ | jquery.js:3854
  | setTimeout (async) |   |  
  | (anonymous) | @ | jquery.js:3872
  | fire | @ | jquery.js:3500
  | add | @ | jquery.js:3559
  | (anonymous) | @ | jquery.js:3892
  | eval | @ | VM195:970
  | obj.<computed> | @ | VM195:180
  | then | @ | jquery.js:3877
  | jQuery.fn.ready | @ | jquery.js:4072
[...]

PR #12476, opened 2022-08-24
fix(networkport): correctly display metrics

stonebuzz:
Correctly display NetworkPort metrics

image

- Fix load of chartist js library 
- Convert bytes to megabytes (dashboard label more readable)

| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #12472

### DIFF
--- a/src/NetworkPortMetrics.php
+++ b/src/NetworkPortMetrics.php
@@ -142,27 +142,24 @@ class NetworkPortMetrics extends CommonDBChild
         $bytes_series = [];
         $errors_series = [];
         $labels = [];
-        $i = 0;
         foreach ($raw_metrics as $metrics) {
             $date = new \DateTime($metrics['date']);
             $labels[] = $date->format(__('Y-m-d'));
-            unset($metrics['id'], $metrics['date'], $metrics[static::$items_id]);
+            unset($metrics['id'], $metrics['date'], $metrics['date_creation'], $metrics['date_mod'], $metrics[static::$items_id]);
 
             $bytes_metrics = $metrics;
             unset($bytes_metrics['ifinerrors'], $bytes_metrics['ifouterrors']);
             foreach ($bytes_metrics as $key => $value) {
                 $bytes_series[$key]['name'] = $this->getLabelFor($key);
-                $bytes_series[$key]['data'][] = $value;
+                $bytes_series[$key]['data'][] = round($value / 1024 / 1024, 0); //convert bytes to megabytes
             }
 
             $errors_metrics = $metrics;
             unset($errors_metrics['ifinbytes'], $errors_metrics['ifoutbytes']);
             foreach ($errors_metrics as $key => $value) {
                 $errors_series[$key]['name'] = $this->getLabelFor($key);
-                $errors_series[$key]['data'][] = $value;
+                $errors_series[$key]['data'][] = round($value / 1024 / 1024, 0); //convert bytes to megabytes
             }
-
-            ++$i;
         }
 
         $bytes_bar_conf = [
@@ -176,9 +173,8 @@ class NetworkPortMetrics extends CommonDBChild
             'distributed' => false
         ];
 
-       //display graph
-        Html::requireJs('charts');
-        echo "<div class='dashboard netports_metrics bytes'>";
+       //display bytes graph
+        echo "<div class='netports_metrics bytes'>";
         echo Widget::multipleLines($bytes_bar_conf);
         echo "</div>";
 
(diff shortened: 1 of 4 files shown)


stonebuzz:
Same error as #11878
From chrome web browser
 
image

src/NetworkPortMetrics.php
```diff
@@ -142,27 +142,24 @@ public function showMetrics(NetworkPort $netport)
...
             foreach ($errors_metrics as $key => $value) {
                 $errors_series[$key]['name'] = $this->getLabelFor($key);
-                $errors_series[$key]['data'][] = $value;
+                $errors_series[$key]['data'][] = round($value / 1024 / 1024, 0); //convert bytes to megabytes
             }
-
-            ++$i;
         }
 
         $bytes_bar_conf = [
```
cconard96, line 165: The label in the `bytes_bar_conf` would need updated to indicate that the data is in megabytes now.

cconard96 (dismissed):

cconard96:
> Same error as #11878 From chrome web browser
> 
> image

For me, it had trouble parsing the chart height. It expected a number or a string starting with a number and ignoring everything else (probably to support something like '100px'?), but was getting a CSS `calc` statement. It didn't make any sense to me though since I hadn't seen any issues with widgets on the dashboards.

stonebuzz:
maybe -> https://github.com/chartist-js/chartist/issues/1099

src/NetworkPortMetrics.php
stonebuzz: fixed

cconard96:
Commenting out the draw listener that contains the animate call didn't resolve the issue. Only removing the height option or specifying a numeric value worked.
https://github.com/glpi-project/glpi/blob/ae93c956786a2021b0b27b300ee999f993167d47/src/Dashboard/Widget.php#L1588

stonebuzz:
> Commenting out the draw listener that contains the animate call didn't resolve the issue. Only removing the height option or specifying a numeric value worked.
> 
> https://github.com/glpi-project/glpi/blob/ae93c956786a2021b0b27b300ee999f993167d47/src/Dashboard/Widget.php#L1588

I confirm

cconard96:
I think I found a solution that would remove the need for a calculated height and seems to allow both the network port metrics and dashboard line charts to display correctly.

1. Add a container div around the `.chart` one so that the automatic height is 100% of this element and not the one that also contains the `.main-label` and `.main-icon` elements (when within a dashboard).
2. Remove the height option when initializing the chart
3. I think the style that uses the calculated height could be removed too. I didn't notice a difference in Chrome after making the other changes.

```diff
diff --git a/src/Dashboard/Widget.php b/src/Dashboard/Widget.php
index d78e8a6e95..1a1aa92d6f 100644
--- a/src/Dashboard/Widget.php
+++ b/src/Dashboard/Widget.php
@@ -1513,20 +1513,14 @@ JAVASCRIPT;
             })";
         }
 
-        $height = "calc(100% - 1px)";
         $legend_options = "";
         if ($p['legend']) {
-            $height = "calc(100% - 40px)";
             $legend_options = "
             Chartist.plugins.legend(),";
         }
 
         $html = <<<HTML
       <style>
-         /** fix chrome resizing height when animating svg (don't know why) **/
-      #{$chart_id} .ct-chart-line {
-         min-height: $height;
-      }
 
       #{$chart_id} {
          background-color: {$p['color']};
@@ -1564,11 +1558,13 @@ JAVASCRIPT;
       {$palette_style}
       </style>
 
-      <div class="card g-chart $class"
-           id="{$chart_id}">
-         <div class="chart ct-chart"></div>
-         <span class="main-label">{$p['label']}</span>
-         <i class="main-icon {$p['icon']}"></i>
+      <div>
+          <div class="card g-chart $class"
+               id="{$chart_id}">
+             <div class="chart ct-chart"></div>
+             <span class="main-label">{$p['label']}</span>
+             <i class="main-icon {$p['icon']}"></i>
+          </div>
       </div>
 HTML;
 
@@ -1585,7 +1581,6 @@ HTML;
             series: {$json_series},
          }, {
             width: '100%',
-            height: '{$height}',
             fullWidth: true,
             chartPadding: {
                right: 40

```

trasher (approved):

orthagh:
> I think I found a solution that would remove the need for a calculated height and seems to allow both the network port metrics and dashboard line charts to display correctly.

I feared that it could broke the advanceddashboard plugin where we had some issue with height but it seems everything work without flaws, so lgtm.

orthagh:
To note, in `Widget.php` there is other methods relying on this calculated height and should be fixed in the same way

stonebuzz:
> To note, in `Widget.php` there is other methods relying on this calculated height and should be fixed in the same way

I've listed :
- getBarsGraph
- pie

orthagh (approved):

cconard96 (approved):

Issue #12484, opened 2022-08-25
10.0.3-dev - Bad display of link icon


WebGregGit:
### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

10.0.3-dev

### Bug description

![image](https://user-images.githubusercontent.com/68922721/186607057-e1b967dd-4272-4fd6-89be-cbcfc8cf977d.png)

The 7th port is connected. But when I open it, I don't know why here I see it disconnected.

![image](https://user-images.githubusercontent.com/68922721/186607789-f981433b-dca7-4d5d-b3eb-37468fa52d86.png)

### Page URL

front/networkport.form.php?id=xx

### Steps To reproduce

1. Go to the Assets > Network devices.
2. Open some switch.
3. Go to the 'Network' ports tab.
4. Choose an active port and enter it.
5. Check the "Connected to" status.

### Your GLPI setup information

GLPI 10.0.3-dev ( => /var/www/html/glpi) (build date: 2022-08-24)
Installation mode: TARBALL
Current language:en_GB

Operating system: Linux chronos 5.4.0-124-generic https://github.com/glpi-project/glpi/issues/140-Ubuntu SMP Thu Aug 4 02:23:37 UTC 2022 x86_64

PHP 7.4.3 (cli) (built: Jun 13 2022 13:43:30) ( NTS )
Copyright (c) The PHP Group
Zend Engine v3.4.0, Copyright (c) Zend Technologies
with Zend OPcache v7.4.3, Copyright (c), by Zend Technologies

stonebuzz:
This icon does not indicate that it is disconnected, it is a button to disconnect it =)

This is a PR to make btn more understandable -> https://github.com/glpi-project/glpi/pull/12485

Best regards

WebGregGit:
Sorry :( 
And thank you.

PR #12485, opened 2022-08-25
fix(ui): btn render to be more understandable

stonebuzz:
Before
image

After

image

| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #12484

### DIFF
--- a/src/NetworkPortInstantiation.php
+++ b/src/NetworkPortInstantiation.php
@@ -928,7 +928,7 @@ class NetworkPortInstantiation extends CommonDBChild
                         _x('button', 'Disconnect'),
                         ['id' => $relations_id],
                         'fa-unlink netport',
-                        'class="btn btn-sm btn-danger"'
+                        'class="btn btn-sm btn-outline-danger"'
                     );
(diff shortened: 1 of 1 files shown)


src/NetworkPortInstantiation.php
```diff
@@ -928,7 +928,7 @@ public static function showConnection($netport, $edit = false)
                         _x('button', 'Disconnect'),
                         ['id' => $relations_id],
                         'fa-unlink netport',
-                        'class="btn btn-sm btn-ghost-danger"'
```
orthagh, line 931: prefer an outline version to avoid attract attention on a secondary button.
i agree ghost was confusing in the current case (confusion mostly on color state with possible action)

stonebuzz: done =)

trasher (approved):

orthagh (approved):

PR #12486, opened 2022-08-25
fix table alias naming in nested join

btry:
While designing a search option I found the new need to mention a linkfield in a nested ['beforejoin' array](https://glpi-developer-documentation.readthedocs.io/en/master/devapi/search.html#join-parameters).

Thre is no place in GLPI where such search option exists, and the search engine fails to build the JOIN chain with such context.

I found that 2 consecutive JOIN are broken because of a misnamed table alias. The inner linkfield is missing.

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #number

### DIFF
--- a/src/Search.php
+++ b/src/Search.php
@@ -5691,6 +5691,9 @@ JAVASCRIPT;
                     if (!isset($tab['joinparams']['nolink']) || !$tab['joinparams']['nolink']) {
                         $cleanrt     = $intertable;
                         $complexjoin = self::computeComplexJoinID($interjoinparams);
+                        if (!empty($interlinkfield) && ($interlinkfield != getForeignKeyFieldForTable($intertable))) {
+                            $intertable .= "_" . $interlinkfield;
+                        }
                         if (!empty($complexjoin)) {
                             $intertable .= "_" . $complexjoin;
                         }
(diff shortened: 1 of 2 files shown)


src/Search.php
```diff
@@ -5691,6 +5691,9 @@ public static function addLeftJoin(
                     if (!isset($tab['joinparams']['nolink']) || !$tab['joinparams']['nolink']) {
                         $cleanrt     = $intertable;
                         $complexjoin = self::computeComplexJoinID($interjoinparams);
+                        if (!empty($interlinkfield) && ($interlinkfield != getForeignKeyFieldForTable($intertable))) {
+                            $intertable .= "_" . $interlinkfield;
+                        }
```
btry, line 5696: This is required to compute the same alias as the computation done in line 5997

orthagh (approved):

trasher:
I'm not confident with this change; and there is no test around it :/

trasher (approved):
Please fix CS

orthagh (approved):

PR #12504, opened 2022-08-26
Delegate validation

btry:
requires:
- #12486

To merge with this PR: 
- https://github.com/glpi-project/doc/pull/105

internal ref: 23396

### Feature 1: select the responsible of the requester to request for validation

When there are several requesters , only the 1st user found (ordered by ID of ticket_user relation) is added in the dropdown

image

### Feature 2: users may choose substitutes to delegate validation to them. 

A date range restricts the effect of the delegation. It is possible to nave no date limitation, only a beginning date , only a end date, or both.

image

image

### Additions for rebase of this PR onto main branch

In main branch I see that we can set a group as validator. The column user_id_actual_validate brings additional value here as it records the user who actually validate the ITIL object.

I added support for resolution of substitute against validator group. 

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #number


install/migrations/update_10.0.2_to_10.0.3/validatorsubstitute.php
```diff
@@ -0,0 +1,94 @@
...
+
+$migration->changeField(
+	'glpi_ticketvalidations',
+	'users_id_validate',
+	'users_id_validate',
+	"int {$default_key_sign} NOT NULL DEFAULT '0'",
+	[
+		'comment' => 'User in charge of the validation',
+	]
+);
```
btry, line 88: Just add a comment in the schema

src/CommonITILObject.php
```diff
@@ -6675,16 +6675,16 @@ public function getTimelineItems(array $options = [])
             class_exists($validation_class) && $params['with_validations']
             && ($validation_class::canView() || $params['bypass_rights'])
         ) {
-            $valitation_obj   = new $validation_class();
```
btry, line 6678: typo in var name. Several lines below changed with this fix

src/CommonITILObject.php
```diff
@@ -6675,16 +6675,16 @@ public function getTimelineItems(array $options = [])
...
-            $validations = $valitation_obj->find([$foreignKey => $this->getID()]);
+            $validation_obj   = new $validation_class();
+            $validations = $validation_obj->find([$foreignKey => $this->getID()]);
             foreach ($validations as $validations_id => $validation) {
-                $canedit = $valitation_obj->can($validations_id, UPDATE);
-                $cananswer = ($validation['users_id_validate'] === Session::getLoginUserID() &&
```
btry, line 6682: check against current user was already implemented on CommonITILValidation::canValidate(). The new feature requires to check date range in a user row then replacing the comparison here by the method, as now a SQL query  is necessary.

src/CommonITILObject.php
```diff
@@ -9168,4 +9168,34 @@ protected function transformActorsInput(array $input): array
...
+        $row = array_pop($rows);
+        if ($row === null) {
+            return null;
+        }
+        $user = User::getById($row['users_id']);
+        if (!($user instanceof User)) {
+            return null;
+        }
+        return $user;
+    }
```
btry, line 9200: To add responsible of the requester of a ITIL object we pick the 1st user added to the ITIL object, sorted by row ID.

src/CommonITILValidation.php
```diff
@@ -197,9 +197,30 @@ public static function canValidate($items_id)
...
                 static::$items_id    => $items_id,
-                'users_id_validate'  => Session::getLoginUserID()
+                'OR' => [
+                    'users_id_validate'  => Session::getLoginUserID(),
+                    'AND' => [
+                        User::getTable() . '.substitution_start_date' => ['<=', $_SESSION['glpi_currenttime']],
+                        User::getTable() . '.substitution_end_date' => ['>=', $_SESSION['glpi_currenttime']],
+                        ValidatorSubstitute::getTable() . '.users_id_substitute' => Session::getLoginUserID(),
+                    ],
+                ]
```
btry, line 223: add the substitutes of the validator to the users allowed to validate an ITIL object. Takes into account the date range set by the delegator.

src/CommonITILValidation.php
```diff
@@ -916,7 +937,7 @@ public function showSummary(CommonDBTM $item)
 
         $colonnes = ['', _x('item', 'State'), __('Request date'), __('Approval requester'),
             __('Request comments'), __('Approval status'),
-            __('Approver'), __('Approval comments'), __('Documents')
+            __('Approver'), __('Actual approver'), __('Approval comments'), __('Documents')
```
btry, line 940: Add a column header to show the user who actually validated an ITIL object

src/CommonITILValidation.php
```diff
@@ -1143,6 +1165,19 @@ public function rawSearchOptions()
...
+            'field'              => 'name',
+            'linkfield'          => 'users_id_actual_validate',
+            'name'               => __('Actual approver'),
+            'datatype'           => 'itemlink',
+            'right'              => [
+                'validate_request',
+                'validate_incident'
+            ]
+        ];
+
```
btry, line 1180: search option to allow sarches against the actual validator of an ITIL object

src/CommonITILValidation.php
```diff
@@ -1287,6 +1322,45 @@ public static function rawSearchOptionsToAdd()
...
+                                    'joinparams'                 => [
+                                        'jointype'                   => 'child',
+                                    ]
+                                ]
+                            ]
+                        ]
+                    ]
+                ],
+            ]
+        ];
```
btry, line 1362: Search option to search an ITIL object by its validator, taking into accounts the substitutes of this validator, and the date range of effect of the delegation. Related to #12486

src/Ticket.php
```diff
@@ -4449,12 +4449,31 @@ public static function showCentralList($start, $status = "process", bool $showgr
...
                     [
-                        'users_id_validate'              => Session::getLoginUserID(),
+                        'OR' => [
+                            'users_id_validate' => Session::getLoginUserID(),
+                            'AND' => [
+                                'glpi_users' . '.substitution_start_date' => ['<=', $_SESSION['glpi_currenttime']],
+                                'glpi_users' . '.substitution_end_date' => ['>=', $_SESSION['glpi_currenttime']],
+                                'glpi_validatorsubstitutes' . '.users_id_substitute' => Session::getLoginUserID(),
+                            ],
+                        ],
```
btry, line 4476: Update search for tickets to validate : take into acount substitutes

src/Ticket.php
```diff
@@ -4758,9 +4777,14 @@ public static function showCentralList($start, $status = "process", bool $showgr
...
-                        $options['criteria'][1]['searchtype'] = 'equals';
-                        $options['criteria'][1]['value']      = Session::getLoginUserID();
+                        $options['criteria'][1]['criteria'][0]['field']      = 59; // validation aprobator
+                        $options['criteria'][1]['criteria'][0]['searchtype'] = 'equals';
+                        $options['criteria'][1]['criteria'][0]['value']      = Session::getLoginUserID();
+                        $options['criteria'][1]['criteria'][0]['link']       = 'AND';
+                        $options['criteria'][1]['criteria'][1]['field']      = 195; // validation aprobator substitute
+                        $options['criteria'][1]['criteria'][1]['searchtype'] = 'equals';
+                        $options['criteria'][1]['criteria'][1]['value']      = Session::getLoginUserID();
+                        $options['criteria'][1]['criteria'][1]['link']       = 'OR';
```
btry, line 4787: Update the search criterias to show the tickets to validate, taking into account the substitutes

btry:
@cconard96 : I updated the date range check to handle unset values. I also added a check to forbid a end date being older than the start date.

front/commonitilvalidation.form.php
```diff
@@ -80,6 +80,39 @@
...
+        if (!$itilObject->getFromDB($_POST[$itemtypeFk])) {
+            Html::back();
+        }
+        $primaryRequester = $itilObject->getPrimaryRequesterUser();
+        if ($primaryRequester === null) {
+            Html::back();
+        }
+        if ($primaryRequester->fields['users_id_supervisor'] == 0) {
+            // TRANS: $1%s us the friendly user name
+            Session::addMessageAfterRedirect(sprintf(__('%1$s does not have a responsible'), $primaryRequester->getFriendlyName()));
```
cconard96, line 103:
```suggestion
            Session::addMessageAfterRedirect(sprintf(__('%1$s does not have a supervisor'), $primaryRequester->getFriendlyName()));
```
Sorry if I wasn't clear, but "Supervisor" is the word that made the most sense to me. In the user form currently, it is labelled "Responsible", but that is an adjective and doesn't make any sense as a label in English.

btry:
Some changes are planned 
- replace the read only multiselect with a table to show start / end dates of delegation
image
- show the name of the responsible in validation request
image
- allow loop between validator substitute and valdiation delegator (currently forbidden by code)

trasher:
Do not forget to change codebase to main.

trasher:
You should change PR to draft

src/CommonITILValidation.php
```diff
@@ -1452,17 +1569,171 @@ public static function rawSearchOptionsToAdd()
...
+                ],
                 'beforejoin'         => [
                     'table'              => static::getTable(),
                     'joinparams'         => [
-                        'jointype'           => 'child'
+                        'jointype'           => 'child',
+                    ]
+                ]
+            ]
+        ];
```
btry, line 1589: :warning: It looks like this search option duplicates the ID 7 a few lines above. In GLPI 9.5, 59 does not exists. This may require  investigation

btry:
Hi, the pr is ready for review @cconard96 @AdrienClairembault @cedric-anne @trasher

src/CommonITILValidation.php
```diff
@@ -770,8 +804,34 @@ public static function getNumberToValidate($users_id)
...
+                        'OR' => [
+                            [
+                                User::getTableField('substitution_end_date') => null,
+                            ],
+                            [
+                                User::getTableField('substitution_end_date') => ['>=', $_SESSION['glpi_currenttime']],
+                            ],
+                        ],
+                    ],
+                    ValidatorSubstitute::getTableField('users_id_substitute')  => $users_id,
```
cedric-anne, line 832: `TicketValidation::getTargetCriteriaForUser()` is used in `Glpi\Dashboard\Provider::nbTicketsGeneric()`, and `ValidatorSubstitute` is not joined there, so query will probaly fail.

There is no need to add joined tables for groups criteria due to usage of subqueries. I guess it is possible to do the same for users criteria.

src/CommonITILValidation.php
```diff
@@ -770,8 +804,34 @@ public static function getNumberToValidate($users_id)
...
+                    static::getTableField('items_id_target') => $users_id,
+                ],
+                'AND' => [
+                    [
+                        'OR' => [
+                            [
+                                User::getTableField('substitution_start_date')  => null,
+                            ],
+                            [
+                                User::getTableField('substitution_start_date')  => ['<=', $_SESSION['glpi_currenttime']],
```
cedric-anne, line 819: You should use `new \QueryExpression('NOW()')`, in order to ensure that date will use the timezone defined in MySQL session.
```suggestion
                                User::getTableField('substitution_start_date')  => ['<=', new \QueryExpression('NOW()')],
```

src/CommonITILValidation.php
```diff
@@ -770,8 +804,34 @@ public static function getNumberToValidate($users_id)
...
+                            ],
+                        ],
+                    ], [
+                        'OR' => [
+                            [
+                                User::getTableField('substitution_end_date') => null,
+                            ],
+                            [
+                                User::getTableField('substitution_end_date') => ['>=', $_SESSION['glpi_currenttime']],
```
cedric-anne, line 828: You should use `new \QueryExpression('NOW()')`, in order to ensure that date will use the timezone defined in MySQL session.
```suggestion
                                User::getTableField('substitution_end_date') => ['>=', new \QueryExpression('NOW()')],
```

src/CommonITILValidation.php
```diff
@@ -783,7 +843,52 @@ final public static function getTargetCriteriaForUser(int $users_id, bool $searc
...
+                                                        User::getTable() => 'id',
+                                                        [
+                                                            'AND' => [
+                                                                [
+                                                                    'OR' => [
+                                                                        [
+                                                                            User::getTable() . '.substitution_start_date' => null,
+                                                                        ],
+                                                                        [
+                                                                            User::getTable() . '.substitution_start_date' => ['<=', $_SESSION['glpi_currenttime']],
```
cedric-anne, line 867: Same here.

src/CommonITILValidation.php
```diff
@@ -783,7 +843,52 @@ final public static function getTargetCriteriaForUser(int $users_id, bool $searc
...
+                                                                        ],
+                                                                    ],
+                                                                ],
+                                                                [
+                                                                    'OR' => [
+                                                                        [
+                                                                            User::getTable() . '.substitution_end_date' => null,
+                                                                        ],
+                                                                        [
+                                                                            User::getTable() . '.substitution_end_date' => ['>=', $_SESSION['glpi_currenttime']],
```
cedric-anne, line 877: Same here.

src/CommonITILObject.php
```diff
@@ -6980,14 +6998,16 @@ public function getTimelineItems(array $options = [])
...
             foreach ($validations as $validations_id => $validation) {
-                /** @var CommonITILValidation $valitation_obj */
-                $valitation_obj->getFromDB($validations_id);
-                $canedit = $valitation_obj->can($validations_id, UPDATE);
-                $cananswer = $valitation_obj->isCurrentUserValidationTarget(true) && ((int) $validation['status'] === CommonITILValidation::WAITING);
-                $request_key = $valitation_obj::getType() . '_' . $validations_id
+
+                /** @var CommonITILValidation $validation_obj */
+                $canedit = $validation_obj->can($validations_id, UPDATE);
+                $cananswer = ($validation_obj->canValidate($this->getID())
```
cedric-anne, line 7007: There are other checks based on `CommonITILValidation::isCurrentUserValidationTarget()` in  `CommonITILValidation::canUpdateItem()` and  `CommonITILValidation::prepareInputForUpdate()`. I think they are obsolete and have to be updated. Maybe they could rely on `CommonITILValidation::canValidate()`, and `CommonITILValidation::isCurrentUserValidationTarget()` could probably be removed then.

src/User.php
```diff
@@ -438,11 +439,23 @@ public function cleanDBonPurge()
...
+        $rows = self::find([
+            'users_id_supervisor' => $this->fields['id']
+        ]);
+        $supervisedUser = new User();
+        foreach ($rows as $row) {
+            $supervisedUser->update([
+                'id' => $row['id'],
+                'users_id_supervisor' => 0,
+            ]);
+        }
```
cedric-anne, line 456: If I remember right, the best way to handle this is to add this relation in `inc/relation.constant.php`. With this, when user is purged, replacement user could be selected. It have to be tested, I am not 100% sure.

```diff
diff --git a/inc/relation.constant.php b/inc/relation.constant.php
index f20d34a399..900abdbc09 100644
--- a/inc/relation.constant.php
+++ b/inc/relation.constant.php
@@ -1410,6 +1410,7 @@ $RELATION = [
             'users_id_validate',
         ],
         '_glpi_useremails'              => 'users_id',
+        'glpi_users'                    => 'users_id_supervisor',
     ],
 
     'glpi_usertitles' => [
```

cedric-anne (commented):
I have not yet finished the review, but here are some remarks.

src/CommonITILObject.php
btry: I see several things to carefully check
- CommonITILValidation::isCurrentUserValidationTarget() may or may not check the groups depending on its bool argument
- git grep shows that all calls to this method pass true as argument. It is therefore useless.
- CollonITILValidation::getTargetCriteriaForUser() also has a bool to enable groups check. No call done with groups check disabled; then this bool is also useless.

```
$ git grep "canValidate("
front/commonitilvalidation.form.php:    if ($validation->canValidate($validation->fields[$validation::$items_id])) {
src/CommonITILObject.php:                $cananswer = ($validation_obj->canValidate($this->getID())
src/CommonITILValidation.php:    public static function canValidate($items_id)
src/CommonITILValidation.php:            && !static::canValidate($item->getID())
src/Ticket.php:                  && TicketValidation::canValidate($this->fields["id"])));
tests/src/CommonITILValidation.php:        $this->boolean($validation::canValidate($itil_items_id))->isFalse();
tests/src/CommonITILValidation.php:        $this->boolean($validation::canValidate($itil_items_id))->isTrue();
tests/src/CommonITILValidation.php:        $this->boolean($validation::canValidate($itil_items_id))->isTrue();
tests/src/CommonITILValidation.php:        $this->boolean($validation::canValidate($itil_items_id))->isFalse();
tests/src/CommonITILValidation.php:        $this->boolean($validation::canValidate($itil_items_id))->isFalse();
tests/src/CommonITILValidation.php:        $this->boolean($validation::canValidate($itil_items_id))->isTrue();
tests/src/CommonITILValidation.php:        $this->boolean($validation::canValidate($itil_items_id))->isTrue();
tests/src/CommonITILValidation.php:        $this->boolean($validation::canValidate($itil_items_id))->isFalse();
```

```
$ git grep "isCurrentUserValidationTarget("
src/CommonITILValidation.php:        $is_target = $this->isCurrentUserValidationTarget(true);
src/CommonITILValidation.php:    public function isCurrentUserValidationTarget(bool $include_groups = true): bool
src/CommonITILValidation.php:        if ($this->isCurrentUserValidationTarget(true) && isset($input["status"])) {
tests/src/CommonITILValidation.php:        $this->boolean($validation->isCurrentUserValidationTarget())->isTrue();
tests/src/CommonITILValidation.php:        $this->boolean($validation->isCurrentUserValidationTarget(true))->isTrue();
tests/src/CommonITILValidation.php:        $this->boolean($validation->isCurrentUserValidationTarget(false))->isFalse();
```

In CommonITILValidation::prepareInputForUpdate() there is a logic problem before changing the call to CommonITILValidation::isCurrentUserValidationTarget(). I need to pass the linked ITIL objects ID but it may be in $this->fields[statkc::$items_id] or in $input[static::$items_id].  $forbif_fields may be empty when reaching the end od CommonITILValidation::prepareInputForUpdate(), where forbidden changes are cleaned up. 

As I don't see a use case where the FK to the ITIL object should change I strongly suggest to initialize $forbid_fields with static::$items_id. Maybe also 'entities_id'.

Let me know your thought before I add a fix.

cedric-anne: Indeed, I think that `static::$items_id` should not changed once item has been created, so you could move it in `$forbid_fields` initial declaration. Same for `entities_id` and `is_recursive` IMHO.

btry: done with 17eb48aa97

src/CommonITILValidation.php
btry: solved with ac6de93d6a

src/CommonITILValidation.php
btry: solved with https://github.com/glpi-project/glpi/commit/ac6de93d6ad23a2b46295229e50cd11dc377fe4b

src/Ticket.php
```diff
@@ -5080,6 +5108,19 @@ public static function showCentralCount(bool $foruser = false, bool $display = t
...
+            $opt['criteria'][1]['criteria'][1]['link']       = 'OR';
+            $opt['criteria'][1]['criteria'][2]['field']      = 196; // validation aprobator group
+            $opt['criteria'][1]['criteria'][2]['searchtype'] = 'equals';
+            $opt['criteria'][1]['criteria'][2]['value']      = 'mygroups'; // Resolved as groups the current user belongs to
+            $opt['criteria'][1]['criteria'][2]['link']       = 'OR';
+            $opt['criteria'][1]['criteria'][3]['field']      = 197; // validation aprobator group
+            $opt['criteria'][1]['criteria'][3]['searchtype'] = 'equals';
+            $opt['criteria'][1]['criteria'][3]['value']      = 'myself'; // Resolved as groups the current user belongs to
+            $opt['criteria'][1]['criteria'][3]['link']       = 'OR';
+            $opt['criteria'][1]['link']       = 'AND';
```
btry, line 5123: There is a need to deduplicate with the hunk L4659-4674

src/CommonITILValidation.php
btry: Shoud be partially solved with bc2048892d  . Check it carefully as I have a doubt with the subquery, but could not yet see it fail to produce the expected results.

src/User.php
btry: It works. Refactored with aa7e22af3f

front/commonitilvalidation.form.php
trasher: see #12707

src/Ticket.php
```diff
@@ -4609,9 +4624,22 @@ public static function showCentralList($start, $status = "process", bool $showgr
                         $options['criteria'][0]['value']      = CommonITILValidation::WAITING;
                         $options['criteria'][0]['link']       = 'AND';
 
-                        $options['criteria'][1]['field']      = 59; // validation aprobator
-                        $options['criteria'][1]['searchtype'] = 'equals';
-                        $options['criteria'][1]['value']      = Session::getLoginUserID();
+                        $options['criteria'][1]['criteria'][0]['field']      = 59; // validation aprobator user
+                        $options['criteria'][1]['criteria'][0]['searchtype'] = 'equals';
+                        $options['criteria'][1]['criteria'][0]['value']      = 'myself'; // Resolved as current user's ID
+                        $options['criteria'][1]['criteria'][0]['link']       = 'AND';
```
cedric-anne, line 4630: Should it be 'OR' ?
```suggestion
                        $options['criteria'][1]['criteria'][0]['link']       = 'OR';
```

templates/components/itilobject/timeline/form_validation.html.twig
```diff
@@ -204,6 +204,7 @@
                     {{ fields.field(
                         '_add_validation',
                         subitem.dropdownValidator({
+                            'parent': item,
```
cedric-anne, line 207: Same modification should proably be made on `templates/components/itilobject/fields/global_validation.html.twig`.

ajax/dropdownValidator.php
```diff
@@ -80,6 +80,31 @@
     }
 
     switch (strtolower($_POST['validatortype'])) {
+        case 'requester_supervisor':
+            // find the supervisor of the requester of the ITIL object
+            $user_id = 0;
+            $itilValidationObjectType = $_POST['validation_class'];
+            if (!class_exists($itilValidationObjectType)) {
```
cedric-anne, line 87:
```suggestion
            if (is_a($itilValidationObjectType, CommonITILValidation::class, true)) {
```

src/CommonITILValidation.php
```diff
@@ -1554,6 +1810,7 @@ public static function dropdownValidator(array $options = [])
         $params = [
             'prefix'            => null,
             'id'                => 0,
+            'parent'            => null,
```
cedric-anne, line 1813: Passing the parent object is a problem, as it result in passing many information in AJAX requests. Only the `parent_id` should be passed (itemtype can be guessed using `static::$itemtype`).

image

templates/pages/admin/user.substitute.html.twig
```diff
@@ -0,0 +1,211 @@
...
+   {% if item.getType() == 'Entity' and item.fields['id'] == 0 %}
+      {% set entity_id = null %}
+   {% else %}
+      {% set entity_id = params['entities_id'] ?? item.getEntityID() ?? session('glpiactive_entity') %}
+   {% endif %}
+
+   {% if is_multi_entities_mode() %}
+      {% set entity_name = get_item_name('Entity', item.getEntityID()) %}
+   {% endif %}
+{% endif %}
```
cedric-anne, line 65: Many variables can be removed here. For instance:
- item has no `is_deleted` flag and `item.isDeleted()` is always false
- `withtemplate`, `noid`, `formtitle`, `friendlyname`, ... parameters are not passed, and some defined variables are probably useless

templates/pages/admin/user.substitute.html.twig
```diff
@@ -0,0 +1,211 @@
...
+
+{% if item.canEdit(user.fields['id']) %}
+   <form name="massaction_{{ rand }}" id="massaction_{{ rand }}" method="post"
+         action="{{ path('/front/massiveaction.php') }}" data-submit-once>
+      <div id="massive_container_{{ rand }}"></div>
+      <input type="hidden" name="_glpi_csrf_token" value="{{ csrf_token() }}" />
+   </form>
+
```
cedric-anne, line 73:
```suggestion
```

templates/pages/admin/user.substitute.html.twig
```diff
@@ -0,0 +1,211 @@
...
+   <form name="asset_form" method="post" action="{{ target }}" {{ formoptions|raw }} enctype="multipart/form-data" data-submit-once>
+      <input type="hidden" name="entities_id" value="{{ entity_id }}" />
+      {% if _request['_in_modal'] is defined and _request['_in_modal'] == "1" %}
+         <input type="hidden" name="_in_modal" value="1"/>
+      {% endif %}
```
cedric-anne, line 78:
```suggestion
```

templates/pages/admin/user.substitute.html.twig
```diff
@@ -0,0 +1,211 @@
...
+         {% set nametype = __('Created from the template %s')|format(template_name) %}
+      {% elseif withtemplate == 1 %}
+         <input type="hidden" name="is_template" value="1" />
+      {% elseif item.isNewItem() %}
+         {% set nametype = __('%1$s - %2$s')|format(__('New item'), nametype) %}
+      {% else %}
+         {% if noid == false and (session('glpiis_ids_visible') or nametype|length == 0) %}
+            {% set nametype = __('%1$s - %2$s')|format(nametype, item.fields['id']) %}
+         {% endif %}
+      {% endif %}
```
cedric-anne, line 94:
```suggestion
```

templates/pages/admin/user.substitute.html.twig
```diff
@@ -0,0 +1,211 @@
...
+
+      {{ call_plugin_hook(constant('Glpi\\Plugin\\Hooks::PRE_ITEM_FORM'), {'item': item, 'options': params}) }}
```
cedric-anne, line 96: Not sure how it should be handled. This hook call should probably be removed.

templates/pages/admin/user.substitute.html.twig
```diff
@@ -0,0 +1,211 @@
...
+
+      {% set rand = random() %}
+      {% set params  = params ?? [] %}
+      {% set target       = params['target'] ?? item.getFormURL() %}
+      {% set withtemplate = params['withtemplate'] ?? '' %}
+      {% set item_type = item.getType() %}
+      {% set item_has_pictures = item.hasItemtypeOrModelPictures() %}
+      {% set field_options = {
+         'locked_fields': item.getLockedFields(),
+      } %}
```
cedric-anne, line 111:
```suggestion
      {% set rand = random() %}
```

cedric-anne (changes requested):

install/migrations/update_10.0.x_to_10.1.0/validatorsubstitute.php
```diff
@@ -0,0 +1,96 @@
...
+ */
+
+$default_charset = DBConnection::getDefaultCharset();
+$default_collation = DBConnection::getDefaultCollation();
+$default_key_sign = DBConnection::getDefaultPrimaryKeySignOption();
+
+if (!$DB->tableExists('glpi_validatorsubstitutes')) {
+    $query = "CREATE TABLE `glpi_validatorsubstitutes` (
+        `id` int {$default_key_sign} NOT NULL AUTO_INCREMENT,
+        `users_id` int {$default_key_sign}  NOT NULL DEFAULT '0' COMMENT 'Delegator user',
```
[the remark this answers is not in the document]
btry: This should have an impact on the search options I added or changed. I need to check before applying this suggestion.

install/mysql/glpi-10.1.0-empty.sql
```diff
@@ -839,14 +839,15 @@ CREATE TABLE `glpi_changevalidations` (
   `users_id` int unsigned NOT NULL DEFAULT '0',
   `changes_id` int unsigned NOT NULL DEFAULT '0',
   `users_id_validate` int unsigned NOT NULL DEFAULT '0',
-  `itemtype_target` varchar(255) NOT NULL,
-  `items_id_target` int unsigned NOT NULL DEFAULT '0',
+  `itemtype_target` varchar(255) NOT NULL COMMENT 'itemtype in charge of the validation',
+  `items_id_target` int unsigned NOT NULL DEFAULT '0' COMMENT 'items id in charge of the validation',
```
btry, line 843:
```suggestion
  `itemtype_target` varchar(255) NOT NULL,
  `items_id_target` int unsigned NOT NULL DEFAULT '0',
```

install/mysql/glpi-10.1.0-empty.sql
```diff
@@ -7480,14 +7482,15 @@ CREATE TABLE `glpi_ticketvalidations` (
   `users_id` int unsigned NOT NULL DEFAULT '0',
   `tickets_id` int unsigned NOT NULL DEFAULT '0',
   `users_id_validate` int unsigned NOT NULL DEFAULT '0',
-  `itemtype_target` varchar(255) NOT NULL,
-  `items_id_target` int unsigned NOT NULL DEFAULT '0',
+  `itemtype_target` varchar(255) NOT NULL COMMENT 'itemtype in charge of the validation',
+  `items_id_target` int unsigned NOT NULL DEFAULT '0' COMMENT 'items id in charge of the validation',
```
btry, line 7486:
```suggestion
  `itemtype_target` varchar(255) NOT NULL,
  `items_id_target` int unsigned NOT NULL DEFAULT '0',
```

install/migrations/update_10.0.x_to_10.1.0/validatorsubstitute.php
```diff
@@ -0,0 +1,96 @@
...
+$validation_tables = ['glpi_ticketvalidations', 'glpi_changevalidations'];
+
+foreach ($validation_tables as $validation_table) {
+    $migration->addfield(
+        $validation_table,
+        'users_id_actual_validate',
+        "int {$default_key_sign} NOT NULL DEFAULT '0'",
+        [
+            'after' => 'timeline_position',
+            'comment' => 'User who did the validation',
```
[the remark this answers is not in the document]
btry: Main branch deprecated users_id_validate It is still present but shoud be replaced with itemtype_target and items_id_target.

When a validation is not done yet, users_id_actual_validate must be 0. I would suggest the following instead.

```suggestion
            'comment' => 'User who did the validation',
            'update' => $DB->quoteName('items_id_target'),
            'condition' => '`status` IN (3, 4) AND `itemtype_target`='User',
```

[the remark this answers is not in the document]
cedric-anne: Indeed, it is better.

src/ValidatorSubstitute.php
```diff
@@ -0,0 +1,346 @@
...
+                return false;
+            }
+
+            // Delete old substitutes which are not in the new substitutes list
+            $substitutes_to_delete = array_diff($old_substitutes, $input['substitutes']);
+            if (count($substitutes_to_delete) > 0) {
+                $success = $success && $validator_substitute->deleteByCriteria([
+                    'users_id' => $validator,
+                    'users_id_substitute' => $substitutes_to_delete,
+                ]);
```
btry, line 213:
```suggestion
                $success = $validator_substitute->deleteByCriteria([
                    'users_id' => $validator,
                    'users_id_substitute' => $substitutes_to_delete,
                ]) && $success;
```

src/ValidatorSubstitute.php
```diff
@@ -0,0 +1,346 @@
...
+        return $success;
+    }
+
+    /**
+     * Get all valdiation substitutes for the given user ID
+     *
+     * @param int $user_id
+     * @return array
+     */
+    public static function getSubstitutes(int $user_id): array
```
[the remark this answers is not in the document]
btry: If you want to make this method non static, I think I should move it in User class (with public visiblity).

src/ValidatorSubstitute.php
```diff
@@ -0,0 +1,346 @@
...
+        return $substitutes;
+    }
+
+    /**
+     * Get all delegators for the given user ID
+     *
+     * @param integer $user_id
+     * @return array
+     */
+    public static function getDelegators(int $user_id): array
```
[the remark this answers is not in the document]
btry: and same as above :)

src/ValidatorSubstitute.php
```diff
@@ -0,0 +1,346 @@
...
+                        ]
+                    ]
+                ],
+            ];
+        }
+
+        $result = $DB->request($request);
+
+        return (count($result) > 0);
+    }
```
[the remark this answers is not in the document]
btry: It is used in Formcreator : https://github.com/pluginsGLPI/formcreator/pull/2968. I need to keep it.

To make it non static, I may consider to move it un User class, as suggested above.

tests/units/ValidatorSubstitute.php
```diff
@@ -0,0 +1,767 @@
...
+    {
+        $this->login('normal', 'normal');
+        yield 'not allowed to edit' => [
+            'input' => [
+                'users_id' => User::getIdByName('glpi'),
+            ],
+            'expected' => [
+                'return' => false,
+                'range' => [],
+                'messages' => ['You cannot change substitutes for this user'],
```
btry, line 324:
```suggestion
                'messages' => ['You cannot change substitutes for this user.'],
```

front/commonitilvalidation.form.php
```diff
@@ -92,6 +92,40 @@
...
```
AdrienClairembault, line 115: From what I understand, responsible can only be used as an adjective in English.
You can be "responsible for something" or "responsible of something" but you can't be "a responsible".

src/ValidatorSubstitute.php
```diff
@@ -0,0 +1,346 @@
...
+                ]);
+            }
+
+            // Add the new substitutes which are not in the old substitutes list
+            $substitutes_to_add = array_diff($input['substitutes'], $old_substitutes);
+            foreach ($substitutes_to_add as $substitute) {
+                $success = $success && $validator_substitute->add([
+                    'users_id' => $validator,
+                    'users_id_substitute' => $substitute,
+                ]);
```
btry, line 222:
```suggestion
                $success = $validator_substitute->add([
                    'users_id' => $validator,
                    'users_id_substitute' => $substitute,
                ]) && $success;
```

src/ValidatorSubstitute.php
```diff
@@ -0,0 +1,346 @@
...
+
+        // Check sanity of substitution date range
+        if ($input['substitution_start_date'] != '' && $input['substitution_end_date'] != '') {
+            if ($input['substitution_end_date'] < $input['substitution_start_date']) {
+                $input['substitution_end_date'] = $input['substitution_start_date'];
+            }
+        } else {
+            $input['substitution_start_date'] = $input['substitution_start_date'] == '' ? 'NULL' : $input['substitution_start_date'];
+            $input['substitution_end_date']   = $input['substitution_end_date'] == '' ? 'NULL' : $input['substitution_end_date'];
+        }
```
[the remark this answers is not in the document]
btry: fixed with : 69f9496bda

src/ValidatorSubstitute.php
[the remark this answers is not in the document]
btry: solved with: 31cdbef57c

src/ValidatorSubstitute.php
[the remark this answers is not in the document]
btry: solved with f316d3af4ab02e2fdc5df8aca26eabb4f4bffe63

src/ValidatorSubstitute.php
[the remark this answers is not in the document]
btry: moved to USer class with f316d3af4ab02e2fdc5df8aca26eabb4f4bffe63

install/migrations/update_10.0.x_to_10.1.0/validatorsubstitute.php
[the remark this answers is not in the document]
btry: As expected this change breaks Central. I don't see any way to change search options 195 and 196 on CommonITILValidation to use the good column name. 

image

src/CommonITILValidation.php
```diff
@@ -770,8 +775,48 @@ public static function getNumberToValidate($users_id)
     final public static function getTargetCriteriaForUser(int $users_id, bool $search_in_groups = true): array
```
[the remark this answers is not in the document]
btry: Factorized with dc30701a70. ADded unit tests for substitution with date range.

front/commonitilvalidation.form.php
btry: An issue has been opened for this topic. As I said to Curtis, this PR is quite big, it is better to focus on the feature and fix this wording problem separately.

src/Ticket.php
```diff
@@ -4300,7 +4300,22 @@ public static function showCentralList($start, $status = "process", bool $showgr
...
+                        'AND' => [
+                            'glpi_ticketvalidations.itemtype_target' => User::class,
+                        ],
+                    ],
+                    'glpi_users' => [
+                        'FKEY' => [
+                            'glpi_validatorsubstitutes' => 'users_id',
+                            'glpi_users' => 'id',
+                        ],
+                    ],
```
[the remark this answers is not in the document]
btry: solved with 605269cff1

src/Ticket.php
btry: This is counter intuitive : field search type and value are the 3 values of a simple expression and link is the logical operator linking several expressions together. 

Link is usually placed after the simple expression but is handled before the simple expression. As a boolean operator is meaningless at the beginning of a complex expression and it is ignored. Actually we can just remove the line.

Same for the 'OR' a few lines below by the way.

src/User.php
```diff
@@ -2882,387 +2883,387 @@ public function showMyForm($target, $ID)
         ) {
             return false;
         }
-        if ($this->getFromDB($ID)) {
-            $rand     = mt_rand();
-            $authtype = $this->getAuthMethodsByID();
+        if (!$this->getFromDB($ID)) {
```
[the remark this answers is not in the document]
btry: removed code refactor with 978c0c7a58

install/migrations/update_10.0.x_to_10.1.0/validatorsubstitute.php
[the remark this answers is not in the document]
btry: as seen together: leaving this topic as is due to a limitation in the search engine which prevents tu rename the column.

templates/pages/admin/user.substitute.html.twig
btry: solved with: cbc0d9afc2

templates/pages/admin/user.substitute.html.twig
btry: solved with: cbc0d9afc2

src/CommonITILValidation.php
btry: fixed in: c8d2c8e92e3015bfd973fff8b70b012e9f62889b

templates/pages/admin/user.substitute.html.twig
btry: after our discussion, I keep it for now. Let's see other reviewers comment.

tests/src/CommonITILValidation.php
```diff
@@ -479,6 +479,19 @@ public function testPrepareInputForUpdate(array $input, array $expected)
...
+            $validation::$items_id => $itilObject->getID(),
+            'users_id_validate' => \Session::getLoginUserID(),
+            'itemtype_target' => \User::class,
+            'items_id_target' => \Session::getLoginUserID(),
+            'status' => $input['status'],
+            'timeline_position' => '1',
+        ]);
+        $this->boolean($validation->isNewItem())->isFalse();
         if (!empty($expected)) {
             // Inject target fields into fields so isCurrentUserValidationTarget will work
```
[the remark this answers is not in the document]
btry: Line numbers moved due to latest changes.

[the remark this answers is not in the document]
cedric-anne: Done in 6465e3295cf368b23a7f3b1aaed91506bce5280d.

src/CommonITILObject.php
cedric-anne: Method removed in 6465e3295cf368b23a7f3b1aaed91506bce5280d.

templates/components/itilobject/timeline/form_validation.html.twig
cedric-anne: Remark is still relevant.

install/migrations/update_10.0.x_to_10.1.0/validatorsubstitute.php
```diff
@@ -0,0 +1,78 @@
...
+            'update' => $DB->quoteName('users_id_validate'),
+            'condition' => "WHERE `status` IN (3, 4)",
+        ]
+    );
+    $migration->addKey($validation_table, 'users_id_actual_validate');
+}
```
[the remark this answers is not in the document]
btry: @cconard96 : do you confirm that you wanted to use users_id_validate to store the user who actually validated a ITIL object ? 

@canne 
Having a BC layer may help  to give time to plugin developers, then this may be a good thing to keep it.

If Curtis confirms your interpretation, then there is  no problem to remove users_id_actual_validate on my PR (just show the other column instead in the HTML table), but we should set users_id_validate to 0 as long as the validation is not complete.

cedric-anne (changes requested):

src/ValidatorSubstitute.php
```diff
@@ -0,0 +1,346 @@
...
+
+    public function canPurgeItem()
+    {
+        if (!isset($this->fields['users_id'])) {
+            return false;
+        }
+        return $this->fields['users_id'] == Session::getLoginUserID();
+    }
+
+    public function canEdit($ID)
```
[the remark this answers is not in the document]
btry: fixed with: 7b14c59dae

removed checks with UserHaveMoreRightsThan because the UI does not allows a user to tune  the substitutes for someone else.

front/validatorsubstitute.form.php
```diff
@@ -36,8 +36,12 @@
...
 
 if (isset($_POST['update'])) {
-    $validator_substitute = new ValidatorSubstitute();
-    if (!$validator_substitute->canEdit($_POST['users_id'] ?? 0)) {
+    $validator_substitute = ValidatorSubstitute::getById($_POST['id'] ?? -1);
+    if (!($validator_substitute instanceof ValidatorSubstitute)) {
+        Html::back();
+    }
+
+    if (!$validator_substitute->canUpdateItem($_POST['users_id'] ?? 0)) {
```
cedric-anne, line 44:
```suggestion
    if (!$validator_substitute->canUpdateItem()) {
```

front/validatorsubstitute.form.php
```diff
@@ -0,0 +1,54 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ * ---------------------------------------------------------------------
+ */
+
+include('../inc/includes.php');
+
+if (isset($_POST['update'])) {
+    $validator_substitute = ValidatorSubstitute::getById($_POST['id'] ?? -1);
```
cedric-anne, line 39: It is not anymore possible to define validator substitutes if there is not one defined yet for user.

btry: should be fixed with: 26f7a5bcb6; code is also simpler now.

templates/pages/admin/user.substitute.html.twig
```diff
@@ -0,0 +1,130 @@
...
+                        __('Start date ')
+                     ) }}
+
+                     {{ fields.datetimeField(
+                        'substitution_end_date',
+                        user.fields['substitution_end_date'],
+                        __('End date ')
+                     ) }}
+
+                     {{ fields.dropdownField(
```
orthagh, line 66: I think substitutes field is the more important one of the form and should be place at the top.

orthagh (changes requested):
One last point, as seen together, i find the feature extension of "responsible" field didn't match the minimal UX prerequisites.
Having only the responsible of the first requester is not understandable at all from a user pov (i understand the technical issue).
This requires documentation at least, but event with, it will be subject to questions in the future i guess.
On my side, i think the feature should be dropped as it's done now is not sufficient or should be enhanced to support responsible of all requesters (with a dedicated list of option like "XXX responsible of YYY).


Another thing, documentation should be updated to:
- explain the subsitutes feature
- explain a responsible user gains view on validations

templates/pages/admin/user.substitute.html.twig
btry: I placed the dates on top because I think that the substitutes will be often set the first time the user will activate the feature. The next times, the user may just update the dates, without changing the substitutes (assument there is not too much turnover in the company). In such use case, dates becomes more important than substitutes.

cedric-anne (approved):

btry:
substitutes tab now handles singular / plural and shows count of substitutes
image

src/ValidatorSubstitute.php
```diff
@@ -0,0 +1,239 @@
...
+    {
+        return _n('Authorized substitute', 'Authorized substitutes', $nb);
+    }
+
+    public function getTabNameForItem(CommonGLPI $item, $withtemplate = 0)
+    {
+        switch ($item->getType()) {
+            case Preference::class:
+                $user = User::getById(Session::getLoginUserID());
+                $nb = count($user->getSubstitutes());
```
cedric-anne, line 50:
```suggestion
                $nb = $_SESSION['glpishow_count_on_tabs'] ? count($user->getSubstitutes()) : 0;
```

src/ValidatorSubstitute.php
```diff
@@ -0,0 +1,239 @@
...
+
+        // Get the old substitutes list
+        $old_substitutes = $user->getSubstitutes();
+
+        // Store the overall sucess of the changes below
+        $success = true;
+
+        if (isset($input['substitutes'])) {
+            if (in_array($input['users_id'], $input['substitutes'])) {
+                Session::addMessageAfterRedirect(__('Cannot add a user as substitute of himself.'), true, ERROR);
```
cconard96, line 191:
```suggestion
                Session::addMessageAfterRedirect(__('A user cannot be their own substitute.'), true, ERROR);
```
Gender neutral language should be used.
"Cannot add a user as substitute of themself." could be acceptable as well but using "themself" isn't very common anymore and the plural form "themselves" wouldn't be correct here.

cconard96 (changes requested):

cconard96 (approved):

orthagh (approved):

cedric-anne (approved):

trasher (approved):

JohnnyMacao:
hi, is this feature usable in glpi 10.0.7? thank you.

PR #12509, opened 2022-08-29
fix(ticket): hide added group when it is hiddenfield in template

Rom1-B:
If the template hides the requesting group, the requesters field is visible, because the template user can add users, but not groups: the desired behavior.

image

If this field already contains a group, it was displayed and the user could delete it.

_Same thing with the other actor fields (observers, assigned to) and in the opposite direction: visible group, hidden user._

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !24765

### DIFF
--- a/templates/components/itilobject/actors/field.html.twig
+++ b/templates/components/itilobject/actors/field.html.twig
@@ -329,6 +329,13 @@
             return openNotifyModal(event);
          }
       }, {capture: true})
+
+      {% if itiltemplate.isHiddenField('_users_id_' ~ actortype) %}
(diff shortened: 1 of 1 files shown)


orthagh (approved):
After some quick searches, i didn't find any other ways to achieve the desired results.
the initial state of the option fix at least the issue with loading time of javascript.

So ok for me.

trasher:
I'll wait @cedric-anne review for this one

cedric-anne (approved):

PR #12513, opened 2022-08-29
Breadcrumbs: fix for dropdowns

AdrienClairembault:
This code rely on detecting null values:
image

https://github.com/glpi-project/glpi/pull/12307 replaced the null value by an empty array, so we need to check with `empty` instead.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |

### DIFF
--- a/src/CommonDropdown.php
+++ b/src/CommonDropdown.php
@@ -216,7 +216,7 @@ abstract class CommonDropdown extends CommonDBTM
         ?string $title = null,
         ?array $menus = null
     ): void {
-        if (is_null($menus)) {
+        if (empty($menus)) {
             $dropdown = new static();
 
             $menus = [


orthagh (approved):

src/CommonDropdown.php
```diff
@@ -216,7 +216,7 @@ public static function displayCentralHeader(
         ?string $title = null,
         ?array $menus = null
     ): void {
-        if (is_null($menus)) {
+        if (empty($menus)) {
```
trasher, line 219: does `empty()` also handle `null` cases? since it's still the default value; it has to be handled

AdrienClairembault: Yes it does: 

image

trasher: :+1:

trasher (approved):

PR #12514, opened 2022-08-29
WIP: add saved search menu entry

btry:
It seems that a menu entry to access saved searches is missing in the user dropdown menu (top right corner)
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #number

### DIFF
--- a/templates/layout/parts/user_header.html.twig
+++ b/templates/layout/parts/user_header.html.twig
@@ -82,7 +82,14 @@
                   {% endif %}
                {% endif %}
 
-               {# @TODO Saved searches panel #}
+               {% if not anonymous %}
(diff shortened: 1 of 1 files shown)


orthagh (changes requested):
As discussed together, this is not sufficient.
We must check about a way to extends the "lists" buttons everywhere if possible.

To be confirmed with another guest

btry:
Obsolete. I found a solution by tuning / fixing the menus in Formcreator. Breadcrumb is properly handled now and this unlock the ability to show the saved searches button.

PR #12524, opened 2022-08-30
Fix object locks

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #11896

The options passed to `ObjectLock::manageObjectLock` did not contain the ID for Tickets at least so they were never used.

### DIFF
--- a/src/CommonDBTM.php
+++ b/src/CommonDBTM.php
@@ -6375,6 +6375,9 @@ class CommonDBTM extends CommonGLPI
             static::displayHelpdeskHeader($title, $menus);
         }
 
+        if (!isset($options['id'])) {
+            $options['id'] = $id;
+        }
         // Show item
         $options['loaded'] = true;
         $item->display($options);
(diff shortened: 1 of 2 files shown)


trasher (approved):

orthagh (commented):
Ok with the fix.

Just a comment (no directly a request for change) to have you opinion:
Unless i miss something, there's only one call for this method (maybe @tomolimo uses it in its plugins). But the signature of the method is very strange.

we get the class of the current item and pass it to the first parameter.
and then we copy the id into an `$options` array.

I think the first parameter could be directly a `CommonDBTM $item`.
Not sure how to manage depreciation of parameters, and how we can manage this clean
but i don't like at all this piece of code.

cconard96:
I pushed a commit for a broader solution since not all lockable items call `display` directly in their front files.
For changing `ObjectLock::manageObjectLock`, it should be easy since the parameters are not typed yet.

If the first parameter is not an instance of CommonDBTM, it can use the ID provided by $options. Otherwise, the ID from the item can be used.

tomolimo:
Hello @cconard96 and @orthagh 
Just to confirm that I don't use this method in my plugins :)
So the signature can be changed if needed.
Thank you
Regards,
Tomolimo

cedric-anne (approved):

PR #12529, opened 2022-08-31
Printers: fix printed pages graph for big datasets

AdrienClairembault:
Page counters get a bit messy with big datasets:

image

Note the labels on the X axis which are unreadable, and the ones on the Y-axis are truncated (it's supposed to go from 0 to 35k in this example).

Proposed changes:

image

* Limit the dataset to 25 items, spread equally across the set
* Change the graph from bars to areas. 
   This help by reducing the Y scale (areas graph seems to start the Y scale from the lowest value (here 15k) instead of 0 and are also easier to read in my opinion
* Rewrite Y axis label to use K or M prefix for big numbers
* Rewrite X axis label to remove the year (this graph will only show the last 365 dates so we can part with this data) and use the month name (more easily readable, especially on small screens if the labels get spread on multiples lines).

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !23397

### DIFF
--- a/src/Dashboard/Widget.php
+++ b/src/Dashboard/Widget.php
@@ -1590,6 +1590,20 @@ HTML;
             chartPadding: {
                right: 40
             },
+            axisY: {
+               labelInterpolationFnc: function(value) {
+                  if (value < 1e3) {
+                     // less than 1K
+                     return value;
+                  } else if (value < 1e6) {
+                     // More than 1k, less than 1M
+                     return value / 1e3 + "K";
+                  } else {
+                     // More than 1M
+                     return value / 1e6 + "M";
+                  }
+               },
+            },
             {$area_options}
             plugins: [
                {$legend_options}
--- a/src/PrinterLog.php
+++ b/src/PrinterLog.php
@@ -121,7 +121,7 @@ class PrinterLog extends CommonDBChild
             ] + $filters
         ]);
 
-        $series = iterator_to_array($iterator);
+        $series = iterator_to_array($iterator, false);
 
         // Reduce the data to 25 points
         $count = count($series);
--- a/src/PrinterLog.php
+++ b/src/PrinterLog.php
@@ -180,8 +180,9 @@ class PrinterLog extends CommonDBChild
             unset($metrics['id'], $metrics['date'], $metrics['printers_id']);
 
             foreach ($metrics as $key => $value) {
-                if ($value > 0) {
-                    $series[$key]['name'] = $this->getLabelFor($key);
+                $label = $this->getLabelFor($key);
+                if ($label && $value > 0) {
+                    $series[$key]['name'] = $label;
                     $series[$key]['data'][] = $value;
                 }
             }
@@ -204,7 +205,15 @@ class PrinterLog extends CommonDBChild
         echo "</div>";
     }
 
-    private function getLabelFor($key)
+    /**
+     * Get the label for a given column of glpi_printerlogs.
+     * To be used when displaying the printed pages graph.
+     *
+     * @param string $key
+     *
+     * @return null|string null if the key didn't match any valid field
+     */
+    private function getLabelFor($key): ?string
     {
         switch ($key) {
             case 'total_pages':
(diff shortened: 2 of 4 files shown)


trasher (approved):

AdrienClairembault:
I've found another issue with date being interpreted as columns:

image

Working on a fix, do not merge yet ;)

AdrienClairembault:
-> Fixed in `3459aa0a9f`

trasher (approved):

src/Dashboard/Widget.php
```diff
@@ -1590,6 +1590,20 @@ private static function getLinesGraph(
             chartPadding: {
                right: 40
             },
+            axisY: {
```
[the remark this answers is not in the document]
AdrienClairembault: I've thought about it, but is there really a case where we want to display huge numbers on the axes ?
For comparison, big numbers dashboards also enforce this behavior.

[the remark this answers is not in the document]
orthagh: ok

orthagh (approved):

src/PrinterLog.php
```diff
@@ -121,7 +121,22 @@ public function getMetrics(Printer $printer, $user_filters = []): array
...
+        $max_size = 25;
+        if ($count > $max_size) {
+            // Keep one row every X entry using modulo
+            $modulo = round($count / $max_size);
+            $series = array_filter(
+                $series,
+                fn($k) => ($k % $modulo) == 0,
+                ARRAY_FILTER_USE_KEY
+            );
+        }
```
cedric-anne, line 137: I think we should ensure that the last entry is always displayed. For instance, if you have 51 entries in the serie, modulo will be 2, and so the last entry will not be displayed.

Also, using round and modulo will not always limit displayed points count to 25 :
- with 37 entries, 37 points will be displayed (`round(37/25) => 1`);
- with 38 entries, 19 points will be displayed (`round(38/25) => 2`);
- with 62 entries, 31 points will be displayed (`round(62/25) => 2`);
- with 63 entries, 21 points will be displayed (`round(63/25) => 3`);

AdrienClairembault: Yeah, this is kind of hacky, feel free to improve it if you can.
Tbh, it feels bad spending time improving this when the graph lib that should handle this itself by picking its data points from the set.
I know @orthagh is trying a PoC to switch to echarts, which would support this feature.

src/PrinterLog.php
```diff
@@ -121,7 +121,22 @@ public function getMetrics(Printer $printer, $user_filters = []): array
...
+
+        // Reduce the data to 25 points
+        $count = count($series);
+        $max_size = 25;
+        if ($count > $max_size) {
+            // Keep one row every X entry using modulo
+            $modulo = round($count / $max_size);
+            $series = array_filter(
+                $series,
+                fn($k) => ($k % $modulo) == 0,
```
cedric-anne, line 134: To be sure that last point will always be display, filtering could be change to this:
```suggestion
                fn($k) => (($k + 1 - $count) % $modulo) == 0,
```

For instance, with 48 entries (last entry index is 47):

using current formula
 - `0 % 2 == 0 => true`
 - `1 % 2 == 0 => false`
 - `2 % 2 == 0 => true`
 - `3 % 2 == 0 => false`
 - ... 
 - `46 % 2 == 0 => true`
 - `47 % 2 == 0 => false`

using proposed formula
 - `( 0 + 1 - 48) % 2 == 0 => false`
 - `( 1 + 1 - 48) % 2 == 0 => true`
 - `( 2 + 1 - 48) % 2 == 0 => false`
 - `( 3 + 1 - 48) % 2 == 0 => true`
 - ... 
 - `(46 + 1 - 48) % 2 == 0 => false`
 - `(47 + 1 - 48) % 2 == 0 => true`

cedric-anne (approved):

PR #12530, opened 2022-08-31
fix(savedsearch): public search visible without right

Rom1-B:
Public saved searches were visible, even on a profile that did not have the right to see them.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !24794


trasher (approved):

trasher:
Would it be possible to add a test?

orthagh (approved):

trasher (approved):

PR #12532, opened 2022-08-31
fix(dashboard): fix draggable item and grid-stack size

stonebuzz:
Drag feature not work on mini dahsboard because ```$grid_rows``` is defined to ```1``` instead of ```2``` (from GLPI 9.5)

Need to fix ```height``` to prevent empty space between ```cell-add``` ```div``` (from ```grid-guide```)

image

| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !24801

### DIFF
--- a/css/standalone/dashboard.scss
+++ b/css/standalone/dashboard.scss
@@ -89,6 +89,7 @@ $break_tablet: 1400px;
         margin-bottom: -10px;
         z-index: 1;
         box-sizing: content-box;
+        height: 100px;
 
         & + .search_page {
             z-index: 2;
--- a/src/Search.php
+++ b/src/Search.php
@@ -163,7 +163,7 @@ class Search
             $itemtype == "Ticket"
             && $default = Glpi\Dashboard\Grid::getDefaultDashboardForMenu('mini_ticket', true)
         ) {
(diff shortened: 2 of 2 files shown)


trasher (approved):

css/standalone/dashboard.scss
```diff
@@ -89,6 +89,7 @@ $break_tablet: 1400px;
         margin-bottom: -10px;
         z-index: 1;
         box-sizing: content-box;
+        height: 100px;
```
orthagh, line 92: is it required ? fixing the mini dashboard height where it should be calculated from it number of cells seems strange.

stonebuzz: Without we have a space at bottom of each cell 
image
and drag feature is completely break
perhaps there is another way ..

orthagh: to be 100% sure, transform the property to `min-height`.

stonebuzz: No change with ```min-height``` property 
image

orthagh (approved):

PR #12582, opened 2022-09-05
fix(inventory): UUID is empty

Rom1-B:
The criterion "is empty" on the UUID was ignored during the execution of the rule.

The addition of the rule "Computer update (by serial + uuid is empty in GLPI)" fixes the case where the asset is created (manually) with a serial and no UUID, because before the default rules created a duplicate.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !24715

### DIFF
--- a/src/RuleImportAsset.php
+++ b/src/RuleImportAsset.php
@@ -331,8 +331,6 @@ class RuleImportAsset extends Rule
 
             case Rule::PATTERN_EXISTS:
             case Rule::PATTERN_DOES_NOT_EXISTS:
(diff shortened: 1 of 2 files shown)


trasher (commented):
You add a new rule, but seems it will be missing on update.

Also, this may need unit tests

orthagh (approved):

orthagh:
> but seems it will be missing on update.

I think we should not mess rules of existing instances.

PR #12602, opened 2022-09-07
Display last inventoried value for locked fields

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !24580

### DIFF
--- a/src/Lockedfield.php
+++ b/src/Lockedfield.php
@@ -165,7 +165,17 @@ class Lockedfield extends CommonDBTM
 
         $locks = [];
         foreach ($iterator as $row) {
-            $locks[] = $row['field'];
+            $value = $row['value'];
+            //if it refers to a ForeignKey get dropdown name
+            if (
+                $row['field'] !== 'entities_id'
+                && isForeignKeyField($row['field'])
+                && is_a($itemtype = getItemtypeForForeignKeyField($row['field']), CommonDropdown::class, true)
+                && null !== $row['value']
+            ) {
+                    $value = Dropdown::getDropdownName(getTableForItemType($itemtype), $row['value'], 0, true, true, $row['value']);
+            }
+            $locks[$row['field']] = $value;
         }
         return $locks;
     }
--- a/src/Lockedfield.php
+++ b/src/Lockedfield.php
@@ -185,7 +185,7 @@ class Lockedfield extends CommonDBTM
                     && isForeignKeyField($row['field'])
                     && is_a($itemtype = getItemtypeForForeignKeyField($row['field']), CommonDropdown::class, true)
                 ) {
-                    $value = Dropdown::getDropdownName(getTableForItemType($itemtype), $row['value'], 0, true, true, $row['value']);
+                    $value = Dropdown::getDropdownName(getTableForItemType($itemtype), $row['value'], 0, true, false, $row['value']);
                 }
                 $locks[$row['field']] = $value;
             }
--- a/src/Lockedfield.php
+++ b/src/Lockedfield.php
@@ -181,10 +181,9 @@ class Lockedfield extends CommonDBTM
                 $value = $row['value'];
                 //if it refers to a ForeignKey get dropdown name
                 if (
-                    $row['field'] !== 'entities_id'
+                    $row['value'] !== null
                     && isForeignKeyField($row['field'])
                     && is_a($itemtype = getItemtypeForForeignKeyField($row['field']), CommonDropdown::class, true)
-                    && null !== $row['value']
                 ) {
                     $value = Dropdown::getDropdownName(getTableForItemType($itemtype), $row['value'], 0, true, true, $row['value']);
                 }
--- a/src/Lockedfield.php
+++ b/src/Lockedfield.php
@@ -138,6 +138,16 @@ class Lockedfield extends CommonDBTM
         return (bool)$item->isDynamic();
     }
 
+    public function getLockedNames($itemtype, $items_id)
+    {
+        return $this->getLocks($itemtype, $items_id, true);
+    }
+
+    public function getLockedValues($itemtype, $items_id)
+    {
+        return $this->getLocks($itemtype, $items_id, false);
+    }
+
     /**
      * Get locked fields
      *
@@ -146,7 +156,7 @@ class Lockedfield extends CommonDBTM
      *
      * return array
      */
-    public function getLocks($itemtype, $items_id)
+    public function getLocks($itemtype, $items_id, bool $fields_only = true)
     {
         global $DB;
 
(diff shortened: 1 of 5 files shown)


templates/components/form/fields_macros.html.twig
```diff
@@ -878,8 +878,12 @@
    {% set locked_mark = '' %}
    {% if options.locked %}
       {% set locked_mark %}
+        {% set locked_title %}{{ __('Field will not be updated from inventory') }}{% endset %}
+        {% if options.locked_value is not empty %}
```
cedric-anne, line 882: `locked_value: null` should be added in options definition, on line 859.

src/Lockedfield.php
```diff
@@ -165,7 +165,7 @@ public function getLocks($itemtype, $items_id)
 
         $locks = [];
         foreach ($iterator as $row) {
-            $locks[] = $row['field'];
+            $locks[$row['field']] = $row['value'];
```
cedric-anne, line 168: `getLocks()` is used at many places.
A new method `getLockedValues()` should probably be declared, to prevent changing method signature on a bugfixes version, and ensure there is no BC breaks possible.

templates/components/form/fields_macros.html.twig
```diff
@@ -702,8 +702,8 @@
       'locked_fields': [],
    }|merge(options) %}
 
-   {% if name in options.locked_fields %}
-      {% set options = options|merge({'locked': true}) %}
+   {% if options.locked_fields[name] is defined %}
+      {% set options = options|merge({'locked': true, 'locked_value': options.locked_fields[name]}) %}
```
cedric-anne, line 706: To prevent any possible BC break, a new option `locked_values` may be added.

```
{% if options.locked_values[name] is defined %}
      {% set options = options|merge({'locked': true, 'locked_value': options.locked_values[name]}) %}
{% elseif name in options.locked_fields %}
      {% set options = options|merge({'locked': true}) %}
{% endif %}
```

cedric-anne (dismissed):

stonebuzz:
For me, ```old_value``` are never saved

need this to get old value

```patch
diff --git a/src/CommonDBTM.php b/src/CommonDBTM.php
index 9a0b38f005..7818771549 100644
--- a/src/CommonDBTM.php
+++ b/src/CommonDBTM.php
@@ -1783,12 +1783,13 @@ class CommonDBTM extends CommonGLPI
                         'itemtype'        => $this->getType(),
                         'items_id'        => $this->fields['id'],
                         'date_creation'   => $_SESSION["glpi_currenttime"],
+                        'value'           => new QueryParam(),
                         'field'           => new QueryParam()
                     ]
                 )
             );
             foreach ($fields as $field) {
-                 $stmt->bind_param('s', $field);
+                 $stmt->bind_param('ss', $this->oldvalues[$field], $field);
                  $res = $stmt->execute();
                 if ($res === false) {
                     if ($DB->errno() != 1062) {

```

stonebuzz:
On update is not working because of method signature
Lock field is in ```key``` now.
image

moreover, and unless I am wrong, it seems that the check to know if the ```itemtype``` is dynamic is based on ```$this->input``` but ```is_dynamic``` key is always passed to ```$this->fields```

templates/components/form/fields_macros.html.twig
trasher: done

src/Lockedfield.php
trasher: I'm not sure this BC breaks anything, all is part of main CommonDBTM calls. To me, the `getLocks` method is wrong from the beginning

templates/components/form/fields_macros.html.twig
trasher: I really do not see how we can have a bc break here.

trasher:
> On update is not working because of method signature Lock field is in `key` now.

This is fixed in last commit

> moreover, and unless I am wrong, it seems that the check to know if the `itemtype` is dynamic is based on `$this->input` but `is_dynamic` key is always passed to `$this->fields`

I'm not sure I understand :/ also not sure it's related to current PR, if there is something wrong; there is|are missing test case :/

src/Lockedfield.php
stonebuzz, line 168:
```suggestion
            $value = $row['value'];
            //if it refers to a ForeignKey get dropdown name
            if (
                $row['field'] !== 'entities_id'
                && isForeignKeyField($row['field'])
                && is_a($itemtype = getItemtypeForForeignKeyField($row['field']), CommonDropdown::class, true)
            ) {
                    $value = Dropdown::getDropdownName(getTableForItemType($itemtype), $row['value']);
            }
            $locks[$row['field']] = $value;
```

stonebuzz (changes requested):
Take care of foreignKey value to get rela name of it

src/Lockedfield.php
cedric-anne: As far as I see, method was legitimate as there is multiple places where we only have to know the locked fields name only, not their values.
The new result is not compatible with the previous one, so if any plugin is already using this method, it will be broken.

Indeed, it is likely that there is actually no plugin that use this method, but we have unwittingly bugged plugins in the past based on this kind of assumption, and unfortunately it sparked a very bitter rhetoric from some plugin developers about us.

templates/components/form/fields_macros.html.twig
trasher: I still not understand how a plugin would have a bc break on that... Seems a real waste of time.

src/Lockedfield.php
trasher: Pffff... If a plugin tries to override that part, I'd say it's normal it breaks. Seems a real nonsense. Plugins can do everything they want because of old GLPOI crap code, that does not means they should or we should take care of.

src/Lockedfield.php
```diff
@@ -165,7 +175,21 @@ public function getLocks($itemtype, $items_id)
...
+            if ($fields_only === true) {
+                $locks[] = $row['field'];
+            } else {
+                $value = $row['value'];
+                //if it refers to a ForeignKey get dropdown name
+                if (
+                    $row['field'] !== 'entities_id'
+                    && isForeignKeyField($row['field'])
+                    && is_a($itemtype = getItemtypeForForeignKeyField($row['field']), CommonDropdown::class, true)
+                    && null !== $row['value']
```
cedric-anne, line 187: 1. From performance point of view, it will probably be more efficient to check null value at first.
2. I do not think that `entities_id` may be locked, but, in this case, we would probably want to display its name too.

```suggestion
                    $row['value'] !== null
                    && isForeignKeyField($row['field'])
                    && is_a($itemtype = getItemtypeForForeignKeyField($row['field']), CommonDropdown::class, true)
```

src/Lockedfield.php
```diff
@@ -165,7 +175,20 @@ public function getLocks($itemtype, $items_id)
...
+                $locks[] = $row['field'];
+            } else {
+                $value = $row['value'];
+                //if it refers to a ForeignKey get dropdown name
+                if (
+                    $row['value'] !== null
+                    && isForeignKeyField($row['field'])
+                    && is_a($itemtype = getItemtypeForForeignKeyField($row['field']), CommonDropdown::class, true)
+                ) {
+                    $value = Dropdown::getDropdownName(getTableForItemType($itemtype), $row['value'], 0, true, true, $row['value']);
```
cedric-anne, line 188: The `$tooltip` arg should be `false`. 
```suggestion
                    $value = Dropdown::getDropdownName(getTableForItemType($itemtype), $row['value'], 0, true, false, $row['value']);
```

cedric-anne (approved):

g-bougard (approved):

PR #12604, opened 2022-09-07
fix(inventory): do not unset data switch LockedField

stonebuzz:
Do not unset data switch ```LockedField```

Data are handled (after) by ```CommonDBTM->update``` to fill into ```value``` column from ```LockedField```

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #number


trasher:
replaced with #12605

PR #12649, opened 2022-09-14
Replace Watcher term with Observer

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #number

The UI currently has both "Watcher" and "Observer" to refer to the same thing. This PR replaces all uses of "Watcher" with "Observer" which is how it is referred to in the code. The only code that refers to this type of actor as a watcher is the property for Groups to allow them to be observers. Since renaming it would be a breaking change with no end-user benefit, I did not rename it in this PR.


orthagh (approved):

trasher:
OK for me, but you have CS and tests to fix ;)

cedric-anne (approved):

trasher (approved):

PR #12653, opened 2022-09-15
fix(inventory): use old_val from lock instead of current inventory value (not handle yet)

stonebuzz:
If an asset have a lock on a ```foreignKey``` like ```manufacturers_id``` field

Inventory set current inventory value instead of lock value.

```json
    "batteries": [
      {
        "capacity": 52945,
        "chemistry": "Li-poly",
        "date": "2015-06-17",
        "manufacturer": "SMP",
        "name": "DELL 0N7T653",
        "real_capacity": 19902,
        "serial": 158,
        "voltage": 7640
      }
```
For example with this batterie, GLPI inventory put ```SMP``` as ```manufacturers_id``` val

| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | [#220](https://github.com/glpi-project/glpi/issues/12654)

### DIFF
--- a/src/Inventory/Asset/InventoryAsset.php
+++ b/src/Inventory/Asset/InventoryAsset.php
@@ -182,7 +182,6 @@ abstract class InventoryAsset
 
         $blacklist = new Blacklist();
         $lockedfield = new Lockedfield();
-        $locks = $lockedfield->getLockedValues($this->item->getType(), $this->item->fields['id'] ?? 0);
 
         $data = $this->data;
         foreach ($data as &$value) {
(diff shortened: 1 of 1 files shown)


orthagh (approved):

Tony710:
Hi @stonebuzz 
Your last commit, fix the bug.
Thanks

stonebuzz:
To conclude, it is not necessary to manage the ```locked fields``` here.

It is necessary to add the ```dropdown``` value  in database to keep in memory the ```ID``` (corresponding to the last value of the inventory)
which is set as old value from ```glpi_lockedfield``` table

trasher:
Last commit does solve the issue, but new values are imported in database for locked relations fields; this is something we want to avoid from the beginning. I've talked about that issue with @stonebuzz; I'll take a look tomorrow to find a "better" solution.

trasher:
Probably replaced with #12670

PR #12655, opened 2022-09-15
No longer throw an exception; closes #12609

### LINKED ISSUE #12609: Issue with stacked network elements
### Bug reporting acknowledgment

Yes, I read it

### Professional support

_No response_

### Describe the bug

Hi ! thank you for your work, I have odd behaviors while importing stacked elements. Sometime GLPI throw this error : Uncaught Exception RuntimeException: Cannot wire a port to itself! in /var/www/glpi.toureiffel.paris/src/Inventory/Asset/NetworkPort.php at line 763
Sometime it is OK but GLPI create duplicated networks elements ( 1 by stacked element )
I tried also with manual import to debug, same issue : I can PM xml files.

Thank you for your help

### To reproduce

glpi-agent ( networkinventory task ) OR from GUI task/import XML

### Expected behavior

A stacked element should not be duplicated , with ports listed like this ( 1/1 -> 2/24 )
( I rembember it was like this with fusion )

### Operating system

Linux

### GLPI Agent version

Other (See additional context below)

### GLPI version

Other (See additional context below)

### GLPIInventory plugin or FusionInventory for GLPI plugin version

Other (See additional context below)

### Additional context

Last agent ( 1.4 ) - GLPI 10.0.0.2 - plugin 1.0.3

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #12609

The main idea throwing an exception was to see when that case was reached, and try to fix it. Since it's been implemented, there were no case such a try was really valid.

So, maybe there is a logic issue when the original array is completed, but when that fails, there is never a real connection to be made - as far as I see.
If inventory results seems OK ignoring self wiring case; it's certainly ok.

### DIFF
--- a/src/Inventory/Asset/NetworkPort.php
+++ b/src/Inventory/Asset/NetworkPort.php
@@ -774,7 +774,8 @@ class NetworkPort extends InventoryAsset
     private function addPortsWiring(int $netports_id_1, int $netports_id_2): bool
     {
         if ($netports_id_1 == $netports_id_2) {
-            throw new \RuntimeException('Cannot wire a port to itself!');
+            //no wiring
+            return false;
         }
 
         $wire = new \NetworkPort_NetworkPort();


stonebuzz:
Seems good to me, no more errors =)

stonebuzz (approved):

cedric-anne (approved):

PR #12661, opened 2022-09-16
fix right item width in timeline (with 2k+ screens)

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

f573edca31b7c2289028681b71b64dd9d33c45fa : 
on large screen (1440p + ) the limitation of 700px (added on #12153) cause the right timeline items to be shifted:
before
image
after 
image

### DIFF
--- a/css/includes/components/itilobject/_timeline.scss
+++ b/css/includes/components/itilobject/_timeline.scss
@@ -100,6 +100,7 @@
         position: relative;
         border-top-left-radius: 0;
         max-width: 700px;
+        min-width: 100%;
         word-break: break-word;
 
         @include media-breakpoint-up(sm) {


templates/layout/parts/saved_searches.html.twig
```diff
@@ -116,8 +116,7 @@ $(function() {
       var target  = tablink.attr('data-bs-target');
       var index   = tablink.closest('.nav-item').index();
 
-      $.get(url, function(data) {
```
AdrienClairembault, line 119: Was this change done on purpose ?

AdrienClairembault (approved):
It shouldn't work according to my understand of css (https://www.w3.org/TR/CSS21/visudet.html#min-max-widths ->the 700px max width would end up ignored).
However functional tests show that the 700px max is still respected so I guess it's OK.

Edit: it's a css 2 doc so maybe not accurate anymore.

cconard96 (approved):
Looks OK even up to 5K width screens

PR #12670, opened 2022-09-16
Fix inventory locks

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

### DIFF
--- a/src/Inventory/Asset/Peripheral.php
+++ b/src/Inventory/Asset/Peripheral.php
@@ -168,7 +168,6 @@ class Peripheral extends InventoryAsset
                     $items_id = $data['found_inventories'][0];
                     $peripheral->getFromDB($items_id);
                     $handled_input = $this->handleInput($val, $peripheral);
-                    //FIXME: handleInput
                     $peripheral->update(Toolbox::addslashes_deep(['id' => $items_id] + $handled_input), false);
                 }
 
--- a/src/Inventory/Asset/Peripheral.php
+++ b/src/Inventory/Asset/Peripheral.php
@@ -163,10 +163,12 @@ class Peripheral extends InventoryAsset
                 $itemtype = 'Peripheral';
                 if ($data['found_inventories'][0] == 0) {
                     // add peripheral
+                    //FIXME: handleInput
                     $handled_input['entities_id'] = $this->entities_id;
                     $items_id = $peripheral->add(Toolbox::addslashes_deep($handled_input), [], false);
                 } else {
                     $items_id = $data['found_inventories'][0];
+                    //FIXME: handleInput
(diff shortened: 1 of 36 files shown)


orthagh (approved):

trasher:
Sorry for the late last commit, I should have seen that before.

src/Inventory/Asset/Peripheral.php
```diff
@@ -163,10 +162,13 @@ public function handle()
...
                 if ($data['found_inventories'][0] == 0) {
                     // add peripheral
-                    $handled_input['entities_id'] = $this->entities_id;
+                    $handled_input = $this->handleInput($val, $peripheral) + ['entities_id' => $this->entities_id];
                     $items_id = $peripheral->add(Toolbox::addslashes_deep($handled_input), [], false);
                 } else {
                     $items_id = $data['found_inventories'][0];
+                    $peripheral->getFromDB($items_id);
+                    $handled_input = $this->handleInput($val, $peripheral);
+                    //FIXME: handleInput
```
cedric-anne, line 171:
```suggestion
```

tests/functionnal/Lockedfield.php
```diff
@@ -602,4 +602,309 @@ public function testLockedDdValue()
...
+        if ($inventory->inError()) {
+            $this->dump($inventory->getErrors());
+        }
+        $this->boolean($inventory->inError())->isFalse();
+        $this->array($inventory->getErrors())->isEmpty();
+
+        //check created databases & instances
+        $this->integer(countElementsInTable(\DatabaseInstance::getTable()))->isIdenticalTo(1);
+
+        //ensure database version has been updated
```
cedric-anne, line 903:
```suggestion
        //make sure manufacturer is still the correct one
```

cedric-anne (approved):
I'm a bit lost after all these changes made on locks management, but it seems OK.

trasher:
> I'm a bit lost after all these changes made on locks management, but it seems OK.

So am I :D We've added tests, we fixed all that together with @stonebuzz; and requester has confirmed this fix his issues.

stonebuzz (approved):

stonebuzz:
Just a remark (this will be part of another PR)
a lock on the OS architecture is not displayed in the "lock" tab

image

image

trasher:
> Just a remark (this will be part of another PR) a lock on the OS architecture is not displayed in the "lock" tab

Lock tab shows only locks for current itemtype... You will only see locks on current computer, never OS devices, etc. But indeed, that is anothher problem, so another PR ;)

Issue #12677, opened 2022-09-17
ticket information alignment


matheusamendola:
### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

10.0.3

### Bug description

some ticket information items have different sizes depending on the text content.

When "Toggle Panels width" is selected, the information overlaps others, as in the second screenshot

In the browser console, the following warning appears in tinymce:
 
"Invalid value passed for the min_height option. The value must be a number."

### Your GLPI setup information

<details><summary>Information about system installation & configuration</summary><pre>GLPI 10.0.3 ( => /var/www/html)
Installation mode: TARBALL
Current language:en_US

</details><details><summary>Server</summary><pre> 
Operating system: Linux gscesvs019 5.4.0-125-generic #141-Ubuntu SMP Wed Aug 10 13:42:03 UTC 2022 x86_64
PHP 8.1.10 apache2handler (Core, FFI, PDO, Phar, Reflection, SPL, SimpleXML, Zend OPcache, apache2handler, apcu, bcmath, bz2,
	calendar, ctype, curl, date, dom, exif, fileinfo, filter, ftp, gd, gettext, hash, iconv, imap, intl, json, ldap, libxml,
	mbstring, mysqli, mysqlnd, openssl, pcre, pdo_mysql, posix, readline, session, shmop, soap, sockets, sodium, standard, sysvmsg,
	sysvsem, sysvshm, tokenizer, xml, xmlreader, xmlrpc, xmlwriter, xsl, zip, zlib)
Setup: max_execution_time="30" memory_limit="128M" post_max_size="8M" safe_mode="" session.save_handler="files"
	upload_max_filesize="2M" 
Software: Apache/2.4.41 (Ubuntu) (Apache/2.4.41 (Ubuntu) Server at csc.sacavalcante.com.br Port 443
)
	Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/105.0.0.0 Safari/537.36
Server Software: Ubuntu 20.04
	Server Version: 10.3.34-MariaDB-0ubuntu0.20.04.1
	Server SQL Mode: STRICT_TRANS_TABLES,ERROR_FOR_DIVISION_BY_ZERO,NO_AUTO_CREATE_USER,NO_ENGINE_SUBSTITUTION
	Host info: Localhost via UNIX socket
	
[...]

PR #12679, opened 2022-09-18
Fix global approval and dropdowns field display

cconard96:
Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #12677

The global approval field in ITIL Objects was not aligned properly if it was supposed to be in the second column.

There was also an issue with the dropdown widths when they had icon buttons. This seemed to be from changing `w-100` class to `w-150` when icons were present which fixed the display in the search engine, but inside a form it caused the dropdown to overflow into the next column.

### DIFF
--- a/src/Dropdown.php
+++ b/src/Dropdown.php
@@ -271,8 +271,7 @@ class Dropdown
         }
 
        // Display comment
-        $icons = "";
-        $icon_count = 0;
+        $icon_array = [];
         if ($params['comments']) {
             $comment_id      = Html::cleanId("comment_" . $params['name'] . $params['rand']);
             $link_id         = Html::cleanId("comment_link_" . $params['name'] . $params['rand']);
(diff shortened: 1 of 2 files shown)


cconard96:
For dropdown size issue, this PR seems to clean up the display nicely:

Before:
Selection_107

After:
Selection_108

src/Dropdown.php
```diff
@@ -389,20 +386,42 @@ public static function show($itemtype, $options = [])
...
+                $kb_link_icon .= '</div>';
+                $icon_array[] = $kb_link_icon;
             }
         }
 
         // Trick to get the "+" button to work with dropdowns that support multiple values
-        if (strlen($icons) == 0 && strlen($add_item_icon) > 0) {
-            $icons .= $add_item_icon;
+        if (count($icon_array) === 0 && $add_item_icon !== '') {
+            $icon_array[] .= $add_item_icon;
```
cedric-anne, line 399:
```suggestion
            $icon_array[] = $add_item_icon;
```

orthagh (approved):

src/Dropdown.php
trasher: ping @cconard96?

trasher (approved):

cedric-anne (approved):

PR #12636, opened 2022-09-13
Feat/logs viewer

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

PR #12687, opened 2022-09-19
fix(historical tab): check READ right to display historical tab

Rom1-B:
The READ right on Administration > Historical displayed/hid the Administration > Logs menu, but the data remained readable in the Historical tabs of the items.

This PR checks this right to display the tab.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !24946


orthagh (changes requested):
as seen together, the `logs` should not be used for checking historical tab.
The former is more targeted to an admin profile as historical can be used by an asset manager.

We need to have a new dedicated right.

To do this, in profile > admin tab, a new paragraph could be created for all type of logs:

- event logs
- log files (cf #12636)
- historical tabs

And so i guess, the change should target main branch

Issue #12690, opened 2022-09-19
Saved search entries visibility issue


Phorms:
### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

10.0.3

### Bug description

**Expected behavior:** Only your own and all plublic saved searches of the current category (tickets, assets, ...) are visible.

**Current behavior:** All saved searches of the current category (tickets, assets, ...) are visible. The private / public option has no impact.

User permissions seems not to affect this behavior.
I'm unaware of any settings that could affect the saved search.
I disabled all plugins, no change.

The system was updated from version 9.5.4 where the filter worked correctly.

Adding a new saved search entry does not create a new entry in the database table "glpi_savedsearches_users". Only a entry in "glpi_savedsearches" is created.

![grafik](https://user-images.githubusercontent.com/14980311/191042410-23a9c697-d079-463e-b481-8c06c239fe13.png)

All three entries are from different users and private.

![grafik](https://user-images.githubusercontent.com/14980311/191044800-18fbd9ad-4f0b-458e-be5b-5226857bb6f2.png)

But, I'm able to see them.

![grafik](https://user-images.githubusercontent.com/14980311/191045008-e4be1c12-c6a4-4b09-9aa4-002e34b88b47.png)

The SQL query received on our SQL server (no filter for 'is_private' field)

`SELECT glpi_savedsearches.*, glpi_savedsearches_users.id AS is_default FROM glpi_savedsearches 
LEFT JOIN glpi_savedsearches_users ON 
(glpi_savedsearches_users.savedsearches_id = glpi_savedsearches.id 
AND glpi_savedsearches.itemtype = glpi_savedsearches_users.itemtype 
AND glpi_savedsearches_users.users_id = '9') 
WHERE (glpi_savedsearches.itemtype = 'Ticket') ORDER BY itemtype, name`

Hope that helps.

### Relevant log output

```shell
access-errors.log - nothing
sql-errors.log - nothing
php-errors.log - nothing
```

### Page URL

/front/ticket.php

### Steps To reproduce

1. Open the ticket list (Assistance ->Tickets)
2. Click on "Lists"
3. Now you see under the new menu "Tickets" your own saved searches AND every other saved search (all saved searches of that category). The private ones have a small lock symbol.

### Your GLPI setup information

<details><summary>Information about system installation and configuration</summary><pre>GLPI 10.0.3 ( => /var/www/glpi)
[...]

PR #12692, opened 2022-09-19
Fix saved search visiblity

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #12690

SavedSearch_User relation class/table is only used to track the default searches for users and not which searches users have access to, so the condition in the JOIN clause doesn't make sense to me. Removing it fixes the visibility so users can only see their own private searches or, if they have the READ right, the public searches.

### DIFF
--- a/src/SavedSearch.php
+++ b/src/SavedSearch.php
@@ -796,15 +796,11 @@ class SavedSearch extends CommonDBTM implements ExtraVisibilityCriteria
             ],
             'FROM'      => $table,
             'LEFT JOIN' => [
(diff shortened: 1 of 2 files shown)


trasher:
I guess a test should be added on that

cedric-anne (approved):

orthagh (approved):

trasher (approved):

PR #12711, opened 2022-09-21
fix(lock): Display all lock available 

stonebuzz:
Actually, GLPI displays only the locks of the current itemtype.

But some data are related to ```Ìtem_XXXX``` relationship who do not have  their own form (like Computer)
and GLPI allow user to update it directly from current itemtype

It is frustrating to see the lock in the ```OS``` tab but not in the ```Lock``` tab

Actually is see two relationship in this case :
- ```Item_OperatingSystem```
- ```Item_Disk```

So this PR :

- compute related ```lock``` (```Item_OperatingSystem``` and ```Item_Disk```) for current ```itemtype```
- display ```itemtype``` and ```value``` from ```glpi_lockedfield``` table
- add ```Item_OperatingSystem``` to ```$CFG_GLPI['inventory_lockable_objects']``` (missing)
- Fix ```Locked items```display (Software / license ```header``` / ```massiveAction``` display if  there is no deleted item)

Before 

image

After
image

If ```Locked items``` part is empty
image

| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #number

### DIFF
--- a/inc/define.php
+++ b/inc/define.php
@@ -491,7 +491,7 @@ $CFG_GLPI['inventory_types'] = [
 
 $CFG_GLPI['inventory_lockable_objects'] = ['Computer_Item',  'Item_SoftwareLicense',
     'Item_SoftwareVersion', 'Item_Disk', 'ComputerVirtualMachine',
-    'NetworkPort', 'NetworkName', 'IPAddress'
+    'NetworkPort', 'NetworkName', 'IPAddress', 'Item_OperatingSystem'
 ];
 
 $CFG_GLPI["kb_types"]              = ['Budget', 'Change', 'Computer',
--- a/inc/define.php
+++ b/inc/define.php
@@ -490,8 +490,12 @@ $CFG_GLPI['inventory_types'] = [
 ];
 
 $CFG_GLPI['inventory_lockable_objects'] = ['Computer_Item',  'Item_SoftwareLicense',
-    'Item_SoftwareVersion', 'Item_Disk', 'ComputerVirtualMachine',
-    'NetworkPort', 'NetworkName', 'IPAddress', 'Item_OperatingSystem'
+    'Item_SoftwareVersion', 'Item_Disk', 'ComputerVirtualMachine','ComputerAntivirus',
+    'NetworkPort', 'NetworkName', 'IPAddress', 'Item_OperatingSystem', 'Item_DeviceBattery', 'Item_DeviceCase',
+    'Item_DeviceControl', 'Item_DeviceDrive', 'Item_DeviceFirmware', 'Item_DeviceGeneric', 'Item_DeviceGraphicCard',
+    'Item_DeviceHardDrive', 'Item_DeviceMemory', 'Item_DeviceMotherboard', 'Item_DeviceNetworkCard', 'Item_DevicePci',
+    'Item_DevicePowerSupply', 'Item_DeviceProcessor', 'Item_DeviceSensor', 'Item_DeviceSimcard', 'Item_DeviceSoundCard',
+    'DatabaseInstance', 'Item_RemoteManagement'
 ];
 
 $CFG_GLPI["kb_types"]              = ['Budget', 'Change', 'Computer',
--- a/src/Lock.php
+++ b/src/Lock.php
@@ -90,6 +90,7 @@ class Lock extends CommonGLPI
             echo "</th>";
             echo "<th>" . $lockedfield->getTypeName() . "</th>";
             echo "<th>" . __('Itemtype') . "</th>";
+            echo "<th>" . __('Link') . "</th>";
             echo "<th>" . __('Last inventoried value')  . "</th></tr>";
 
             $subquery = [];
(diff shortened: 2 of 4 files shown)


src/Lock.php
```diff
@@ -88,31 +88,81 @@ public static function showForItem(CommonDBTM $item)
...
+                            $lockedfield->getTable() . '.itemtype'  => $itemtype,
+                            $lockedfield->getTable() . '.is_global' => 1
                         ]
                     ]
                 ]
             ]);
 
+            //get other locked related to curren itemtype
+            //where GLPI allows to update their data
+            $inventory_lockable_objects = [
```
trasher, line 116: I do not understand why you limit to those 2 itemtypes only? It should be all of `inventory_lockable_objects` as far as I know, isn't it?

inc/define.php
```diff
@@ -491,7 +491,7 @@
 
 $CFG_GLPI['inventory_lockable_objects'] = ['Computer_Item',  'Item_SoftwareLicense',
     'Item_SoftwareVersion', 'Item_Disk', 'ComputerVirtualMachine',
-    'NetworkPort', 'NetworkName', 'IPAddress'
+    'NetworkPort', 'NetworkName', 'IPAddress', 'Item_OperatingSystem'
```
trasher, line 494: Sounds like several other possible itemtypes are missing here aswell, lika all item/devices, computer antivirus, and so on.

src/Lock.php
stonebuzz: Agree with that 
inventory_lockable_objects need to be review, some objects are not in this list

inc/define.php
stonebuzz: agree =)

src/Lock.php
trasher: Yes. Keep in mind `inventory_lockable_objects` has been designed for global locks, the dynamic lock mechanism can be done on any field of any object.

stonebuzz:
Redefined ```inventory_lockable_objects``` with all objects that have the ```is_dynamic``` column 
 ex : ```item_DeviceProcessor```, ```DatabaseInstance```, ```ComputerAntivirus```.
and can be linked to an asset
by ```ìtemtype_id``` (ie: ```computers_id```) or ```itemtype / items_id```

stonebuzz:
image

src/Lock.php
```diff
@@ -85,34 +85,107 @@ public static function showForItem(CommonDBTM $item)
...
+                    ) {
+                        $query['WHERE'][] = [getTableForItemType($lockable_itemtype) . '.' . $lockable_object::$items_id => $ID];
+                    } elseif (property_exists($lockable_object, 'items_id_1') && getForeignKeyFieldForItemType($itemtype) == $lockable_object::$items_id_1) {
+                        $query['WHERE'][] = [getTableForItemType($lockable_itemtype) . '.' . $lockable_object::$items_id_1 => $ID];
+                    } else {
+                        // ::$items_id != getForeignKeyFieldForItemType($itemtype)
+                        // ex: get ComputerVirtualMachine from NetworkEquipement
+                        continue;
+                    }
+                }
```
cedric-anne, line 168: You could probably use result of `CommonDBConnexity::getSQLCriteriaToSearchForItem()` here.
```suggestion
                if (is_a($lockable_itemtype, CommonDBConnexity::class, true)) {
                    $connexity_criteria = $lockable_itemtype::getSQLCriteriaToSearchForItem($itemtype, $ID);
                    if ($connexity_criteria === null) {
                        continue;
                    }
                    $query['WHERE'][] = $connexity_criteria['WHERE'];
                }
```

trasher (approved):

src/Lock.php
stonebuzz: does not work with this method, the generated ```where``` clause contains ambiguous column

```sql
  Error: Column 'items_id' in where clause is ambiguous
  Backtrace :
  src/DBmysqlIterator.php:110                        
  src/DBmysql.php:1048                               DBmysqlIterator->execute()
  src/Lock.php:158                                   DBmysql->request()
  src/Lock.php:660                                   Lock::showForItem()
  src/CommonGLPI.php:689                             Lock::displayTabContentForItem()
  ajax/common.tabs.php:113                           CommonGLPI::displayStandardTab()
```

cedric-anne: Maybe the method could be updated to user `table.field` instead of only `field` on its `WHERE` conditions.

stonebuzz: I think it's going to be quite difficult

```sql
  Error: Unknown column 'glpi_computers.computers_id' in 'where clause'
  Backtrace :
  src/DBmysqlIterator.php:110                        
  src/DBmysql.php:1048                               DBmysqlIterator->execute()
  src/Lock.php:183                                   DBmysql->request()
  src/Lock.php:684                                   Lock::showForItem()
  src/CommonGLPI.php:689                             Lock::displayTabContentForItem()
  ajax/common.tabs.php:113                           CommonGLPI::displayStandardTab()
  {"user":"2@stanislas-asus-desktop"} 
```

cedric-anne (approved):

PR #12749, opened 2022-09-23
fix(Notification Template Translation): list url 404

Rom1-B:
In the translations of the notification templates. the link refers to: ".../front/notificationtemplateetranslation.php" (which does not exist) instead of ".../front/notificationtemplate.form.php?id=xx"

image

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !24997

### DIFF
--- a/front/notificationtemplatetranslation.form.php
+++ b/front/notificationtemplatetranslation.form.php
@@ -45,6 +45,10 @@ if (!isset($_GET["id"])) {
 
 $language = new NotificationTemplateTranslation();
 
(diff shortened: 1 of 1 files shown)


cedric-anne (approved):
This is ugly, but this is not the only place that is using this hack, so OK.

trasher (approved):

PR #12792, opened 2022-09-28
feat(inventory): add dedicated right for locked field

stonebuzz:
Add dedicated right for locked field (only ```CREATE``` (for global lock) and ```PURGE``` are needed)

image

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #number

### DIFF
--- a/src/Lockedfield.php
+++ b/src/Lockedfield.php
@@ -249,11 +249,6 @@ class Lockedfield extends CommonDBTM
         return ['update', 'clone'];
     }
 
-    public static function canPurge()
-    {
-        return Session::haveRight(static::$rightname, PURGE);
-    }
-
     public function prepareInputForAdd($input)
     {
         return $this->prepareInput($input);
@@ -288,13 +283,6 @@ class Lockedfield extends CommonDBTM
         return true;
     }
 
-    public static function canCreate()
-    {
-        if (static::$rightname) {
-            return Session::haveRight(static::$rightname, CREATE);
-        }
-        return false;
-    }
 
     /**
      * List of itemtypes/fields that can be locked globally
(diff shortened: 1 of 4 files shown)


orthagh (approved):

cedric-anne:
Can you check why the `PURGE` right is not placed in same column as `PURGE` right of other objects ?

stonebuzz:
Fixed 
image

trasher (approved):

src/Lockedfield.php
```diff
@@ -251,7 +251,7 @@ public function getForbiddenStandardMassiveAction()
 
     public static function canPurge()
     {
-        return Session::haveRight(static::$rightname, UPDATE);
+        return Session::haveRight(static::$rightname, PURGE);
     }
```
cedric-anne, line 255: This method is probably useless now, as the default implementation can be used now.

src/Lockedfield.php
```diff
@@ -291,7 +291,7 @@ public function showForm($ID, array $options = [])
     public static function canCreate()
     {
         if (static::$rightname) {
-            return Session::haveRight(static::$rightname, UPDATE);
+            return Session::haveRight(static::$rightname, CREATE);
         }
         return false;
     }
```
cedric-anne, line 297: This method is probably useless now, as the default implementation can be used now.

cedric-anne (changes requested):

src/Lockedfield.php
stonebuzz: indeed

src/Lockedfield.php
stonebuzz: indeed

cedric-anne (approved):

PR #12821, opened 2022-09-30
Fix undefined cas_version

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #12814


trasher (approved):

cedric-anne (changes requested):
I think the problem is that this entry is not present in `install/empty_data.php`. Indeed, it has been added in GLPI 9.3 -> 9.4 migration, but is not present in a fresh installation.

cedric-anne (approved):

PR #12880, opened 2022-10-05
fix(mini dashboard): card display issue

Rom1-B:
When we zoomed in, the maps went beyond the frame.

image

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !25101


cconard96:
This seems to break resizing cards again.
Resize a card to only be in the first row, save, and then try resizing it back to take up both rows or try moving it from the first row to the second one.

orthagh:
replaced by #12938

PR #12884, opened 2022-10-05
Feat/dashboard palettes

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

Include #12591 

- [x] user can choose a palette for a card
- [x] plugin developers can add new palettes

image


cconard96:
I see the palette selection elements but they are hidden by default.
Am I missing some way to use the palettes?

Selection_117

orthagh:
Good catch, i did only my tests on existing cards, not the scenario of adding a new one.
I will check later i can support also addition process

cconard96 (approved):

cedric-anne (approved):

PR #12944, opened 2022-10-11
feat(lock): add additional info

stonebuzz:
add additional info about ```locked fields``` and ```locked items```

image

image

| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #number

### DIFF
--- a/src/Lock.php
+++ b/src/Lock.php
@@ -270,7 +270,7 @@ class Lock extends CommonGLPI
         ]);
         $header |= $results['header'];
 
-        echo "<div class='alert alert-primary d-flex align-items-center' role='alert'>";
+        echo "<div class='alert alert-primary d-flex align-items-center mb-4' role='alert'>";
         echo "<i class='fas fa-info-circle fa-xl'></i>";
         echo "<span class='ms-2'>";
         echo __("A locked items is a manually deleted items (example: a monitor).");
--- a/src/Lock.php
+++ b/src/Lock.php
@@ -253,7 +253,6 @@ class Lock extends CommonGLPI
                 echo "</tbody>";
                 echo "</table>";
             }
-
         }
         echo "</br><div width='100%'>";
         echo "<form method='post' id='lock_form' name='lock_form' action='" . Toolbox::getItemTypeFormURL(__CLASS__) . "'>";
--- a/src/Lock.php
+++ b/src/Lock.php
@@ -73,27 +73,16 @@ class Lock extends CommonGLPI
             return false;
         }
 
+        echo "<div class='alert alert-primary d-flex align-items-center' role='alert'>";
+        echo "<i class='fas fa-info-circle fa-xl'></i>";
+        echo "<span class='ms-2'>";
+        echo __("Here the locked fields are displayed.<br> A locked field is a manually modified field.
+        <br>The automatic inventory will no longer modify this field, unless you unlock it.");
+        echo "</span>";
+        echo "</div>";
+
         $lockedfield = new Lockedfield();
         if ($lockedfield->isHandled($item)) {
-            $rand = mt_rand();
-            Html::openMassiveActionsForm('mass' . __CLASS__ . $rand);
-            $massiveactionparams = [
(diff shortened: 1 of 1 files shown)


src/Lock.php
```diff
@@ -252,6 +269,14 @@ public static function showForItem(CommonDBTM $item)
         ]);
         $header |= $results['header'];
 
+        echo "<div class='alert alert-primary d-flex align-items-center' role='alert'>";
+        echo "<i class='fas fa-info-circle fa-xl'></i>";
+        echo "<span class='ms-2'>";
+        echo __("Here the locked items are displayed.<br> A locked items is a manually deleted items (example: a monitor).
```
trasher, line 275:
```suggestion
        echo __("A locked item is a manually deleted item (example: a monitor).
```

src/Lock.php
```diff
@@ -156,87 +145,115 @@ public static function showForItem(CommonDBTM $item)
...
-                echo "<td class='left'>" . $row['value'] . "</td>";
-                echo "</tr>\n";
+                echo "</table>";
+                $massiveactionparams['ontop'] = false;
+                Html::showMassiveActions($massiveactionparams);
+                Html::closeForm();
+            } else {
+                echo "<table class='tab_cadre_fixehov'>";
+                echo "<tbody>";
+                echo "<tr><th colspan='5' class='center'>" . __('Locked fields') . "</th></tr>";
```
trasher, line 251:
```suggestion
                echo "<tr><th colspan='5' class='center'>" . _n('Locked field', 'Locked fields') . "</th></tr>";
```

src/Lock.php
```diff
@@ -73,27 +73,16 @@ public static function showForItem(CommonDBTM $item)
             return false;
         }
 
+        echo "<div class='alert alert-primary d-flex align-items-center' role='alert'>";
+        echo "<i class='fas fa-info-circle fa-xl'></i>";
+        echo "<span class='ms-2'>";
+        echo __("Here the locked fields are displayed.<br> A locked field is a manually modified field.
```
trasher, line 79:
```suggestion
        echo __("A locked field is a manually modified field.
```

src/Lock.php
trasher: Maybe should you use anther concatenated string for second sentence instead of using an html tag in translations.

stonebuzz:
image

src/Lock.php
stonebuzz: fixed

src/Lock.php
stonebuzz: fixed

src/Lock.php
stonebuzz: fixed

src/Lock.php
```diff
@@ -73,27 +73,17 @@ public static function showForItem(CommonDBTM $item)
             return false;
         }
 
+        echo "<div class='alert alert-primary d-flex align-items-center' role='alert'>";
```
trasher, line 76: Should not this messages be hidden if there are no locked fields (condition in iterator count)?

src/Lock.php
```diff
@@ -252,6 +270,15 @@ public static function showForItem(CommonDBTM $item)
         ]);
         $header |= $results['header'];
 
+        echo "<div class='alert alert-primary d-flex align-items-center' role='alert'>";
```
trasher, line 273: Should not this message be hidden if itemtype is not computer?

trasher (commented):
OK for me appart of that :)

src/Lock.php
stonebuzz: nop, because ```Item_SoftwareVersion``` and ```Item_SoftwareLicense``` are displayed (if available) regardless of the itemtype

trasher (approved):

orthagh (commented):
Seems good for me, but maybe add a little spacing between the 2nd message and the above table

PR #12973, opened 2022-10-13
Improve dashboards performances

AdrienClairembault:
I've been notified that dashboard performances are not great on heavy GLPI databases and I'm looking for potential solutions.

Here are the potential fixes I've found so far:
- [x] Remove duplicated AJAX request
- [x] Restore browser cache
- [x] Split AJAX requests

1\) Duplicated AJAX request

I've found that `getCardsAjax` was being called twice when loading/refreshing a dashboard
image
image

2\) Restore browser cache

Browser cache has been broken since #9370 as `POST` request can't be cached.
At that time, I'd propose to revert to a `GET` request (#9548) but it wasn't possible due to the request size and no solution was found at the time

Is it possible to find a solution for this now ?

I've heard of a new `QUERY` request type that allow cached content of unlimited size but I couldn't find anything regarding browser support (it is quite recent) so it probably isn't usable yet.

The obvious solution would be to lighten the request and get it under the `GET` limit.
The current request content looks like this:
image
It seems to me that it contains mainly data that is already known to the back-end in case of a dashboard refresh (widget color, grid stack ids, ...).
image
I've also noted that the `c_cache_key` data seems to be unused as I can only find the code that set the value:
image

I may be missing some information but it seems to me that a full dashboard refresh should only send the dashboard ID and the current filters to the backend.
Doing so would reduce the query size drastically.
The other information may be needed for some other actions (editing a card ? I'm not sure) but not for the load/refresh action.

3\) Split AJAX requests

In the first iteration of the dashboards every card had its own AJAX request.
This implementation was changed because it would trigger too many request for some servers and it seems now that we are now using a single request for all the cards.

This is problematic because if you have a very slow card (due to a complex SQL query or a heavy database) then nothing is displayed in the meantime.

Maybe some cards should be flagged as "slow" (store the last 10 executions time then do an average) and be run in their own AJAX request in this case.

This mean you would have a "main" AJAX request with all the quick data and a few specific requests for the slow loading dashboards.

---

What are you thoughts about it @orthagh @cedric-anne @trasher @cconard96 ?
Specifically:
- is my assumption for 2 correct and that we could keep the request payload super light as none of these data are really needed ?
- is 3 doable ? am I correct to assume all dashboard are always loaded by a single ajax query at this moment ?
- any other potential solutions I'm missing ? 

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !25025

### DIFF
--- a/src/Dashboard/Grid.php
+++ b/src/Dashboard/Grid.php
@@ -35,6 +35,7 @@
 
 namespace Glpi\Dashboard;
 
+use DateInterval;
 use DBConnection;
 use Dropdown;
 use Glpi\Application\View\TemplateRenderer;
@@ -116,10 +117,10 @@ class Grid
     /**
      * Init dashboards cards
      * A define.php constant (GLPI_AJAX_DASHBOARD) exists to control how the cards should be loaded
-     *  - if true: only the parent block will be initialized and the content will be load by ajax
-     *    pros: if a widget fails, only this one will crash
-     * - else: load all html
-     *    pros: better perfs
+     *  - if true: load all cards in seperate ajax request
+     *    pros: slow cards wont impact the others
+     * - else: load all cards in a single ajax request
+     *    pros: less strain for the server
      *
      * @return void
      */
@@ -143,16 +144,11 @@ class Grid
                 'card_id' => $card_id
             ];
 
-            if (GLPI_AJAX_DASHBOARD) {
-                $card_html    = <<<HTML
+            $card_html    = <<<HTML
             <div class="loading-card">
                <i class="fas fa-spinner fa-spin fa-3x"></i>
             </div>
 HTML;
-            } else {
-                $card_html = $this->getCardHtml($card_id, ['args' => $card_options]);
-            }
-
            // manage cache
             $dashboard_key = $this->current;
             $footprint = sha1(serialize($card_options) .
@@ -245,6 +241,8 @@ HTML;
      */
     public function show(bool $mini = false)
     {
+        global $GLPI_CACHE;
+
         $rand = mt_rand();
 
         if (!self::$embed && !$this->dashboard->canViewCurrent()) {
@@ -279,7 +277,12 @@ HTML;
             }
         }
 
-       // prepare all available cards
+        // Force clear the cards cache in debug mode
+        if ($_SESSION['glpi_use_mode'] === Session::DEBUG_MODE) {
+            $GLPI_CACHE->delete('getAllDasboardCards');
+        }
+
+        // prepare all available cards
         $cards = $this->getAllDasboardCards();
         $cards_json = json_encode($cards);
 
@@ -927,12 +930,10 @@ HTML;
 
             if ($use_cache) {
                 // browser cache
-                if (GLPI_AJAX_DASHBOARD) {
-                    header_remove('Pragma');
-                    header('Cache-Control: public');
-                    header('Cache-Control: max-age=' . $cache_age);
-                    header('Expires: ' . gmdate('D, d M Y H:i:s \G\M\T', time() + $cache_age));
-                }
+                header_remove('Pragma');
+                header('Cache-Control: public');
+                header('Cache-Control: max-age=' . $cache_age);
+                header('Expires: ' . gmdate('D, d M Y H:i:s \G\M\T', time() + $cache_age));
 
                 // server cache
                 $dashboard_cards = $GLPI_CACHE->get($cache_key, []);
@@ -1104,36 +1105,36 @@ HTML;
      */
     public function getAllDasboardCards($force = false): array
     {
-        global $CFG_GLPI;
+        global $CFG_GLPI, $GLPI_CACHE;
 
-       // anonymous fct for adding relevant filters to cards
-        $add_filters_fct = static function ($itemtable, $DB) {
-            $add_filters = [];
-            if ($DB->fieldExists($itemtable, "ititlcategories_id")) {
-                  $add_filters[] = "itilcategory";
-            }
-            if ($DB->fieldExists($itemtable, "requesttypes_id")) {
-                 $add_filters[] = "requesttype";
-            }
-            if ($DB->fieldExists($itemtable, "locations_id")) {
-                $add_filters[] = "location";
-            }
-            if ($DB->fieldExists($itemtable, "manufacturers_id")) {
-                $add_filters[] = "manufacturer";
-            }
-            if ($DB->fieldExists($itemtable, "groups_id_tech")) {
-                $add_filters[] = "group_tech";
-            }
-            if ($DB->fieldExists($itemtable, "users_id_tech")) {
-                $add_filters[] = "user_tech";
-            }
-
-            return $add_filters;
-        };
-
-        static $cards = null;
+        $cards = $GLPI_CACHE->get('getAllDasboardCards');
 
         if ($cards === null || $force) {
+            // anonymous fct for adding relevant filters to cards
+            $add_filters_fct = static function ($itemtable, $DB) {
+                $add_filters = [];
+                if ($DB->fieldExists($itemtable, "ititlcategories_id")) {
+                    $add_filters[] = "itilcategory";
+                }
+                if ($DB->fieldExists($itemtable, "requesttypes_id")) {
+                    $add_filters[] = "requesttype";
+                }
+                if ($DB->fieldExists($itemtable, "locations_id")) {
+                    $add_filters[] = "location";
+                }
+                if ($DB->fieldExists($itemtable, "manufacturers_id")) {
+                    $add_filters[] = "manufacturer";
+                }
+                if ($DB->fieldExists($itemtable, "groups_id_tech")) {
+                    $add_filters[] = "group_tech";
+                }
+                if ($DB->fieldExists($itemtable, "users_id_tech")) {
+                    $add_filters[] = "user_tech";
+                }
+
+                return $add_filters;
+            };
+
             $cards = [];
             $menu_itemtypes = $this->getMenuItemtypes();
             $DB_read = DBConnection::getReadConnection();
@@ -1204,7 +1205,7 @@ HTML;
                 ];
             }
 
-           // add multiple width for Assets itemtypes grouped by their foreign keys
+            // add multiple width for Assets itemtypes grouped by their foreign keys
             $assets = array_merge($CFG_GLPI['asset_types'], ['Software']);
             foreach ($assets as $itemtype) {
                 $fk_itemtypes = [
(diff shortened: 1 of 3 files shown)


cconard96:
The QUERY method looks like it's just a spec proposal and hasn't been accepted yet. Mozilla MDN had no information about it from what I see and not even Chrome which usually adds new features  even when they are just proposals hasn't added support as far as I know.

cedric-anne:
IMHO, doing separe request for each card is not a good idea from a performance point of view, as it will require to re-init GLPI on each request, including plugins, and that will result in a too important overhead.

I am not sure it is still relevant to send more than the dashboard ID and filter values in the request, so we should probably try to limit the aount of data that is sent and put back a GET query. It will, at least, prevent issues with CSRF token expiration when autorefresh is activated (see #12791).

I am not sure how dashboards are used, but I guess the same dashboard is shared between multiple users, so the best way to optimize them is probably to improve caching of data on server side.

orthagh:
1) seems ok

2) ok for returning to GET and check how to reduce query parameters.
For refreshing, it's possible we doesn't need full attributes.
It may be me who has been overprotective at the time of developing the feature but i had an hard time with some parameters loss (especially with markdown cards)


3) I disagree on  the previous statement regarding each queries executing separately.
Initially, the dashboard was done thus to **favor UX** (and i missed big data to be cautious about degraded performance)
BUT, there was, from the start a constant to disable ajax mode : GLPI_AJAX_DASHBOARD (enabled by default)
I'm not sure when we "reduced" the calls to one single query (#9370, i guess i was not sufficiently cautious) but i'm in favor of returning to how it was before.
And set GLPI_AJAX_DASHBOARD to 0 to still have the one query mode by default.
At least, that the current behavior of 1 ajax query matches GLPI_AJAX_DASHBOARD == 0.

One last thing, i didn't had the time yet to check how to improve that, but the `includes.php` take too much time for ajax queries.
We must find ways to make it faster (cache/shortcuts/disable some loads, i don't yet), i think doing ajax queries for a dashboard is still the good way to do.

AdrienClairembault:
> I'm not sure when we "reduced" the calls to one single query (https://github.com/glpi-project/glpi/pull/9370, i guess i was not sufficiently cautious) but i'm in favor of returning to how it was before.

That would be a good news because it would partly solve the 2 point if `GLPI_AJAX_DASHBOARD` is enabled, as single card payload would be way under the GET limit even with the current "extra" data

AdrienClairembault:
I've implemented the suggested new behavior for `GLPI_AJAX_DASHBOARD`.
If acceptable, it would solve 2 (partly, maybe another PR to reduce the data but it would be lesser priority if at least one mode has a working cache) & 3.

If you wish to do some functional validation, I'd suggest using a custom dashboard to simulate a slow request like this:

image

AdrienClairembault:
> One last thing, i didn't had the time yet to check how to improve that, but the includes.php take too much time for ajax queries.
We must find ways to make it faster (cache/shortcuts/disable some loads, i don't yet), i think doing ajax queries for a dashboard is still the good way to do.

Regarding this point it didn't seem to be that bad on my side (but my instance is quite light - some tests on real data would be more representative).
-> `includes.php` takes 0.02s to load with no plugins activated and between 0.1s and 0.2s with 20+ plugins.
It seems loading plugins would end up being up to 70% of the execution time in my case.
If we really want to do something specifics for dashboards we could skip loading plugins (with a few exception like advanced dashboards obviously).

cconard96:
> If we really want to do something specifics for dashboards we could skip loading plugins (with a few exception like advanced dashboards obviously).

It would be difficult to know ahead of time which plugins add cards to the dashboard unless we either keep a manifest of used hooks in the DB after the plugins are enabled or defer plugin loading until we know which cards are going to be loaded (figure out the plugins from the card data) and only load those ones.

AdrienClairembault:
> As seen together with @orthagh, some performances could be gained by improving `getAllDashboard()`. Loading all cards for each AJAX request (that request only one specific card) isn't very efficient.
> 
> -> I'll try to look into it.

Done [0019f94](https://github.com/glpi-project/glpi/pull/12973/commits/0019f94985561d55f525ee8afc5876f5b68da17f).

Before:
image

After:
image

I'll deal with the conflicts once everything is approved.

orthagh:
So a summary, without plugins, we have now the same performance in AJAX or not.
For a first pass, i think the pr is good to merge.

We may investigate the plugin issue and check if we can store those which have dashboard hooks and only load them (exclude those which didn't interact with dashboard)

orthagh (approved):

src/Dashboard/Grid.php
```diff
@@ -279,7 +276,16 @@ public function show(bool $mini = false)
...
 
-       // prepare all available cards
+        // Force clear the cards cache everytime the page is loaded
+        // This is needed because some plugins might add cards at runtime (like
+        // advanceddashboards) and are not aware of this new cache
+        // In the future (TODO 10.1 ?) plugins should clear this cache entry
+        // themselves as needed
+        $cache_manager = new CacheManager();
+        $GLPI_CACHE = $cache_manager->getCoreCacheInstance();
+        $GLPI_CACHE->delete('getAllDasboardCards');
```
cedric-anne, line 286: 1. `$GLPI_CACHE` should be accessed from its global declaration.
2. If we clean this cache entry everytime a plugin is activated/deactivated, there is no reason to clear this entry on each page (except on debug mode, as it can correspond to a "dev" environment).

```suggestion
        // In debug mode, force clear the cards cache everytime the page is loaded.
        // This is needed because some plugins might add cards at runtime (like
        // advanceddashboards) and are not aware of this new cache
        //
        // FIXME: When we will be able to distinguish "dev" and "prod" environments,
        // do this only in "dev" context.
        global $GLPI_CACHE;
        if ($_SESSION['glpi_use_mode'] === Session::DEBUG_MODE) {
            $GLPI_CACHE->delete('getAllDasboardCards');
        }
```

src/Dashboard/Grid.php
```diff
@@ -1106,6 +1110,14 @@ public function getAllDasboardCards($force = false): array
     {
         global $CFG_GLPI;
 
+        $cache_manager = new CacheManager();
+        $GLPI_CACHE = $cache_manager->getCoreCacheInstance();
```
cedric-anne, line 1114:
```suggestion
        global $CFG_GLPI, $GLPI_CACHE;
```

src/Dashboard/Grid.php
```diff
@@ -1106,6 +1110,14 @@ public function getAllDasboardCards($force = false): array
...
+
+        $cards = $GLPI_CACHE->get('getAllDasboardCards', null);
+        if ($cards && !$force) {
```
cedric-anne, line 1117: 1. `null` is the default value of the param.
2. It is better to use strict checks.
```suggestion
        $cards = $GLPI_CACHE->get('getAllDasboardCards');
        if ($cards !== null && !$force) {
```

cconard96 (approved):

src/Dashboard/Grid.php
orthagh: > If we clean this cache entry everytime a plugin is activated/deactivated, there is no reason to clear this entry on each page (except on debug mode, as it can correspond to a "dev" environment).

The plugin `advanceddashboard` can add new cards at glance (after its installation)

src/Dashboard/Grid.php
```diff
@@ -1189,266 +1159,302 @@ public function getAllDasboardCards($force = false): array
...
-            if (is_array($more_cards)) {
-                $cards = array_merge($cards, $more_cards);
-            }
+        $more_cards = Plugin::doHookFunction(Hooks::DASHBOARD_CARDS);
+        if (is_array($more_cards)) {
+            $cards = array_merge($cards, $more_cards);
         }
 
+        // Set cards specifications to cache
+        $GLPI_CACHE->set('getAllDasboardCards', $cards, new DateInterval('PT5M'));
```
cedric-anne, line 1457: Why only 5 minutes ?

src/Dashboard/Grid.php
AdrienClairembault: Yes that's what I tried to imply in the big comment block.
We can reword the comment if that wasn't clear for you @cedric-anne ?

cedric-anne: Solution could be to store into GLPI core cache the core cards only and delegate to each plugin the responsibility of implementing a caching logic.
```php
$core_cards = $GLPI_CACHE->get('getAllDasboardCards');
if ($core_cards === null || $force) {
   // build cards
}

$GLPI_CACHE->set('getAllDasboardCards');

$more_cards = Plugin::doHookFunction(Hooks::DASHBOARD_CARDS);
$cards = is_array($more_cards) ? array_merge($core_cards, $more_cards) : $core_cards;
```

src/Dashboard/Grid.php
AdrienClairembault: The cache will be cleared on each reload anyway (see other comments above) so I don't think the 5 minute limit will be reached.

If we "break" plugin compatibility in 10.1 and remove the cache clear on page load then 5 min seems a fair limit for unexpected cases (e.g. someone running a patch that add a new card type and might no think of running the clear cache command).
This cache save about 0.2s of computations, which is a big deal in multiple simultaneous AJAX calls but shouldn't hurt once every 5 min.

src/Dashboard/Grid.php
orthagh: we discussed that with @AdrienClairembault.
And for a bugfixes, i'm not in favor of this solution.
I took the example of advanceddashboard, but maybe other plugin do the same.

And for one call, we don't lost so much time.
So the current solution is ok i think

src/Dashboard/Grid.php
AdrienClairembault: As discussion, changed to 1 day in the last commit.

src/Dashboard/Grid.php
AdrienClairembault: Done [6f6044b](https://github.com/glpi-project/glpi/pull/12973/commits/6f6044b0c9df97d86558fdff2ebbf90abae661de).

src/Dashboard/Grid.php
```diff
@@ -1442,11 +1443,12 @@ public function getAllDasboardCards($force = false): array
                     'content' => __("Toggle edit mode to edit content"),
                 ]
             ];
+            $GLPI_CACHE->set('getAllDasboardCards', $cards, new DateInterval('P1D'));
```
cedric-anne, line 1446: Could probably be cached without time restriction. Unless GLPI is updated, there is no reason to have a different result, and when GLPI is updated, cache is cleared anyway.

```suggestion
            $GLPI_CACHE->set('getAllDasboardCards', $cards);
```

AdrienClairembault: Done.

cedric-anne (approved):

AdrienClairembault:
Rebased to fix conflicts.

cedric-anne (approved):

cconard96 (approved):

orthagh (approved):

trasher (approved):

PR #12997, opened 2022-10-17
fix(inventory): fix front right and add new right for global lock

stonebuzz:
Use ```UPDATE``` for ```create``` and ```purge``` global lock

Use ```CREATE``` for ```read``` and ```create``` standard lock

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #number

### DIFF
--- a/front/lockedfield.form.php
+++ b/front/lockedfield.form.php
@@ -37,7 +37,7 @@ use Glpi\Event;
 
 include('../inc/includes.php');
 
-Session::checkRight("locked_field", READ);
+Session::checkRight("locked_field", CREATE);
 
 if (!isset($_GET["id"])) {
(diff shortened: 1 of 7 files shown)


orthagh (changes requested):
As seen toogether, i think only an UPDATE right is required for global locks.

I don't see a scenario where an admin will requires to have fin rights.
I you need to be add a global right, you need probably also to delete.

Also, i think UPDATE is fine both in UI and technical pov, no need for a specific right

orthagh (approved):

trasher (approved):

PR #13013, opened 2022-10-18
CommonITIL: add 'Entity' meta search criteria

AdrienClairembault:
Add entities as a new meta criteria for CommonITIL items:

image

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !25017

### DIFF
--- a/src/CommonITILObject.php
+++ b/src/CommonITILObject.php
@@ -4292,7 +4292,7 @@ abstract class CommonITILObject extends CommonDBTM
             case 'time_to_own':
                 return 'IF(' . $DB->quoteName($table . '.' . $type) . ' IS NOT NULL
             AND ' . $DB->quoteName($table . '.status') . ' <> ' . self::WAITING . '
-            AND ((' . $DB->quoteName($table . '.takeintoaccountdate') . ' IS NOT NULL AND 
+            AND ((' . $DB->quoteName($table . '.takeintoaccountdate') . ' IS NOT NULL AND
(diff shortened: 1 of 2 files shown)


cedric-anne (commented):
`CommonITILObject::getAllTypesForHelpdesk()` method is supposed to return items that can be linked to an ITIL object. I do not really like the idea of deflecting it to add some meta criteria.

Maybe we could simply use a global rule like:
```php
if ($item->isField('entities_id') && !($item instanceof Entity)) {
    $types[Entity::class] = Entity::getTypeName(1);
}
```

AdrienClairembault:
> `CommonITILObject::getAllTypesForHelpdesk()` method is supposed to return items that can be linked to an ITIL object. I do not really like the idea of deflecting it to add some meta criteria.

I agree but I didn't find a better solution as it is called directly from `Search::getMetaItemtypeAvailable`:

image

Since `CommonITILObject::getAllTypesForHelpdesk()` also handle sorting the entries, it means either:
- "Hacking" the new entry through a parameter as I've done
- Add the new entry after the method call and sort again (core first then plugins)

Open to suggestions on this point (without rewriting too many lines as this target the BF branch).

cedric-anne:
In search engine, display of these values is made by `Dropdown::showItemTypes()` that is already sorting values.

cedric-anne (approved):

trasher (approved):

PR #13043, opened 2022-10-21
fix(ticket): rejectsolution raise 2 notifications

Rom1-B:
When a ticket was resolved, if its solution is rejected, it sent the rejection notification AND the follow-up notification.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !25098

### DIFF
--- a/src/ITILFollowup.php
+++ b/src/ITILFollowup.php
@@ -241,8 +241,6 @@ class ITILFollowup extends CommonDBChild
             'date' => $this->fields['date'],
         ]);
 
-        $donotif = !isset($this->input['_disablenotif']) && $CFG_GLPI["use_notifications"];
-
        // Check if stats should be computed after this change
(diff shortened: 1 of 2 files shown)


cedric-anne (approved):

cedric-anne (commented):
It is technically OK, but maybe it is intended to have both notifications sent. If we want to keep both, maybe in this case the followup notification should be sent before the rejectsolution one.

Rom1-B:
> It is technically OK, but maybe it is intended to have both notifications sent. If we want to keep both, maybe in this case the followup notification should be sent before the rejectsolution one.

IMHO, the event is the rejection of the solution, the fact that it creates a follow-up in the timeline, it is our internal way, but it is not the event.

cedric-anne (approved):

PR #13050, opened 2022-10-24
fix(ui): fix the display of the icon

stonebuzz:
Fix the display of the icon

Before : 
image

After : 
image

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #number

### DIFF
--- a/src/Dropdown.php
+++ b/src/Dropdown.php
@@ -884,8 +884,8 @@ class Dropdown
                 // templates for select2 dropdown
                 $js = <<<JAVASCRIPT
                 $(function() {
-                    formatFormIcon{$rand} = function(icon) {
-                        if (!icon.id) {
+                    const formatFormIcon = function(icon) {
+                        if (!icon.id || icon.id == '0') {
                             return icon.text;
                         }
                         var img = '<span><img alt="" src="{$CFG_GLPI['typedoc_icon_dir']}/'+icon.id+'" />';
@@ -894,8 +894,8 @@ class Dropdown
                     };
                     $("#dropdown_{$myname}{$rand}").select2({
                         width: '60%',
-                        templateSelection: formatFormIcon{$rand},
-                        templateResult: formatFormIcon{$rand}
+                        templateSelection: formatFormIcon,
+                        templateResult: formatFormIcon
                     });
                 });
 JAVASCRIPT;
--- a/src/Dropdown.php
+++ b/src/Dropdown.php
@@ -890,7 +890,7 @@ class Dropdown
                         }
                         var img = '<span><img alt="" src="{$CFG_GLPI['typedoc_icon_dir']}/'+icon.id+'" />';
                         var label = '<span>'+icon.text+'</span>';
-                        return $(img+'&nbsp'+label);
+                        return $(img+'&nbsp;'+label);
                     };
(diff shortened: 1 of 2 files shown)


trasher (approved):

cedric-anne:
The best solution would be to have an icon in each dropdown option.

stonebuzz:
> The best solution would be to have an icon in each dropdown option.

image

orthagh (approved):

src/Dropdown.php
```diff
@@ -863,18 +863,43 @@ public static function dropdownIcons($myname, $value, $store_path, $display = tr
...
+                // templates for select2 dropdown
+                $js = <<<JAVASCRIPT
+                $(function() {
+                    formatFormIcon{$rand} = function(icon) {
+                        if (!icon.id) {
+                            return icon.text;
+                        }
+                        var img = '<span><img alt="" src="{$CFG_GLPI['typedoc_icon_dir']}/'+icon.id+'" />';
+                        var label = '<span>'+icon.text+'</span>';
+                        return $(img+'&nbsp'+label);
```
cedric-anne, line 893:
```suggestion
                        return $(img+'&nbsp;'+label);
```

cedric-anne (changes requested):

src/Dropdown.php
```diff
@@ -863,18 +863,43 @@ public static function dropdownIcons($myname, $value, $store_path, $display = tr
...
                         $options
                     )
                 );
+
+                global $CFG_GLPI;
+
+                // templates for select2 dropdown
+                $js = <<<JAVASCRIPT
+                $(function() {
+                    formatFormIcon{$rand} = function(icon) {
```
cedric-anne, line 887: Will prevent function to be declared in global scope.
```suggestion
                    const formatFormIcon = function(icon) {
```

src/Dropdown.php
```diff
@@ -863,18 +863,43 @@ public static function dropdownIcons($myname, $value, $store_path, $display = tr
...
+                        if (!icon.id) {
```
cedric-anne, line 888: `'0'` string is not skipped for now.
```suggestion
                        if (!icon.id || icon.id == '0') {
```

src/Dropdown.php
```diff
@@ -863,18 +863,43 @@ public static function dropdownIcons($myname, $value, $store_path, $display = tr
...
+                            return icon.text;
+                        }
+                        var img = '<span><img alt="" src="{$CFG_GLPI['typedoc_icon_dir']}/'+icon.id+'" />';
+                        var label = '<span>'+icon.text+'</span>';
+                        return $(img+'&nbsp;'+label);
+                    };
+                    $("#dropdown_{$myname}{$rand}").select2({
+                        width: '60%',
+                        templateSelection: formatFormIcon{$rand},
+                        templateResult: formatFormIcon{$rand}
```
cedric-anne, line 898:
```suggestion
                        templateSelection: formatFormIcon,
                        templateResult: formatFormIcon
```

cedric-anne (changes requested):

cedric-anne (approved):

PR #13079, opened 2022-10-27
fix(ticket): fill auto-assign when no tech defined (only for tech) from API

stonebuzz:
Re-implementation of a feature removed in GLPI 10

Fill auto-assign when no tech defined (usefull from API)

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !25351

### DIFF
--- a/src/CommonITILObject.php
+++ b/src/CommonITILObject.php
@@ -321,6 +321,7 @@ abstract class CommonITILObject extends CommonDBTM
                     'items_id'          => $user['users_id'],
                     'itemtype'          => 'User',
                     'text'              => $name,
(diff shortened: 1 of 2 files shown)


trasher (approved):

cedric-anne:
I am not sure about side effects. As far as I remember we validated removal of this behaviour on API.

stonebuzz:
> I am not sure about side effects. As far as I remember we validated removal of this behaviour on API.

So, this option no longer work on GLPI 10 through API

image

cedric-anne:
Replaced by #13305.

PR #13099, opened 2022-10-28
feat(inventory): add new option to not change asset states

stonebuzz:
Add new option to avoid the change of ```states_id``` (when is set manually before inventory)

image

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #number

### DIFF
--- a/src/Inventory/Asset/MainAsset.php
+++ b/src/Inventory/Asset/MainAsset.php
@@ -645,12 +645,12 @@ abstract class MainAsset extends InventoryAsset
         $entities_id = $this->entities_id;
         $val->is_dynamic = 1;
         $val->entities_id = $entities_id;
-
-        if ($items_id != 0 && $this->states_id_default != '-1') {
-            $val->states_id = $this->states_id_default ?? 0;
+        $default_states_id = $this->states_id_default ?? 0;
+        if ($items_id != 0 && $default_states_id != '-1') {
+            $val->states_id = $default_states_id;
         } elseif ($items_id == 0) {
             //if create mode default states_id can't be '-1' put 0 if needed
-            $val->states_id = $this->states_id_default > 0 ? $this->states_id_default : 0;
+            $val->states_id = $default_states_id > 0 ? $default_states_id : 0;
         }
 
         // append data from RuleImportEntity
--- a/src/Inventory/Asset/MainAsset.php
+++ b/src/Inventory/Asset/MainAsset.php
@@ -645,7 +645,10 @@ abstract class MainAsset extends InventoryAsset
         $entities_id = $this->entities_id;
         $val->is_dynamic = 1;
         $val->entities_id = $entities_id;
-        $val->states_id = $this->states_id_default ?? 0;
+
+        if ($this->states_id_default ?? 0 != '-1') {
+            $val->states_id = $this->states_id_default ?? 0;
+        }
 
         // append data from RuleImportEntity
         foreach ($this->ruleentity_data as $attribute => $value) {
--- a/src/Inventory/Asset/MainAsset.php
+++ b/src/Inventory/Asset/MainAsset.php
@@ -646,7 +646,9 @@ abstract class MainAsset extends InventoryAsset
         $val->is_dynamic = 1;
         $val->entities_id = $entities_id;
 
-        if ($this->states_id_default != '-1') {
+        if ($items_id != 0 && $this->states_id_default != '-1') {
+            $val->states_id = $this->states_id_default ?? 0;
+        } elseif ($items_id == 0) {
             $val->states_id = $this->states_id_default ?? 0;
         }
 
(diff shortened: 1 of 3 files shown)


src/Inventory/Asset/MainAsset.php
```diff
@@ -645,7 +645,10 @@ public function rulepassed($items_id, $itemtype, $rules_id, $ports_id = 0)
         $entities_id = $this->entities_id;
         $val->is_dynamic = 1;
         $val->entities_id = $entities_id;
-        $val->states_id = $this->states_id_default ?? 0;
+
+        if ($this->states_id_default =! '-1') {
+            $val->states_id = $this->states_id_default ?? 0;
```
trasher, line 650:
```suggestion
        if ($this->states_id_default ?? 0 =! '-1') {
            $val->states_id = $this->states_id_default ?? 0;
```

stonebuzz: done

src/Inventory/Asset/MainAsset.php
```diff
@@ -645,7 +645,10 @@ public function rulepassed($items_id, $itemtype, $rules_id, $ports_id = 0)
         $entities_id = $this->entities_id;
         $val->is_dynamic = 1;
         $val->entities_id = $entities_id;
-        $val->states_id = $this->states_id_default ?? 0;
+
+        if ($this->states_id_default ?? 0 != '-1') {
```
cedric-anne, line 649: Maybe `0` value should be forced when we are in a new asset creation context, as this `0` value could probably be checked by other rules or plugins in other pieces of code.

src/Inventory/Asset/MainAsset.php
```diff
@@ -645,7 +645,13 @@ public function rulepassed($items_id, $itemtype, $rules_id, $ports_id = 0)
...
         $val->is_dynamic = 1;
         $val->entities_id = $entities_id;
-        $val->states_id = $this->states_id_default ?? 0;
+
+        if ($items_id != 0 && $this->states_id_default != '-1') {
+            $val->states_id = $this->states_id_default ?? 0;
+        } elseif ($items_id == 0) {
+            //if create mode default states_id can't be '-1' put 0 if needed
+            $val->states_id = $this->states_id_default > 0 ? $this->states_id_default : 0;
+        }
```
cedric-anne, line 654:
```suggestion
        $default_states_id = $this->states_id_default ?? 0;
        if ($items_id != 0 && $default_states_id != '-1') {
            $val->states_id = $default_states_id;
        } elseif ($items_id == 0) {
            //if create mode default states_id can't be '-1' put 0 if needed
            $val->states_id = $default_states_id > 0 ? $default_states_id : 0;
        }
```

src/Inventory/Conf.php
```diff
@@ -450,12 +450,15 @@ public function showConfigForm()
         echo "</label>";
         echo "</td>";
         echo "<td>";
+
+        var_dump($config['states_id_default']);
```
orthagh, line 454:
```suggestion
```

src/Inventory/Asset/MainAsset.php
```diff
@@ -689,6 +695,8 @@ public function rulepassed($items_id, $itemtype, $rules_id, $ports_id = 0)
             $this->agent->fields['entities_id'] = $entities_id;
         }
 
+
+
```
orthagh, line 699:
```suggestion
```

src/Inventory/Conf.php
```diff
@@ -120,7 +120,7 @@ class Conf extends CommonGLPI
         'component_control'              => 1,
         'component_battery'              => 1,
         'component_simcard'              => 1,
-        'states_id_default'              => 0,
+        'states_id_default'              => -1,
```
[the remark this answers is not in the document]
stonebuzz: people who add manually, then use the native inventory will not go to this option (but they won't have a problem)

People who only do native inventory, without touching this option will have the status empty (as if the default value was '----')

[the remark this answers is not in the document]
trasher: That's the point: it changes the default value. Are we sure it won't be a problem that status will be kept in 10.0.4; but was set to default "0" value previousely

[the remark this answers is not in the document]
stonebuzz: i see
agree with you

trasher (approved):

orthagh (approved):

TheRoby:
It's ok for me !
Thank you !
(but ... : the "do not change", is not translate ;)
And i think we have the same issue with network equipement (GLPI Inventory / config, network inventory)

But, again, very thank you !

stonebuzz:
> It's ok for me ! Thank you ! (but ... : the "do not change", is not translate ;) And i think we have the same issue with network equipement (GLPI Inventory / config, network inventory)
> 
> But, again, very thank you !

Hi @TheRoby 

translation are done with Transifex =)

Best regards

PR #13101, opened 2022-10-28
try to fix restoration of actor after a form reload

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes


orthagh:
too much issue to manage after merge of #12929, we should revert it for now 
and think about a real rework on ticket backend. There is too much data mess curently

PR #13110, opened 2022-10-29
feat: human search order

po1vo:
Human sort order for computers and network equipment.
Simple order by name:

```
p100
p1000
pc01
pc02
pc03
pc1
pc101
pc11
pc12
pc13
pc21
pc22
```
Human sort order:

```
p100
p1000
pc1
pc01
pc02
pc03
pc11
pc12
pc13
pc21
pc22
pc101
```

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | N/A

### DIFF
--- a/src/Search/Provider/SQLProvider.php
+++ b/src/Search/Provider/SQLProvider.php
@@ -2803,6 +2803,11 @@ final class SQLProvider implements SearchProviderInterface
                     case "glpi_ipaddresses.name":
                         $criterion = "INET6_ATON(`$table$addtable`.`$field`) $order";
                         break;
+                    case "glpi_computers.name":
+                    case "glpi_networkequipments.name":
+                        $_field = "`$table$addtable`.`$field`";
+                        $criterion = "SOUNDEX($_field),LENGTH($_field),$_field $order";
+                        break;
                 }
             }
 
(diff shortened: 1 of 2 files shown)


orthagh (commented):
Some comments:

- Afaik, `SOUNDEX` function only works properly with english language. As GLPI is internationalized, a solution for other languages would be appreciated. I don't find any good suggestion btw.
- I think all `name` column could benefit from this not only Computer and NetworkEquipment

cconard96:
SOUNDEX use would also mean that words that are spelled differently but sound the same in English would end up next to each other.
For example, sorting by a user's first name, you would have "Lewis" and "Luis" with the same SOUNDEX value.

It is hard to tell for certain how assets are named by everyone to know how this could affect their searches.

po1vo:
I admit this kind of sort works best with a "string + index" kind of item names (such as computer names), for real human names probably not so much. Regarding `SOUNDEX` not working right with international names, true, however i don't expect anybody to name their computers using non-Latin characters.

With all that said, may be make the human sort order an option?

cconard96:
> i don't expect anybody to name their computers using non-Latin characters.

GLPI is used in countries that don't use the Latin alphabet like Russia, Japan, UAE, etc. I don't even know how that function would work in those cases. Some languages may even use a mix of Latin and non-Latin characters.

trasher:
I also think proposed solution is not correct; and it's probably really useless to make that an option.

PR #13234, opened 2022-11-08
feat(core): install / update handle rules with xml files

stonebuzz:
Add capability to manage ```rules``` import on install and on update with xml files

On fresh install , inject all xml files from ```src/Ressources/xml/```

A new button  ```Reset``` will make it possible to delete specific rule (ie : ```RuleImportAsset```) and  inject standard / native xml file

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #number


src/Rule.php
```diff
@@ -3495,4 +3495,45 @@ public function prepareInputForClone($input)
...
+        $files = array_diff(scandir($base_dir, 1), array('..', '.'));
+        foreach ($files as $rule_file) {
+            if (($subtype != null && str_contains($rule_file, $subtype)) || is_null($subtype)) {
+                $_FILES['xml_file'] =
+                [
+                    "size" => filesize($base_dir . $rule_file),
+                    "tmp_name" => $base_dir . $rule_file,
+                    "error" => UPLOAD_ERR_OK
+                ];
+                RuleCollection::previewImportRules();
```
cedric-anne, line 3534: Current rule import process is made to be used from UI. I think it should be reactored a bit to be used from here.
1. A method could be used to extract rules specifications from a XML file (e.g. `function extractRulesFromFile(string $path): array`).
2. A method could be used to import rules from given specifications (e.g. `function importRules(array $rules)`).

inc/based_config.php
```diff
@@ -59,6 +59,7 @@
         'GLPI_CONFIG_DIR'      => GLPI_ROOT . '/config', // Path for configuration files (db, security key, ...)
         'GLPI_VAR_DIR'         => GLPI_ROOT . '/files',  // Path for all files
         'GLPI_MARKETPLACE_DIR' => GLPI_ROOT . '/marketplace', // Path for marketplace plugins
+        'GLPI_RESSOURCE_DIR'   => GLPI_ROOT . '/src/Ressources', // Path for marketplace plugins
```
cedric-anne, line 62: This directory should not be customizable. There is no need to declare a constant for it.

stonebuzz: done

src/Rule.php
stonebuzz: done

src/RuleCriteria.php
```diff
@@ -389,7 +389,8 @@ public static function match(RuleCriteria &$criterion, $field, &$criterias_resul
             return true;
         }
 
-        $pattern = trim($pattern);
+        //since 8.1 PHP Deprecated: trim(): Passing null to parameter #1 ($string) of type string is deprecated
+        $pattern = trim((string) $pattern);
```
cedric-anne, line 393:
```suggestion
        $pattern = trim($pattern ?? '');
```

src/Rule.php
```diff
@@ -3495,4 +3495,42 @@ public function prepareInputForClone($input)
...
+            $rules = $rule->find($input);
+            foreach ($rules as $data) {
+                $rule->delete($data);
+            }
+        }
+
+        $base_dir = GLPI_ROOT . '/src/Ressources/Rules/';
+        $files = array_diff(scandir($base_dir, 1), array('..', '.'));
+        foreach ($files as $rule_file) {
+            if (($subtype != null && str_contains($rule_file, $subtype)) || is_null($subtype)) {
```
cedric-anne, line 3527:
```suggestion
            if ($subtype === null || str_contains($rule_file, $subtype)) {
```

front/rule.common.php
```diff
@@ -53,9 +53,10 @@
     $rulecollection->changeRuleOrder($_POST["id"], $_POST["action"], $_POST['condition']);
     Html::back();
 } else if (isset($_POST["reinit"]) || isset($_GET['reinit'])) {
-   //reinitialize current rules
+    $subtype = isset($_POST["reinit"]) ? $_POST["reinit"] : $_GET['reinit'];
```
cedric-anne, line 56: Code is not used.
```suggestion
```

cedric-anne (commented):
I do not really like the idea to reuse the current rule import methods as they are expected some session/request values that are not appropriate in the case we are trying to handle.
Also, reused code is a bit complex as it is handling some features we will not have to support in rules init (unless I miss something), like mapping with dropdown values.

We already have a simplier logic in `RuleImportAsset::initRules()` that creates rules from an array, and we have to keep this logic to create rules defined by `ruleImportAsset_addGlobalCriteria` hook (unless we remove support of this hook that does not seems to be used anymore, and, as far as I understand the code, is completely broken).

I think the cleaner solution would be do do this kind of logic:
1) extract array from XML file;
2) call the hook (is still appropriate), and convert result to same format as format returned by XML);
3) create rules as it is done in `RuleImportAsset::initRules()`, by moving this logic in a `protected function createRulesFromArray(array $rules)` method in `RuleCollection` class.

src/Ressources/Rules/RuleAsset.xml
```diff
@@ -0,0 +1,105 @@
...
+      <ranking>1</ranking>
+      <name>Domain user assignation</name>
+      <description></description>
+      <match>AND</match>
+      <is_active>1</is_active>
+      <comment></comment>
+      <is_recursive>1</is_recursive>
+      <uuid>glpi_rule_rule_asset_domain_user_assignation</uuid>
+      <condition>3</condition>
+      <date_creation></date_creation>
```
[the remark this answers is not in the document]
stonebuzz: how to do it? ```date_creation``` is always different switch GLPI installation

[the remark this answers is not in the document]
cedric-anne: I mean "should be removed".

install/migrations/update_9.5.x_to_10.0.0/native_inventory.php
```diff
@@ -161,7 +161,7 @@
 
 if (countElementsInTable(Rule::getTable(), ['sub_type' => 'RuleImportAsset']) === 0) {
     //default rules.
-    RuleImportAsset::initRules(false, false, true);
+    RuleImportAsset::initRules(false, false, true, RuleImportAsset::class);
```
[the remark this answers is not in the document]
cedric-anne: Fixed by #13365 .

cedric-anne:
Replaced by #13821.

PR #13252, opened 2022-11-09
fix(rules): add missing or use hardcoded uuid

stonebuzz:
Add missing ```uuid``` from ```RuleImportAsset```

Use ```hardcoded``` ```uuid``` for ```Rules```

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #number


orthagh (approved):

trasher (approved):

cedric-anne (approved):
Having this kind of values for UUID does not seems to be a problem, but, IMHO, we should try to have an initialization logic that create uniques values without having to define them in default data.

PR #13257, opened 2022-11-09
fix namespaced tab not correctly set as active

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

Found while discuting with @mary-Clb about plugin development tutorial, set active tab doesn't work correctly with namespaced tabs.
Not sure it's the good way, but i don't have time today to check for a better solution.


orthagh:
Example : `front/marketplace.php?itemtype=Glpi\Marketplace\View&glpi_tab=Glpi\Marketplace\View$1`

cedric-anne (approved):

trasher (approved):

PR #13283, opened 2022-11-14
CommonITIL: restrict location dropdown to correct entity

AdrienClairembault:
Given the 3 following entities;

image

The location dropdown does not restrict the values correctly:

image

After fix:

image

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !25523

### DIFF
--- a/templates/components/itilobject/fields_panel.html.twig
+++ b/templates/components/itilobject/fields_panel.html.twig
@@ -212,7 +212,9 @@
                   item.fields['locations_id'],
                   'Location'|itemtype_name,
                   field_options|merge({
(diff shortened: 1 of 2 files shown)


orthagh (approved):

cedric-anne (approved):

AdrienClairembault:
Some tests were failing because they need at least one valid location to exists and were relying on invalid values.
Fixed by adding a new location in the correct entity.

trasher (approved):

PR #13289, opened 2022-11-14
Dropdowns with complex joins

btry:
requires SQL DISTINCT keyword to deduplicate entries; see commit 40400792cc48e25fe6015f97e8a983154998cc5a

In Formcreator's incoming feature to filter dropdowns (GLPI object or dropdown question types) items may e duplicated when a search filter is implemented.

To reproduce : 

Use Formcreator branch support/2.14.0 with PR https://github.com/pluginsGLPI/formcreator/pull/2978 (and dependencies in GLPI; see its description)

- Create 2 or more computers
- Create an appliance 
- Add several computers aas items in the appliance
- Create a form with a question of type GLPI Object / Appliance 
- Add a search filter on this question which allows the appliance to show
- Preview the form; develop the dropdown of the question
- See that the appliance appears as many times as items linked to it.

The problem comes from the lack of DISTINCT in the query which feeds the dropdown, via Dropdown::getDropdownValue() because such filter adds LEFT JOIN clauses. 

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #number


cedric-anne (approved):

trasher (approved):
Sounds good to me, but I cannot say if that can cause any issue

PR #13316, opened 2022-11-16
Fix satisfaction search options

AdrienClairembault:
A few issues on the satisfaction search options :
* It's missing from changes
* It does not show values above 5 stars (we can go to 10 with the new config)
* It disappears when reloading the search results

1\)  It's missing from changes
-> Added it to changes. 
They don't share the same searchoptions id as the original ids are sadly not available for changes.
Instead they match the ticket's ids + 200 (kinda hacky - any plan to drop number based search options ids in the future ?).

2\) It does not show values above 5 stars (we can go to 10 with the new config)
-> Rewrote `displaySatisfaction` to take into account entity config

3\) It disappears when reloading the search results
-> Rewrote `displaySatisfaction` to use javascript to trigger the rateit library.
The previous method (relying on data-xxxx attributes) does not seem to trigger on ajax reloads.

image

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !25160

### DIFF
--- a/src/CommonITILSatisfaction.php
+++ b/src/CommonITILSatisfaction.php
@@ -284,21 +284,19 @@ abstract class CommonITILSatisfaction extends CommonDBTM
             $value = $max_rate;
         }
 
-        if ($use_js) {
-            $rand = mt_rand();
-            $out = "<input type='hidden' id='backing_$rand'>";
-            $out .= "<div id='rateit_$rand' class='rateit'></div>";
-            $out .= Html::scriptBlock("
-                $(function () {
-                    $('#rateit_$rand').rateit({
-                        max: $max_rate,
-                        resetable: false,
-                        value: $value,
-                        readonly: true,
-                    });
+        $rand = mt_rand();
+        $out = "<input type='hidden' id='backing_$rand'>";
+        $out .= "<div id='rateit_$rand' class='rateit'></div>";
+        $out .= Html::scriptBlock("
+            $(function () {
+                $('#rateit_$rand').rateit({
+                    max: $max_rate,
+                    resetable: false,
+                    value: $value,
+                    readonly: true,
                 });
-            ");
-        }
+            });
+        ");
 
         return $out;
     }
--- a/src/CommonITILSatisfaction.php
+++ b/src/CommonITILSatisfaction.php
@@ -49,6 +49,7 @@ abstract class CommonITILSatisfaction extends CommonDBTM
     public const TYPE_EXTERNAL = 2;
 
     abstract public static function getConfigSufix(): string;
+    abstract public static function getSearchOptionIDOffset(): int;
 
     public static function getTypeName($nb = 0)
     {
@@ -354,8 +355,11 @@ abstract class CommonITILSatisfaction extends CommonDBTM
         return $itemtype::getFormURLWithID($satisfaction->fields[$itemtype::getForeignKeyField()]) . '&forcetab=' . $itemtype::getType() . '$3';
     }
 
-    public static function rawSearchOptionsToAdd(string $table, $base_id = 0)
+    public static function rawSearchOptionsToAdd()
     {
+        $base_id = static::getSearchOptionIDOffset();
+        $table = static::getTable();
+
         $tab[] = [
             'id'                 => 'satisfaction',
             'name'               => __('Satisfaction survey')
--- a/src/CommonITILSatisfaction.php
+++ b/src/CommonITILSatisfaction.php
@@ -215,6 +215,20 @@ abstract class CommonITILSatisfaction extends CommonDBTM
             }
         }
 
+        if (array_key_exists('satisfaction', $input) && $input['satisfaction'] >= 0) {
+            $item = static::getItemtype();
+            $item = new $item();
+            $fkey = static::getIndexName();
+            if ($item->getFromDB($this->fields[$fkey] ?? $this->fields[$fkey])) {
+                $max_rate = Entity::getUsedConfig(
+                    'inquest_config',
+                    $item->fields['entities_id'],
+                    'inquest_max_rate' . static::getConfigSufix()
+                );
+                $input['satisfaction_scaled_to_5'] = $input['satisfaction'] / ($max_rate / 5);
+            }
+        }
+
         return $input;
     }
 
@@ -251,7 +265,7 @@ abstract class CommonITILSatisfaction extends CommonDBTM
      *
      * @param int $value Between 0 and 10
      **/
-    public static function displaySatisfaction($value, $entities_id)
+    public static function displaySatisfaction($value, $entities_id, $use_js = true)
     {
         if (is_null($value)) {
             return "";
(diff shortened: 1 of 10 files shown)


src/Search/Provider/SQLProvider.php
```diff
@@ -5269,7 +5269,21 @@ public static function giveItem(
 
                 case 'glpi_ticketsatisfactions.satisfaction':
                     if ($html_output) {
-                        return \TicketSatisfaction::displaySatisfaction($data[$ID][0]['name']);
+                        $ticket = Ticket::getById($data['raw']['id']);
```
cedric-anne, line 5272: Is it possible to get entity id from search option additionnal fields? With your solution, if you display many tickets, it will produce many additionnal queries.

src/CommonITILSatisfaction.php
```diff
@@ -331,4 +353,76 @@ public static function getFormURLWithID($id = 0, $full = true)
         $itemtype = static::getItemtype();
         return $itemtype::getFormURLWithID($satisfaction->fields[$itemtype::getForeignKeyField()]) . '&forcetab=' . $itemtype::getType() . '$3';
     }
+
+    public static function rawSearchOptionsToAdd(string $table, $base_id = 0)
```
cedric-anne, line 357: I do not like the idea of changing base signature for this. It will force developpers to pass a new param to get expected SO ids. Maybe using an abstract getSearchOptionStartingIndex could be a better solution.

AdrienClairembault: Ok, I'll look into adding `getSearchOptionStartingIndex`.
Is there any plan to move away from these number based search options id in the future ?

src/Search/Provider/SQLProvider.php
AdrienClairembault: I'm not sure it can be done with the `additionalfields` parameter as we need the entities_id field from the ticket/change table, not the satisfaction table

```diff
diff --git a/src/CommonITILSatisfaction.php b/src/CommonITILSatisfaction.php
index 5c41f84940..212e05343e 100644
--- a/src/CommonITILSatisfaction.php
+++ b/src/CommonITILSatisfaction.php
@@ -412,7 +412,8 @@ abstract class CommonITILSatisfaction extends CommonDBTM
             'massiveaction'      => false,
             'joinparams'         => [
                 'jointype'           => 'child'
-            ]
+            ],
+            'additionalfields'   => ['entities_id'],
         ];

         $tab[] = [
(
```

image

src/CommonITILSatisfaction.php
cedric-anne: > Is there any plan to move away from these number based search options id in the future ?

We could use some kind of nameid (e.g. `id`, `name`, `satisfaction-rate`, ...), but we would have to find relevant name for every search option, and code update would be a real pain. I am not against such a change, but I am not sure anyone will be vonlunteer for this if we decide to do it.

AdrienClairembault:
(I'm aware of the tests failure, working on it)

src/CommonITILSatisfaction.php
cedric-anne: Maybe using something like this (not tested):
```diff
diff --git a/src/Search.php b/src/Search.php
index bd3ff9c92e..83df8fe645 100644
--- a/src/Search.php
+++ b/src/Search.php
@@ -3933,16 +3933,22 @@ JAVASCRIPT;
             && count($searchopt[$ID]["additionalfields"])
         ) {
             foreach ($searchopt[$ID]["additionalfields"] as $key) {
+                if (preg_match('/^TABLE\./', $key) === 1) {
+                    $key = preg_replace('/^TABLE\./', '', $key);
+                    $additionalfield_field = $DB->quoteName($orig_table . '.' . $key);
+                } else {
+                    $additionalfield_field = $DB->quoteName($table . $addtable . '.' . $key);
+                }
                 if (
                     $meta
                     || (isset($searchopt[$ID]["forcegroupby"]) && $searchopt[$ID]["forcegroupby"])
                 ) {
-                    $ADDITONALFIELDS .= " IFNULL(GROUP_CONCAT(DISTINCT CONCAT(IFNULL(`$table$addtable`.`$key`,
+                    $ADDITONALFIELDS .= " IFNULL(GROUP_CONCAT(DISTINCT CONCAT(IFNULL($additionalfield_field,
                                                                          '" . self::NULLVALUE . "'),
                                                    '" . self::SHORTSEP . "', $tocomputeid)ORDER BY $tocomputeid SEPARATOR '" . self::LONGSEP . "'), '" . self::NULLVALUE . self::SHORTSEP . "')
                                     AS `" . $NAME . "_$key`, ";
                 } else {
-                    $ADDITONALFIELDS .= "`$table$addtable`.`$key` AS `" . $NAME . "_$key`, ";
+                    $ADDITONALFIELDS .= "$additionalfield_field AS `" . $NAME . "_$key`, ";
                 }
             }
```

src/CommonITILSatisfaction.php
```diff
@@ -212,6 +215,20 @@ public function prepareInputForUpdate($input)
             }
         }
 
+        if (array_key_exists('satisfaction', $input) && $input['satisfaction'] >= 0) {
```
cedric-anne, line 218: Should probably done also in `prepareInputForAdd`.

src/CommonITILSatisfaction.php
```diff
@@ -248,18 +265,40 @@ public function post_UpdateItem($history = 1)
...
-            $value = 10;
+        if ($value > $max_rate) {
+            $value = $max_rate;
         }
 
-        $out = "<div class='rateit' data-rateit-value='$value' data-rateit-ispreset='true'
-               data-rateit-readonly='true'></div>";
+        if ($use_js) {
+            $rand = mt_rand();
+            $out = "<input type='hidden' id='backing_$rand'>";
```
cedric-anne, line 289: Is this used ?

src/CommonITILSatisfaction.php
AdrienClairembault: The inquest is created by the backend when the ticket is closed, users then update the values.

src/CommonITILSatisfaction.php
```diff
@@ -212,6 +215,20 @@ public function prepareInputForUpdate($input)
             }
         }
 
+        if (array_key_exists('satisfaction', $input) && $input['satisfaction'] >= 0) {
+            $item = static::getItemtype();
+            $item = new $item();
+            $fkey = static::getIndexName();
+            if ($item->getFromDB($this->fields[$fkey] ?? $this->fields[$fkey])) {
```
cedric-anne, line 222:
```suggestion
            if ($item->getFromDB($input[$fkey] ?? $this->fields[$fkey])) {
```

src/CommonITILSatisfaction.php
```diff
@@ -248,18 +265,38 @@ public function post_UpdateItem($history = 1)
...
-        if ($value > 10) {
-            $value = 10;
+        if ($value > $max_rate) {
+            $value = $max_rate;
         }
 
-        $out = "<div class='rateit' data-rateit-value='$value' data-rateit-ispreset='true'
-               data-rateit-readonly='true'></div>";
+        $rand = mt_rand();
+        $out = "<input type='hidden' id='backing_$rand'>";
```
cedric-anne, line 288:
```suggestion
```

src/CommonITILSatisfaction.php
AdrienClairembault: Nice catch, forgot to change to first variable name here.

cedric-anne (approved):

AdrienClairembault:
Rebased for conflicts

cedric-anne (approved):

orthagh (approved):

trasher (approved):

PR #13329, opened 2022-11-17
Fix empty cases search

btry:
Recent changes break Formcreator unit tests.

- When a plugin hook generates an empty fragment of SQL query (SELECT, WHERE). Hooks often add SQL depending on the itemtype. If itemtype does not patches anything, then the function returns an empty string
example 
```php
function plugin_formcreator_addDefaultSelect($itemtype) {
   switch ($itemtype) {
      case PluginFormcreatorIssue::class:
         return "`glpi_plugin_formcreator_issues`.`itemtype`, ";
   }
   return "";
}
```

- When a search is done with the meta value 'mygroups', it may be resolved into an empty array, later used un a SQL IN statement. The search engine should ignore the clause in most cases.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #number

### DIFF
--- a/src/Search/Provider/SQLProvider.php
+++ b/src/Search/Provider/SQLProvider.php
@@ -1214,16 +1214,25 @@ final class SQLProvider implements SearchProviderInterface
                 if ($val === 'mygroups') {
                     switch ($searchtype) {
                         case 'equals':
+                            if (count($_SESSION['glpigroups']) < 1) {
+                                return [];
+                            }
                             return [
                                 "$table.id" => $_SESSION['glpigroups']
                             ];
 
                         case 'notequals':
+                            if (count($_SESSION['glpigroups']) < 1) {
+                                return [];
+                            }
                             return [
                                 "$table.id" => ['NOT IN', $_SESSION['glpigroups']]
                             ];
 
                         case 'under':
+                            if (count($_SESSION['glpigroups']) < 1) {
+                                return [];
+                            }
                             $groups = $_SESSION['glpigroups'];
                             foreach ($_SESSION['glpigroups'] as $g) {
                                 $groups += getSonsOf($inittable, $g);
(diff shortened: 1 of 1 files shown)


src/Search/Provider/SQLProvider.php
```diff
@@ -1208,16 +1214,25 @@ public static function getWhereCriteria($nott, $itemtype, $ID, $searchtype, $val
                 if ($val === 'mygroups') {
                     switch ($searchtype) {
                         case 'equals':
+                            if (count($_SESSION['glpigroups']) < 1) {
```
cconard96, line 1217:
```suggestion
                            if (count($_SESSION['glpigroups']) === 0) {
```

src/Search/Provider/SQLProvider.php
```diff
@@ -1208,16 +1214,25 @@ public static function getWhereCriteria($nott, $itemtype, $ID, $searchtype, $val
...
+                                return [];
+                            }
                             return [
                                 "$table.id" => $_SESSION['glpigroups']
                             ];
 
                         case 'notequals':
+                            if (count($_SESSION['glpigroups']) < 1) {
```
cconard96, line 1225:
```suggestion
                            if (count($_SESSION['glpigroups']) === 0) {
```

src/Search/Provider/SQLProvider.php
```diff
@@ -1208,16 +1214,25 @@ public static function getWhereCriteria($nott, $itemtype, $ID, $searchtype, $val
...
+                                return [];
+                            }
                             return [
                                 "$table.id" => ['NOT IN', $_SESSION['glpigroups']]
                             ];
 
                         case 'under':
+                            if (count($_SESSION['glpigroups']) < 1) {
```
cconard96, line 1233:
```suggestion
                            if (count($_SESSION['glpigroups']) === 0) {
```

src/Search/Provider/SQLProvider.php
```diff
@@ -1228,6 +1243,9 @@ public static function getWhereCriteria($nott, $itemtype, $ID, $searchtype, $val
                             ];
 
                         case 'notunder':
+                            if (count($_SESSION['glpigroups']) < 1) {
```
btry, line 1246:
```suggestion
                            if (count($_SESSION['glpigroups']) === 0) {
```

cedric-anne (approved):

cconard96 (approved):

trasher (approved):

PR #13335, opened 2022-11-18
[DBAL] Replace SQL query strings with QueryFunction

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

First PR to replace SQL query strings with QueryFunction as an abstraction layer. The PR work was getting very big so I stopped here and would continue in other PRs. I also replaced some hard-coded backticks/quotes but that wasn't the main goal.


src/Api/API.php
```diff
@@ -1272,7 +1272,7 @@ protected function getItems($itemtype, $params = [], &$totalcount = 0)
         }
 
        // get result full row counts
-        $count_query = "SELECT COUNT(*) FROM {$DB->quoteName($table)} $join WHERE $where";
+        $count_query = "SELECT " . QueryFunction::count('*') . " FROM {$DB::quoteName($table)} $join WHERE $where";
         $totalcount = $DB->query($count_query)->fetch_row()[0];
```
cedric-anne, line 1276: IMHO, in this kind of cases, we should probably replace usage of `$DB->query()` by `$DB->request()`.

src/QueryFunction.php
```diff
@@ -80,6 +80,43 @@ public function getValue()
             );
             $output .= $alias;
             return $output;
+        } else if ($this->name === 'GROUP_CONCAT') {
```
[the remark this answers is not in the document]
cconard96: I wouldn't want it to stay in the long-term either especially if/when multiple databases are supported. The alternate solution used was a function that returned an array of functions with optional output callbacks depending on the current platform.

Separate classes for any function that is more than just a simple list of parameters (and in the same order for every platform) could work well, but static methods on QueryFunction should be how they get used.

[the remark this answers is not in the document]
cconard96: I'm curious what other people think. There are a lot of different functions being used and some have different parameter orders on different platforms or unique syntax beyond a simple list of parameters. Would it be best to handle them within a single method using conditionals, formatting callables as [seen here](https://github.com/cconard96/glpi/blob/fe1ea2d0c398a800ce65a837ebdb412050797ef7/src/QueryFunction.php#L131-L224) or separate classes for each function.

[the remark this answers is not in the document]
trasher: +1 to move `Query*` in their own namespace. It's been added before we begin to use namespaces.

On the `QueryFunction` split, I agree with Cédric. Indeed, in a near future, a switch/case will be difficult to understand and maintain; and I guess dedicated classes will be simplier to deal with.

cedric-anne (changes requested):

src/Console/Task/UnlockCommand.php
```diff
@@ -102,20 +103,16 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
                     )
                 ],
                 'FROM'   => CronTask::getTable(),
                 'WHERE'  => [
                     'state' => CronTask::STATE_RUNNING,
-                    new QueryExpression(
-                        'UNIX_TIMESTAMP(' .  $this->db->quoteName('lastrun') . ') + ' . $delay
-                        . ' <  UNIX_TIMESTAMP(NOW())'
+                    QueryFunction::unixTimestamp(
+                        $this->db::quoteName('lastrun') . " + $delay < " . QueryFunction::unixTimestamp(QueryFunction::now())
```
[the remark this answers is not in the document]
cconard96: unixTimestamp takes an expression already and returns a QueryFunction instance that wouldn't be re-quoted.

Evaluation:
1. QueryFunction::unixTimestamp(QueryFunction::now() -> UNIX_TIMESTAMP(NOW())
2. QueryFunction::unixTimestamp($this->db::quoteName('lastrun') . " + $delay < " . UNIX_TIMESTAMP(NOW())) -> UNIX_TIMESTAMP(`lastrun` + $delay < UNIX_TIMESTAMP(NOW()))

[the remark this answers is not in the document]
cedric-anne: `` UNIX_TIMESTAMP(lastrun + $delay < UNIX_TIMESTAMP(NOW())) `` does not seems valid. It should probably be `` UNIX_TIMESTAMP(lastrun + $delay) < UNIX_TIMESTAMP(NOW()) ``

cedric-anne (approved):
It is OK for me as a first step.

On this kind of subject, that could lead to massive code change, I would prefer that we go through a preliminary stage of validation of technical specifications. It will probably make the review smoother to do things in smaller steps.

For future steps, I propose to do a separate PR for each SQL function.

src/Api/API.php
trasher, line 1: Since old API will be deprecated in 10.1; is it really useful to make so much changes here?

src/DBAL/Function/Count.php
```diff
@@ -0,0 +1,56 @@
...
+ * ---------------------------------------------------------------------
+ */
+
+namespace Glpi\DBAL\Function;
+
+use Glpi\DBAL\QueryFunction;
+
+/**
+ * Class for formatting an COUNT SQL function
+ * @interal Not for direct use. Use {@link QueryFunction} instead.
```
trasher, line 42:
```suggestion
 * @internal Not for direct use. Use {@link QueryFunction} instead.
```

(several other places to be changed)

src/Api/API.php
cedric-anne: REST API will not be deprecated. It is still the ony way to do "low level" operations.

src/DBAL/Function/Cast.php
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * Class for formatting an CAST SQL function
+ * @interal Not for direct use. Use {@link QueryFunction} instead.
+ */
+final class Cast implements Formatter
+{
+    public static function format(array $parameters): string
+    {
+        [$expression, $type] = $parameters;
```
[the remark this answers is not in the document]
cedric-anne: That was also the purpose of my comment:
image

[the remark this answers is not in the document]
trasher: Indeed, and I already agreed with you... I no longer had that in mind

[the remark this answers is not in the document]
trasher: As discuseed, to be changed later; let's try to validate/integrate this one as is.

cconard96:
Having to quote plain values and wrap them in a `QueryExpression` while identifiers get passed as a simple string seems odd, but it doesn't bother me that much overall.

I'm still not a fan of having so many class initializations directly in a criteria array or having 30+ classes to represent all the SQL functions being used (my biggest issue) when most are the same simple string concatenation to build the SQL. My opinion of that isn't going to change. In fact, I would have preferred not having to create any objects to handle SQL functions, but I had no choice because they had to avoid being quoted as identifiers in `DBmysqlIterator`. At least it was a single class being initialized, hidden behind static functions all from the same class.

cedric-anne:
> Having to quote plain values and wrap them in a `QueryExpression` while identifiers get passed as a simple string seems odd, but it doesn't bother me that much overall.

The fact is that it is how it works, for instance, when you pass a list of fields to select. We should try to keep something that matches existing code.

> I'm still not a fan of having so many class initializations directly in a criteria array or having 30+ classes to represent all the SQL functions being used (my biggest issue) when most are the same simple string concatenation to build the SQL. My opinion of that isn't going to change. In fact, I would have preferred not having to create any objects to handle SQL functions, but I had no choice because they had to avoid being quoted as identifiers in `DBmysqlIterator`. At least it was a single class being initialized, hidden behind static functions all from the same class.

Another solution is have something like this:
```php
// `QueryExpression` with `alias` support
class QueryExpression
{
    private string $expression;
    private string $alias;

    public function __construct(string $expression, ?string $alias = null)
    {
        $this->expression = $expression;
        $this->alias = $alias;
    }

    public function getValue()
    {
        $sql = $this->expression;
        if (!empty($this->alias)) {
            $sql .= ' AS ' . $DB::quoteName($this->alias);
        }
        return $sql;
    }
}

// `QueryFunction` that contains only helper methods
class QueryFunction
{
    public static function groupconcat(string|QueryExpression $expression, ?string $separator = null, bool $distinct = false, ?string $order_by = null, ?string $alias = null)
    {
        global $DB;

        $sql = 'GROUP_CONCAT(';
        if ($distinct) {
            $sql .= 'DISTINCT ';
        }
        $sql .= $DB::quoteName($expression);
        if ($order_by !== null) {
            $sql .= (new DBmysqlIterator(null))->handleOrderClause($order_by);
        }
        if ($separator !== null) {
            $sql .= ' SEPARATOR ' . $DB::quoteValue($separator);
        }
        $sql .= ')';

        return new QueryExpression($sql, $alias);
    }

    public static function __callStatic(string $name, array $arguments)
    {
        $expression = strtoupper($name)
            . '('
            . implode(',', array_map([$DB, 'quoteName'], $arguments[0]))
            . ')';

        $alias = $arguments[1] ?? null;

        return new QueryExpression($expression, $alias);
    }
}
```

With this, only complex methods have to be defined, and other methods can be handled by the `__callStatic` magic method, and we will only need a single class to handle this.

cconard96:
The alternate solution proposed seems good to me. Less code than there is even now and would just require some phpdoc to declare the dynamic methods for the majority of sql functions.

src/DBAL/QueryFunction.php
```diff
@@ -0,0 +1,357 @@
...
+    }
+
+    /**
+     * Build an IFNULL SQL function call
+     * @param string|QueryExpression $expression Expression to check
+     * @param string|QueryExpression $value Value to return if expression is null
+     * @param string|null $alias Function result alias (will be automatically quoted)
+     * @return QueryExpression
+     */
+    public static function ifNull(string|QueryExpression $expression, string|QueryExpression $value, ?string $alias = null): QueryExpression
```
cedric-anne, line 139: Maybe it should follow naming rule applied to magic methods.
Same remark for `nullIf`, `timestampDiff` and `dateDiff`.
```suggestion
    public static function ifnull(string|QueryExpression $expression, string|QueryExpression $value, ?string $alias = null): QueryExpression
```

src/DBAL/QueryFunction.php
```diff
@@ -0,0 +1,357 @@
...
+    }
+
+    /**
+     * Build a FROM_UNIXTIME SQL function call
+     * @param string|QueryExpression $expression Expression to convert
+     * @param string|QueryExpression|null $format Function result alias
+     * @param string|null $alias Function result alias (will be automatically quoted)
+     * @return QueryExpression
+     */
+    public static function fromUnixTimestamp(string|QueryExpression $expression, string|QueryExpression $format = null, ?string $alias = null): QueryExpression
```
cedric-anne, line 272: It should reflect the exact name of the MySQL method.
```suggestion
    public static function fromUnixtime(string|QueryExpression $expression, string|QueryExpression $format = null, ?string $alias = null): QueryExpression
```

src/DBAL/QueryFunction.php
```diff
@@ -0,0 +1,357 @@
...
+     * Build a DATE_FORMAT SQL function call
+     * @param string|QueryExpression $expression Expression to format
+     * @param string|QueryExpression $format Format to use
+     * @param string|null $alias Function result alias (will be automatically quoted)
+     * @return QueryExpression
+     */
+    public static function dateFormat(string|QueryExpression $expression, string|QueryExpression $format, ?string $alias = null): QueryExpression
+    {
+        return self::getExpression('DATE_FORMAT', [$expression, $format], $alias);
+    }
```
cedric-anne, line 291: According to MySQL documentation (https://dev.mysql.com/doc/refman/8.0/en/date-and-time-functions.html#function_date-format), `$format` will always be a string surrounded by `'`. We could automatize quoting of this parameter to make method usage simplier.

```suggestion
    public static function dateFormat(string|QueryExpression $expression, string $format, ?string $alias = null): QueryExpression
    {
        global $DB;
        $format = new QueryExpression($DB::quoteValue($format));
        return self::getExpression('DATE_FORMAT', [$expression, $format], $alias);
    }
```

src/DBAL/QueryFunction.php
```diff
@@ -0,0 +1,357 @@
...
+     * @param string|QueryExpression $expression Expression to pad
+     * @param string|QueryExpression $length Length to pad to
+     * @param string|QueryExpression $pad_string String to pad with
+     * @param string|null $alias Function result alias (will be automatically quoted)
+     * @return QueryExpression
+     */
+    public static function lpad(string|QueryExpression $expression, string|QueryExpression $length, string|QueryExpression $pad_string, ?string $alias = null): QueryExpression
+    {
+        return self::getExpression('LPAD', [$expression, $length, $pad_string], $alias);
+    }
```
cedric-anne, line 304: I think `$length` will always be an integer here, and `$pad_string` will probably always be a string value.

```suggestion
    public static function lpad(string|QueryExpression $expression, int $length, string $pad_string, ?string $alias = null): QueryExpression
    {
        global $DB;
        $length = new QueryExpression((string)$length);
        $pad_string = new QueryExpression($DB::quoteValue($pad_string));
        return self::getExpression('LPAD', [$expression, $length, $pad_string], $alias);
    }
```

src/DBAL/QueryFunction.php
```diff
@@ -0,0 +1,357 @@
...
+     * @param string|null $alias Function result alias (will be automatically quoted)
+     * @return QueryExpression
+     */
+    public static function round(string|QueryExpression $expression, string|QueryExpression $precision = null, ?string $alias = null): QueryExpression
+    {
+        if ($precision === null) {
+            $precision = new QueryExpression('0');
+        }
+        return self::getExpression('ROUND', [$expression, $precision], $alias);
+    }
```
cedric-anne, line 319: Precision will always be an integer here.

```suggestion
    public static function round(string|QueryExpression $expression, int $precision = 0, ?string $alias = null): QueryExpression
    {
        global $DB;
        $precision = new QueryExpression((string)$precision);
        return self::getExpression('ROUND', [$expression, $precision], $alias);
    }
```

cedric-anne (changes requested):

cedric-anne (approved):
Seems OK, but with a so huge diff, it is hard to stay focused on all the changes.

PR #13386, opened 2022-11-23
replace setup > general by twig template

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes


orthagh:
New settings page:
image

templates/pages/setup/general/general_setup.html.twig
```diff
@@ -0,0 +1,337 @@
...
+    ) }}
+
+    {{ fields.dropdownNumberField(
+        'ajax_limit_count',
+        config['ajax_limit_count'],
+        __("Don't show search engine in dropdowns if the number of items is less than"),
+        inline_field_options|merge({
+            'min': 1,
+            'max': 200,
+            'toadd': [__('Never')],
```
cedric-anne, line 162: Maybe it is more secure to force key value (not tested).
```suggestion
            'toadd': {'0': __('Never')},
```

templates/pages/setup/general/general_setup.html.twig
```diff
@@ -0,0 +1,337 @@
...
+            'width': 'auto',
+        })
+    ) }}
+
+    {% set allow_search_view_options = inline_field_options %}
+    {% if not config['allow_search_view'] %}
+        {% set allow_search_view_options = allow_search_view_options|merge({
+            'readonly': true,
+            'disabled': true,
+        }) %}
```
cedric-anne, line 194: I know it was done like this previously, but why ? `Items seen` criteria and global search are disctinct features.

templates/pages/setup/general/general_setup.html.twig
```diff
@@ -0,0 +1,337 @@
...
+        inline_field_options
+    ) }}
+
+    {% set lock_options = field_options %}
+    {% if not config['lock_use_lock_item'] %}
+        {% set lock_options = lock_options|merge({
+            'readonly': true,
+            'disabled': true,
+        }) %}
+    {% endif %}
```
cedric-anne, line 272: It forces user to first set 'Yes' , then save to be able to configure the feature. Maybe a small javascript listener could enhance UX.

templates/pages/setup/general/general_setup.html.twig
orthagh: I don't know, i'll remove the condition

cconard96:
I'm thinking the Global Search option is reliant on Items Search option because it does use a hidden Items Seen criteria. The difference is it searches across multiple itemtypes instead of one.

cedric-anne:
> I'm thinking the Global Search option is reliant on Items Search option because it does use a hidden Items Seen criteria. The difference is it searches across multiple itemtypes instead of one.

I pushed a commit to allow global search even when "view" search is not allowed.

cedric-anne (approved):

PR #13412, opened 2022-11-25
fix(document): show filename

Rom1-B:
When a document is uploaded from a ticket, its name looks like "`Document Ticket <ID>`". If this ticket has multiple documents, they all have the same name.

The idea is to display the file name on mouse-over

After:
image

| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !25684


orthagh (approved):

cedric-anne (commented):
Maybe another solution would be to add filename in displayed name, like this:
 - when `entry_i['name']` is empty => `name = entry_i['filename']`
 - when `entry_i['filename']` is empty => `name = entry_i['name']`
 - when both are not empty => `name = entry_i['name'] ~ '(' ~ entry_i['filename'] ~ ')'`

cedric-anne (approved):

trasher (approved):

PR #13527, opened 2022-12-05
Hide locked fields for deleted relation

AdrienClairembault:
From the "Locks" tab of a monitor, I can see the computer locked fields AND the locked fields of items linked to the computer:

image

@stonebuzz tell me that this is done on purpose (to make it easier to browse all locked fields from the parent computer) so no issues at this point.

However, the linked fields are still shown even if the relation is deleted:

image

They should be hidden in the case (the fields are irrelevant because the items are no longer linked).

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !25474

### DIFF
--- a/src/Lock.php
+++ b/src/Lock.php
@@ -819,7 +819,7 @@ class Lock extends CommonGLPI
             echo "<th width='10'></th>";
             echo "<th>" . DatabaseInstance::getTypeName(Session::getPluralNumber()) . "</th>";
             echo "<th>" . __('Name') . "</th>";
-            echo "<th>" . __('Version') . "</th>";
+            echo "<th>" . _n('Version', 'Versions', 1) . "</th>";
             echo "<th>" . __('Automatic inventory') . "</th>";
             echo "</tr>";
         }
--- a/src/Lock.php
+++ b/src/Lock.php
@@ -136,6 +136,11 @@ class Lock extends CommonGLPI
                         continue;
                     }
                     $query['WHERE'][] = $connexity_criteria['WHERE'];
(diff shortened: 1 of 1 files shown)


stonebuzz (changes requested):
This filter could be generalized and used by

- the first ```if``` statement -> ```instanceof CommonDBConnexity```

For ```Item_Devicexxx```,```NetworkPort```, ```NetworkName```, ```IPAddress```, ```Antivirus```, ```VirtualMachine```, ```Software``` ...

- the second ```elsif``` statement -> ```$CFG_GLPI['directconnect_types']```

For ```Monitor```, ```Peripheral```, ```Phone```, ```Printer```

- the last ```elseif``` statement -> ```isField($itemtype) && isField($items_id)```

For ```DatabaseInstance```

if I'm not mistaken, all these objects have the ```is_deleted``` field but it could be checked with

```php
if ($lockable_object->isField('is_deleted')) {

}
```

AdrienClairembault:
DatabaseInstance done in [94110ff](https://github.com/glpi-project/glpi/pull/13527/commits/94110ffdcd9031416b07cb0e73897ec127bc4b51).

Before:

image

The DB is deleted but its locked fields are still displayed AND it's missing from the second table that indicate deleted items.

After:

image

stonebuzz (approved):

cedric-anne (approved):

trasher (approved):

PR #13547, opened 2022-12-07
Fix analysis of raw plugin joins

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

Fixes parsing of join strings to handle table aliases and also gets the full join condition and coverts it to a QueryExpression instead of trying to parse it. Support for QueryExpression join conditions was added to the iterator.

### DIFF
--- a/src/DBmysqlIterator.php
+++ b/src/DBmysqlIterator.php
@@ -723,6 +723,8 @@ class DBmysqlIterator implements SeekableIterator, Countable
                 }
                 return $fkey;
             }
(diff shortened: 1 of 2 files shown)


cedric-anne (approved):

btry (approved):
Looks good : Formcreator's tests are now green (locally)

trasher (approved):

PR #13588, opened 2022-12-12
feat(timeline): add capability to change the layout of button

stonebuzz:
Add user pref option to split or not timeline action button.

Standard layout (```merged```)
image

New layout (```splitted```)
image

NB : do not hide buttons when form (```task```, ```followup```, ```solution``` ... ) are open

| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #number

### DIFF
--- a/templates/components/itilobject/timeline/filter_timeline.html.twig
+++ b/templates/components/itilobject/timeline/filter_timeline.html.twig
@@ -81,7 +81,7 @@
    }) %}
 {% endif %}
 
-<div class="filter-timeline float-end position-relative">
+<div class="filter-timeline float-end position-relative ms-auto">
    <span data-bs-toggle="tooltip" data-bs-placement="top" title="{{ __('Timeline filter') }}">
       <button type="button"
             class="btn btn-icon btn-ghost-secondary open-timeline-filter-popover"
--- a/templates/components/itilobject/timeline/filter_timeline.html.twig
+++ b/templates/components/itilobject/timeline/filter_timeline.html.twig
@@ -81,7 +81,7 @@
    }) %}
 {% endif %}
 
-<div class="filter-timeline float-end position-relative ms-auto">
+<div class="filter-timeline position-relative ms-auto">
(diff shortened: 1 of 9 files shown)


stonebuzz:
image

AdrienClairembault:
I think you should try to flex the container : 

image

Without this I get a 2 row display despite having enough space on my screen to fit all buttons:

image

AdrienClairembault:
Also, an open question: maybe we could gain space by shortening the labels ?
The first button is very short "Réponse" but the others are more verbose like "Ajouter une solution" instead of just "Solution".
I'm afraid that there wont be enough space on the screen for some languages.

AdrienClairembault (changes requested):
See comments.

AdrienClairembault:
I think validation was a different color in GLPI 9.5 (if you want to keep it consistent).
Some kind of lime green if I recall correctly ?

AdrienClairembault (approved):

templates/components/itilobject/timeline/filter_timeline.html.twig
```diff
@@ -81,7 +81,7 @@
    }) %}
 {% endif %}
 
-<div class="filter-timeline float-end position-relative">
+<div class="filter-timeline float-end position-relative ms-auto">
```
orthagh, line 84: i guess the float-end class is not needed anymore

stonebuzz: ```ms-auto``` is always needed

AdrienClairembault: yes but not `float-end` 😉

stonebuzz: Oo, sorry !

orthagh (approved):

cedric-anne (approved):

PR #13592, opened 2022-12-13
Increase masonry breakpoints to avoid overflow

AdrienClairembault:
image

It seems that showing 3 columns at 1400px is a bit too tight and wont show the full data (especially when using a real database and not a dev one).

You can find more "extreme" examples in the internal ref.

Fixed by adding a new 1600px breakpoint and using it for masonry content.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !25856


AdrienClairembault:
> I'm not a big fan to redeclare whole boostrap breakpoints (with indeed same values) Is it not possible to extends them to add xxxl entry ?

I've tried defining only the xxxl and it didn't work.
Maybe there is a special syntax for it but I wasn't able to find it (@cedric-anne thoughts ?)

cedric-anne:
> > I'm not a big fan to redeclare whole boostrap breakpoints (with indeed same values) Is it not possible to extends them to add xxxl entry ?
> 
> I've tried defining only the xxxl and it didn't work. Maybe there is a special syntax for it but I wasn't able to find it (@cedric-anne thoughts ?)

It does not seems to be possible to add only a new value here. All the map have to be redefined.

AdrienClairembault:
New commit to fix some visual issue.

Before:
image

After:
image

AdrienClairembault:
I've also noticed this issue with the planning card:

image

However since this is HTML generated by an outside lib (fullcalendar) I have no idea how to fix it.

AdrienClairembault:
And d1f959a480 for tasks.

Before:
image


After:
image

orthagh:
> I'm considering going to 2 column max for any resolution.
> Would that be ok @orthagh ?

Yes, the old design (without masonry) was on two columns.
So no problems to fix it by return to this number.

If you remove the hack on bootstrap breakpoints, event more.

AdrienClairembault:
Done.

cedric-anne (approved):

orthagh (approved):

trasher (approved):

PR #13595, opened 2022-12-13
enhance the entity name when selecting all

Rom1-B:
When selecting the active entities, the title is the same if you display all or the 1st entity in tree structure.

Example with this structure:

image

If we choose "select all" or "entity1" in recursive, we display `Root entity > entity1 (tree structure)` 

Before (both case):
image

After ("select all" case):
image

After ("entity1" case):
image

| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !25861

### DIFF
--- a/src/Session.php
+++ b/src/Session.php
@@ -413,7 +413,7 @@ class Session
                 $active
             );
             $_SESSION["glpiactive_entity_shortname"] = getTreeLeafValueName("glpi_entities", $active);
-            if ($is_recursive || $ID == "all") {
+            if ($is_recursive) {
                 //TRANS: %s is the entity name
                 $_SESSION["glpiactive_entity_name"]      = sprintf(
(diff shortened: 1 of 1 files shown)


trasher (approved):

orthagh (commented):
a case, you must check.  Selecting "All" is not the same as selecting the upper entity with subs.
A user may have several entities associated to the same profile.
In this particular case, selecting "All" let display all entities of its profile not the only the upper one

Rom1-B:
Yes, that's right (I explained it wrong), in my example "select all" should display entity1 and entity2.
The goal is to distinguish more explicitly the 2 cases.

orthagh (approved):

Issue #13609, opened 2022-12-14
Discovery printer network port


periodistacg:
### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

10.0.6-dev

### Bug description

Printer management port id changes each discovery
[printer.xml.zip](https://github.com/glpi-project/glpi/files/10227852/printer.xml.zip)

trasher:
I confirm the ports is deleted and recreated, not updated; but I have no idea if that's expected/normal

stonebuzz:
already seen / fix here -> https://github.com/glpi-project/glpi/pull/13326/files#diff-556766af2a2604805cb30ffb9fbbe579656168e6e408a93b9f7769ba495e546c (i think)

I will make a separate PR

trasher:
Seems to be fixed on my side with #13610

periodistacg:
It is ok if is not connected to a network equipment.
When you have a printer connected to a network equipment and do a discovery, the printer port change to management.
I am seeing that snmp inventory change the id of the printer port also each inventory even if connected.

stonebuzz:
Hi @periodistacg 
can you test with this new commit -> https://github.com/glpi-project/glpi/pull/13610/commits/d3f20b5c64e01c11863b8db68bb7ebc495cf98b9

stonebuzz:
Can you provide inventory file from ```snmp inventory``` step.

Best regards

periodistacg:
I think it is working fine. I will try the complete process

periodistacg:
The inventory changes the id printer port and still connected the printer
A new discovery does not disconnect the printer and still the id printer port (ok)
I put the inventory printer file
[printerinventory.zip](https://github.com/glpi-project/glpi/files/10235710/printerinventory.zip)

stonebuzz:
Greate =) 

last request, I changed my method can you revert ```d3f20b5``` and try https://github.com/glpi-project/glpi/pull/13610/commits/114257c9055098558970247373d2998f3057df73 ?

Sorry for the inconvenience, this new ```commit``` has less impact on the inventory

stonebuzz:
Thank you for your time

periodistacg:
Exactly the same (with 114257c):
The inventory changes the id printer port and keeps connected the printer
A new discovery does not disconnect the printer and keeps the id printer port (ok)

stonebuzz:
great
[...]

PR #13610, opened 2022-12-14
fix(snmp): do not delete management port each time

stonebuzz:
Do not delete management port at each inventory

Because ```handleUpdate``` unset management network port after handle it (and ```handleDeletesManagementPorts``` thinks that there is no)

https://github.com/stonebuzz/glpi/blob/88d839ade9f0fc28c0fb6f51d252dd0ad78a0b9f/src/Inventory/Asset/InventoryNetworkPort.php#L370-L380

| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |https://github.com/glpi-project/glpi/issues/13609

### DIFF
--- a/src/Inventory/Asset/Printer.php
+++ b/src/Inventory/Asset/Printer.php
@@ -36,6 +36,7 @@
 
 namespace Glpi\Inventory\Asset;
 
+use Blacklist;
 use CommonDBTM;
 use Glpi\Toolbox\Sanitizer;
 use IPAddress;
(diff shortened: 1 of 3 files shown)


trasher (approved):

src/Inventory/Asset/Printer.php
```diff
@@ -365,8 +365,8 @@ public function handleMetrics()
      */
     public static function needToBeUpdatedFromDiscovery(CommonDBTM $item, $val)
     {
-        if (property_exists($val, 'ips') && isset($val->ips[0])) {
-            $ip = $val->ips[0];
+        if (property_exists($val, 'ip')) {
```
trasher, line 368: Are you sure there is always a `$val->ip` if there is a `$val->ips[0]`? I'm really unsure

stonebuzz: I will wait for @periodistacg feedback on these tests.
If so, I already have another solution 
Filter all ```ips``` with ```Blacklist```

Because for now this method only check first entry of ```ips```

```xml
<IPS>
  <IP>0.0.0.0</IP>  <-- this is use for check if printer need to be updated
  <IP>10.100.51.207</IP>
  <IP>127.0.0.1</IP>
</IPS>
```
and ```Inventory``` understands that he must update the ```network``` part

trasher (approved):

cedric-anne (approved):

PR #13622, opened 2022-12-15
fix(notification): prevent twice notification on close

stonebuzz:
prevent twice notification on close :

Actually GLPI send two notifications on closing ```CommonITILObject```
- one for followup added
- one for closed status 

skip ```add_followup``` when the ```_close``` (added from ```prepareInputForAdd```) ```key``` is found

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | Internal ref !25923

### DIFF
--- a/src/ITILFollowup.php
+++ b/src/ITILFollowup.php
@@ -253,7 +253,9 @@ class ITILFollowup extends CommonDBChild
 
         $this->updateParentStatus($this->input['_job'], $this->input);
 
-        $donotif = !isset($this->input['_disablenotif']) && $CFG_GLPI["use_notifications"];
+        $donotif = !isset($this->input['_disablenotif']) && $CFG_GLPI["use_notifications"]
+        //skip the notification if it comes from the solution approval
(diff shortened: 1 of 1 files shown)


Rom1-B:
The follow-up notification would no longer be sent, whether the closing notification is active or not?

cedric-anne (commented):
We already had unwanted side effects while trying to handle this using existing input keys (see #13043, reverted by #13280). If we want to do something reliable here, we should set something like `$this->input['_skip_followup_notification'] = true` in `Glpi\Features\ParentStatus::updateParentStatus()` and check it in `ITILFollowup::post_addItem()`.

Also, the problem is that two distinct notification are sent.
1. They may have distinct targets. So, in edge cases, a user may be a target of `add_followup` notification but not a target of `closed` notification. So, in this case, if we remove the  `add_followup` notification, user will receive no notification.
2. The `closed` notification template may contains only summary information and not contains followups. In this case, administrator will want to have both notification sent.

PR #13634, opened 2022-12-16
fix(inventory): prevent double 'Dropdown::import()'

stonebuzz:
Prevent double ```Dropdown::importExternal()```

Which generates ```Manufacturer``` and ```SoftwareCategory``` with numbers

The first from ```src/Inventry/Asset/Software.php``` (which returns an ID as value)

https://github.com/glpi-project/glpi/blob/56f60f6e62bc9094551a0c469cd104ceeb05011c/src/Inventory/Asset/Software.php#L155-L163

The second from  ```src/Inventry/Asset/Software.php``` (which import ID previously generated)
https://github.com/glpi-project/glpi/blob/56f60f6e62bc9094551a0c469cd104ceeb05011c/src/Inventory/Asset/InventoryAsset.php#L253

N.B:
Sometime a number is added as a ```Manufacturer```
In my case I have memories with a number in the ```MANUFACTURER``` XML node 
this case is therefore legitimate

```xml
    <MEMORIES>
      <CAPACITY>8192</CAPACITY>
      <CAPTION>ChannelA-DIMM0</CAPTION>
      <DESCRIPTION>DIMM</DESCRIPTION>
      <MANUFACTURER>1315</MANUFACTURER>
      <MODEL>BLT8G3D1608DT1TX0.</MODEL>
      <NUMSLOTS>1</NUMSLOTS>
      <SERIALNUMBER>A0154896</SERIALNUMBER>
      <SPEED>1600</SPEED>
      <TYPE>DDR3</TYPE>
    </MEMORIES>
```

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #12933 #13459

Issue #13459, opened 2022-11-30
Generating  Random numbers in Manufactures drop down menu


blackpearl-2022:
### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

10.0.5

### Bug description

I am having an issue with generating random numbers in the manufacturers drop-down menu and also in the software categories. After deleting these, it will again create automatically after some time. This occurs the same in the software category also
I am using the latest build of 10.0.5 and glpi agent 1.4 with ubuntu server 22.04 lts

![1](https://user-images.githubusercontent.com/118801689/204790659-83c75f35-a83b-4f88-8fa7-79c6c30f4eea.JPG)
![2](https://user-images.githubusercontent.com/118801689/204790685-9267fc6b-8fe2-4051-8522-b9d797377bd8.JPG)

### Your GLPI setup information

GLPI 10.0.5 ( => /var/www/html/glpi)
Installation mode: TARBALL
Current language:en_US

PR #13639, opened 2022-12-16
fix(inventory): prevent double 'Dropdown::import()' 

stonebuzz:
Alternative to https://github.com/glpi-project/glpi/pull/13634

Prevent double ```Dropdown::importExternal()```

Which generates ```Manufacturer``` and ```SoftwareCategory``` with numbers

The first from ```src/Inventry/Asset/Software.php``` (which returns an ID as value)

https://github.com/glpi-project/glpi/blob/56f60f6e62bc9094551a0c469cd104ceeb05011c/src/Inventory/Asset/Software.php#L155-L163

The second from  ```src/Inventry/Asset/Software.php``` (which import ID previously generated)
https://github.com/glpi-project/glpi/blob/56f60f6e62bc9094551a0c469cd104ceeb05011c/src/Inventory/Asset/InventoryAsset.php#L253

Or 
https://github.com/glpi-project/glpi/blob/56f60f6e62bc9094551a0c469cd104ceeb05011c/src/Inventory/Asset/InventoryAsset.php#L257

N.B:
Sometime a number is added as a ```Manufacturer```
In my case I have memories with a number in the ```MANUFACTURER``` XML node 
this case is therefore legitimate

```xml
    <MEMORIES>
      <CAPACITY>8192</CAPACITY>
      <CAPTION>ChannelA-DIMM0</CAPTION>
      <DESCRIPTION>DIMM</DESCRIPTION>
      <MANUFACTURER>1315</MANUFACTURER>
      <MODEL>BLT8G3D1608DT1TX0.</MODEL>
      <NUMSLOTS>1</NUMSLOTS>
      <SERIALNUMBER>A0154896</SERIALNUMBER>
      <SPEED>1600</SPEED>
      <TYPE>DDR3</TYPE>
    </MEMORIES>
```

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #12933 #13459

### DIFF
--- a/src/Inventory/Asset/InventoryAsset.php
+++ b/src/Inventory/Asset/InventoryAsset.php
@@ -252,6 +252,7 @@ abstract class InventoryAsset
                         );
                     } else if (isset($foreignkey_itemtype[$key])) {
                         //if SoftwareCategory Or Manufacturer and integer do not process importExternal
+                        //dropdown already handle from Inventory/asset/Software.php
                         if (
                             $foreignkey_itemtype[$key] == SoftwareCategory::class
                             || ($foreignkey_itemtype[$key] == Manufacturer::class && is_int($value->$key))
@@ -263,6 +264,7 @@ abstract class InventoryAsset
                     } else if ($key !== 'entities_id' && $key !== 'states_id' && isForeignKeyField($key) && is_a($itemtype = getItemtypeForForeignKeyField($key), CommonDropdown::class, true)) {
                         $foreignkey_itemtype[$key] = $itemtype;
                         //if SoftwareCategory Or Manufacturer and integer do not process importExternal
+                        //dropdown already handle from Inventory/asset/Software.php
                         if (
                             $foreignkey_itemtype[$key] == SoftwareCategory::class
                             || ($foreignkey_itemtype[$key] == Manufacturer::class && is_int($value->$key))
--- a/src/Inventory/Asset/InventoryAsset.php
+++ b/src/Inventory/Asset/InventoryAsset.php
@@ -251,16 +251,22 @@ abstract class InventoryAsset
                             ['manufacturer' => $manufacturer_name]
                         );
                     } else if (isset($foreignkey_itemtype[$key])) {
-                        //if integer, drodpown already import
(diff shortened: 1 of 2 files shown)


cedric-anne (approved):
I am OK with this workaround, but for me, the most reliable solution would be to:

1. Store raw values in different keys (e.g. `$val->manufacturer` that would contain the raw value, and `$val->manufacturers_id` that would contains the ID, or null if import has not be done yet).
2. In `handleLinks`, parse all keys that validates `isForeignKeyField($key)`and, if value is null, call `Dropdown::importExternal()` based on raw key value.

With this kind of logic, it would not be possible anymore to mix raw values and IDs in the same property key.

I guess it has many impact on code, so I do not know if it is possible to do such a change in 10.0/bugfixes branch.

cedric-anne:
Maybe it also fixes https://github.com/glpi-project/glpi/issues/12202.

stonebuzz:
Work with

rule dictionnaries 

ex : assign another manufacturer
image

ex : assign another software category (
image

cedric-anne (dismissed):

trasher (dismissed):

trasher:
I'm completely lost in PRs, issues and comments... Please add tests (or tell me how to be sure to reproduce all cases and I'll add tests myself).

trasher:
As said on #13642; relying on `is_int()` sounds a very bad solution. 

Handling properly relations is kind of an headache because it's been modified since the original implementation, and not the better way (mostly my bad, but it's too late/complicated to change that now).

In inventory files logic, `*_id` must be the raw original value, and not any corresponding ID. Corresponding IDs are stored in `known_links` property array. This has been changed because it was almost impossible to properly deal with a mix of those values to get older bugs fixed.

I'm working on an alternative PR using "correct" logic.

trasher:
> I'm working on an alternative PR using "correct" logic.

#13669

stonebuzz:
I close in favor of https://github.com/glpi-project/glpi/pull/13669

PR #13641, opened 2022-12-16
Fix anonymous file upload

AdrienClairembault:
The changes introduced in https://github.com/glpi-project/glpi/commit/edb815973a4359c05d9ad1ae85706b79b128487e broke file upload for formcreator on public forms.

These changes revert this behavior for the `fileupload.php` endpoint, which mean we only check the `valid_id` session's data, which is generated correctly by formcreator in the context of public forms.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |

### DIFF
--- a/front/document.send.php
+++ b/front/document.send.php
@@ -38,7 +38,8 @@ use Glpi\Inventory\Conf;
 include('../inc/includes.php');
 
 if (!$CFG_GLPI["use_public_faq"]) {
-    Session::checkLoginUser();
+    // Might be displaying image for anonymous users on formcreator
+    Session::checkValidSessionId(true);
 }
 
 $doc = new Document();
--- a/ajax/fileupload.php
+++ b/ajax/fileupload.php
@@ -41,7 +41,7 @@ use Glpi\Application\ErrorHandler;
 
 include('../inc/includes.php');
 
-Session::checkLoginUser();
+Session::checkValidSessionId(true);
 
 // Ensure warnings will not break ajax output.
 ErrorHandler::getInstance()->disableOutput();
--- a/ajax/getFileTag.php
+++ b/ajax/getFileTag.php
@@ -44,7 +44,7 @@ include('../inc/includes.php');
 header('Content-type: application/json');
 Html::header_nocache();
 
-Session::checkLoginUser();
+Session::checkValidSessionId(true);
 
 if (isset($_POST['data'])) {
     $response = [];
--- a/src/Session.php
+++ b/src/Session.php
@@ -871,9 +871,12 @@ class Session
      *
      * @see https://wiki.php.net/rfc/strict_sessions
      *
+     * @param bool $allow_anonymous If enabled, we will only check if
+     *                              $_SESSION['valid_id'] is set correctly
+     *
      * @return void|true
      **/
-    public static function checkValidSessionId()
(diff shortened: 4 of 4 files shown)


btry (approved):

AdrienClairembault:
Last commit is needed if the form contain a "description" fields that contains images.
Without it the image wont load and the session will be destroyed (thus preventing the form to be submitted as the session CSRF data will be wiped).

image

front/document.send.php
```diff
@@ -38,7 +38,8 @@
 include('../inc/includes.php');
 
 if (!$CFG_GLPI["use_public_faq"]) {
-    Session::checkLoginUser();
+    // Might be displaying image for anonymous users on formcreator
+    Session::checkValidSessionId(true);
```
cedric-anne, line 42: I probably miss something here.

`front/document.send.php` endpoint can be called using 2 distinct params.
 - Using the `file` param, only dumps, user pictures and inventory files can be served. Are you using this case ?
 - Using the `docid` param, `Document::canViewFile()` method will be used to validate that user has right to see the file. If there is no rights in session, will this method be able to return `true` ?

cedric-anne (changes requested):

front/document.send.php
AdrienClairembault: Full URL: `document.send.php?docid=11&itemtype=PluginFormcreatorQuestion&items_id=7`

I guess it falls under this condition:

image

Which leads to `PluginFormcreatorForm ::canView()`, which is always true:

image

cedric-anne (approved):

orthagh (changes requested):
I'm not in favor of this pr atm.
Don't merge it, this requires discussions

AdrienClairembault:
I'm thinking about other potentials solutions.

The cleanest way to do it would be through specific endpoints created by Formcreator that explicitly allow anonymous users.
However it isn't very practical as it would be hard to maintain (at least 4 endpoints would need to be maintained for things that should realistically be handled by GLPI core, not a plugin).

Another option would be to allow plugins to add a specific session flag to bypass the new check.
For example, setting `$_SESSION['do_not_check_profiles] = true;`
It feels dirty but it would be: 
* Extremely simple
* Let each plugins implement it anywhere they need it (no need to update core code once it's done)
* Do not worsen GLPI security (core)
* If done rights in plugins, it should not worsen their security either as it should be done only on pages that would be accessible without an account (thus not checking for a profile make sense and should not be exploitable)

We could even go on step further and name the parameter `$_SESSION['do_not_check_session]`, which mean Formcreator wouldn't even need to simulate a session anymore by setting some values like `glpiid`.

trasher:
Will not be merged, as discussed

trasher:
This is a plugin issue, and should be fixed in plugin. GLPI Core does not allow anonymous uploads anywhere.

This may be discussed, yes, but this is not the right place :)

trasher:
(we've talked about that in our weekly meeting)

cedric-anne:
It seems that people are tempted to apply this patch to fix `formcreator` anonymous form attachment functionality.
Anonymous uploads cannot be safe in GLPI context, so any patch that would fix this feature is a security flaw by nature.

PR #13642, opened 2022-12-16
fix(inventory): prevent 'Dropdown::import()' for values defined by rules (already an ID)

stonebuzz:
Prevent adding rule value as new ```locations_id``` (because ```RuleImportEntityCollection``` returns an ID for ```ForeignKey```)

Same behavior for  ```users_id_tech``` and  ```groups_id_tech```

Force ```int``` type when inventory manage data returned by ```RuleImportEntityCollection```

```int``` type is usefull to check if value is a real ID

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | https://github.com/glpi-project/glpi-inventory-plugin/issues/276

### DIFF
--- a/src/Inventory/Asset/InventoryAsset.php
+++ b/src/Inventory/Asset/InventoryAsset.php
@@ -239,7 +239,11 @@ abstract class InventoryAsset
                 if (!isset($this->known_links[$known_key])) {
                     $entities_id = $this->entities_id;
                     if ($key == "locations_id") {
-                        $this->known_links[$known_key] = Dropdown::importExternal('Location', $value->$key, $entities_id);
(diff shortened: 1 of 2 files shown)


trasher:
#13668 seems a better solution

trasher:
> #13668 seems a better solution

Mainly because relying on `is_int()` sounds a very bad solution, and also because new fix change less code :p

stonebuzz:
I close in favor of  https://github.com/glpi-project/glpi/pull/13668

PR #13652, opened 2022-12-19
fix(ticket): cloning, unexpected automatic task

Rom1-B:
When cloning a ticket with tasks with durations, the created ticket had an automatic completed task, like this:

image

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !25928

### DIFF
--- a/src/CommonITILObject.php
+++ b/src/CommonITILObject.php
@@ -9284,4 +9284,10 @@ abstract class CommonITILObject extends CommonDBTM
 
         return $input;
     }
+
+    public function prepareInputForClone($input)
+    {
+        unset($input['actiontime']);
+        return $input;
+    }
 }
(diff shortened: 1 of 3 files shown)


trasher (approved):

trasher:
Please add a test

Rom1-B:
I do not see the cases where this auto task should appear?

trasher:
Test should ensure there is nothing more added than expected

trasher (approved):

cedric-anne (changes requested):
Same problem will present on `Problem` and `Change` classes.

Also, I did not tested it, but maybe the following code would be more relevant:
```php
    public function prepareInputForClone($input)
    {
        unset($input['actiontime']);
        return $input;
    }
```

cedric-anne (approved):

trasher (approved):

PR #13678, opened 2022-12-28
Allow to add pending reasons when item is already pending

AdrienClairembault:
It is currently impossible to add a pending reason to a pending ticket:

image

After changes:

image

Note that if the ticket is pending AND already have a pending reason, the input will still not be shown to avoid overriding the active pending reason.

image

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !26069

### DIFF
--- a/src/CommonITILObject.php
+++ b/src/CommonITILObject.php
@@ -445,6 +445,7 @@ abstract class CommonITILObject extends CommonDBTM
             'canupdate'               => $canupdate,
             'canpriority'             => $canupdate,
             'canassign'               => $canupdate,
+            'has_pending_reason'      => PendingReason_Item::getForItem($this),
         ]);
 
         return true;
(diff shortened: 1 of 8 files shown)


Rom1-B:
For the 3rd screenshot, after adding a manual pending follow up, do auto follow ups work?

AdrienClairembault:
> For the 3rd screenshot, after adding a manual pending follow up, do auto follow ups work?

Do you suspect any issue with it ? It does not impact auto follow as far as I know.

Rom1-B:
Yes, currently, when a ticket has an active pending reason and a follow-up is added (pending or not), the pending reason stops, there are no more auto generated follow-ups.
If the new follow-up is pending, either the waiting reason continues to work or it is stopped, but in this case, the dropdown of the reasons should be displayed to be able to activate a new one.

AdrienClairembault:
Understood, I'll do some testing and report back ;)

cedric-anne (changes requested):

AdrienClairembault:
> Yes, currently, when a ticket has an active pending reason and a follow-up is added (pending or not), the pending reason stops, there are no more auto generated follow-ups. If the new follow-up is pending, either the waiting reason continues to work or it is stopped, but in this case, the dropdown of the reasons should be displayed to be able to activate a new one.

Fixed in `1d4e8cd66d`.

We should probably add tests for this part, will look into it.

cedric-anne (approved):

Rom1-B (approved):

trasher (approved):

PR #13736, opened 2023-01-04
Changes: Fix notifications setting for assigned techs

AdrienClairembault:
It seems that when adding a tech to a change in GLPI 9.5 notifications are always disabled by default:

image

It does work correctly for tickets so I suppose some magic code add the correct value somewhere but has been forgotten for changes.

It seems to be me the easiest way to fix it is to always display the notification dropdown, like for requester and observers:

image

This way the value is sent with the form and the notification usage is enabled/disabled accordingly.

In GLPI 10 this part was rewritten due to the new UI and seems to work fine.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !26009

### DIFF
--- a/ajax/dropdownItilActors.php
+++ b/ajax/dropdownItilActors.php
@@ -106,13 +106,14 @@ if (isset($_POST["type"])
 
             if ($CFG_GLPI["notifications_mailing"]) {
                echo "<br><span id='notif_user_$rand'>";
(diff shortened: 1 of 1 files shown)


trasher:
9.5 will receive security fixes only

PR #13740, opened 2023-01-04
fix(queuednotification): delete, restore, purge and event logs

flegastelois:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | N/A

"Put in trashbin", "Delete permanently", "Restore" buttons in the QueuedNotification form have no effect (displays an empty form), this PR fixes that.

We also add log Events.

### DIFF
--- a/front/queuednotification.form.php
+++ b/front/queuednotification.form.php
@@ -37,6 +37,8 @@
  * @since 0.85
  */
 
+use Glpi\Event;
+
 include('../inc/includes.php');
 
 Session::checkRight('queuednotification', READ);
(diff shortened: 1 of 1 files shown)


cedric-anne (approved):

trasher (approved):

flegastelois:
Small point of attention: the Events display is OK without adding specific code in Event::logArray, but not for the line:

image

Which is however completely identical in DB:
image

orthagh (approved):

PR #13755, opened 2023-01-06
feat(rules): display criterias/actions on list

flegastelois:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | N/A

When you have a large number of rules, it sometimes takes a long time to have to go through them to find in which rule a particular criterion/action has been added.

This PR therefore makes it possible to display them in the list, thus making it possible to do a search (ctrl+f).

I don't know which branch is better (bug fixes or main).

Before

image

After

image

Perhaps an improvement: rather than having the display forced into the code, we could imagine two "toggle" buttons next to the "Actions" button (with persistent in-session choice).

image

### DIFF
--- a/front/rule.common.php
+++ b/front/rule.common.php
@@ -148,7 +148,7 @@ Html::header(
 );
 
 $rulecollection->display([
-    'criterias' => 1,
-    'actions'   => 1,
+    'display_criterias' => true,
+    'display_actions'   => true,
 ]);
 Html::footer();
--- a/front/rule.common.php
+++ b/front/rule.common.php
@@ -147,5 +147,8 @@ Html::header(
     $rulecollection->menu_option
 );
 
-$rulecollection->display();
+$rulecollection->display([
+    'criterias' => 1,
+    'actions'   => 1,
+]);
 Html::footer();
--- a/src/Rule.php
+++ b/src/Rule.php
@@ -2029,7 +2029,7 @@ class Rule extends CommonDBTM
             echo "<td>";
             foreach ($RuleCriterias->getRuleCriterias($this->fields['id']) as $RuleCriteria) {
                 $to_display = $this->getMinimalCriteria($RuleCriteria->fields);
-                echo implode('<i class="fas fa-caret-right mx-1"></i>', $to_display) . '<br />';
+                echo "<span class='glpi-badge mb-1'>" . implode('<i class="fas fa-caret-right mx-1"></i>', $to_display) . '</span><br />';
             }
             echo "</td>";
         }
@@ -2040,7 +2040,7 @@ class Rule extends CommonDBTM
             echo "<td>";
             foreach ($RuleAction->getRuleActions($this->fields['id']) as $RuleAction) {
                 $to_display = $this->getMinimalAction($RuleAction->fields);
-                echo implode('<i class="fas fa-caret-right mx-1"></i>', $to_display) . '<br />';
+                echo "<span class='glpi-badge mb-1'>" . implode('<i class="fas fa-caret-right mx-1"></i>', $to_display) . '</span><br />';
             }
             echo "</td>";
         }
--- a/src/Rule.php
+++ b/src/Rule.php
@@ -1999,8 +1999,8 @@ class Rule extends CommonDBTM
         // bool $display_criterias = false,
         // bool $display_actions = false
     ) {
-    $display_criterias = func_get_args()[5] ?? false;
-    $display_actions = func_get_args()[6] ?? false;
+        $display_criterias = func_get_args()[5] ?? false;
+        $display_actions = func_get_args()[6] ?? false;
         $canedit = (self::canUpdate() && !$display_entities);
         echo "<tr class='tab_bg_1' data-rule-id='" . $this->fields['id'] . "'>";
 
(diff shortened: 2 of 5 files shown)


trasher:
As far as I can see, defazulkt behavior is still the same, therefore it's OK for me to target 10.0/bf branch

orthagh:
tweaked the display a little to have badge design
image

src/Rule.php
```diff
@@ -1986,9 +1986,18 @@ public function cleanDBonPurge()
...
+     **/
+    public function showMinimalForm(
+        $target,
+        $first = false,
+        $last = false,
+        $display_entities = false,
+        $active_condition = 0,
+        $display_criterias = 0,
+        $display_actions = 0
+    ) {
```
cedric-anne, line 2000: Adding new parameter on a public method, even optional, is a bad idea on a bugfix version.

A change on this method will trigger a fatal error on `statecheck` plugin `Fatal error: Declaration of PluginStatecheckRule::showMinimalForm($target, $first = false, $last = false, $display_entities = false, $active_condition = 0) must be compatible with Rule::showMinimalForm($target, $first = false, $last = false, $display_entities = false, $active_condition = 0, $display_criterias = 0, $display_actions = 0)`.

A solution to handle this is to do something like this:
```suggestion
        $active_condition = 0
        // FIXME Uncomment this in GLPI 10.1
        // bool $display_criterias = false,
        // bool $display_actions = false
    ) {
    $display_criterias = func_get_args()[5] ?? false;
    $display_actions = func_get_args()[6] ?? false;
```

front/rule.common.php
```diff
@@ -147,5 +147,8 @@
     $rulecollection->menu_option
 );
 
-$rulecollection->display();
+$rulecollection->display([
+    'criterias' => 1,
+    'actions'   => 1,
```
cedric-anne, line 152: 1. Boolean values seems more appropriate.
2. Names could be more meaningful.

```suggestion
    'display_criterias' => true,
    'display_actions'   => true,
```

src/Rule.php
```diff
@@ -2271,6 +2302,18 @@ public function showMinimalCriteriaForm($fields, $canedit, $rand)
...
+        $text  = "<td $addtotd>" . $to_display['criterion'] . "</td>";
+        $text .= "<td $addtotd>" . $to_display['condition'] . "</td>";
+        $text .= "<td $addtotd>" . $to_display['pattern'] . "</td>";
+        return $text;
+    }
+
+    /**
+     * @param $fields
+     **/
+    public function getMinimalCriteria($fields)
```
cedric-anne, line 2316: 1. This method seems to be used only locally, it should be declared as private.
2. Types hinting should be added.

```suggestion
    private function getMinimalCriteria(array $fields): array
```

src/Rule.php
```diff
@@ -2280,18 +2323,30 @@ public function getMinimalCriteriaText($fields, $addtotd = '')
...
+        $text  = "<td $addtotd>" . $to_display['field'] . "</td>";
+        $text .= "<td $addtotd>" . $to_display['type'] . "</td>";
+        $text .= "<td $addtotd>" . $to_display['value'] . "</td>";
+        return $text;
+    }
+
+    /**
+     * @param $fields
+     **/
+    public function getMinimalAction($fields)
```
cedric-anne, line 2349:
```suggestion
    private function getMinimalAction(array $fields): array
```

src/RuleCollection.php
```diff
@@ -467,6 +467,8 @@ public function showListRules($target, $options = [])
         $p['condition'] = 0;
         $p['_glpi_tab'] = $options['_glpi_tab'];
         $rand           = mt_rand();
+        $p['criterias'] = 0;
+        $p['actions']   = 0;
```
cedric-anne, line 471:
```suggestion
        $p['display_criterias'] = false;
        $p['display_actions']   = false;
```

src/RuleCollection.php
```diff
@@ -477,9 +479,19 @@ public function showListRules($target, $options = [])
...
+            ) {
+                $p[$param] = $options[$param];
+            }
+        }
+
+        $rule              = $this->getRuleClass();
+        $display_entities  = ($this->isRuleRecursive()
+                            && ($p['inherited'] || $p['childrens']));
+        $display_criterias = $p['criterias'];
+        $display_actions   = $p['actions'];
```
cedric-anne, line 494:
```suggestion
        $display_criterias = $p['display_criterias'];
        $display_actions   = $p['display_actions'];
```

cedric-anne (changes requested):

front/rule.common.php
trasher: BTW, `criteria` is plural for `criterion`, please, no extra `s` ;p

src/Rule.php
```diff
@@ -1986,9 +1986,21 @@ public function cleanDBonPurge()
...
+        $first = false,
+        $last = false,
+        $display_entities = false,
+        $active_condition = 0
+        // FIXME Uncomment this in GLPI 10.1
+        // bool $display_criterias = false,
+        // bool $display_actions = false
+    ) {
+    $display_criterias = func_get_args()[5] ?? false;
+    $display_actions = func_get_args()[6] ?? false;
```
cedric-anne, line 2003:
```suggestion
        $display_criterias = func_get_args()[5] ?? false;
        $display_actions = func_get_args()[6] ?? false;
```

cedric-anne (approved):

trasher (approved):

PR #13832, opened 2023-01-18
Show only assets with specific states in assistance

AdrienClairembault:
Add a new setting for assets' states:

image

It's enabled by default and allow to filter assets that are shown in GLPI's assistance depending on their states:

image

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !26043

### DIFF
--- a/install/migrations/update_10.0.x_to_10.1.0/state_show_in_assistance.php
+++ b/install/migrations/update_10.0.x_to_10.1.0/state_show_in_assistance.php
@@ -1,0 +1,41 @@
+<?php
+
+/**
+ * ---------------------------------------------------------------------
+ *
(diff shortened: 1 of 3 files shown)


cedric-anne (commented):
It seems OK, but you should add a search option for this new field.

orthagh (commented):
The features exists somehow for some itemtypes (direct filter instead of filter of their state).
Software, Appliance, DatabaseInstance, SoftwareLicense itemtype have a `is_helpdesk_visible` field.
We may have the same name for this new filtering.

orthagh (approved):

cedric-anne (approved):

PR #13849, opened 2023-01-20
ProjectTemplate: show/edit name after creation

AdrienClairembault:
Project template name is visible on creation in a special header:

image

But after creation this header is no longer displayed (since global changes made in GLPI 10.0.4) and thus we can't edit it anymore :

image

I've added the field at the top to fix this:

image

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !26272

### DIFF
--- a/src/Project.php
+++ b/src/Project.php
@@ -1559,6 +1559,19 @@ class Project extends CommonDBTM implements ExtraVisibilityCriteria
                 echo "<td colspan='2'>&nbsp;</td>";
             }
             echo "</tr>";
+        } elseif ($is_template & !$this->isNewItem()) {
+            // Show template name after creation (creation is already handled by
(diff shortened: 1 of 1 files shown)


orthagh (approved):

cedric-anne (approved):

trasher:
Is that limited to Projects only? How are handled all other objects that can be based on a template?

trasher:
> Is that limited to Projects only? How are handled all other objects that can be based on a template?

Nevermind, I did not pay attention that concerns a not yet migrated to twig object.

trasher (approved):

PR #13863, opened 2023-01-24
fix(Socket): Socket can only be linked to one Cable

stonebuzz:
```Socket``` can only be linked to one ```Cable```

So we need to exclude all ```Socket``` already linked from ```dropdown```

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #13851

### DIFF
--- a/ajax/cable.php
+++ b/ajax/cable.php
@@ -66,7 +66,7 @@ switch ($action) {
                     'itemtype'           => $_GET['itemtype'],
                     'items_id'           => $_GET['items_id']
                 ],
-                'used'         => Socket::getSocketAlreadyLinked($_GET['itemtype'], $_GET['items_id']),
+                'used'         => (int)$_GET['items_id'] > 0 ? Socket::getSocketAlreadyLinked($_GET['itemtype'], (int)$_GET['items_id']),
                 'displaywith'  => ['itemtype', 'items_id', 'networkports_id'],
             ]);
         }
--- a/ajax/cable.php
+++ b/ajax/cable.php
@@ -61,7 +61,7 @@ switch ($action) {
             (isset($_GET['itemtype']) && class_exists($_GET['itemtype']))
             && isset($_GET['items_id'])
         ) {
-            Socket::dropdown(['name'         =>  $_GET['dom_name'], 
+            Socket::dropdown(['name'         =>  $_GET['dom_name'],
                 'condition'    => ['socketmodels_id'   => $_GET['socketmodels_id'] ?? 0,
                     'itemtype'           => $_GET['itemtype'],
                     'items_id'           => $_GET['items_id']
--- a/ajax/cable.php
+++ b/ajax/cable.php
@@ -61,12 +61,12 @@ switch ($action) {
             (isset($_GET['itemtype']) && class_exists($_GET['itemtype']))
             && isset($_GET['items_id'])
         ) {
-            Socket::dropdown(['name'         =>  $_GET['dom_name'],
+            Socket::dropdown(['name'         =>  $_GET['dom_name'], 
                 'condition'    => ['socketmodels_id'   => $_GET['socketmodels_id'] ?? 0,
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: templates/pages/assets/cable.html.twig [hunk 1]
```diff
@@ -132,7 +132,8 @@
                               'socketmodels_id': item.fields['socketmodels_id_endpoint_' ~ side],
                               'itemtype': item.fields['itemtype_endpoint_' ~ side],
                               'items_id': item.fields['items_id_endpoint_' ~ side]
-                           }
+                           },
+                           'used': call('Glpi\\Socket::getSocketAlreadyLinked', [ item.fields['itemtype_endpoint_' ~ side], item.fields['items_id_endpoint_' ~ side]])
```
cedric-anne, line 136:
```suggestion
                           'used': call('Glpi\\Socket::getSocketAlreadyLinked', [item.fields['itemtype_endpoint_' ~ side], item.fields['items_id_endpoint_' ~ side]])
```

cedric-anne (approved):

ajax/cable.php
```diff
@@ -66,6 +66,7 @@
                     'itemtype'           => $_GET['itemtype'],
                     'items_id'           => $_GET['items_id']
                 ],
+                'used'         => Socket::getSocketAlreadyLinked($_GET['itemtype'], $_GET['items_id']),
```
cedric-anne, line 69:
```suggestion
                'used'         => (int)$_GET['items_id'] > 0 ? Socket::getSocketAlreadyLinked($_GET['itemtype'], (int)$_GET['items_id']),
```

templates/pages/assets/cable.html.twig
```diff
@@ -132,7 +132,8 @@
                               'socketmodels_id': item.fields['socketmodels_id_endpoint_' ~ side],
                               'itemtype': item.fields['itemtype_endpoint_' ~ side],
                               'items_id': item.fields['items_id_endpoint_' ~ side]
-                           }
+                           },
+                           'used': call('Glpi\\Socket::getSocketAlreadyLinked', [item.fields['itemtype_endpoint_' ~ side], item.fields['items_id_endpoint_' ~ side]])
```
cedric-anne, line 136: Maybe it will fix the issue reported. To be tested.
```suggestion
                           'used': item.fields['items_id_endpoint_' ~ side] is not empty ?call('Glpi\\Socket::getSocketAlreadyLinked', [item.fields['itemtype_endpoint_' ~ side], item.fields['items_id_endpoint_' ~ side]] : [])
```

cedric-anne (changes requested):

ajax/cable.php
```diff
@@ -66,6 +66,7 @@
                     'itemtype'           => $_GET['itemtype'],
                     'items_id'           => $_GET['items_id']
                 ],
+                'used'         => (int)$_GET['items_id'] > 0 ?? Socket::getSocketAlreadyLinked($_GET['itemtype'], (int)$_GET['items_id']),
```
cedric-anne, line 69:
```suggestion
                'used'         => (int)$_GET['items_id'] > 0 ? Socket::getSocketAlreadyLinked($_GET['itemtype'], (int)$_GET['items_id']) : [],
```

cedric-anne (approved):

cedric-anne:
I added a commit to ensure that current value is still available. Without this commit, user was not able to select back previously used value if he changed it by mistake:
image

orthagh (approved):

trasher (approved):

RicardoBarboza:
In many cases, such as in industrial automation, we have situations where several cables are connected to a single socket. Therefore, it would be very useful to allow this situation to occur and be handled by GLPI.

Regards,

Ricardo Barboza

PR #13869, opened 2023-01-24
feat(networkdevice): keep IP from inventory

stonebuzz:
Actually [converter](https://github.com/glpi-project/inventory_format/) merge IP and IPS in same node

From GLPI side, we cannot distinguish between the management IP and other IPS

image

And ```GLPIInentory``` use bad IP to do ```NetInventory```

This PR try to handle IP and IPS

Need https://github.com/glpi-project/inventory_format/pull/89

TU have been prepared for this

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !26397


src/Inventory/Asset/InventoryNetworkPort.php
```diff
@@ -476,8 +476,11 @@ private function handleUpdates()
                             if ($ip_data == $db_ip_data) {
                                 unset($ips[$ip_key]);
                                 unset($db_addresses[$db_ip_key]);
-                             //result found in db, useless to continue
-                                break 1;
+                                //result found in db, useless to continue
```
g-bougard, line 479: Should we keep the comment if you remove the `break` line ?

stonebuzz: done

cedric-anne (commented):
This seems ok, but, honestly, I am a bit lost about how we should handle ips/network stuffs.

trasher:
> This seems ok, but, honestly, I am a bit lost about how we should handle ips/network stuffs.

On my side... I'm totally lost, from a while :D

cedric-anne (changes requested):
We discussed about it IRL.

We probably should handle the case where a network equipement would answer to SNMP ping on many addresses. In this case, each address should probably be considered as a management port.

g-bougard:
Indeed the management port in GLPI will only set an ip to be used by netinventory task from glpiinventory. IMHO, if netdiscovery reaches one device from different ip ranges, only one ip there is required. So if one appears during a netdiscovery task and is overwritten by the same task some time later with another ip for another range, we should just assume the last ip is sufficient to inventory the device in netinventory. We don't need to inventory 2 times de same device from 2 different ips.

Maybe "Management port" could be renamed or annotated to reflect it is used in GLPI for SNMP inventory and is not what network administrators can think a network management port is for their all-day job.

stonebuzz:
Closed in favor of : https://github.com/glpi-project/glpi/pull/13935

PR #13871, opened 2023-01-25
feat(plugins): event log, add user

flegastelois:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | N/A

### DIFF
--- a/src/Plugin.php
+++ b/src/Plugin.php
@@ -828,7 +828,11 @@ class Plugin extends CommonDBTM
                 Plugin::class,
                 3,
                 "setup",
-                $message
+                sprintf(
+                    __('Plugin %1$s has been uninstalled by %2$s.'),
+                    $this->fields['name'],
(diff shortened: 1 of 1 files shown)


src/Plugin.php
```diff
@@ -1018,7 +1024,11 @@ public function activate($ID)
                     Plugin::class,
                     3,
                     "setup",
-                    $log_message
+                    sprintf(
+                        __('Plugin %1$s has been activated by %2$s.'),
+                        $this->fields['name'],
+                        User::getNameForLog(Session::getLoginUserID(true))
```
cedric-anne, line 1030: This will probably fail with an error if you try to activate the plugin from console command.

flegastelois: Seems ok, without any error.

image

image

trasher (approved):

PR #13883, opened 2023-01-26
improving the `password_update` right label

Rom1-B:
The right label was not specific enough, because the right only allows you to change your own password and not those of other users.

image

| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !25947

### DIFF
--- a/src/Profile.php
+++ b/src/Profile.php
@@ -2989,7 +2989,7 @@ class Profile extends CommonDBTM
             'id'                 => '4',
             'table'              => 'glpi_profilerights',
             'field'              => 'rights',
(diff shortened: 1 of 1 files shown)


cedric-anne (changes requested):
You should change the search option label too.

cedric-anne (approved):

orthagh (approved):

trasher (approved):

PR #13916, opened 2023-01-30
improve kanban actors stack

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  |/no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

While browsing kanban for answering an incoming change, i though stack of actors in kanban can be slightly improved:

**From**:
image

**To**:
image

PS: i didn't had the time to check, but global kanban in ticket doesn't save columns anymore on my instance. @cconard96 Do you have an idea ?

### DIFF
--- a/ajax/getUserPicture.php
+++ b/ajax/getUserPicture.php
@@ -72,10 +72,11 @@ foreach ($_GET['users_id'] as $user_id) {
                 $path = User::getURLForPicture($user->fields['picture']);
             }
             $img = Html::image($path, [
-                'title'  => getUserName($user_id),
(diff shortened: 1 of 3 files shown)


trasher (approved):

cedric-anne (approved):

cconard96:
> PS: i didn't had the time to check, but global kanban in ticket doesn't save columns anymore on my instance. @cconard96 Do you have an idea ?

#13918 should fix this

cconard96 (approved):

PR #13918, opened 2023-01-30
fix kanban load with linked tickets

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -


orthagh:
It doesn't seem to fix my issue reported in #13916.

I see the states seem to be correctly loaded by ajax:
image

But the display stay empty on reload

cedric-anne:
@cconard96 Could you explain what is the issue supposed to be fixed here? I do not know how to test the fix.

cconard96:
When tickets have other tickets linked as child tickets, they show up as steps inside the Kanban card based on status. For example, if there are 3 child tickets and 1 is solved, it shows 1 / 3 steps completed.

The feature implementation was incomplete though, or had been accidentally removed, and the status of those linked tickets wasn't filled, so when it tried getting that data for child tickets, it broke the loading of the Kanban.

cedric-anne:
@cconard96 I cannot validate this fix as the `/ajax/kanban.php?action=get_column&itemtype=Ticket&items_id=0&column_field=status&column_id=1` request timeouts after 60 seconds.

cconard96:
Do you have an enormous number of new tickets?

cedric-anne:
> Do you have an enormous number of new tickets?

image

cedric-anne (commented):
I still have a timeout after 60 seconds on Kanban view with this patch and #14333, I suspect an infinite loop somewhere in PHP code.

cedric-anne (approved):

PR #13920, opened 2023-01-30
Printers: fix "Page counters" graph

AdrienClairembault:
Oversight of #13775, sometimes widgets are displayed without a dashboard.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #13882

### DIFF
--- a/src/Dashboard/Widget.php
+++ b/src/Dashboard/Widget.php
@@ -656,7 +656,12 @@ HTML;
 
         $js = <<<JAVASCRIPT
       $(function () {
-         var chart = new Chartist.Pie(Dashboard.getActiveDashboard().element.find('#{$chart_id} .chart')[0], {
+         if (Dashboard.getActiveDashboard()) {
+            var target = Dashboard.getActiveDashboard().element.find('#{$chart_id} .chart')[0];
+         } else {
+            var target = '#{$chart_id} .chart';
+         }
+         var chart = new Chartist.Pie(target, {
             labels: {$labels},
             series: {$series},
          }, {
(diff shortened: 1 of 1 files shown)


AdrienClairembault:
@stonebuzz can you make sure it fixes your issue with `NetworkPort` too ?

trasher (approved):

cedric-anne (commented):
If you target items using an id (i.e. a selector starting with a `#`), then, unless I miss something, as ids are meant to be unique, there should be no difference between `Dashboard.getActiveDashboard().element.find('#{$chart_id} .chart')[0]` and `#{$chart_id} .chart`.

If there are many elements with the same id in the page, it should probably be fixed.

trasher:
> If you target items using an id (i.e. a selector starting with a `#`), then, unless I miss something, as ids are meant to be unique, there should be no difference between `Dashboard.getActiveDashboard().element.find('#{$chart_id} .chart')[0]` and `#{$chart_id} .chart`.
> 
> If there are many elements with the same id in the page, it should probably be fixed.

This is not about duplicated ids, but something missing (not loaded?) : `Dashboard.getActiveDashboard() is undefined`

cedric-anne:
> > If you target items using an id (i.e. a selector starting with a `#`), then, unless I miss something, as ids are meant to be unique, there should be no difference between `Dashboard.getActiveDashboard().element.find('#{$chart_id} .chart')[0]` and `#{$chart_id} .chart`.
> > If there are many elements with the same id in the page, it should probably be fixed.
> 
> This is not about duplicated ids, but something missing (not loaded?) : `Dashboard.getActiveDashboard() is undefined`

I just mean that if if we know the id of the element, then `Dashboard.getActiveDashboard()` is useless here, in any case.

trasher:
Looks like that does fix network port metrics

image

trasher:
As well a printer page counters

image

trasher:
> I just mean that if if we know the id of the element, then `Dashboard.getActiveDashboard()` is useless here, in any case.

Oh, OK, I did not understand that... In that case, I do not know, I'm not sure how JS Dashboard is intended to work

AdrienClairembault:
> If there are many elements with the same id in the page, it should probably be fixed.

There is indeed multiple items with the same id on the page when you have two tabs (dashboards + all).
Each widget have a fixed id that's stored in the database, and will be used for in each tab.

I was able to fix the multiple tabs issue without directly altering these duplicates ids so I didn't investigate further into the possibles solutions (but you are welcome to).

stonebuzz:
```Printer```
image

```NetworkPort```
image

cedric-anne (approved):
OK as a workaround, but duplicated IDs should be fixed in a future version.

PR #13926, opened 2023-01-30
2FA Login

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

- [x] Base feature
- [ ] Backup codes
- [ ] Administration (policy control)


orthagh (changes requested):
Some issues i found during 1st tests:

- when enabling "enforcing 2FA" and directly logout/login, i has been asked for a code but i didn't setup my application. There should a scenario where a user doesn't setup should have the registering form and not the seondary login process.
- maybe linked, the initial redirection to `?mfa_setup=1 ` redirect to base root of my localhost domain . I am in a subfolder (http://localhost/glpi/.../) and i need to rewrite manually the url.
- some UX enhancements to do:
   - the registering page could be prettier and guide more the user, some example
      - https://www.beyondtrust.com/docs/remote-support/resources/images/2fa/2fa-replace-21-3_thumb_0_0.png
      - https://dribbble.com/shots/15641874--2FA
      - https://miro.medium.com/v2/format:webp/1*En1MJwdo-U8Rl-voliThfA.png
   - same for recovery code, enhance UI please
   - pasting (with Ctrl+v) a 6 digits code in the intended input should fill all the inputs, not only the first one
   - [use a backup code] button should be on the same line as others ([cancel])
   - Auto-submiting the form after filling the 6-digits code is good, but depsite it's not usefull, a [Submit] button should be displayed to have a comprehensive design (and provide a fallback if something break)
- security tab in Entity miss an icon
- i didn't see this tab in Profile/group (this has been specificaly asked by the customer)
- not sure where is the bug, but the manual registering (with copy/pasting the key in my Authy app) doesn't to be functional. The provided 6-digits code is not accepted by glpi this way. By using QR-code, it's fine.

orthagh:
Another things:

- a cli command to manage a user 2FA could be usefull. Imagine you lost the 2fa for the super-admin account, you're fucked (need to alter things in DB)
- the searchoption for 2FA status in user list may be searchable

cconard96:
> - the searchoption for 2FA status in user list may be searchable

No support for searching virtual search options.

src/Security/TOTPManager.php
```diff
@@ -0,0 +1,379 @@
...
+
+final class TOTPManager
+{
+    public const SECRET_LENGTH_BITS = 160;
+
+    public const CODE_LENGTH_DIGITS = 6;
+
+    public const CODE_VALIDITY_SECONDS = 30;
+
+    public const CODE_ALGORITHM = 'sha1';
```
flegastelois, line 49: SHA1 should not be avoided? since 2015, it is an algorithm that is considered insecure.

cconard96: Yes it should be. However, Google Authenticator doesn't work when I tried SHA-256. I spent a lot of time trying to figure out why it wasn't working with it.

cconard96: An important note is that the RFC for TOTP (rfc6238) is based on HOTP (rfc4226) which specifically mentions SHA-1. By default, authenticator apps seem to use SHA-1 by default for compatibility reasons. When you use a QR code instead of a manually entered secret, it is actually a `otpauth` URI which can contain an `algorithm` parameter to specify the use of a different algorithm. For some reason, the Google Authenticator still ignores the parameter in the URI (I verified it was set correctly). I don't know why we should expect them to support a parameter of a URI they formalized themselves...
https://github.com/google/google-authenticator/wiki/Key-Uri-Format#algorithm

As for the real world implications, there are people who know way more than me on the subject. However, from what I understand, TOTP uses HMAC (HMAC-SHA-1 in this case) which does reduce the risks of using SHA-1 as the secret key is assumed to be unknown by anyone except the application and client.

cconard96:
>     There should a scenario where a user doesn't setup should have the registering form and not the seondary login process.

That is already the case. I broke the enforced setup twig template with a rouge parenthesis. In some odd cases (bugs), the 2FA code request screen can show instead of the enforced setup screen (asks users to set up 2FA or skip if within the grace period), but that is not a normal behavior and the user can cancel out of it if it ever ends up in that state.

>     * not sure where is the bug, but the manual registering (with copy/pasting the key in my Authy app) doesn't seems to be functional. The provided 6-digits code is not accepted by glpi this way. By using QR-code, it's fine.

I can't find any proper documentation on how the Authy app works. I'm guessing that the default algorithm is different than SHA-1. When using the QR code, you can specify the algorithm used to generate the TOTP codes. Unfortunately, Google Authenticator is a very popular authenticator and they formalized the URI spec used in QR code setups, but ignore their own algorithm parameter and only support SHA-1,

cconard96:
Some technical changes made:
1. Fixed redirects, missing icon, and missing enforcement configurations
2. Added logic to automatically determine the algorithm used by the authenticator in case it isn't sha1 like we expected. When setting the secret, all supported algorithms are checked (sha1 and sha256 currently) and the matching one is stored in the user's 2FA config for future use. This should improve compatibility when using the manual secret entry.

orthagh:
I confirm the authy registering and the 1st enroll are succesfull now :+1:

cconard96:
UX changes made.

Regular/optional 2FA Setup:
Selection_128

Enforced setup during login (in grace period):
Selection_131

Enforced setup during login (no/outside grace period):
Selection_130

2FA code request:
Selection_132

2FA backup code
Selection_133
 request:

2FA backup code generation:
Selection_129

orthagh:
Seems very good.
To be picky, i'll suggest:

- to avoid repetitions ("2FA setup", "2FA is required for your account")
- to add some vertical spacing around your numbered paragraphs (this is valid for any explanation paragraph, keep them "aerated")
- to avoid rounded borders on digit inputs not placed on the edges or **BETTER**:
   - use spacing between digits (keep rounded borders)
   - increase greatly font and input sizes (this is the primary control, we want it to be the attraction of the page)
   - ex:

image

cconard96:
Done

orthagh (changes requested):
found some issues while testing correctly:

- copy button right of code (when registering) doesn't seem to work (both chrome and firefox) and a js error is present
  ```
  Uncaught (in promise) TypeError: Failed to execute 'write' on 'Clipboard': The object must have a callable @@iterator property.
  ```
- .card html issue:
    image
- I didn't understand [reset 2FA] button purpose in preferences, nothing seems to change after clicking to it


UI/UX pov, nothing more to say, great job !

Tomorrow, I'll check/review Entity/Profile/Group enforcing, now i need to go

cconard96:
>     * I didn't understand [reset 2FA] button purpose in preferences, nothing seems to change after clicking to it
It is supposed to display the setup form again without actually disabling 2FA first. It shows in place of the Disable 2FA button when 2FA is mandatory. It is bugged because the GET parameter that is set is not available when the tab content loads.

src/Config.php
```diff
@@ -3415,6 +3427,42 @@ public function showFormSecurity()
...
+        Dropdown::showNumber('2fa_grace_days', ['value' => $CFG_GLPI["2fa_grace_days"],
+            'min'   => 0,
+            'max'   => 365,
+            'rand'  => $rand
+        ]);
+        if (!empty($CFG_GLPI['2fa_grace_date_start']) && (int) $CFG_GLPI['2fa_enforced'] === 1 && (int) $CFG_GLPI['2fa_grace_days'] > 0) {
+            $days_left = (new \Glpi\Security\TOTPManager())->getGracePeriodDaysLeft();
+            if ($days_left >= 0) {
+                $end_ts = strtotime($CFG_GLPI['2fa_grace_date_start']) + ($CFG_GLPI['2fa_grace_days'] * DAY_TIMESTAMP);
+                echo '<br><span>' . __('2FA grace period started: ') . Html::convDateTime($CFG_GLPI['2fa_grace_date_start']) . '</span>';
```
cedric-anne, line 3457: Value position may change in the sentence depending on the locale.
```suggestion
                echo '<br><span>' . sprintf(__('2FA grace period started on %s'), Html::convDateTime($CFG_GLPI['2fa_grace_date_start'])) . '</span>';
```

src/Console/Security/DisableTFACommand.php
```diff
@@ -0,0 +1,46 @@
...
+use Symfony\Component\Console\Question\ConfirmationQuestion;
+
+class DisableTFACommand extends AbstractCommand
+{
+    protected function configure()
+    {
+        parent::configure();
+        $this->setName('security:disable_2fa');
+        $this->setDescription(__('Disable 2FA for a user'));
+        $this->addArgument('login', InputArgument::REQUIRED, 'Username');
```
cedric-anne, line 18:
```suggestion
        $this->addArgument('login', InputArgument::REQUIRED, __('Username'));
```

src/Console/Security/DisableTFACommand.php
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $username = $input->getArgument('login');
+        $user = new \User();
+        if (!$user->getFromDBbyName($username)) {
+            $output->writeln("<error>" . __("User $username not found") . "</error>");
```
cedric-anne, line 26:
```suggestion
            $output->writeln("<error>" . sprintf(__("User %s not found"), $username) . "</error>");
```

src/Console/Security/DisableTFACommand.php
```diff
@@ -0,0 +1,46 @@
...
+        if ($totp_manager->get2FAEnforcement($user->getID())) {
+            $output->writeln("<info>" . __("2FA is enforced for this user. They will be required to set it up again the next time they log in.") . "</info>");
+        }
+        $helper = $this->getHelper('question');
+        $question = new ConfirmationQuestion(__('Are you sure you want to disable 2FA for this user?'), false);
+        if (!$helper->ask($input, $output, $question)) {
+            return 0;
+        }
+        $totp_manager->disable2FAForUser($user->getID());
+        $output->writeln("<info>" . __("2FA disabled for user $username") . "</info>");
```
cedric-anne, line 43:
```suggestion
        $output->writeln("<info>" . sprintf(__("2FA disabled for user %s"), $username) . "</info>");
```

src/Profile.php
```diff
@@ -2150,6 +2156,58 @@ public function showFormSetupHelpdesk($openform = true, $closeform = true)
...
+        echo "<div class='spaced'>";
+        if ($canedit) {
+            echo "<form method='post' name=form action='" . Toolbox::getItemTypeFormURL(__CLASS__) . "' data-track-changes='true'>";
+        }
+
+        echo "<table class='tab_cadre_fixe'>";
+
+        $rand = mt_rand();
+
+        echo "<tr><th colspan='2'>" . __('2FA') . "</th></tr>";
```
cedric-anne, line 2178: Same remark.

src/Security/TOTPManager.php
```diff
@@ -0,0 +1,495 @@
...
+                if (!empty($profiles)) {
+                    $enforced = $DB->request([
+                        'SELECT' => ['2fa_enforced'],
+                        'FROM' => 'glpi_profiles',
+                        'WHERE' => [
+                            'id' => $profiles,
+                            '2fa_enforced' => 1
+                        ]
+                    ])->count() > 0;
+                }
```
cedric-anne, line 397: This implies that a user that has a limited profile, that does not enforce 2FA, as its default profile and an admin profile, that enforces 2FA, and that is used really often, then the user will be forced to use 2FA on both profiles. Is it possible to activate the 2FA only on profile selection?

templates/pages/2fa/2fa_status.html.twig
```diff
@@ -0,0 +1,133 @@
...
+            {{ __('2FA grace period ends %s')|format(grace_period_end|relative_datetime) }}
+         {% elseif enforcement == constant('Glpi\\Security\\TOTPManager::ENFORCEMENT_MANDATORY') %}
+            {{ __('Mandatory') }}
+         {% else %}
+            {{ __('Optional') }}
+         {% endif %}
+      </h2>
+   </div>
+   <div id="backup_codes" class="d-none py-2" style="width: fit-content;">
+      <h3 class="mb-2">{{ __('Backup codes (This is the only time these will be shown)') }}</h3>
```
cedric-anne, line 48: Maybe the `This is the only time these will be shown` message should be shown in a :warning: block.

src/Config.php
```diff
@@ -3415,6 +3427,42 @@ public function showFormSecurity()
...
+        echo '</td>';
+        echo '<td>';
+        echo '<label for="dropdown_2fa_grace_days' . $rand . '">';
+        echo __('2FA grace period (in days)');
+        echo '</label>';
+        echo '</td>';
+        echo '<td>';
+        Dropdown::showNumber('2fa_grace_days', ['value' => $CFG_GLPI["2fa_grace_days"],
+            'min'   => 0,
+            'max'   => 365,
```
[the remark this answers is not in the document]
cconard96: That's not really for me to say. It is up to the IT/Security team.

src/Auth.php
```diff
@@ -942,14 +960,84 @@ public function login($login_name, $login_password, $noauto = false, $remember_m
...
+        // is not present on the DB, so we add him.
+        // if not, we update him.
+        if (empty($mfa_pre_auth) && empty($login_name) && empty($login_password)) {
+            // Bad login
+            return false;
+        }
+        if ($mfa_pre_auth || $this->validateLogin($login_name, $login_password, $noauto, $login_auth)) {
+            // Check MFA
+            $totp = new TOTPManager();
+            $web_access = !isAPI() && !isCommandLine();
```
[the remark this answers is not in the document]
cconard96: I don't know the CalDAV feature at all, but basic authentication probably shouldn't be supported in the API at all and can be disabled with the pending OAuth server and High-Level API feature PRs.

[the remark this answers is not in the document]
cedric-anne: We discussed about it with @orthagh .

CalDAV and API security of API/CalDAV endpoints will be enforced in dedicated PRs. We added corresponding tasks in the roadmap.

I pushed a commit that fixes CalDAV endpoint authentication.

src/Security/TOTPManager.php
cedric-anne: We discussed about it with @orthagh.

Requesting a OTP code on login as long as user has access to an entity that enforces 2FA or is part of a group/profile that enforces 2FA is the expected behaviour.

The only thing to add it a :information_source: block on corresponding configuration forms to clearly explains this.

cedric-anne (commented):
Looks OK to me. I would not have time to do extensive testing right now since I will be off for a week.

src/Security/TOTPManager.php
```diff
@@ -0,0 +1,495 @@
...
+        }
+
+        $enforced = false;
+        if ((int) $CFG_GLPI['2fa_enforced'] === 1) {
+            $enforced = true;
+        } else {
+            // Check entity configuration
+            $entities = PermissionManager::getInstance()->getAllEntities($users_id);
+            foreach ($entities as $entity) {
+                if ((int) \Entity::getUsedConfig('2fa_enforcement_strategy', $entity, '', true) === 1) {
```
orthagh, line 380: the `true` value for default make the `get2FAEnforcement` method to return enforced=true when no enforcement has been set at all.

And so, i didn't setup the enforcement while testing the feature on an instance where the pr hasn't been tested at all, and it asked to register the 2fa application

orthagh (changes requested):

front/login.php
```diff
@@ -86,7 +90,17 @@
 
 
 // now we can continue with the process...
-if ($auth->login($login, $password, (isset($_REQUEST["noAUTO"]) ? $_REQUEST["noAUTO"] : false), $remember, $login_auth)) {
+if (isset($_REQUEST['cancel'])) {
```
orthagh, line 93: should we use more precise keys like `totp_cancel`, `totp_code`, `backup_code` when using POST variables between redirections ?

src/Config.php
```diff
@@ -3415,6 +3427,42 @@ public function showFormSecurity()
         echo '</td>';
         echo '</tr>';
 
+        echo '<tr class="tab_bg_1">';
```
orthagh, line 3430: as we target main, we may take advantage to move to twig template for this whole security form

src/Entity.php
```diff
@@ -2641,6 +2645,71 @@ public static function showUiCustomizationOptions(Entity $entity)
...
+            echo "<form method='post' name=form action='" . Toolbox::getItemTypeFormURL(__CLASS__) . "' data-track-changes='true'>";
+        }
+
+        echo "<table class='tab_cadre_fixe'>";
+
+        Plugin::doHook(Hooks::PRE_ITEM_FORM, ['item' => $entity, 'options' => []]);
+
+        $rand = mt_rand();
+
+        echo "<tr><th colspan='2'>" . __('Two-factor authentication (2FA)') . "</th></tr>";
```
orthagh, line 2675: same, can we use twig for this part ?

src/Group.php
```diff
@@ -642,6 +647,49 @@ public function showLDAPForm($ID)
         $this->showFormButtons($options);
     }
 
+    /**
+     * @param $ID
+     **/
+    public function showSecurityForm($ID)
+    {
+        $options = [];
```
orthagh, line 655: same, twig ?

src/Profile.php
```diff
@@ -2150,6 +2156,60 @@ public function showFormSetupHelpdesk($openform = true, $closeform = true)
...
     }
 
+    /**
+     * Print the Security form for a profile
+     *
+     * @param $openform     boolean  open the form (true by default)
+     * @param $closeform    boolean  close the form (true by default)
+     **/
+    public function showFormSecurity($openform = true, $closeform = true)
+    {
```
orthagh, line 2166: same twig ?

also, the form seem (maybe there's difference) to be the same between group, profile and entity, can we factorize the display ?

src/Security/PermissionManager.php
```diff
@@ -0,0 +1,88 @@
...
+ */
+
+namespace Glpi\Security;
+
+use Profile;
+use Profile_User;
+use ProfileRight;
+use QueryExpression;
+
+final class PermissionManager
```
orthagh, line 43: can you comment above the class its general purpose, it's not that clear at read

orthagh (changes requested):
in addition, you should add to User class the `2fa` entry to `$undisclosedFields` property. This to avoid leak of key, or backup codes

src/Security/PermissionManager.php
cconard96: done. for reference, this was borrowed from the pending PR #11649

orthagh: 👍🏽

orthagh (approved):

cedric-anne (changes requested):
When a user enables the 2FA on its own account, from its preferences page, backup codes are not shown after 2FA activation. They should.

cedric-anne:
Tests are failing and should be fixed.

cedric-anne (approved):

trasher (approved):

PheitosaBr:
Dear all, congratulations on the solution. How can I install it inside my GLPI?

PR #13979, opened 2023-02-01
Update RuleLocationCollection.php - sort order and breadcrumb trail

SylvieCoz:
- Order Location Rules by name rather than ID
- Add this rule collection to the breadcrumb trail

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #number

### DIFF
--- a/src/RuleLocationCollection.php
+++ b/src/RuleLocationCollection.php
@@ -39,7 +39,9 @@ class RuleLocationCollection extends RuleCollection
 {
     public $stop_on_first_match = true;
     public static $rightname    = 'rule_location';
-
+    public $orderby             = 'name';
+    public $menu_option         = 'location';
+    
     public function getTitle()
     {
         return __("Location rules");
--- a/src/RuleLocationCollection.php
+++ b/src/RuleLocationCollection.php
@@ -39,9 +39,7 @@ class RuleLocationCollection extends RuleCollection
 {
     public $stop_on_first_match = true;
     public static $rightname    = 'rule_location';
-    public $orderby             = 'name';
     public $menu_option         = 'location';
-    
     public function getTitle()
     {
         return __("Location rules");


src/RuleLocationCollection.php
```diff
@@ -39,7 +39,9 @@ class RuleLocationCollection extends RuleCollection
 {
     public $stop_on_first_match = true;
     public static $rightname    = 'rule_location';
-
+    public $orderby             = 'name';
```
cedric-anne, line 42: I do not know why it is still possible to change the field used to define rules display order, but it completely breaks the ability to sort them. Indeed, while it is still possible to use the drag&drop feature (see image), their ranking is changed but they are always loaded and evaluated based on a name sorting (and reloading the page always show them ordered by name)
image

```suggestion
```

src/RuleLocationCollection.php
```diff
@@ -39,7 +39,9 @@ class RuleLocationCollection extends RuleCollection
...
+    public $menu_option         = 'location';
+
```
cedric-anne, line 44:
```suggestion

```

src/RuleLocationCollection.php
SylvieCoz: Cédric, you're right : I missed the drag&drop feature, so forget the adding of **$orderby** suggestion
FYI, I took the **$orderby** instruction from the RuleRightCollection class (need to be corrected as well ?)

I still suggest the adding of **$menu_option** to have the breadcrumb trail updated (see image)
image

cedric-anne (approved):

trasher (approved):

cedric-anne:
Thanks for the contribution.

src/RuleLocationCollection.php
cedric-anne: > FYI, I took the **$orderby** instruction from the RuleRightCollection class (need to be corrected as well ?)

#13988 will remove ability to override default order field.

PR #13986, opened 2023-02-02
fix(document): entry name in database

Rom1-B:
The name of the document in the database loaded from a ticket was not determined in the same way:

- If the document was loaded directly (Action: `Add a document`), the ticket name was used
- If it is in a followup, the file name was used.

_IMO, users look at documents more from a ticket than from the document menu, so the name based on the source is less relevant than the original file._

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !26561

### DIFF
--- a/src/Document.php
+++ b/src/Document.php
@@ -241,21 +241,6 @@ class Document extends CommonDBTM
             && ($item = getItemForItemtype($input["itemtype"]))
             && ($input["items_id"] > 0)
         ) {
-            $typename = $item->getTypeName(1);
-            $name     = NOT_AVAILABLE;
-
-            if ($item->getFromDB($input["items_id"])) {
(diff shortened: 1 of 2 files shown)


cedric-anne (changes requested):
This code is here from a long. The main difference between GLPI 9.5 and GLPI 10.0 is that, in the timeline, we now shows the document name instead of the filename.

So, for me:
 - this PR should target main branch,
 - another PR should be made to display the filename instead of the document name in the timeline (will fix #12974).

cedric-anne (approved):

trasher (approved):

PR #14040, opened 2023-02-09
Upgrade tabler icons to v2.2

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

Based on the changelogs, no icon was removed and the icons that had been renamed don't seem to be used by GLPI or any of the plugins I have downloaded. There have been hundreds of new icons added since 1.68 and many of them are useful for a plugin I am working on.

If BC is still a concern, I guess CSS rules could be added so that the old names of any renamed icons that exist in 1.68 can still work.


cedric-anne (approved):
There are really few icons from 1.68- that have been renamed, so it is OK for me.

orthagh (approved):

trasher (approved):

PR #14077, opened 2023-02-14
add key on level column for CommonTreeDropdowns

btry:
For itemtypes using commontreedropdown items may show very slowly because of unoptimized SQL queries. 

This may happen with Formcreator, when a question is designed to show only a subtree of existing items, and when there are a thousand of items in the database.

Adding an index on level columns shows a speed gain of 50%.

internal ref !26669

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #

### DIFF
--- a/install/migrations/update_10.0.6_to_10.0.7/indexes.php
+++ b/install/migrations/update_10.0.6_to_10.0.7/indexes.php
@@ -1,0 +1,60 @@
+<?php
+
+/**
+ * ---------------------------------------------------------------------
+ *
+ * GLPI - Gestionnaire Libre de Parc Informatique
+ *
+ * http://glpi-project.org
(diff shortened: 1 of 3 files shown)


trasher (approved):

orthagh (approved):

cedric-anne (changes requested):
A check should be added in `DatabaseKeysChecker` to validates that `level` field is always indexed when table refeers to a `CommonTreeDropdown` item.

trasher:
ping @btry ?

cedric-anne (approved):

PR #14084, opened 2023-02-14
Rework All Tab

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

Proof of concept to utilize the tab content already loaded instead of forcibly getting the content of every tab again for the All tab. This should eliminate all issues with duplicate IDs and duplicate initialization of JavaScript components, and simplify the server-side handling of this tab. This also allows the user to start filling a form in a tab, and finish it in the All tab. The downside is that loading the All tab may take longer because the loading of previously unloaded tabs would be done in individual requests.

New functionality when the All tab is selected:
1. Load the content of every other tab without forcing a reload so that only the unloaded tabs are retrieved from the server
2. Manually add the `active` and `show` classes to the tab panels to show all the tabs
When selecting a tab other than the All tab, the superfluous `active` and `show` classes are removed.

In some cases, there seems to be some race conditions where the classes are not set as expected. This PR is just a proof of concept and would require some additional work for a proper implementation.

### DIFF
--- a/src/Ajax.php
+++ b/src/Ajax.php
@@ -360,7 +360,7 @@ JAVASCRIPT;
             }
             echo  "</div>"; // .tab-content
             echo "</div>"; // .container-fluid
-            $js = "
+            $js = <<<JS
          var loadTabContents = function (tablink, force_reload = false) {
             var url = tablink.attr('href');
             var target = tablink.attr('data-bs-target');
@@ -418,7 +418,7 @@ JAVASCRIPT;
                $('#$tabdiv_id li a').eq($(this).val()).tab('show');
             });
          });
-         ";
+JS;
 
             echo Html::scriptBlock($js);
         }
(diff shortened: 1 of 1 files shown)


orthagh:
> This should eliminate all issues with duplicate ID

Just for that, I'm strongly in favor of this pr. ping @AdrienClairembault 

> The downside is that loading the All tab may take longer

Seems a minor cons regarding the pros

Also the current changes (despites somes remaining bugs), seems very simple and well understandable

AdrienClairembault:
Looks good to me too :)

Just need to be sure we don't break anything on the dashboards/personal views/group views as I've add issues in the past with them that needed specifics fixes for the all tabs.

cconard96:
> 

I've not seen any issues with the central page with this PR.

AdrienClairembault (approved):
Tested on my side, looks good :)

orthagh (approved):

cedric-anne (approved):

PR #14100, opened 2023-02-16
fix(search): satisfaction column empty in search result

Rom1-B:
In the ticket search, the satisfaction column was empty after a search, but complete when reloading the whole page.

The star rating depends on the JS lib "rateit" which only applies to its loading. However, when we refresh only the results table, the JS was not reloaded and therefore the stars do not appear.

Before:
image

After:
image

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !26728

### DIFF
--- a/js/modules/Search/Table.js
+++ b/js/modules/Search/Table.js
@@ -199,6 +199,7 @@ window.GLPI.Search.Table = class Table extends GenericView {
                 this.getElement().trigger('search_refresh', [this.getElement()]);
                 this.hideLoadingSpinner();
                 this.shiftSelectAllCheckbox();
+                this.reloadJS('rateit');
             }, () => {
                 handle_search_failure();
             });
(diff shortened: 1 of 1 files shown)


orthagh:
@cconard96 you're the most capable of reviewing this PR I guess

js/modules/Search/Table.js
```diff
@@ -199,6 +199,7 @@ window.GLPI.Search.Table = class Table extends GenericView {
                 this.getElement().trigger('search_refresh', [this.getElement()]);
                 this.hideLoadingSpinner();
                 this.shiftSelectAllCheckbox();
+                this.reloadJS('rateit');
```
cconard96, line 202: It looks like this can be simplified to just:
ajax_container.find('.rateit').rateit();

No need to reload the rateit library as it is available when the page loads. The issue is that the library doesn't re-initialize the components automatically after the page is done loading so anything added by AJAX requests needs manually initialized.

cedric-anne (approved):

trasher (approved):

trasher:
Tetss are failing (I do not know why)

orthagh (approved):

AdrienClairembault:
Note that is already fixed in main : https://github.com/glpi-project/glpi/pull/13316

> 3) It disappears when reloading the search results
-> Rewrote displaySatisfaction to use javascript to trigger the rateit library.
The previous method (relying on data-xxxx attributes) does not seem to trigger on ajax reloads.

trasher:
I agree we have to be consistent across bf and main branch. Maybe the solution already done in main should be preferred if it does not break tests :)
Anyways, current PR cannot be merged until tests are fixed.

cedric-anne (approved):

Issue #14088, opened 2023-02-15
[inventory] Duplicate entry 'Computer-' for key 'unicity' - and error 500 from the agent


TheRoby:
### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

10.0-f00a4a1

### Bug description

Hi,
the inventory of some computers does not work anymore (error 500 in the logs)
i don't know if it's glpi or glpi-agent concern ...

### Relevant log output

```shell
php-errors.log :
[2023-02-15 10:15:38] glpiphplog.ERROR:   *** PHP User Error (256): Duplicate entry 'Computer-4324-60150' for key 'unicity' in /var/www/glpi/src/DBmysql.php at line 1969
  Backtrace :
  src/DBmysql.php:1969                               trigger_error()
  src/Inventory/Asset/Software.php:878               DBmysql->executeStatement()
  src/Inventory/Asset/Software.php:445               Glpi\Inventory\Asset\Software->storeAssetLink()
  src/Inventory/Asset/MainAsset.php:874              Glpi\Inventory\Asset\Software->handle()
  src/Inventory/Asset/MainAsset.php:794              Glpi\Inventory\Asset\MainAsset->handleAssets()
  src/RuleImportAsset.php:969                        Glpi\Inventory\Asset\MainAsset->rulepassed()
  src/Rule.php:1519                                  RuleImportAsset->executeActions()
  src/RuleCollection.php:1625                        Rule->process()
  src/Inventory/Asset/MainAsset.php:571              RuleCollection->processAllRules()
  src/Inventory/Inventory.php:706                    Glpi\Inventory\Asset\MainAsset->handle()
  src/Inventory/Inventory.php:341                    Glpi\Inventory\Inventory->handleItem()
  src/Inventory/Request.php:360                      Glpi\Inventory\Inventory->doInventory()
  src/Inventory/Request.php:90                       Glpi\Inventory\Request->inventory()
  src/Agent/Communication/AbstractRequest.php:359    Glpi\Inventory\Request->handleAction()
  src/Agent/Communication/AbstractRequest.php:271    Glpi\Agent\Communication\AbstractRequest->handleJSONRequest()
  front/inventory.php:92                             Glpi\Agent\Communication\AbstractRequest->handleRequest()
  plugins/glpiinventory/front/communication.php:72   include_once()
  plugins/glpiinventory/index.php:45                 include_once()

[...]

PR #14105, opened 2023-02-17
Fix unbreakable space issues in soft name during inventory

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #14088

### DIFF
--- a/src/Inventory/Asset/Software.php
+++ b/src/Inventory/Asset/Software.php
@@ -110,7 +110,7 @@ class Software extends InventoryAsset
 
             //If the software name exists and is defined
             if (property_exists($val, 'name') && $val->name != '') {
-                $val->name = trim(preg_replace('/\s+/', ' ', $val->name));
+                $val->name = trim(preg_replace('/\s+/u', ' ', $val->name));
 
                 $res_rule       = [];
 
--- a/tests/functionnal/Glpi/Inventory/Assets/Software.php
+++ b/tests/functionnal/Glpi/Inventory/Assets/Software.php
@@ -836,4 +836,112 @@ class Software extends AbstractInventoryAsset
 
         $computer->deleteByCriteria(['id' => $first_computer['id']]);
     }
+
+    public function testDuplicatedSoft()
+    {
+        global $DB;
+        $this->login();
+
+        $computer = new \Computer();
+        $soft = new \Software();
+        $version = new \SoftwareVersion();
+        $item_version = new \Item_SoftwareVersion();
+
+        //inventory with a software name containing an unbreakable space
+        $xml_source = "<?xml version=\"1.0\" encoding=\"UTF-8\" ?>
+<REQUEST>
+  <CONTENT>
+    <SOFTWARES>
+      <ARCH>i586</ARCH>
+      <FROM>registry</FROM>
+      <GUID>Office15.STANDARD</GUID>
+      <INSTALLDATE>18/08/2020</INSTALLDATE>
+      <NAME>Microsoft Office Standard 2013</NAME>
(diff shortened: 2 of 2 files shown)


src/Inventory/Asset/Software.php
```diff
@@ -110,7 +110,7 @@ public function prepare(): array
 
             //If the software name exists and is defined
             if (property_exists($val, 'name') && $val->name != '') {
-                $val->name = trim(preg_replace('/\s+/', ' ', $val->name));
+                $val->name = trim(preg_replace('/\s+/u', ' ', $val->name));
```
stonebuzz, line 113: The problem with handling it this way is that software with the same name (but with a different GUID) merges

We can't handle ```<GUID>``` easily (straightening the data seems very complicated to me), maybe we can just keep an unbreakable space in DB (actually database stores it and returns it as a normal space).

tests/functionnal/Glpi/Inventory/Assets/Software.php
```diff
@@ -836,4 +836,112 @@ public function testSoftwareWithHtmlentites($path)
...
+              <SSN>ggheb7ne7</SSN>
+            </BIOS>
+            <VERSIONCLIENT>FusionInventory-Agent_v2.3.19</VERSIONCLIENT>
+          </CONTENT>
+          <DEVICEID>test-pc002</DEVICEID>
+          <QUERY>INVENTORY</QUERY>
+        </REQUEST>";
+
+        $this->doInventory($xml_source, true);
+    }
```
stonebuzz, line 946:
```suggestion
        $manufacturer = new \Manufacturer();
        $this->boolean($manufacturer->getFromDBByCrit(['name' => 'Microsoft Corporation']))->isTrue();

        //we have 1 software & versions for Microsoft Office Standard - Microsoft Corporation
        //Microsoft Office Standard - Microsoft Corporation
        $softs = $soft->find(['name' => 'Microsoft Office Standard 2013', 'manufacturers_id' => $manufacturer->fields['id']]);
        $this->integer(count($softs))->isIdenticalTo(2);


	      //15.0.4569.1506
        $versions = $version->find(['name' => '15.0.4569.1506']);
        $this->integer(count($versions))->isIdenticalTo(2);

        foreach ($versions as $version_data) {
            $this->boolean($item_version->getFromDBByCrit([
                "itemtype" => "Computer",
                "items_id" => $computers_id,
                "softwareversions_id" => $version_data['id']
            ]))->isTrue();
        }
    }
```

stonebuzz: With this, TU failed and reveals that the software has merged

stonebuzz (changes requested):

tests/functionnal/Glpi/Inventory/Assets/Software.php
trasher: As discussed; we cannot keep both software. Unbreakable space cannot be used from mysql database because any query on it will return both standard and unbreakable space.

In previous versions - as in original plugin - we keep only one of those two.

src/Inventory/Asset/Software.php
trasher: I agree the best solution would be to rely on the `GUID` to keep both software, but it's a new feature since they've always been merged. And also, software inventory is not simple, that new feature would take time to be implemented.

orthagh (approved):

stonebuzz (approved):

PR #14167, opened 2023-02-24
Fix skipping 2FA setup in grace period

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

The code is validated on the server side so having the code input labeled as required isn't needed. This fixes the issue with skipping the 2FA setup when in a grace period.

### DIFF
--- a/index.php
+++ b/index.php
@@ -131,6 +131,9 @@ if (!file_exists(GLPI_CONFIG_DIR . "/config_db.php")) {
         Toolbox::manageRedirect($redirect);
     }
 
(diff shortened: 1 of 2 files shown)


orthagh (approved):

cedric-anne (commented):
Maybe the `formnovalidate` attribute should have been placed on the `Skip` button instead.

cedric-anne (approved):

trasher (approved):

PR #14182, opened 2023-02-28
IITLTimeline: hide "answer" button if not allowed to add followups

AdrienClairembault:
User that aren't allowed to add followups still have the "answer" button in the ITIL timeline.

Rights:

image

Before:

image

Note that "answer" is shown but not "add task" despite the user rights indicating the opposite.

After:

image

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !26945

### DIFF
--- a/templates/components/itilobject/footer.html.twig
+++ b/templates/components/itilobject/footer.html.twig
@@ -45,21 +45,27 @@
    <div class="buttons-bar d-flex py-2">
       <div class="col {{ timeline_btns_cls }} ps-3 timeline-buttons d-flex">
          {% if not item.isNewItem() %}
-            {% set default_action_data = timeline_itemtypes|first %}
+            {% set main_actions_itemtypes = timeline_itemtypes|filter((v, k) => v.hide_in_menu is not defined or v.hide_in_menu != true) %}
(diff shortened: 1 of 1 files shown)


orthagh (commented):
At first read, the feeling i have : There is too much behavior coding in the template.
Please, defer the maximum of the kind of code in the `showForm` method of Ticket|CommonItilObject

AdrienClairembault:
> At first read, the feeling i have : There is too much behavior coding in the template. Please, defer the maximum of the kind of code in the `showForm` method of Ticket|CommonItilObject

The code changes doesn't add any new "behavior coding", it just move some code to allow the visibility checks to be handled a bit sooner.
Before, the check was done after the first button was already displayed, which would makes it to always visible.
Now it's done before any buttons are displayed.

If you want to rewrite the template, it should probably be done in main instead.

cedric-anne (approved):

orthagh:
Ok my bad, i stop the read of the pr after the second changes.

PR #14231, opened 2023-03-02
fix(html): add some icons

flegastelois:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | N/A

### DIFF
--- a/src/RuleAsset.php
+++ b/src/RuleAsset.php
@@ -48,6 +48,10 @@ class RuleAsset extends Rule
         return __('Business rules for assets');
     }
 
+    public static function getIcon()
+    {
+        return "ti ti-package";
+    }
 
     public function maybeRecursive()
     {
--- a/src/RuleCriteria.php
+++ b/src/RuleCriteria.php
@@ -99,7 +99,7 @@ class RuleCriteria extends CommonDBChild
 
     public static function getIcon()
     {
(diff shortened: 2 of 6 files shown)


src/RuleCriteria.php
```diff
@@ -97,6 +97,11 @@ public static function getTypeName($nb = 0)
         return _n('Criterion', 'Criteria', $nb);
     }
 
+    public static function getIcon()
+    {
+        return "ti ti-adjustments";
```
orthagh, line 102:
```suggestion
        return "ti ti-list-check";
```

src/RuleAsset.php
```diff
@@ -48,6 +48,10 @@ public function getTitle()
         return __('Business rules for assets');
     }
 
+    public static function getIcon()
+    {
+        return "ti ti-package";
```
orthagh, line 53:
```suggestion
        return "ti ti-player-play";
```

flegastelois: "ti-package" choice was to keep consistency with the menu to "Asset" menu.

orthagh (approved):

trasher (approved):

PR #14232, opened 2023-03-02
feat(global search): sort items by number of results

Rom1-B:
Sort items in the global search results by number of results

Before:
image

After:
image

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !26078

### DIFF
--- a/front/search.php
+++ b/front/search.php
@@ -43,7 +43,7 @@ if (!$CFG_GLPI['allow_search_global']) {
 }
 if (isset($_GET["globalsearch"])) {
     $searchtext = trim($_GET["globalsearch"]);
-    $all_data = [];
+    $no_result = [];
 
     echo "<div class='search_page search_page_global flex-row flex-wrap'>";
     foreach ($CFG_GLPI["globalsearch_types"] as $itemtype) {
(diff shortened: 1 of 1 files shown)


cconard96 (approved):

orthagh (approved):

cconard96:
Would it be better to just hide the empty lists and add a note somewhere indicating which types were searched against but had no results?
My worry is that the global search doesn't allow for more specific searches currently, so more results of one type doesn't necessarily mean the results of that type are more relevant.

trasher:
I agree with Curtis, the number of results does not mean they're relevant.
Also, it can be confusing on several successive search that itemtype order has changed.

 I also agree to hide empty lists.

Rom1-B:
Currently, the 30 item types without apparent sorting logic (not even alphabetical). I'm not against having the same sorting all the time, but which one?
I think that if an item type has more results, there is a higher probability that it contains what we are looking for and therefore appears at the top.

trasher:
> I think that if an item type has more results, there is a higher probability that it contains what we are looking for and therefore appears at the top.

No, having many results never implies that it contains the one you are looking for.

Maybe the way all of that is displayed may be reviewed also, with kind of filtering/direct access.

Any sort on this page will always be wrong for someone; itemtype name is not more relevant than number of results; buut it's at least consistent between two queries.

cconard96 (approved):

trasher (approved):

orthagh (approved):

cedric-anne (approved):

PR #14238, opened 2023-03-02
Improve form for the "Add new task" massive action

AdrienClairembault:
Changes summary:
* Improve "Add new task" form in massive actions
* Increase task content textarea size 
* Remove width restriction on massive actions form

###  1) Improve "Add new task" form in massive actions

The 'Add new task" massive action form didn't support all the options from the "original" add task form found in the timeline:

image

image

To improve this, I've exported the HTML code responsible for the "timeline version" of this form in a separate twig file (form_task_main_form.html.twig) which can then be included in the timeline and in massive actions.

This way we use the same form for both endpoints, minus a few special behaviors that must be handled with the template parameters:
* `no_form` as the massive action page already has a form defined with a specific target
* `disable_pending_reasons` to hide the pending reason toggle on the massive actions page as it may lead to accidental changes on the tickets' statuses.

Massive action form after changes:

image

###  2) Increase task content textarea size 

We have a lot of options on the right side which leads to wasted space on the left:

image

I think it look better if we use the same space on both columns:

image

To do this, I've added a new `editor_height` option to the `textareaField` macro.

This also affect tasks on the timeline:

image

image

###  3) Remove width restriction on massive actions form

It seems the massive action page had a weird 90% width restriction:

image

I've removed it as it didn't seem intentional, probably a remnant from the old UI.

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !24844 FR75


AdrienClairembault:
Tests failure can be ignored as it does not seem related to the changes:

image

trasher (approved):

templates/components/itilobject/timeline/form_task_main_form.html.twig
```diff
@@ -0,0 +1,370 @@
...
+            </button>
+
+            {% if subitem.can(subitem.fields['id'], constant('PURGE')) %}
+                <button class="btn btn-outline-danger me-2" type="submit" name="purge"
+                        onclick="return confirm('{{ __('Confirm the final deletion?') }}');">
+                <i class="fas fa-trash-alt"></i>
+                <span>{{ _x('button', 'Delete permanently') }}</span>
+                </button>
+            {% endif %}
+            {{ pending_reasons|raw }}
```
cedric-anne, line 362: Should not be printed if `disable_pending_reasons` is `true`.

templates/components/itilobject/timeline/form_task_main_form.html.twig
```diff
@@ -0,0 +1,370 @@
...
+                        <span>{{ __('Plan this task') }}</span>
+                    </button>
+                {% endif %}
+                <div id="viewplan{{ rand }}"></div>
+                </div>
+            </div>
+        </div>
+    </div>
+
+    {% set pending_reasons %}
```
cedric-anne, line 294: Should not be computed if `disable_pending_reasons` is `true`.

cedric-anne (changes requested):
`form_task_main_form.html.twig` expect many variables to be defined, like `formoptions`, `params`, `rand`, `can_read_kb`, `kb_id_toload`, `candedit`, `can_update_kb`, `nokb` and `add_html_form`, but these are not defined by caller template.
Either caller template should pass them explicitely, either `form_task_main_form.html.twig` should define default values explicitely in its beginning, to make maintenance of this template easier.

cedric-anne (approved):

orthagh (approved):

PR #14340, opened 2023-03-21
fix duplicated predefined actors

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #13601

Not 100% sure this is a good fix for the issue. The predefined user actor seems to be in the main $params array as well as $params['_predefined_fields'] so it is added twice.


cconard96:
This PR is indeed wrong.

PR #14345, opened 2023-03-21
Do not load duplicated authorizations

AdrienClairembault:
Users may have "obsolete" authorizations:

image

For example in the image above I already have a recursive authorization on the Root entity but I also have two distinct authorizations on sub entities.
These authorizations are useless as I can already see any data in entities below root but they are still loaded into GLPI's session, causing some displays issues in the entity selector:

image

To fix this I'm doing some safety checks before setting the session data, clearing obsolete authorizations as needed.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |

### DIFF
--- a/src/Entity.php
+++ b/src/Entity.php
@@ -65,6 +65,14 @@ class Entity extends CommonTreeDropdown
     const AUTO_ASSIGN_CATEGORY_HARDWARE  = 2;
 
     /**
+     * Possible relations between entities
+     */
+    public const RELATION_SELF      = 0;
+    public const RELATION_PARENT    = 1;
+    public const RELATION_CHILD     = 2;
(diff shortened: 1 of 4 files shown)


stonebuzz (approved):

AdrienClairembault:
Note: pending discussion with @orthagh to confirm this isn't some kind of feature that we want to keep.

orthagh:
> obsolete

redundant, no ?

> causing some displays issues in the entity selector:

can you precise which issues you encounter.

Not against the changes, but i'm not entirely certain that some people use (abuse) this feature to have shortcuts to some entities

PR #14352, opened 2023-03-21
add missing softwarelicense cache fields

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

SoftwareLicense is a CommonTreeDropdown class and there are two cache fields that are expected to be present for all those classes, but are missing here. A SQL error is thrown after purging a Software because of these missing fields.

### DIFF
--- a/install/migrations/update_10.0.6_to_10.0.7/softwarelicense.php
+++ b/install/migrations/update_10.0.6_to_10.0.7/softwarelicense.php
@@ -1,0 +1,47 @@
+<?php
+
+/**
+ * ---------------------------------------------------------------------
+ *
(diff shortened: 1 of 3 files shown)


cedric-anne (commented):
`Glpi\System\Diagnostic\DatabaseSchemaConsistencyChecker::getMissingFields()` class can probably be update to check that tables related to `CommonTreeDropdown` items have this field.

This would prevent such issue to reappear on another class in the future.

cedric-anne (approved):

trasher (approved):

PR #14378, opened 2023-03-23
Transferring tickets between entities

fabioneres:
Hello!
It would be interesting if GLPI had in its new versions the resource to create new types of business rules for tickets, such as rules for transferring tickets between entities. Example:

Criterion: technical group is IT
Action: Entity is IT

In this way, it would automate and make the transfer simpler, so that it would be done in an automated way, as normally whoever is going to pass a ticket to another entity does not need to know the information present in the current transfer screen, which was visibly designed for assets.


cconard96:
Suggestions should be made on the [UserEcho site](https://glpi.userecho.com/).
Pull requests are solely for code contributions.
Issues are for bug reports.
Community support can be found on the [Community forum](https://forum.glpi-project.org/)

PR #14404, opened 2023-03-28
Fix missing history entry when adding a user to a group

AdrienClairembault:
New `Group_User` relations are not shown in GLPI history.

These history entries are handled automatically by `CommonDBRelation::post_addItem()` and the `Group_User` class was overriding the function without calling the parent method.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !27336

### DIFF
--- a/src/Group_User.php
+++ b/src/Group_User.php
@@ -918,6 +918,8 @@ class Group_User extends CommonDBRelation
     {
         global $DB;
 
+        parent::post_purgeItem();
+
        // remove user from plannings
         $groups_id  = $this->fields['groups_id'];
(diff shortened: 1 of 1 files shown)


cedric-anne (changes requested):
OK for me, but could you add the same for `post_purgeItem` method? Parent method is empty for now, but it may change in the future.

cedric-anne (approved):

trasher (approved):

PR #14418, opened 2023-03-30
drop global var for uploads indexation

btry:
This change is required to allow a user to add files to a file field which is rendered with pre-uploaded files.

This use case may happen in Formcreator when a validator validates answers containing a file question. This PR is required for https://github.com/pluginsGLPI/formcreator/pull/3226

The goal is to not start the uploads index from zero, but counts the existing inputs related to a file upload field instead.

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #number

### DIFF
--- a/js/fileupload.js
+++ b/js/fileupload.js
@@ -130,8 +130,8 @@ var handleUploadedFile = function (files, files_data, input_name, container, edi
  * @param      {String}  input_name    Name of generated input hidden (default filename)
  * @param      {Object}  container     The fileinfo container
  */
-var fileindex = 0;
 var displayUploadedFile = function(file, tag, editor, input_name, filecontainer) {
+    var fileindex = $('input[name^="_'+input_name+'["]').length;
     var ext = file.name.split('.').pop();
 
     var p = $('<p></p>')
@@ -169,8 +169,6 @@ var displayUploadedFile = function(file, tag, editor, input_name, filecontainer)
     $('<span class="ti ti-circle-x pointer"></span>').click(function() {
         deleteImagePasted(elementsIdToRemove, tag.tag, editor);
     }).appendTo(p);
-
-    fileindex++;
 };
 
 /**


js/fileupload.js
```diff
@@ -130,8 +130,8 @@ var handleUploadedFile = function (files, files_data, input_name, container, edi
  * @param      {String}  input_name    Name of generated input hidden (default filename)
  * @param      {Object}  container     The fileinfo container
  */
-var fileindex = 0;
 var displayUploadedFile = function(file, tag, editor, input_name, filecontainer) {
+    var fileindex = $('input[name^="_'+input_name+'["]').length;
```
cedric-anne, line 134: I may be wrong, but I think the `[` that is inside the value has to be escaped.
```suggestion
    var fileindex = $('input[name^="_'+input_name+'\["]').length;
```

btry: I did the following test 

I have a file upload field with 2 filles in the DOM.
I duplicated an input and edited it to have it not indexed

image

Then I tried to count the inputs
image

Without the opening bracket, i find 3 inputs, as expected. If I add the bracked, with and without escaping, the count is 2, as expected. I think this is not necessary to escape it.

cedric-anne (approved):
I confirm it fixes an issue.

It is really easy to reproduce:
 - open ticket creation form,
 - write some contents and paste an image inside it,
 - change category,
 - add some contents and paste another image inside it,
 - submit the form.
-> First image remains in base64 and no document is created for it.

DOM before applying this fix (0 index is used twice, so first image inputs are overriden):
image

DOM after applying this fix (there are correct indexes):
image

Tested on both Chrome and Firefox.

js/fileupload.js
cedric-anne: Indeed, I guess it is not required as value is inside quotes.

trasher (approved):

PR #14511, opened 2023-04-12
Improve plugins init performances

AdrienClairembault:
I've noticed a while ago that the performances cost of plugins seemed to be a bit too high.
I can't tell if it was an exponential scale or not but the more plugins you had the slower GLPI would get.

I've decided to take a quick look at it today and managed to cut the plugins initialization process time by half.
There may be more optimizations to be found but it's a good first step.

This was done by moving some repetitive code that would be run multiple time with the same input into class properties that are only loaded once.

To benchmark the performances on my side, I've enabled about 20 plugins and added the code below to time the `plugin::init()` function, which is called once per page or ajax request and will load all the plugins.
```diff
diff --git a/src/Plugin.php b/src/Plugin.php
index 5ee1f541b1..d2592d808c 100644
--- a/src/Plugin.php
+++ b/src/Plugin.php
@@ -229,6 +229,7 @@ class Plugin extends CommonDBTM
     public function init(bool $load_plugins = false, array $excluded_plugins = [])
     {
         global $DB;
+        $start_time = microtime(true);

         self::$plugins_state_checked = false;
         self::$activated_plugins = [];
@@ -270,6 +271,9 @@ class Plugin extends CommonDBTM
             // For plugins which require action after all plugin init
             Plugin::doHook(Hooks::POST_INIT);
         }
+
+        $end_time = microtime(true);
+        Toolbox::logDebug($end_time - $start_time);
     }

```

Then I went on the plugin page (could be any other page), pressed F5 and saved the logs.
Each logs seems to contain 4 entries, which would mean we have 1 page being loaded + 3 ajax requests.

First, here are the executions time in `10.0/bugfixes`:

image

The average time seems to be 0.1s second which is quite big as the total load time of the page is about 0.27ms (so plugins init account for almost half my execution time in a localhost environnement). 

And here are the exections time after applying the optimizations found in this PR:

image

We are down to around 0.05s which is a great improvement for instances with a lot of plugins, especially for things like loading dashboards where there may be a 20+ ajax requests at the same time,.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |

### DIFF
--- a/src/Plugin.php
+++ b/src/Plugin.php
@@ -124,6 +124,13 @@ class Plugin extends CommonDBTM
      */
     private static $loaded_plugins = [];
 
+    /**
+     * Store additional infos for each plugins
+     *
+     * @var array
+     */
+    private array $plugins_informations;
+
     public static function getTypeName($nb = 0)
     {
         return _n('Plugin', 'Plugins', $nb);
@@ -458,10 +465,9 @@ class Plugin extends CommonDBTM
      */
     public function checkStates($scan_inactive_and_new_plugins = false, array $excluded_plugins = [])
     {
-
         $directories = [];
 
-       // Add known plugins to the check list
+        // Add known plugins to the check list
         $condition = $scan_inactive_and_new_plugins ? [] : ['state' => self::ACTIVATED];
         $known_plugins = $this->find($condition);
         foreach ($known_plugins as $plugin) {
@@ -469,7 +475,7 @@ class Plugin extends CommonDBTM
         }
 
         if ($scan_inactive_and_new_plugins) {
-           // Add found directories to the check list
+            // Add found directories to the check list
             foreach (PLUGINS_DIRECTORIES as $plugins_directory) {
                 if (!is_dir($plugins_directory)) {
                     continue;
@@ -486,10 +492,28 @@ class Plugin extends CommonDBTM
             }
         }
 
-       // Prevent duplicated checks
+        // Prevent duplicated checks
         $directories = array_unique($directories);
 
-       // Check all directories from the checklist
+        // Build plugin info data once before iterating to save performances
+        $plugins_informations = [];
+        $plugins_directories = new DirectoryIterator(GLPI_ROOT . '/plugins');
+        foreach ($plugins_directories as $plugin_directory) {
+            $plugin_name = $plugin_directory->getFilename();
+
+            if (
+                in_array($plugin_name, ['.svn', '.', '..'])
+                || !is_dir($plugin_directory->getRealPath())
+            ) {
+
+                continue;
+            }
+
+            $info = $this->getInformationsFromDirectory($plugin_name);
+            $plugins_informations[$plugin_name] = $info;
+        }
+        $this->plugins_informations = $plugins_informations;
+        // Check all directories from the checklist
         foreach ($directories as $directory) {
             if (in_array($directory, $excluded_plugins)) {
                 continue;
@@ -513,9 +537,8 @@ class Plugin extends CommonDBTM
 
         $plugin = new self();
 
-        $informations = $this->getInformationsFromDirectory($plugin_key);
+        $informations = $this->plugins_informations[$plugin_key] ?? [];
         $new_specs    = $this->getNewInfoAndDirBasedOnOldName($plugin_key);
-
         $is_already_known = $plugin->getFromDBByCrit(['directory' => $plugin_key]);
         $is_loadable      = !empty($informations);
         $is_replaced      = $new_specs !== null;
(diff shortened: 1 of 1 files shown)


src/Plugin.php
```diff
@@ -486,10 +492,28 @@ public function checkStates($scan_inactive_and_new_plugins = false, array $exclu
...
         }
 
-       // Prevent duplicated checks
+        // Prevent duplicated checks
         $directories = array_unique($directories);
 
-       // Check all directories from the checklist
+        // Build plugin info data once before iterating to save performances
+        $plugins_informations = [];
+        $plugins_directories = new DirectoryIterator(GLPI_ROOT . '/plugins');
```
cconard96, line 500: Missing support for plugins added from the Marketplace?

cedric-anne:
Do you have Opcache extension installed/activated?

src/Plugin.php
AdrienClairembault: I'm not sure, the original code was only targeting /plugins too so maybe its on purpose ?

AdrienClairembault:
> Do you have Opcache extension installed/activated?

image

AdrienClairembault:
Tests failed because I didn't noticed that `Plugin::checkPluginState()` was public and call be called independently.
Latest commit should fix the issue.

AdrienClairembault:
I've added marketplace support in the latest commit but it would be nice to understand how the original code was able to work with only `/plugins`

cconard96:
> I've added marketplace support in the latest commit but it would be nice to understand how the original code was able to work with only `/plugins` (or maybe it didn't work).

Was the original code from `getNewInfoAndDirBasedOnOldName`?
I think it predates the Marketplace and I'm not entirely sure how replacements are handled for Marketplace plugins (if at all).

src/Plugin.php
```diff
@@ -124,6 +124,13 @@ class Plugin extends CommonDBTM
      */
     private static $loaded_plugins = [];
 
+    /**
+     * Store additional infos for each plugins
+     *
+     * @var array
+     */
+    private ?array $plugins_informations = null;
```
trasher, line 132:
```suggestion
    private ?array $plugins_information = null;
```

"information" is uncountable

AdrienClairembault: Done.

AdrienClairembault:
Web tests are failing due to some kind of missing plugin warning, need to investigate.

cconard96 (approved):

trasher (approved):

AdrienClairembault:
`getInformationsFromDirectory` could be renamed too but not in 10/bf since it's a public method.

trasher (approved):

cedric-anne (changes requested):

cedric-anne:
`Plugin::getInformationsFromDirectory()` should probably use information already fetched by your new method.

AdrienClairembault:
> `Plugin::getInformationsFromDirectory()` should probably use information already fetched by your new method.

The new method use `getInformationsFromDirectory` so we might end up with an infinite reference loop ;)

Rom1-B (approved):

src/Plugin.php
```diff
@@ -500,6 +506,40 @@ public function checkStates($scan_inactive_and_new_plugins = false, array $exclu
         self::$plugins_state_checked = true;
     }
 
+    /**
+     * Build plugin info data once to save performances
```
cedric-anne, line 510:
```suggestion
     * Get information for all plugins found on filesystem.
```

src/Plugin.php
```diff
@@ -500,6 +506,40 @@ public function checkStates($scan_inactive_and_new_plugins = false, array $exclu
...
+     *
+     * @return array
+     */
+    private function getPluginInformation(): array
```
cedric-anne, line 514:
```suggestion
    private function getPluginsInformation(): array
```

src/Plugin.php
```diff
@@ -510,14 +550,12 @@ public function checkStates($scan_inactive_and_new_plugins = false, array $exclu
      */
     public function checkPluginState($plugin_key)
     {
-
         $plugin = new self();
 
-        $informations = $this->getInformationsFromDirectory($plugin_key);
+        $information = $this->getPluginInformation()[$plugin_key];
```
cedric-anne, line 555: A plugin can be present in DB but not on filesystem.

```suggestion
        $information = $this->getPluginsInformation()[$plugin_key] ?? [];
```

src/Plugin.php
```diff
@@ -728,27 +766,16 @@ public function checkPluginState($plugin_key)
...
-            if (
-                in_array($plugin_directory->getFilename(), ['.svn', '.', '..'])
-                || !is_dir($plugin_directory->getRealPath())
-            ) {
-                continue;
-            }
-
-            $informations = $this->getInformationsFromDirectory($plugin_directory->getFilename());
-            if (array_key_exists('oldname', $informations) && $informations['oldname'] === $oldname) {
+        foreach ($this->plugins_information as $plugin_name => $information) {
```
cedric-anne, line 773:
```suggestion
        foreach ($this->getPluginsInformation() as $plugin_name => $information) {
```

orthagh (approved):

cedric-anne (approved):

PR #14526, opened 2023-04-14
added hook for dashboard filter front

bessantoy:
Hello, 

I am a developer at IT Gouvernance and we develop several plugins on GLPI. There is a problem that we regularly encountered, in fact, when we wanted to add a filter using the DASHBOARD_FILTERS hook, we also had to modify the Filter.php file in order to write a static function there that calls Filter::displayList so that the dropdown can be displayed on the dashboard. To overcome this problem I added a new Hook and modified the Grid:: getFilterHtml function so that we can provide the front of the filter directly from the plugin. We tested this feature and it works.

Regards,

BESSIERES Anthony from IT Gouvernance


bessantoy:
Hello, 
I allow myself to remind you about this pull request, please consider it.

orthagh (commented):
I agree with hook addition proposal.
But i found the parameter and the return odd, the core function ask the html of a specific filter_id, and the hook return all the filter and it's required to check the specific filter exist in the hook.

If i would have develop myself the hook, i would have set the same input/output as the above core call

So 2 parameters:
- string $filter id // name of the filter
- mixed (string/array) $filter_values // value(s) send when a filter is updated in the grid

And the return should be directly the html of the filter provided by the plugin.

The hook name should be singural, i suggest `dashboard_html_filter`

So in summary, the core should not care about testing the filter_id (new lines added from 1061 to 1062) as the current function ask for the html of a precise filter_id.

**I let other core developers give also their opinion, wait for it before doing any changes.**

bessantoy:
I updated my solution so now the core doesn't take care about testing filter_id as the current function ask for the html of a precise filter_id. Waiting for your feedbacks.

Regards

orthagh:
The last commit seems wrong :)

bessantoy:
I reviewed my code following your advice, I implemented an abstract class representing a filter, all native filters extend this class and are in the \Dashboard\Filter folder, I modified the Filter.php file, the getAll() function now returns the classes representing the native filters as well as those retrieved by the DASHBOARD_FILTERS hook. Filter.php still contains the functions getForDashboard() and addForDashboard() which I have not modified
I also modified the Grid.php file to consider the new filter representation.

We can now add new filters in our plugin using the DASHBOARD_FILTERS hook like : 
 
`$PLUGIN_HOOKS[Hooks::DASHBOARD_FILTERS]['exemple'] = [
    MyFirstFilter::class,
    MySecondFilter::class
];`

and

class MyFirstFilter extends Dashboard\Filters\AbstractFilter
{
    public function getName(): string
    {
        return __("My first filter");
    }
    
    ...
}

Waiting for your feedbacks.

Regards,

Anthony from IT Gouvernance.

bessantoy:
I cleaned my dirty commits and pushed one commit containing all my modifications

src/Dashboard/Filters/AbstractFilter.php
```diff
@@ -0,0 +1,159 @@
...
+
+      JAVASCRIPT;
+      $field .= Html::scriptBlock($js);
+
+      return self::field($fieldname, $field, $label, $value !== null);
+  }
+
+
+   
+}
```
orthagh, line 159:
```suggestion
}

```

missing EOF

src/Plugin/Hooks.php
```diff
@@ -143,6 +143,7 @@ class Hooks
     // Dashboard hooks
     const DASHBOARD_CARDS   = 'dashboard_cards';
     const DASHBOARD_FILTERS = 'dashboard_filters';
+    const DASHBOARD_HTML_FILTER = 'dashboard_html_filter';
```
orthagh, line 146:
```suggestion  
```

src/Dashboard/Filters/DatesFilter.php
```diff
@@ -0,0 +1,106 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ * ---------------------------------------------------------------------
+ */
+
+namespace Glpi\Dashboard\Filters;
+
+use Glpi\Dashboard\Filters\AbstractFilter;
```
orthagh, line 38: in the same namespace, an import is not required (you have the same thing in all below files, please fix also)

src/Dashboard/Filter.php
```diff
@@ -34,19 +34,18 @@
...
+use Glpi\Dashboard\Filters\DatesFilter;
+use Glpi\Dashboard\Filters\ItilCategoryFilter;
+use Glpi\Dashboard\Filters\LocationFilter;
+use Glpi\Dashboard\Filters\ManufacturerFilter;
+use Glpi\Dashboard\Filters\RequestTypeFilter;
+use Glpi\Dashboard\Filters\StateFilter;
+use Glpi\Dashboard\Filters\TicketTypeFilter;
+use Glpi\Dashboard\Filters\GroupTechFilter;
+use Glpi\Dashboard\Filters\UserTechFilter;
+use Glpi\Dashboard\Filters\DatesModFilter;
```
orthagh, line 48:
```suggestion
use Glpi\Dashboard\Filters\{
    DatesFilter,
    ItilCategoryFilter,
    LocationFilter,
    ManufacturerFilter,
    RequestTypeFilter,
    StateFilter,
    TicketTypeFilter,
    GroupTechFilter,
    UserTechFilter,
    DatesModFilter
};
```

Not a mandatory change, but fyi, you can have a shorter syntax for importing several classes of the same namespace)

orthagh (changes requested):
Apart some sugar syntax remark and a miss, seems very good !

src/Plugin/Hooks.php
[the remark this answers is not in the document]
bessantoy: well, it's not

bessantoy:
It seems clean now

AdrienClairembault:
I think you also need to take these two functions into account : `getFiltersCriteria` and `getSearchFiltersCriteria`.
Without this two function I don't think any plugin filter can work properly.

The  `getFiltersCriteria` function explain to GLPI how to apply the filter (by adding the needed content to the SQL query).

image

On the other hand, `getSearchFiltersCriteria` is the function used to display the corrects search criteria that will be used when an user click on a widget and is taken to the search page.

image

There is also `getAllDasboardCards` but I'm not sure how these one are used, maybe @orthagh can help us here

image

bessantoy:
So I considered the functions getSearchFiltersCriteria and getFiltersCriteria for the new representation of filters. Each Filter must now implement 2 new functions returning their crtieria and their searchcriteria

AdrienClairembault (changes requested):
Looks good overall, I would just avoid using `$DB = DBConnection::getReadConnection();` in each filter as `DBConnection::getReadConnection()` can be a bit slow when using sql replication (see https://github.com/glpi-project/glpi/pull/12965).

Maybe initializing `$DB` once and then passing it as a parameter would be better for performances.

bessantoy:
i fixed DatesFilter and DatesModFilter classes that i had badly implemented, i also fixed a syntax error.
I tested each filter one by one and everything works.

orthagh:
CI says you still have some issues to fix :)

src/Dashboard/Filters/ItilCategoryFilter.php
```diff
@@ -0,0 +1,123 @@
...
+ * ---------------------------------------------------------------------
+ */
+
+namespace Glpi\Dashboard\Filters;
+
+use Session;
+use ITILCategory;
+use DBmysql;
+
+class ItilcategoryFilter extends AbstractFilter
```
cedric-anne, line 42:
```suggestion
class ItilCategoryFilter extends AbstractFilter
```

orthagh (approved):

cconard96 (approved):

trasher (approved):
Seems good to me, I'll wait for @AdrienClairembault validation before merge since he follows this development and I did not

src/Dashboard/Filter.php
```diff
@@ -58,235 +60,39 @@ class Filter extends \CommonDBChild
...
+            GroupTechFilter::class,
+            UserTechFilter::class,
         ];
 
-        $more_filters = Plugin::doHookFunction(Hooks::DASHBOARD_FILTERS);
-        if (is_array($more_filters)) {
-            $filters = array_merge($filters, $more_filters);
+        if (isset($PLUGIN_HOOKS[Hooks::DASHBOARD_FILTERS])) {
+            foreach ($PLUGIN_HOOKS[Hooks::DASHBOARD_FILTERS] as $hook_filters) {
+                foreach ($hook_filters as $filter) {
```
cedric-anne, line 86: Please verify here that corresponding plugin is active. Sometimes, plugins are registering their hook without being active.

src/Dashboard/Filter.php
```diff
@@ -58,235 +60,39 @@ class Filter extends \CommonDBChild
...
+                    $more_filters[] = $filter;
+                }
+            }
         }
+        $filters = array_merge($filters, $more_filters);
 
         return $filters;
```
cedric-anne, line 85: Previously, filters were returned as key => value. Now they will only be a set of values. It may break some plugins, like `advanceddashboards`.

Impacted piece of code:
```
'filters'    => array_keys(Filter::getAll()),
```

src/Dashboard/Filter.php
```diff
@@ -58,235 +60,39 @@ class Filter extends \CommonDBChild
...
-    /**
-     * Get HTML for a dates range filter
-     *
-     * @param string|array $values init the input with these values, will be a string if empty values
-     * @param string $fieldname how is named the current date field
-     *                         (used to specify creation date or last update)
-     *
-     * @return string
-     */
-    public static function dates($values = "", string $fieldname = 'dates'): string
```
cedric-anne, line 97: Public methods should not be removed on a bugfixes version. At least, it should trigger a deprecation notice and maybe return a mocked value.

cedric-anne (changes requested):

src/Dashboard/Filter.php
bessantoy: So do you mean that i should change this : `$filters = [
            DatesFilter::class,
            ...
        ];` to `$filters = [
            DatesFilter::getId() => DatesFilter::class,
           ...
        ];`

cedric-anne: I do not know how it is supposed to be used. Maybe @AdrienClairembault could help here.

src/Dashboard/Filter.php
```diff
@@ -58,182 +62,198 @@ class Filter extends \CommonDBChild
...
+        );
+    }
+
+    /**
+     * @deprecated 10.0.0
+     */
+    public static function dates($values = "", string $fieldname = 'dates'): string
+    {
+        Toolbox::deprecated(__METHOD__ . ' is deprecated. Use ' . DatesFilter::class . ' instead.');
+        return null;
```
cedric-anne, line 155:
```suggestion
        return "";
```

cedric-anne, line 155: Would this work?

```suggestion
        return DatesFilter::getHtml($values, $fieldname);
```

bessantoy: I think

src/Dashboard/Grid.php
```diff
@@ -1246,7 +1263,7 @@ public function getAllDasboardCards($force = false): array
                     'State',
                     'Entity',
                     'Manufacturer',
-                    'Location',
+                    LocationFilter::getId(),
```
AdrienClairembault, line 1266:
```suggestion
                    'Location',
```
I think this one should be reverted, does not seem related to filters.

src/Dashboard/Grid.php
```diff
@@ -1408,17 +1425,17 @@ public function getAllDasboardCards($force = false): array
...
 
             foreach (
                 [
-                    'ITILCategory' => __("Top ticket's categories"),
+                    ItilCategoryFilter::getId() => __("Top ticket's categories"),
                     'Entity'       => __("Top ticket's entities"),
-                    'RequestType'  => __("Top ticket's request types"),
-                    'Location'     => __("Top ticket's locations"),
+                    RequestTypeFilter::getId()  => __("Top ticket's request types"),
+                    LocationFilter::getId()     => __("Top ticket's locations"),
```
AdrienClairembault, line 1438: Same for these 3, does not seem related to filters.

src/Dashboard/Provider.php
```diff
@@ -1625,376 +1627,39 @@ public static function formatMonthyearDates(string $monthyear): array
...
-                        'link'       => 'AND',
-                        'field'      => 8, // group tech
-                        'searchtype' => 'equals',
-                        'value'      => $groups_id
-                    ];
-                }
+        foreach ($filters as $filter) {
+            $filter_criteria = $filter::getSearchCriteria($DBread, $table, $apply_filters);
+            if (!empty($filter_criteria)) {
+                $s_criteria = array_merge($s_criteria, $filter_criteria);
```
AdrienClairembault, line 1641: Search criteria are not working because they is one unneeded array level:

image

Can be fixed like this:

```suggestion
                array_push($s_criteria, ...$filter_criteria);
```

After fix:

image

AdrienClairembault (changes requested):
There are also some inconsistencies with the the `getHtml` function.

In the abstract class, it does not have any parameters:

image

Most filters implements it with a single parameter:

image

The `Datefilter` implements it with two parameters :

image

The function seems to be called with only one parameter in the `Grid` class:

image

AdrienClairembault:
One little functional issue that may be difficult to deal with.

Changing a custom filter does not trigger a refresh of the displayed cards.

This is because we only refresh cards that are marked as "impacted" by the filter:

image

These special `filter-*` classes are set depending on the data shown below, which may be difficult to set for each plugins filters as you need to do it for each existing cards

image

I think we should either always reload all cards or do some kind of generics checks (like it's done when adding the sql criteria where we check if a specific field exist in the table).

Any thoughts on this @orthagh @cedric-anne ?

AdrienClairembault (approved):
Ok for me in glpi main

cedric-anne:
@bessantoy 

Could I push some commits to your branch, or would you prefer me to open a new PR based on a branch of mine, including your commits?
I want to push some minor changes on changelog, methods signatures, ... and it will probably faster for everyone if I push them directly.

bessantoy:
> @bessantoy
> 
> Could I push some commits to your branch, or would you prefer me to open a new PR based on a branch of mine, including your commits? I want to push some minor changes on changelog, methods signatures, ... and it will probably faster for everyone if I push them directly.

Of course you can

cedric-anne:
Replaced by #14717.

Thanks for the contribution.

PR #14552, opened 2023-04-18
Fix missing ticket actors in content templates

AdrienClairembault:
Tickets actors information are missing when using this task template as a predefined field:

image

image

After fix:

image

The fix was simply moving the `handleXXXXTemplateInput()` functions after the `updateActors()` call to make sure we are working with the latest data.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !27636

### DIFF
--- a/src/CommonITILObject.php
+++ b/src/CommonITILObject.php
@@ -1894,20 +1894,20 @@ abstract class CommonITILObject extends CommonDBTM
 
     public function post_updateItem($history = 1)
     {
-       // Handle "_tasktemplates_id" special input
-        $this->handleTaskTemplateInput();
-
-       // Handle "_itilfollowuptemplates_id" special input
-        $this->handleITILFollowupTemplateInput();
-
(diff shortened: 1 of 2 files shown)


cedric-anne (approved):

trasher (approved):

laurixia:
This happens to me but with the observer in a tracking template. The variable does not take the value of the observer when I create the ticket and go through the assignment rule to the ticket.

Follow Template:

1github

Rule:

2github


Example of the ticket generated with the empty variable

3github

cedric-anne:
@laurixia This seems to be another issue. Anyway, by commenting a ticket that was closed more than one year ago, many people will probably miss your comment. If you want someone to investigate on your issue, please open a new one, and give all required information.

PR #14570, opened 2023-04-20
Ensure rights have search options to allow historical logging

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #14481

Only some permission changes in profiles are being logged in the historical data. Rights must have a search option for their changes to be logged. This PR adds a test to ensure a search option exists for at least the basic rights that are displayed as a checkbox matrix. Most of this PR is moving the rights arrays from each `showForm*` method into a single, separate method so they would be known by the test.

### DIFF
--- a/src/Profile.php
+++ b/src/Profile.php
@@ -778,6 +778,356 @@ class Profile extends CommonDBTM
         return true;
     }
 
+    /**
+     * Get all rights to display for a specific form and interface.
+     *
+     * This is only used for GLPI core rights and not rights added by plugins.
+     *
+     * @param string $form The tab/form name
+     * @phpstan-param non-empty-string $form
+     * @param string $interface The interface name
+     * @phpstan-param 'central'|'helpdesk' $interface
+     * @return array
+     * @phpstan-return array<string, array{
+     *     rights?: array{}, itemtype?: class-string<CommonDBTM>, label: string, field: string, group: string, scope: string
+     * }>
+     * @internal BC not guaranteed. Only public so it can be used in tests to ensure search options are made for all rights.
+     */
+    public static function getRightsForForm(string $form, string $interface): array
+    {
+        // Helper function to streamline rights definition
+        /**
+         * @param class-string<CommonDBTM> $itemtype
+         * @param string|array $interface
+         * @param string $field
+         * @param array $options
+         * @return array
+         */
+        $fn_get_rights = static function (string $itemtype, $interface, string $field = null, array $options = []) {
+            $options = array_replace([
+                'group' => null,
+                'scope' => 'entity'
+            ], $options);
+            $rights = Profile::getRightsFor($itemtype, $interface);
+            if ($field === null) {
+                $field = $itemtype::$rightname;
+            }
+            return [
+                'rights' => $rights,
+                'label'  => $itemtype::getTypeName(Session::getPluralNumber()),
+                'field'  => $field,
+                'interface' => !is_array($interface) ? [$interface] : $interface,
+                'group' => $options['group'],
+                'scope' => $options['scope']
+            ];
+        };
+
+        static $all_rights = null;
+
+        if ($all_rights === null) {
+            $all_rights['tracking'] = [
+                $fn_get_rights(Ticket::class, 'helpdesk'),
+                $fn_get_rights(ITILFollowup::class, 'helpdesk'),
+                $fn_get_rights(TicketTask::class, 'helpdesk'),
+                $fn_get_rights(TicketValidation::class, 'helpdesk'),
+                [
+                    'itemtype'  => 'TicketTemplate',
+                    'label'     => _n('Template', 'Templates', Session::getPluralNumber()),
+                    'field'     => 'itiltemplate',
+                    'interface' => 'central',
+                    'group'     => 'itilobjects',
+                    'scope'     => 'entity'
+                ],
+                $fn_get_rights(PendingReason::class, 'central', null, ['group' => 'itilobjects']),
+                $fn_get_rights(Ticket::class, 'central', null, ['group' => 'tickets']),
+                $fn_get_rights(TicketCost::class, 'central', null, ['group' => 'tickets']),
+                $fn_get_rights(TicketRecurrent::class, 'central', null, ['group' => 'tickets']),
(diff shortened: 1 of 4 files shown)


src/Profile.php
```diff
@@ -778,6 +778,363 @@ public function showForm($ID, array $options = [])
...
+     * @phpstan-param non-empty-string $form
+     * @param string $interface The interface name
+     * @phpstan-param 'central'|'helpdesk' $interface
+     * @return array
+     * @phpstan-return array<string, array{
+     *     rights?: array{}, itemtype?: class-string<CommonDBTM>, label: string, field: string, group: string, scope: string
+     * }>
+     * @internal BC not guaranteed. Only public so it can be used in tests to ensure search options are made for all rights.
+     */
+    public static function getRightsForForm(string $form, string $interface): array
```
cedric-anne, line 796: Usage of this method could probably be simplified in order to not have to filter results.
```suggestion
    public static function getRightsSpecs(?string $group = null): array
```

`$group` could be something like '{$interface}/{$form}(/{$section})', for instance:
- `helpdesk/tracking`
- `helpdesk/tools`
- `central/assets`
- `central/tools/general`
- `central/tools/projects`

With this, usage in profile form would be simple as:
```php
$this->displayRightsChoiceMatrix(self::getRightsSpecs('central/tools/projects'), $matrix_options);
```
and usage from tests could be simple as 
:
```php
$rights = \Profile::getRightsSpecs(); // No $group param passed => get all possible rights
foreach ($rights as $right) {
    ...
}
```

cconard96: I'm not sure about `helpdesk/tracking` and `central/tools/general` style for getting rights instead of parameters. I don't think it really fits with anything else in the code. If anything, I think another parameter should be added for the section/group and do the filtering within the function itself and change the default group from `null` to `general`. Then, `null` would return all sections/groups for a specific form and interface.

I'd like to have other opinions on this though.

src/Profile.php
```diff
@@ -778,6 +778,363 @@ public function showForm($ID, array $options = [])
...
+                'interface' => !is_array($interface) ? [$interface] : $interface,
+                'group' => $options['group'],
+                'scope' => $options['scope']
+            ];
+        };
+
+        static $all_rights = null;
+
+        if ($all_rights === null) {
+            $all_rights['tracking'] = [
```
cedric-anne, line 828: IMHO, declaration could be easier to read if it was a 3 levels array, like this:
```php
$all_rights = [
    'central' => [
        'tracking' => [
            'tickets' => [
                $fn_get_rights(Ticket::class),
                $fn_get_rights(TicketCost::class),
                $fn_get_rights(TicketRecurrent::class),
            ],
            'followups_tasks' => [...],
            ...
        ]
    ]
]
```

orthagh: I agree with that.
Finding where adding a new right will be easier so

orthagh: method getRightsForForm should permits also to filter by the 2 first level, and interface filtering should be in a third level

orthagh:
One last question, why some right are declared directly on not with fn_get_rights ? wrong label ?
maybe we can take the opportunity to move/correct their respective classes ?

cconard96:
> One last question, why some right are declared directly on not with fn_get_rights ? wrong label ? maybe we can take the opportunity to move/correct their respective classes ?

Some of the calculated values provided by the helper function don't match what was already being used. For example, the Templates field in Assistance uses TicketTemplate. The calculated label would be "Ticket templates" even though it is being used to represent all ITIL templates. Using ITILTemplate as the class instead would have an error if `getTypeName` was called because it relies on being called from a child class (could be changed).

The Contract permission field uses the Contract class, but the label is `Contracts / Suppliers`.

cconard96:
Some of those cases could be fixed by making changes to the classes. I don't know how it would affect BC though.

cedric-anne:
> Some of those cases could be fixed by making changes to the classes. I don't know how it would affect BC though.

Maybe you could add the ability to pass the `label` option to the `fn_get_rights()` function in GLPI 10.0 to handle this, and then change labels directly in classes in GLPI 10.1 to remove this specific handling.

orthagh:
One thing we discussed with Cédric is the scope field, currently it's wrongly set for a bunch of types (agent, unmanaged, maybe some others). But maybe we should drop completely the thing. I don't see the point of knowing  a right apply to a entity or global itemtype.
This exists also in list of dropdowns, but it's better displayed (with an icon and an helper) and at least seems right to be displayed in the context

orthagh (approved):

src/Profile.php
```diff
@@ -778,6 +778,334 @@ public function showForm($ID, array $options = [])
...
+                        'general' => [
+                            $fn_get_rights(KnowbaseItem::class, 'helpdesk'),
+                            $fn_get_rights(ReservationItem::class, 'helpdesk'),
+                            $fn_get_rights(Reminder::class, 'helpdesk'),
+                            $fn_get_rights(RSSFeed::class, 'helpdesk'),
+                        ],
+                    ],
+                    'setup' => [
+                        'general' => [
+                            $fn_get_rights(Config::class, 'central', ['scope' => 'entity']),
```
cedric-anne, line 1082: This one should not be here.

```suggestion
```

cedric-anne (changes requested):

src/Profile.php
```diff
@@ -778,6 +778,334 @@ public function showForm($ID, array $options = [])
...
+                            $fn_get_rights(Change::class, 'central'),
+                            $fn_get_rights(ChangeValidation::class, 'central'),
+                            $fn_get_rights(RecurrentChange::class, 'central'),
+                        ],
+                    ],
+                    'tools' => [
+                        'general' => [
+                            $fn_get_rights(Reminder::class, 'central'),
+                            $fn_get_rights(RSSFeed::class, 'central'),
+                            $fn_get_rights(SavedSearch::class, 'central'),
```
cedric-anne, line 869: Should we add `Public` term? @orthagh 

Before:
image

After:
image

cconard96:
All labels you mentioned were labels fixed to match what they were before.

cedric-anne (approved):

PR #14579, opened 2023-04-23
select2 dropdownAutowidth workaround for webkit

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #13433

When using a browser running Webkit, the first time you open a select2 dropdown the auto width does not apply. This seems like a bug in the Webkit engine. This PR adds a workaround when it detects an AppleWebKit user agent to close and then reopen the dropdown when the user tries opening it the first time.

Since this only happens for WebKit, I had to test this using Cypress with the experimental Webkit support. You should be able to test this fix using this branch which includes this fix along with a Cypress + Webkit setup:
https://github.com/cconard96/glpi/tree/fix/13433

1. Start Cypress with `npm run test:e2e`
2. Select E2E Testing option
3. Select Webkit and then click "Start E2E Testing in WebKit"
4. Select the "test.cy.js" spec. After it is finished, you should be able to use GLPI manually to test the fix.
Since manual interactions seem to glitch out sometimes, the quickest/easiest way to test is to check the language dropdown.

### DIFF
--- a/js/common.js
+++ b/js/common.js
@@ -37,6 +37,21 @@
 
 var timeoutglobalvar;
 
+if (navigator.userAgent.indexOf('AppleWebKit') !== -1) {
+    // Workaround for select2 dropdownAutowidth not applying until the second time the dropdown is opened
+    // See: https://github.com/glpi-project/glpi/issues/13433 and https://github.com/select2/select2/issues/4678
+    const original_select2_fn = $.fn.select2;
+    $.fn.select2 = function (options) {
+        const result = original_select2_fn.apply(this, arguments);
+        if (typeof options === 'object') {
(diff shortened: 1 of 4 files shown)


cedric-anne (approved):

trasher (approved):

trasher:
> ok for me, the only suggestion i would is could we isolate these kind of fixes in a dedicated file (like webkit_fix.js)

ping @cconard96 ?

cedric-anne (approved):

orthagh (approved):

PR #14601, opened 2023-04-27
fix: missing category when change created from ticket

Rom1-B:
When creating a change from a ticket, the category of the original ticket was not automatically included in the new change.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !27779

### DIFF
--- a/src/CommonITILObject.php
+++ b/src/CommonITILObject.php
@@ -483,12 +483,21 @@ abstract class CommonITILObject extends CommonDBTM
         if (in_array($this->getType(), ['Change', 'Problem']) && $tickets_id) {
             $ticket->getFromDB($tickets_id);
 
-            $options['content']             = $ticket->fields['content'];
-            $options['name']                = $ticket->fields['name'];
-            $options['impact']              = $ticket->fields['impact'];
-            $options['urgency']             = $ticket->fields['urgency'];
-            $options['priority']            = $ticket->fields['priority'];
-            if (isset($options['_tickets_id']) && !isset($options['_saved']['itilcategories_id'])) {
+            foreach ([
+                'content',
+                'name',
+                'impact',
+                'urgency',
+                'priority',
+                'time_to_resolve',
+                'entities_id',
+            ] as $field) {
+                if (!isset($options['_saved'][$field])) {
+                    $options[$field] = $ticket->fields[$field];
+                }
+            }
+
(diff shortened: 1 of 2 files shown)


templates/components/form/link_existing_or_new.html.twig
```diff
@@ -74,7 +74,7 @@
          </div>
          <div class="col-auto ms-4">
             {% if create_link %}
-               {% set target_form_path = (target_itemtype|itemtype_form_path ~ "?_" ~ source_itemtype|itemtype_foreign_key ~ "=" ~ source_items_id) %}
+               {% set target_form_path = (target_itemtype|itemtype_form_path ~ "?" ~ source_itemtype|itemtype_foreign_key ~ "=" ~ source_items_id) %}
```
cconard96, line 77: When creating a ticket from a project task, the foreign key needs the underscore in front. That seems to be the only case, and must have been the first case I migrated. Perhaps the code in the Ticket class should be changed to handle both `projecttasks_id` and `_projecttasks_id` in addition to this change.

cedric-anne (changes requested):
As discussed IRL, all fields that are copied from initial Tickets must be predefined only on initial load, to ensure that they are not reset when form is reloaded after category change.

src/CommonITILObject.php
```diff
@@ -483,12 +483,22 @@ protected function setPredefinedFields(ITILTemplate $tt, array &$options, array
         if (in_array($this->getType(), ['Change', 'Problem']) && $tickets_id) {
             $ticket->getFromDB($tickets_id);
 
-            $options['content']             = $ticket->fields['content'];
-            $options['name']                = $ticket->fields['name'];
-            $options['impact']              = $ticket->fields['impact'];
-            $options['urgency']             = $ticket->fields['urgency'];
-            $options['priority']            = $ticket->fields['priority'];
-            if (isset($options['tickets_id'])) {
+            $fields = [
```
orthagh, line 486:
```suggestion
            // copy fields from original ticket, only when fields are not already set by the user (contained in _saved array)
            $fields = [
```

add a comment to clarify the feature, like seen together, it was hard without explanation to find how this works

trasher (approved):

PR #14646, opened 2023-05-03
fix finding form button to lock

btry:
Logic error when searching for form button to lock.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | none

### DIFF
--- a/js/common.js
+++ b/js/common.js
@@ -1414,7 +1414,7 @@ function blockFormSubmit(form, e) {
 
     // if submitter is not a button, find the first submit button with add or update as the name
     if (submitter === null || !submitter.is('button')) {
-        submitter = submitter.find('button[name="add"]:first, button[name="update"]:first');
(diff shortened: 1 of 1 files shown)


js/common.js
```diff
@@ -1414,7 +1414,7 @@ function blockFormSubmit(form, e) {
 
     // if submitter is not a button, find the first submit button with add or update as the name
     if (submitter === null || !submitter.is('button')) {
-        submitter = submitter.find('button[name="add"]:first, button[name="update"]:first');
+        submitter = form.find('button[name="add"]:first, button[name="update"]:first');
```
btry, line 1417: if submitter is null, as stated in the enclosing if () assertion, then this line fails.

cconard96 (approved):

trasher (approved):

cedric-anne (approved):

PR #14648, opened 2023-05-03
fix: The last update user not updated with anonymous users

Rom1-B:
When an anonymous user answers on a ticket the field corresponding to the last author of update (users_id_lastupdater) does not change and keeps the value of the last GLPI user, which is false.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !27829


Rom1-B:
@cedric-anne,
Is it when a user updates an item? It doesn't seem to me that an anonymous user can edit?

orthagh (changes requested):
edit: bad review, i remove my comment

PR #14675, opened 2023-05-09
fix deprecation warning (found in 10.0.7)

btry:
When viewing a list of tickets, debug mode enabled, I may get several errors:

```PHP Deprecated function (8192): strlen(): Passing null to parameter #1 ($string) of type string is deprecated in /var/www/glpi/src/DbUtils.php at line 1580```

I don't have access to backtrace, but the fix seems obvious.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #number

### DIFF
--- a/src/DbUtils.php
+++ b/src/DbUtils.php
@@ -1606,10 +1606,10 @@ final class DbUtils
             $id_visible = $_SESSION["glpiis_ids_visible"];
         }
 
-        if ($realname !== null && strlen($realname) > 0) {
+        if (strlen($realname ?? '') > 0) {
             $formatted = $realname;
 
-            if ($firstname !== null && strlen($firstname) > 0) {
+            if (strlen($firstname ?? '') > 0) {
                 if ($order == User::FIRSTNAME_BEFORE) {
                     $formatted = $firstname . " " . $formatted;
                 } else {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/DbUtils.php [hunk 1]
```diff
@@ -1609,7 +1609,7 @@ public function formatUserName($ID, $login, $realname, $firstname, $link = 1, $c
         if ($realname !== null && strlen($realname) > 0) {
             $formatted = $realname;
 
-            if (strlen($firstname) > 0) {
+            if ($firstname !== null && strlen($firstname) > 0) {
```
trasher, line 1612: Maybe more simple?
```suggestion
            if (strlen($firstname ?? '') > 0) {
```

cedric-anne (approved):

src/DbUtils.php
btry: Why not, but I used the same style as L1609. Then I think I should change both to be the same. OK for you, @trasher  ?

src/DbUtils.php
```diff
@@ -1609,7 +1609,7 @@ public function formatUserName($ID, $login, $realname, $firstname, $link = 1, $c
         if ($realname !== null && strlen($realname) > 0) {
```
btry, line 1609:
```suggestion
        if (strlen($realname ?? '') > 0) {
```

cedric-anne (approved):

trasher (approved):

PR #14678, opened 2023-05-09
Fix fields ambiguity in `CommonDBTM::canUnrecurs()`

kabassanov:
…eld is present in the secondary table

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #number

### DIFF
--- a/src/CommonDBTM.php
+++ b/src/CommonDBTM.php
@@ -2550,7 +2550,7 @@ class CommonDBTM extends CommonGLPI
                                 $items_id_field = reset($items_id_matches);
                             }
                             $or_criteria[] = [
-                                $itemtype_field => $this->getType(),
+                                $tablename . "." . $itemtype_field => $this->getType(),
                                 $tablename . "." . $items_id_field => $this->getID(),
                             ];
                         } else {
--- a/src/CommonDBTM.php
+++ b/src/CommonDBTM.php
@@ -2551,7 +2551,7 @@ class CommonDBTM extends CommonGLPI
                             }
                             $or_criteria[] = [
                                 $itemtype_field => $this->getType(),
-                                $items_id_field => $this->getID(),
+                                $tablename . "." . $items_id_field => $this->getID(),
                             ];
                         } else {
                             // Relation based on single foreign key
--- a/src/CommonDBTM.php
+++ b/src/CommonDBTM.php
@@ -2556,7 +2556,7 @@ class CommonDBTM extends CommonGLPI
                         } else {
                             // Relation based on single foreign key
                             $or_criteria[] = [
-                                $field => $this->getID(),
+                                $tablename.".".$field => $this->getID(),
                             ];
                         }
                     }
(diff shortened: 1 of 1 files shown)


src/CommonDBTM.php
```diff
@@ -2556,7 +2556,7 @@ public function canUnrecurs()
                         } else {
                             // Relation based on single foreign key
                             $or_criteria[] = [
-                                $field => $this->getID(),
+                                $tablename.".".$field => $this->getID(),
```
cconard96, line 2559:
```suggestion
                                $tablename . "." . $field => $this->getID(),
```

cconard96:
Is this a fix for a specific bug you noticed?

kabassanov:
> Is this a fix for a specific bug you noticed?

Yes it is. There was an error while displaying group form:

[2023-05-09 12:14:58] glpisqllog.ERROR: DBmysql::query() in /usr/share/glpi/src/DBmysql.php line 370
  *** MySQL query error:
  SQL: SELECT COUNT(*) AS cpt FROM `glpi_problemtasks`, `glpi_tasktemplates` WHERE (((`groups_id_tech` = '1234'))) AND `glpi_problemtasks`.`tasktemplates_id` = `glpi_tasktemplates`.`id` AND  NOT (`glpi_tasktemplates`.`entities_id` IN ('
10', '19'))
  Error: Column 'groups_id_tech' in where clause is ambiguous

By the way there are some other pieces of code with the same "Relation based on single foreign key". Probably need to change them also.

cconard96:
What group form?

cconard96:
Issue occurs on the Group item form when the Child Entities checkbox is checked. Error is triggered during the `canUnrecurs` check.

trasher (approved):

cedric-anne (changes requested):
Seems OK, but same change should be done on lines 2553 and 2554.

cconard96:
> Why not, but I'm not sure how it works exactly. I supposed that sending the item type would somehow define the table to use... Isn't it true?

No. This is just translated directly to a SQL query.
```php
$or_criteria[] = [
    $itemtype_field => $this->getType(),
    $items_id_field => $this->getID(),
];
```
becomes something like:
OR (`itemtype` = 'Computer' AND `items_id` = 32)

No guesses are made about which table the fields come from.

src/CommonDBTM.php
```diff
@@ -2551,12 +2551,12 @@ public function canUnrecurs()
                             }
                             $or_criteria[] = [
                                 $itemtype_field => $this->getType(),
```
trasher, line 2553:
```suggestion
                                $tablename . "." . $itemtype_field => $this->getType(),
```

orthagh (approved):

cedric-anne (approved):

orthagh (approved):

PR #14708, opened 2023-05-15
fix(Computer): fix antivirus not propagate from template

stonebuzz:
```ComputerAntivirus``` item (attached to template) are no longer propagate when Computer is added from template.

This PR fix this.

| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !27980

### DIFF
--- a/src/Computer.php
+++ b/src/Computer.php
@@ -74,7 +74,8 @@ class Computer extends CommonDBTM
             Computer_Item::class,
             Notepad::class,
             KnowbaseItem_Item::class,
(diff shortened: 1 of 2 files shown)


stonebuzz:
> As far as I remember, we already have tests on computer cloning. If I'm right, that case should be added.

indeed, TU added for this case

trasher (approved):

PR #14718, opened 2023-05-16
fix(CommonITILObject): display requester instead of users id recipient

stonebuzz:
Actually, when a user create a ticket for another (see example "glpi" create a ticket for "tech") 

image

Display author attached to main description is related to ```users_id_recipient```

some users are confused about this display (one to be honest)

In GLPI version 9.5 we have a dedicated method to display ```requester``` if

- there is only one requester
- it is not an anonymous user

```php

      $user = new User();
      $display_requester = false;
      $requesters = $this->getUsers(CommonITILActor::REQUESTER);
      if (count($requesters) === 1) {
         $requester = reset($requesters);
         if ($requester['users_id'] > 0) {
            // Display requester identity only if there is only one requester
            // and only if it is not an anonymous user
            $display_requester = $user->getFromDB($requester['users_id']);
         }
      }
```

This PR try to resettle this behavior from twig template

I just noticed something on 9.5

If there are several requesters the identity becomes "Requesters" and the image is the default one

image

should I reproduce the same behaviour on this part? it seems to me heavy to carry on in ```picture.html.twig```

| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !27988

### DIFF
--- a/templates/components/itilobject/timeline/main_description.html.twig
+++ b/templates/components/itilobject/timeline/main_description.html.twig
@@ -31,7 +31,23 @@
  # ---------------------------------------------------------------------
  #}
 
-{% set users_id = (item.fields['users_id_recipient'] > 0 ? item.fields['users_id_recipient'] : 0) %}
+{# Compute users_id (requester) to display identity / thumbnail if #}
+{# - there is only one requester #}
+{# - it is not an anonymous user #}
+{# else diaply a blank thumbnail without identity #}
+
+{% set array_user = item.getUsers(constant('CommonITILActor::REQUESTER')) %}
+{% if array_user|length == 1 and array_user[0].users_id > 0 %}
+   {% set display_requester = get_item('User', array_user[0].users_id) is not null ? 1 : 0 %}
+   {% if display_requester %}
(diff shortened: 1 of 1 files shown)


trasher:
I'm a bit lost in all this but proposal sounds correct.

About the multiple requesters question, what is displayed with current PR if there are several requesters?

cedric-anne:
I thought we decided in GLPI 10.0 to display the real author here instead of the requester, so I am not sure we should put back the GLPI 9.5 behaviour.

orthagh:
To add to the @cedric-anne comment

> In GLPI version 9.5 we have a dedicated method to display requester

Yes and we moved from this in 10.0.0 because it's wrong.
We can have multiple requester but only one author, and the current behavior let people know a technician open the ticket (and not the requester)

> one to be honest

If you can't convince him this is a bad behavior and we wanted to move on, can you provide a patch only for him ?

cedric-anne:
Also, this information can be updated by changing the author in edit mode:
image

PR #14731, opened 2023-05-17
feat(inventory): handle several IPV4 or IPV6

stonebuzz:
Add capability to manage several IPV4 and IPV6 from one ```NetworkPort```

This particularity has been seen from ```SCCM``` plugin 

see this thread -> https://github.com/pluginsGLPI/sccm/pull/98#issuecomment-1548021772

image

which generates the following XML:

```xml
<NETWORKS>
    <IPADDRESS>169.254.41.148</IPADDRESS>
    <IPADDRESS>192.168.0.225</IPADDRESS>
    <IPADDRESS>10.50.78.89</IPADDRESS>
    <IPADDRESS6>fe80::c0b2:cf62:cba8:8d8</IPADDRESS6>
    <DESCRIPTION>Intel(R) Ethernet Connection (7) I219-V</DESCRIPTION>
    <IPMASK>255.255.255.0, 255.255.255.0, 255.255.255.128, 64</IPMASK>
    <IPDHCP/>
    <IPGATEWAY>10.50.78.19</IPGATEWAY>
    <MACADDR>04:92:26:D7:AC:51</MACADDR>
</NETWORKS>

<NETWORKS>
    <IPADDRESS>192.168.1.133</IPADDRESS>
    <IPADDRESS6>fe80::9bbb:c6ae:3cd5:d9fb</IPADDRESS6>
    <IPADDRESS6>2a0c:5a80:140d:4400:c480:da92:e839:ec9d</IPADDRESS6>
    <IPADDRESS6>2a0c:5a80:140d:4400:70f7:45d7:71fd:4d85</IPADDRESS6>
    <IPADDRESS6>2a0c:5a80:140d:4400:c3a:ccfe:ecdc:42b4</IPADDRESS6>
    <IPADDRESS6>2a0c:5a80:140d:4400:89d2:18f3:d43:2dfb</IPADDRESS6>
    <DESCRIPTION>Intel(R) Ethernet Connection (7) I219-LM</DESCRIPTION>
    <IPMASK>255.255.255.0, 64, 128, 128, 128, 64</IPMASK>
    <IPDHCP>192.168.1.1</IPDHCP>
    <IPGATEWAY>192.168.1.1, fe80::1</IPGATEWAY>
    <MACADDR>9C:7B:EF:B3:50:96</MACADDR>
</NETWORKS>
```

Code developed with great care by @tolemac 

I just added the unit tests

Need : https://github.com/glpi-project/inventory_format/pull/111

Best regards

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |

### DIFF
--- a/src/Inventory/Asset/NetworkCard.php
+++ b/src/Inventory/Asset/NetworkCard.php
@@ -197,27 +197,45 @@ class NetworkCard extends Device
                 }
 
                 if (isset($this->ports[$portkey])) {
-                    if (property_exists($val_port, 'ip') && $val_port->ip != '') {
-                        if (!in_array($val_port->ip, $this->ports[$portkey]->ipaddress)) {
-                             $this->ports[$portkey]->ipaddress[] = $val_port->ip;
+                    if (!property_exists($this->ports[$portkey], "ipaddress") || !is_array($this->ports[$portkey]->ipaddress)) {
+                        $this->ports[$portkey]->ipaddress = [];
+                    }
+                    if (property_exists($val_port, "ip")) {
+                        if (is_array($val_port->ip)) {
+                            $this->ports[$portkey]->ipaddress = array_merge($this->ports[$portkey]->ipaddress, $val_port->ip);
+                        } else {
+                            array_push($this->ports[$portkey]->ipaddress, $val_port->ip);
                         }
                     }
-                    if (property_exists($val_port, 'ipaddress6') && $val_port->ipaddress6 != '') {
-                        if (!in_array($val_port->ipaddress6, $this->ports[$portkey]->ipaddress)) {
-                            $this->ports[$portkey]->ipaddress[] = $val_port->ipaddress6;
+                    if (property_exists($val_port, 'ipaddress6')) {
+                        if (is_array($val_port->ipaddress6)) {
+                            $this->ports[$portkey]->ipaddress = array_merge($this->ports[$portkey]->ipaddress, $val_port->ipaddress6);
+                        } else {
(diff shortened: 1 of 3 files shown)


g-bougard (approved):

g-bougard:
Just as a side note, actually GLPI-Agent is indeed providing one network interface by found ip and there's only one ipv4 or one ipv6 by network interface. So the list notion is created at the netowrk interface level by the agent. Also each interface also includes ipmask or ipmaskv6 and ipsubnet or ipsubnetv6. These last information seems not provided by SCCM plugin. Anyway, I don't know if they are used/useful in GLPI. But this means we may also have to support them as list.
In GLPI UI, all ips related to one interface are aggregated on one interface so this should give the same result than having all ipv4 & ipv6 on one interface.

trasher:
This seems to fix a plugin only issue; as far as I understand this cannot happen with "standard" inventory.

Seeing @g-bougard comment; this would require more changes - therefore I wonder if it's better to fix that plugin side.

stonebuzz:
> Just as a side note, actually GLPI-Agent is indeed providing one network interface by found ip and there's only one ipv4 or one ipv6 by network interface. So the list notion is created at the netowrk interface level by the agent. Also each interface also includes ipmask or ipmaskv6 and ipsubnet or ipsubnetv6. These last information seems not provided by SCCM plugin. Anyway, I don't know if they are used/useful in GLPI. But this means we may also have to support them as list. In GLPI UI, all ips related to one interface are aggregated on one interface so this should give the same result than having all ipv4 & ipv6 on one interface.

PR #14736, opened 2023-05-19
Add simple API response formatter middleware

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

Add simple response middleware for new API to optionally transform the JSON output into CSV or XML format to ease the integration of external systems.

### DIFF
--- a/src/Api/HL/Controller/AdministrationController.php
+++ b/src/Api/HL/Controller/AdministrationController.php
@@ -37,6 +37,7 @@ namespace Glpi\Api\HL\Controller;
 
 use Entity;
 use Glpi\Api\HL\Doc as Doc;
+use Glpi\Api\HL\Middleware\ResultFormatterMiddleware;
 use Glpi\Api\HL\Route;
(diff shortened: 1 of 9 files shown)


orthagh:
Is there an actual usage currently within the ui ? like in swagger ui ?

cconard96:
I've changed the documentation generator so that the format dropdown in the response sections will change the Accept header.

This isn't the final version of the PR though. I think the middleware should be restricted more rather than working for any route that can return a JSON response.

orthagh (approved):

trasher (approved):

cconard96:
I don't understand why it fails on those lines. I understand that it could be treating the attribute declaration as a function, but it doesn't make sense that it hates this specific case when there are plenty of other ones spanning multiple lines with multiple parameters on the same line.

cedric-anne (approved):

PR #14764, opened 2023-05-23
Improve kanban performances 3

AdrienClairembault:
Following #14525 and #14599.

I've found another bottleneck that was slowing down the loading of the kanban page.

Each columns are loaded by an ajax request, which will find all the tickets of that column.
Then, for each tickets found, we load some extra data (actors and linked tickets).

This quickly become a costly operation as it require at least ~10 extra SQL requests for each ticket to load this extra data (so for 100 tickets, you run 1000 extra queries).

I've improved the performances massively by bulk loading all the extra data using a few SQL queries.
Then for each ticket we can just reach for the data in the preloaded arrays.

The small downside is that we use a bit more of memory but it seems reasonable (8mb for ~500 tickets).
If memory become a problem for massive databases (which would probably require ~30k open tickets to break a 256mb limit) , we can always split the process in X tickets batches.

Regarding the benchmark, I've used a test database with around 130k tickets.
Most tickets are closed, the biggest non-closed column had ~550 tickets:

image
 
Before the changes, the biggest column would take almost 2 seconds to load.
After the changes, it only take around 0.3 seconds.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !27631

### DIFF
--- a/ajax/kanban.php
+++ b/ajax/kanban.php
@@ -242,6 +242,7 @@ if (($_POST['action'] ?? null) === 'update') {
     header("Content-Type: application/json; charset=UTF-8", true);
     echo json_encode($itemtype::getAllKanbanColumns($_REQUEST['column_field']));
 } else if ($_REQUEST['action'] === 'get_column') {
+    session_write_close();
     $checkParams(['column_id', 'column_field', 'items_id']);
     header("Content-Type: application/json; charset=UTF-8", true);
     $column = $itemtype::getKanbanColumns($_REQUEST['items_id'], $_REQUEST['column_field'], [$_REQUEST['column_id']]);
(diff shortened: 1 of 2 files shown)


cconard96 (approved):

cedric-anne (commented):
`steps` data seems to have disappeared, is this intended?

AdrienClairembault:
> `steps` data seems to have disappeared, is this intended?

No, it seems I've put in in `_links` instead by mistake.

cedric-anne (approved):
It is hard to read, but it seems OK.

AdrienClairembault:
> It is hard to read, but it seems OK.

Code could probably be split into multiple functions (which could then be unit tested) but I wanted to keep things simple for a bugfixes PR.

orthagh (approved):

PR #14766, opened 2023-05-24
Rules Engine for Problems

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

Add the remaining missing ITIL Object rule engine.
https://glpi.userecho.com/communities/1/topics/2180-business-rules-for-problems

- [ ] De-duplicate ITIL rules code. Handle more in CommonITILObject now.


src/RuleProblem.php
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+class RuleProblem extends RuleCommonITILObject
+{
+    // From Rule
+    public static $rightname = 'rule_problem';
+
+    public function getTitle()
+    {
+        return __('Business rules for changes');
```
trasher, line 43:
```suggestion
        return __('Business rules for problems');
```

orthagh (commented):
Maybe hard to achieve with Ticket class in mind, but my first thought has been : could we move all the stuff in PrepareInputForXXX method in CommonITILObject.

Otherwise, if we can confirm all the new stuff only alter $input variable, maybe we can move to a new generic CommonITILObject::processRules($input).

cedric-anne (approved):
Diff is pretty hard to read, but it seems OK.

cconard96:
> Not a new issue, but while I reviewed the pr, I checked the profiles and image
> 
> This is weird, they have the same value (1024) but several columns are generated. In 10.0/bf with only ticket and asset, they are grouped under the same column. Could this started with #14570 ?

Possibly. It is mostly OK on the tagged 10.0.7 commit. Although, the tooltip for the "Parent Business" right is "Business rules for ticket (entity parent)"

cconard96:
The real bug is what is happening in 10.0/bugfixes currently due to the long name of the right for RuleAsset being "Business rules for tickets (entity parent)". I guess it was a bad copy-paste. The expected behavior of the rights matrix is to separate the values based on the long label. This is why the asset and ticket rules have Parent Business in the same column.

This could be "fixed" by using a generic long label between the ITIL and Asset rules.

orthagh:
I'm not sure we want the "(entity parent)" for asset rules.
Currently, I will be in pain to explain what it means.

cedric-anne:

cedric-anne (approved):

trasher (approved):

orthagh (approved):

PR #14768, opened 2023-05-24
 Make business rules work with "is inventory" criteria 

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #14609

As draft as I'm not entirely sure; see FIXMEs in tests.

### DIFF
--- a/src/Inventory/Asset/InventoryAsset.php
+++ b/src/Inventory/Asset/InventoryAsset.php
@@ -425,7 +425,7 @@ abstract class InventoryAsset
 
     protected function handleInput(\stdClass $value, ?CommonDBTM $item = null): array
     {
-        $input = [];
+        $input = ['_auto' => 1];
         $locks = [];
 
         if ($item !== null) {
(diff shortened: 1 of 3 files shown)


Mirkk:
Hi, we had the same issue as in #14609 and this small code change fixes the issue for us: the state is changed appropriately now.
GLPI 10.0.7
It would be perfect if this push request is put to the next GLPI version, thanks!

g-bougard:
Hi @trasher

I was making tests with real datas to understand how we can fix the [glpi-agent issue related to AzureAD users computer/user linking](https://github.com/glpi-project/glpi-agent/issues/476).

I added some debug on my GLPI 10.0.9 test instance and finally found none of the business rules for assets could match. So this totally prevents computer/user linking.

Now after I applied this PR, I can at least have a computer/user linking on a simple case.

So I can confirm wihtout this PR the "is inventory" condition which is on all "business rules for assets" rules never matches.

stonebuzz:
Fix internal ref !29764 too

trasher:
Replaced with #15695

PR #14783, opened 2023-05-26
Fixed style and placement of plugin uninstall button

FragDev:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | - 

Following pull request #14746, the style and placement of the uninstallation button has been modified.
The button is no longer the same color and is offset.

Actual :

Expected :

### DIFF
--- a/src/Plugin.php
+++ b/src/Plugin.php
@@ -2674,12 +2674,12 @@ class Plugin extends CommonDBTM
                         $output .= TemplateRenderer::getInstance()->render('components/plugin_uninstall_modal.html.twig', [
                             'plugin_name' => $plugin->getField('name'),
                             'modal_id' => 'uninstallModal' . $plugin->getField('directory'),
-                            'open_btn' => '<span class="fas fa-fw fa-folder-minus fa-2x pointer"
-                                                 data-bs-toggle="modal"
-                                                 data-bs-target="#uninstallModal' . $plugin->getField('directory') . '"
-                                                 title="' . __s("Uninstall") . '">
-                                               <span class="sr-only">' . __s("Uninstall") . '</span>
-                                           </span>',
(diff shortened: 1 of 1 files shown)


src/Plugin.php
```diff
@@ -2674,12 +2674,12 @@ public static function getSpecificValueToDisplay($field, $values, array $options
...
-                                                 data-bs-target="#uninstallModal' . $plugin->getField('directory') . '"
-                                                 title="' . __s("Uninstall") . '">
-                                               <span class="sr-only">' . __s("Uninstall") . '</span>
-                                           </span>',
+                            'open_btn' => '<a class="pointer"><span class="fas fa-fw fa-folder-minus fa-2x"
+                                                  data-bs-toggle="modal"
+                                                  data-bs-target="#uninstallModal' . $plugin->getField('directory') . '"
+                                                  title="' . __s("Uninstall") . '">
+                                                  <span class="sr-only">' . __s("Uninstall") . '</span>
+                                              </span></a>&nbsp;',
```
orthagh, line 2682: This seems wrong, could you use margin modifier (`me-1` for example) ?

FragDev: `me-1` could also work, although it doesn't produce exactly the same spacing.
But I've based it on the previous version, which uses a non-breaking space.
What makes the most sense?

orthagh: Never use nbsp for design/alignment.
If the previous design use this, it's still wrong (maybe me, I take the blame)

If the alignment it's still not ok with margin end, apply also on the other control (install button)

orthagh (approved):

cedric-anne (approved):

PR #14785, opened 2023-05-26
Kanban: add missing join

AdrienClairembault:
Issue introduced in #14333: `Search::addDefaultWhere()`  is used without the needed `JOIN` clause, which lead to invalid SQL requests.

This affect any user who can't see all the tickets and thus will have some conditions on external tables (linked user, groups, validation, ...) returned by `Search::addDefaultWhere()`.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !27631

### DIFF
--- a/src/DBmysqlIterator.php
+++ b/src/DBmysqlIterator.php
@@ -669,6 +669,7 @@ class DBmysqlIterator implements SeekableIterator, Countable
             }
 
             foreach ($jointables as $jointablekey => $jointablecrit) {
+                // QueryExpression support, can be removed once Search::getDefaultJoin no longer returns raw SQL
                 if ($jointablecrit instanceof QueryExpression) {
                     $query .= $jointablecrit->getValue();
                     continue;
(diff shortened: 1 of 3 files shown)


src/DBmysqlIterator.php
```diff
@@ -669,6 +669,11 @@ public function analyseJoins(array $joinarray)
             }
 
             foreach ($jointables as $jointablekey => $jointablecrit) {
+                if ($jointablecrit instanceof QueryExpression) {
```
trasher, line 672: New cases in this class must be tested

AdrienClairembault: Done.

AdrienClairembault:
It would be nice to also test `getDataToDisplayOnKanban` to ensure its stability but I can't spare the time at this moment...

orthagh (approved):

PR #14799, opened 2023-05-30
fix(ComputerVM): recovers old behaviour when inventory handle a missing VM

stonebuzz:
As discuss internally

Drop current behavior :

- When a ```VM``` disappears from an ```ESX``` inventory, ```Computer``` and ```ComputerVirtualMachine``` are moved to the recycle bin
- When this ```VM``` is remounted from another ```ESX```, "```Computer``` is restored, and a new ```ComputerVirtualMachine``` link is created

To respect old behaviour :

- When a ```VM``` disappears from an ```ESX``` inventory, ```Computer``` is no longer trashed, ```ComputerVirtualMachine``` is purged
- When this ```VM``` comes back up from another ```ESX```, a ```ComputerVirtualMachine```  link is created

| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !26293


trasher (approved):
OK for me. PR title should certainly be changed

cedric-anne (approved):

liedekef:
Just to add as a comment: the new behavior caused issues for a proxmox cluster (which doesn't have an ESX inventory): when a VM migrated to another host in the cluster, the new behavior caused the VM to go to trash (and a link remained in both the VM towards the host is was running on and vise versa). This caused pingpong issues: when the inventory runs on the VM, it was set to active, but when the inventory then ran on a host it migrated away from, the VM was put in trash etc ... the old behavior is much more consistent.

trasher:
See #14928.

stonebuzz:
Maybe we can completely purged link between host and VM

stonebuzz:
> Maybe we can completely purged link between host and VM

this is already the case

liedekef:
The referred issue https://github.com/glpi-project/glpi/issues/14928 is solved by this change, so all is good for me.

PR #14800, opened 2023-05-30
feat(Inventory): put OS as Software

stonebuzz:
See from here : https://glpi.userecho.com/communities/1/topics/2179-return-os-to-software-list

Put ```OS``` as ```Software``` (as before)

I'm doing this PR to get a clearer idea of the changes involved

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #number

### DIFF
--- a/src/Inventory/Asset/Software.php
+++ b/src/Inventory/Asset/Software.php
@@ -94,6 +94,31 @@ class Software extends InventoryAsset
         $with_manufacturer = [];
         $without_manufacturer = [];
 
+        //put OS as software
+        $main_asset_raw_data = $this->getMainAsset()->getRawData();
+        if (
+            property_exists($main_asset_raw_data, 'content')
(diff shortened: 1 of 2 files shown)


src/Inventory/Asset/Software.php
```diff
@@ -94,6 +94,31 @@ public function prepare(): array
         $with_manufacturer = [];
         $without_manufacturer = [];
 
+        //put OS as software
+        $main_asset_raw_data = $this->getMainAsset()->getRawData();
+        if (
+            property_exists($main_asset_raw_data, 'content')
+            && property_exists($main_asset_raw_data->content, 'operatingsystem')
```
trasher, line 101: OperatingSystem data is already available as `$this->extra_data[OperatingSystem::class]`

PR #14849, opened 2023-06-04
Add cartridge levels survey for printers by vcartand notification 

kabassanov:
| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #number

### DIFF
--- a/front/printercartridgelevelalert.php
+++ b/front/printercartridgelevelalert.php
@@ -1,0 +1,46 @@
+<?php
+
+/**
+ * ---------------------------------------------------------------------
+ *
+ * GLPI - Gestionnaire Libre de Parc Informatique
+ *
+ * http://glpi-project.org
+ *
+ * @copyright 2015-2023 Teclib' and contributors.
+ * @copyright 2003-2014 by the INDEPNET Development Team.
+ * @licence   https://www.gnu.org/licenses/gpl-3.0.html
+ *
+ * ---------------------------------------------------------------------
+ *
+ * LICENSE
+ *
+ * This file is part of GLPI.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ * ---------------------------------------------------------------------
+ */
+
+include('../inc/includes.php');
+
+Session::checkRight("printer", READ);
+Session::checkRight("cartridge", READ);
+
+Html::header(PrinterCartridgeLevelAlert::getTypeName(Session::getPluralNumber()), $_SERVER['PHP_SELF'], "tools", "printercartridgelevelalert");
+
+PrinterCartridgeLevelAlert::title();
+PrinterCartridgeLevelAlert::displayAlerts();
+
+Html::footer();
--- a/install/migrations/update_10.0.7_to_10.0.8/cartridgelevelalert.php
+++ b/install/migrations/update_10.0.7_to_10.0.8/cartridgelevelalert.php
@@ -75,7 +75,8 @@ $result = $DB->query($query_id) or die ($DB->error());
 $notification = $DB->result($result, 0, 'id');
 
 // Insert notification template
-$DB->queryOrDie("INSERT INTO `glpi_notificationtemplates` VALUES ('2023-05-26 09:44:46',NULL, 'Printer Cartridge Levels', 'PrinterCartridgeLevelAlert', '2023-05-26 09:44:46','',NULL)");
+$DB->queryOrDie("INSERT INTO `glpi_notificationtemplates` ( `date_creation`, `id`, `name`, `itemtype`, `date_mod`, `comment`, `css`) 
+            VALUES ('2023-05-26 09:44:46',NULL, 'Printer Cartridge Levels', 'PrinterCartridgeLevelAlert', '2023-05-26 09:44:46','',NULL)");
 
 // Get notification template ID
 $query_id = "SELECT `id`
--- a/install/migrations/update_10.0.7_to_10.0.8/cartridgelevelalert.php
+++ b/install/migrations/update_10.0.7_to_10.0.8/cartridgelevelalert.php
@@ -92,7 +92,7 @@ $DB->queryOrDie($query);
 
 // Insert translation for the notification
 $query = "INSERT INTO `glpi_notificationtemplatetranslations`                                 VALUES(NULL, " . $notificationtemplate . ", '','##cartridge.action## : ##cartridge.entity##',
-##FOREACHcartridges##
+'##FOREACHcartridges##
 ##lang.cartridge.printer## : ##cartridge.printer##
 ##lang.cartridge.entity## : ##cartridge.entity##
 ##lang.cartridge.item## : ##cartridge.item##
--- a/install/migrations/update_10.0.7_to_10.0.8/cartridgelevelalert.php
+++ b/install/migrations/update_10.0.7_to_10.0.8/cartridgelevelalert.php
@@ -39,7 +39,7 @@
  */
 
 $migration->addField('glpi_cartridgeitems', 'type_tag', "varchar(255)");
-$migration->addField('glpi_cartridgeitems', `warn_level` int unsigned NOT NULL DEFAULT '0');
+$migration->addField('glpi_cartridgeitems', 'warn_level' int unsigned NOT NULL DEFAULT '0');
 $migration->addKey('glpi_cartridgeitems', 'type_tag', 'type_tag');
 
 $migration->addField('glpi_entities', 'printer_cartridge_levels_alert_repeat', "int NOT NULL DEFAULT '-2'");
--- a/install/migrations/update_10.0.7_to_10.0.8/cartridgelevelalert.php
+++ b/install/migrations/update_10.0.7_to_10.0.8/cartridgelevelalert.php
@@ -55,7 +55,7 @@ if (empty($crontask->find(['name' => 'PrinterCartridgeLevelAlert']))) {
             'state' => CronTask::STATE_DISABLE,
             'mode' => CronTask::MODE_EXTERNAL
         ]
-        );
+    );
 }
 if (!$cron_added) {
         die("Can't add PrinterCartridgeLevelAlert cron");
--- a/install/migrations/update_10.0.7_to_10.0.8/cartridgelevelalert.php
+++ b/install/migrations/update_10.0.7_to_10.0.8/cartridgelevelalert.php
@@ -120,5 +120,5 @@ $DB->queryOrDie($query);
 
 // Insert Entity Administrator as default target for low level cartridge notifications
 $query = "INSERT INTO `glpi_notificationtargets` (`items_id`, `type`, `notifications_id`)
-            VALUES (11, 1, " . $notification_id . ");";
+            VALUES (11, 1, " . $notification . ");";
 $DB->query($query);
--- a/install/migrations/update_10.0.7_to_10.0.8/cartridgelevelalert.php
+++ b/install/migrations/update_10.0.7_to_10.0.8/cartridgelevelalert.php
@@ -53,9 +53,9 @@ if (empty($crontask->find(['name' => 'PrinterCartridgeLevelAlert']))) {
         1 * DAY_TIMESTAMP,
         [
             'state' => CronTask::STATE_DISABLE,
-            'mode' => CronTask::MODE_EXTERNAL,
+            'mode' => CronTask::MODE_EXTERNAL
         ]
-    )
+        );
 }
 if (!$cron_added) {
         die("Can't add PrinterCartridgeLevelAlert cron");
--- a/install/migrations/update_10.0.7_to_10.0.8/cartridgelevelalert.php
+++ b/install/migrations/update_10.0.7_to_10.0.8/cartridgelevelalert.php
@@ -39,7 +39,7 @@
  */
 
 $migration->addField('glpi_cartridgeitems', 'type_tag', "varchar(255)");
-$migration->addField('glpi_cartridgeitems', 'warn_level' int unsigned NOT NULL DEFAULT '0');
+$migration->addField('glpi_cartridgeitems', 'warn_level', "int unsigned NOT NULL DEFAULT '0'");
 $migration->addKey('glpi_cartridgeitems', 'type_tag', 'type_tag');
 
 $migration->addField('glpi_entities', 'printer_cartridge_levels_alert_repeat', "int NOT NULL DEFAULT '-2'");
--- a/install/migrations/update_10.0.7_to_10.0.8/cartridgelevelalert.php
+++ b/install/migrations/update_10.0.7_to_10.0.8/cartridgelevelalert.php
@@ -75,12 +75,12 @@ $result = $DB->query($query_id) or die ($DB->error());
 $notification = $DB->result($result, 0, 'id');
 
 // Insert notification template
-$DB->queryOrDie("INSERT INTO `glpi_notificationtemplates` VALUES ('2023-05-26 09:44:46',NULL, 'Printer Cartridge level', 'PrinterCartridgeLevelAlert', '2023-05-26 09:44:46','',NULL)");
+$DB->queryOrDie("INSERT INTO `glpi_notificationtemplates` VALUES ('2023-05-26 09:44:46',NULL, 'Printer Cartridge Levels', 'PrinterCartridgeLevelAlert', '2023-05-26 09:44:46','',NULL)");
 
 // Get notification template ID
 $query_id = "SELECT `id`
                 FROM `glpi_notificationtemplates`
-                WHERE `itemtype`='PrinterCartridgeLevelAlert' AND `name` = 'Printer Cartridge level'";
+                WHERE `itemtype`='PrinterCartridgeLevelAlert' AND `name` = 'Printer Cartridge Levels'";
 $result = $DB->query($query_id) or die ($DB->error());
 $notificationtemplate = $DB->result($result, 0, 'id');
 
--- a/install/migrations/update_10.0.7_to_10.0.8/cartridgelevelalert.php
+++ b/install/migrations/update_10.0.7_to_10.0.8/cartridgelevelalert.php
@@ -55,7 +55,7 @@ if (empty($crontask->find(['name' => 'PrinterCartridgeLevelAlert']))) {
             'state' => CronTask::STATE_DISABLE,
             'mode' => CronTask::MODE_EXTERNAL,
         ]
-    );
+    )
 };
 if (!$cron_added) {
         die("Can't add PrinterCartridgeLevelAlert cron");
--- a/install/migrations/update_10.0.7_to_10.0.8/cartridgelevelalert.php
+++ b/install/migrations/update_10.0.7_to_10.0.8/cartridgelevelalert.php
@@ -56,7 +56,7 @@ if (empty($crontask->find(['name' => 'PrinterCartridgeLevelAlert']))) {
             'mode' => CronTask::MODE_EXTERNAL,
         ]
     )
-};
+}
 if (!$cron_added) {
         die("Can't add PrinterCartridgeLevelAlert cron");
     }
--- a/install/migrations/update_10.0.7_to_10.0.8/cartridgelevelalert.php
+++ b/install/migrations/update_10.0.7_to_10.0.8/cartridgelevelalert.php
@@ -42,7 +42,7 @@ $migration->addField('glpi_cartridgeitems', 'type_tag', "varchar(255)");
 $migration->addField('glpi_cartridgeitems', 'warn_level', "int NOT NULL DEFAULT '0'");
 $migration->addKey('glpi_cartridgeitems', 'type_tag', 'type_tag');
 
-$migration->addField('glpi_entities', 'printer_cartridge_levels_alert_repeat, "int NOT NULL DEFAULT '-2'");
+$migration->addField('glpi_entities', 'printer_cartridge_levels_alert_repeat', "int NOT NULL DEFAULT '-2'");
 
 // Add crontask for low level notifications
 $crontask = new CronTask();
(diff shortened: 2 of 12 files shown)


install/migrations/update_10.0.7_to_10.0.8/cartridgelevelalert.php
```diff
@@ -0,0 +1,124 @@
...
+    );
+}
+if (!$cron_added) {
+        die("Can't add PrinterCartridgeLevelAlert cron");
+}
+
+// Insert notification
+$query = "INSERT INTO `glpi_notifications` (`name`, `entities_id`, `itemtype`, `event`, `is_recursive`, `is_active`)
+            VALUES ('Printer Cartridge Levels', 0, 'PrinterCartridgeLevelAlert', 'alert', 1, 1)";
+$DB->queryOrDie($query);
```
cconard96, line 67: Try to avoid raw queries. You can use insertOrDie here.

This applies to the other queries here. The only type of query that should need raw SQL currently is when creating tables.

src/Html.php
```diff
@@ -1426,7 +1426,7 @@ public static function getMenuInfos()
                 'title' => __('Administration'),
                 'types' => [
                     'User', 'Group', 'Entity', 'Rule',
-                    'Profile', 'QueuedNotification', 'Glpi\\Event', 'Glpi\Inventory\Inventory'
+                    'Profile', 'QueuedNotification', 'PrinterCartridgeLevelAlert', 'Glpi\\Event', 'Glpi\Inventory\Inventory'
```
cconard96, line 1429: I'm not sure having this directly in the menu is a good idea.

src/PrinterCartridgeLevelAlert.php
```diff
@@ -0,0 +1,328 @@
...
+    * @param $entities
+    *        $repeat for already notified management
+    *
+    * @return string
+    */
+    private static function query($entities, $repeat = null)
+    {
+        global $DB;
+
+        $query = "SELECT c.id as cartridge, i.id as cartridgeitem, p.id as printer, p.entities_id as entity, l.value as cartridgelevel, a.id as alertID, a.date as alertDate
```
cconard96, line 82: No raw queries please. This method could return an array of criteria to use with `$DB->request` instead.

cconard96 (changes requested):
Not an extensive review. Just a quick check of the code from the mobile GitHub app.

It may also be helpful to have a description of the changes/feature.

src/Html.php
kabassanov: This is the behavior introduced by the AdditionalAlerts plugin. Having a summary page that allows to have a list of installed cartridges in a low level state is very usefull IMHO.

src/PrinterCartridgeLevelAlert.php
kabassanov: I need some help. I was unable to correctly create a full array SQL construction.

trasher: You can find documentaiton at: https://glpi-developer-documentation.readthedocs.io/en/master/devapi/database/dbiterator.html

kabassanov: > You can find documentaiton at: https://glpi-developer-documentation.readthedocs.io/en/master/devapi/database/dbiterator.html

My concern is about extra characters like quotes that the parser adds by itself and that breaks sql syntax. Will try to take a look once again...

src/Printer_CartridgeInfo.php
```diff
@@ -63,14 +63,50 @@ public function getInfoForPrinter(Printer $printer)
...
+                    'FKEY'   => [
+                        'glpi_cartridges'        => 'cartridgeitems_id',
+                        'glpi_cartridgeitems'        => 'id'
+                    ]
+                ]
+            ],
+            'WHERE'     => [
+                'glpi_cartridges.printers_id = ' . $id,
+                'glpi_cartridges.date_out is not null'
+            ]
```
cconard96, line 89:
```suggestion
            'WHERE'     => [
                'glpi_cartridges.printers_id' => $id,
                'NOT' => ['glpi_cartridges.date_out' => null]
            ]
```

src/Printer_CartridgeInfo.php
```diff
@@ -63,14 +63,50 @@ public function getInfoForPrinter(Printer $printer)
...
     {
         $info = $this->getInfoForPrinter($printer);
 
+        $mapping = $this->getCartridgeTagInfoForPrinter($printer);
+
         echo "<h3>" . $this->getTypeName(Session::getPluralNumber()) . "</h3>";
 
         echo "<table class='tab_cadre_fixehov'>";
-        echo "<thead><tr><th>" . __('Property') . "</th><th>" . __('Value') . "</th></tr></thead>";
+        echo "<thead><tr><th>" . __('Property') . "</th><th>" . _n('Cartridge model', 'Cartridge models', 1) . "</th><th> " . __('Value') . "</th></tr></thead>";
```
cconard96, line 109:
```suggestion
        echo "<thead><tr><th>" . __('Property') . "</th><th>" . CartridgeItem::getTypeName(1) . "</th><th> " . __('Value') . "</th></tr></thead>";
```

cconard96 (changes requested):

src/PrinterCartridgeLevelAlert.php
```diff
@@ -0,0 +1,364 @@
...
+    * @param $entities
+    *        $repeat for already notified management
+    *
+    * @return string
+    */
+    private static function query($entities, $repeat = null)
+    {
+        global $DB;
+
+        $where = "c.date_out is NULL AND ( ( l.value REGEXP '^[0-9]+$' AND l.value <= i.warn_level ) OR ( l.value like 'WARNING' OR l.value like 'BAD' )) AND l.printers_id = p.id AND p.entities_id IN ($entities)";
```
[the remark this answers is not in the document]
trasher: Also, using `LIKE` with a simple text value does not make sense, it's easier to simply use equals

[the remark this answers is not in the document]
kabassanov: I need to check it attentively. Will do it tomorrow. BTW, I wonder if "l.printers_id = p.id" is really usefull. I get the same results without it, but it could be related to my data set... When could the entity be empty? Is it just to prevent errors in the future?

Need also to check again, but last time I had some troubles with this kind of expressions:   ['l.value' => ['<=' => 'i.warn_level']] as the parser considered that the 'i.warn_level' was a string and not a table field...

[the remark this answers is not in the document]
cconard96: > Operators notations like `['LIKE' => 'BAD']` is wrong; correct syntax is `['LIKE', 'BAD']` (see https://glpi-developer-documentation.readthedocs.io/en/master/devapi/database/dbiterator.html#operators)

I was thinking it was intentional for a case insensitive check, but it would be best to avoid the possibility of the case being different between records.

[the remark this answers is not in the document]
trasher: Seems like equals is case insensitive as well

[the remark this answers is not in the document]
kabassanov: > Seems like equals is case insensitive as well

The case insensivity comes from the DB itself by default. There are particular treatments to do if you want to work with case-sensitive operations.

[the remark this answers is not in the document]
kabassanov: Guys,

using  ```'p.entities_id' => $entities``` leads to ```AND `p`.`entities_id` = ''19', '24', '22', '25', '23'' AND ``` which is not correct from MariaDB POV.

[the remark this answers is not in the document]
kabassanov: BTW, ```AND (`l`.`value` IN ('i.warn_level'))) OR``` is what I was talking about yesterday evening...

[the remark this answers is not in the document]
trasher: > Guys,
> 
> using `'p.entities_id' => $entities` leads to `` AND `p`.`entities_id` = ''19', '24', '22', '25', '23'' AND `` which is not correct from MariaDB POV.

For sur, because you call `$query  = self::query($_SESSION["glpiactiveentities_string"]);`; which is... a sstring, and not an array.

> 
> `'p.entities_id' => [$entities]` could be better if... in `` p`.`entities_id` IN (''19', '24', '22', '25', '23'') `` the constructor was not putting each entity id between quotes...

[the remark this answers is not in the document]
kabassanov: > > using `'p.entities_id' => $entities` leads to `` AND `p`.`entities_id` = ''19', '24', '22', '25', '23'' AND `` which is not correct from MariaDB POV.
> 
> For sure, because you call `$query = self::query($_SESSION["glpiactiveentities_string"]);`; which is... a sstring, and not an array.

OK, my fault, I'll change it to ```$query  = self::query($_SESSION["glpiactiveentities"]); ```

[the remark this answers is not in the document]
kabassanov: > BTW, `` AND (`l`.`value` IN ('i.warn_level'))) OR `` is what I was talking about yesterday evening...

But this is still blocking...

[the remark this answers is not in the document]
kabassanov: OK, with some small changes it seems to be correct now. Launching tests...

src/NotificationTargetPrinterCartridgeLevelAlert.php
```diff
@@ -0,0 +1,90 @@
...
+{
+    public function getEvents()
+    {
+        return ['alert' => __('Printer cartridge level alarm')];
+    }
+
+    public function addDataForTemplate($event, $options = [])
+    {
+        $events = $this->getAllEvents();
+        $this->data['##cartridge.action##'] = __($events[$event]);
```
cconard96, line 50: No need to translate this as it is already done in the `getEvents` method.

src/NotificationTargetPrinterCartridgeLevelAlert.php
```diff
@@ -0,0 +1,90 @@
...
+            if (!isset($this->data[$tag])) {
+                $this->data[$tag] = $values['label'];
+            }
+        }
+    }
+
+    public function getTags()
+    {
+        $tags = [
+            'cartridge.printer'        => _n('Printer', 'Printers', Session::getPluralNumber()),
```
cconard96, line 72:
```suggestion
            'cartridge.printer'        => Printer::getTypeName(Session::getPluralNumber()),
```

src/PrinterCartridgeLevelAlert.php
```diff
@@ -0,0 +1,383 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ * ---------------------------------------------------------------------
+ */
+
+/**
+ *  PrinterCartridgeLevelAlert class
+ *
+ * @ since version 0.84
```
cconard96, line 39: This isn't correct

src/PrinterCartridgeLevelAlert.php
```diff
@@ -0,0 +1,383 @@
...
+        );
+    }
+
+    /**
+    * @param $entities
+    *        $repeat for already notified management
+    *
+    * @return string
+    */
+    private static function query($entities, $repeat = null)
```
cconard96, line 78:
```suggestion
    * @param array $entities
    * @param int $repeat for already notified management
    *
    * @return array
    */
    private static function query(array $entities, int $repeat = 0): array
```

src/PrinterCartridgeLevelAlert.php
```diff
@@ -0,0 +1,383 @@
...
+                        ['l.value' => ['REGEXP', '^[0-9]+$']],
+                        ['l.value' => ['<=', new QueryExpression($DB::quoteName('i.warn_level')) ] ]
+                    ],
+                    'OR' => [
+                        ['l.value' => 'WARNING'],
+                        ['l.value' => 'BAD']
+                    ]
+                ]
+            ];
+            if ($repeat) {
```
cconard96, line 98:
```suggestion
            if ($repeat > 0) {
```

src/PrinterCartridgeLevelAlert.php
```diff
@@ -0,0 +1,383 @@
...
+                ],
+                'WHERE'        => $WHERE,
+                'ORDERBY'      => [
+                    'p.name'
+                ]
+            ];
+
+            return $query;
+        } else {
+            return "";
```
cconard96, line 157:
```suggestion
            return [];
```

src/PrinterCartridgeLevelAlert.php
```diff
@@ -0,0 +1,383 @@
...
+        }
+    }
+
+    /**
+    * @param $data
+    *
+    * @return string
+    */
+    public static function prepareBodyValues($data)
```
cconard96, line 166:
```suggestion
    /**
    * @param array $data
    *
    * @return array
    */
    public static function prepareBodyValues(array $data): array
```

src/PrinterCartridgeLevelAlert.php
```diff
@@ -0,0 +1,383 @@
...
+        $result['cartridge.alertDate'] = $data["alertDate"];
+        return $result;
+    }
+
+    /**
+    * @param $data
+    *
+    * @return string
+    */
+    private static function displayBody($data)
```
cconard96, line 198:
```suggestion
    * @param array $data
    *
    * @return string
    */
    private static function displayBody(array $data): string
```

src/PrinterCartridgeLevelAlert.php
```diff
@@ -0,0 +1,383 @@
...
+        $body .= "<td>" . $tmp['cartridge.level'] . "</td>";
+        $body .= "<td>" . $tmp['cartridge.alertDate'] . "</td>";
+        return $body;
+    }
+
+
+    /**
+    *
+    */
+    public static function displayAlerts()
```
cconard96, line 214:
```suggestion
    /**
    * @return void
    */
    public static function displayAlerts(): void
```

src/PrinterCartridgeLevelAlert.php
cconard96: The method name could be confusing since it returns the HTML rather than displaying it.

src/PrinterCartridgeLevelAlert.php
```diff
@@ -0,0 +1,383 @@
...
+        ];
+    }
+
+    /**
+     * Cron action on cartridges : alert if a stock is behind the threshold
+     *
+     * @param CronTask $task CronTask for log, display information if NULL? (default NULL)
+     *
+     * @return void
+     **/
```
cconard96, line 306: PHPDoc description seems wrong for this method.
Param type should be CronTask|null since it can accept null.
Return type is declared to be void but the method always returns 0.

src/Printer_CartridgeInfo.php
```diff
@@ -63,14 +63,52 @@ public function getInfoForPrinter(Printer $printer)
         return $info;
     }
 
+    public function getCartridgeTagInfoForPrinter(Printer $printer)
```
cconard96, line 66: Could the parameter be changed to just be the printer ID? This way, if the method is used for something else we don't need to create a printer object if we already have the ID.

cconard96 (changes requested):

src/NotificationTargetPrinterCartridgeLevelAlert.php
trasher: Also, only strings must be translated, not PHP expressions

src/PrinterCartridgeLevelAlert.php
kabassanov: Are you talking about "@ since version 0.84"? I thought this was related to a requirement since this version, but if it simply means that the class exists since 0.84, I'll delete it.

trasher: Yes, you can remove this line.

src/PrinterCartridgeLevelAlert.php
kabassanov: I simply copied/pasted some code from cronCartridge and it seems to have the same errors :) . Some mess in cronConsumable also... Will change mine :)

src/Printer_CartridgeInfo.php
kabassanov: I can do this, but as the calling function receives already a printer object as parameter, I wonder if it is useful... Do you want me to change it however?

stonebuzz:
Hi @kabassanov 

any news about this feature?

Best regards

front/printercartridgelevelalert.php
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * ---------------------------------------------------------------------
+ */
+
+include('../inc/includes.php');
+
+Session::checkRight("printer", READ);
+Session::checkRight("cartridge", READ);
+
+Html::header(PrinterCartridgeLevelAlert::getTypeName(Session::getPluralNumber()), $_SERVER['PHP_SELF'], "tools", "printercartridgelevelalert");
```
stonebuzz, line 41:
```suggestion
Html::header(PrinterCartridgeLevelAlert::getTypeName(Session::getPluralNumber()), $_SERVER['PHP_SELF'], "admin", "printercartridgelevelalert");
```

install/migrations/update_10.0.7_to_10.0.8/cartridgelevelalert.php
```diff
@@ -0,0 +1,138 @@
...
+ * @var Migration $migration
+ */
+
+$migration->addField('glpi_cartridgeitems', 'type_tag', "varchar(255)");
+$migration->addField('glpi_cartridgeitems', 'warn_level', "int unsigned NOT NULL DEFAULT '0'");
+$migration->addKey('glpi_cartridgeitems', 'type_tag', 'type_tag');
+
+$migration->addField('glpi_entities', 'printer_cartridge_levels_alert_repeat', "int NOT NULL DEFAULT '-2'");
+
+// Add crontask for low level notifications
```
stonebuzz, line 47: The auto task must also be added in the case of a fresh install too  (see ```install/empty_data.php```)

install/migrations/update_10.0.7_to_10.0.8/cartridgelevelalert.php
```diff
@@ -0,0 +1,138 @@
...
+            'state' => CronTask::STATE_DISABLE,
+            'mode' => CronTask::MODE_EXTERNAL
+        ]
+    );
+    if (!$cron_added) {
+        die("Can't add PrinterCartridgeLevelAlert cron");
+    }
+}
+
+// Insert notification
```
stonebuzz, line 64: Notifications must also be added in the case of a fresh install too (see ```install/empty_data.php```)

kabassanov:
Hi @stonebuzz,

I'm not aware of anything I should modify in the PR. So I'm just waiting...

src/Entity.php
```diff
@@ -2166,6 +2177,20 @@ public static function showNotificationOptions(Entity $entity)
...
+            'inherit_parent' => (($ID > 0) ? 1 : 0)
+        ]);
+
+        if ($entity->fields['printer_cartridge_levels_alert_repeat'] == self::CONFIG_PARENT) {
+            $tid = self::getUsedConfig('printer_cartridge_levels_alert_repeat', $entity->getField('entities_id'));
+            self::inheritedValue(self::getSpecificValueToDisplay('printer_cartridge_levels_alert_repeat', $tid), true);
+        }
+
+        echo "</td></tr>";
+
```
stonebuzz, line 2193: originally, the "Cartridge" option in Entity > notification mainly concerned stock management, 
adding an option here for cartridge level alerts bothers me.

image


For me, it's necessary to add a new "line" "cartridge level alerts" and possibly rename the first "cartridge" line as "Cartridge stock".

src/Inventory/Asset/Cartridge.php
```diff
@@ -41,7 +41,7 @@
 
 class Cartridge extends InventoryAsset
 {
-    public function knownTags(): array
+    public function knownTags(bool $withstates = true): array
```
stonebuzz, line 44: Can you explain this change ?

src/Item_Ticket.php
```diff
@@ -198,6 +198,7 @@ public static function itemAddForm(Ticket $ticket, $options = [])
         ];
 
         $opt = [];
+        unset($options['entities_id']);
```
stonebuzz, line 201: Why this ?

stonebuzz (changes requested):

kabassanov:
Let me check your comments one by one...

front/printercartridgelevelalert.php
kabassanov: Hmmm, I'm not sure why I inserted the menu item into the Administration menu. For me it should be in the tools. If so, I must change this elsewhere...

src/Inventory/Asset/Cartridge.php
kabassanov: I changed this function in order to reuse it for the cartridge type mapping to the type inventory process return. As originally it was creating Max, Used and Remainng instances for each type and this information is not usefull for the second usage I introduced, I just added a parameter defaulting to the previous behaviour suppressing it.

src/Item_Ticket.php
kabassanov: Wrong piece of code related to: https://github.com/glpi-project/glpi/pull/15450 
This bug was solved in another way. I've just suppress the line.

src/Printer_CartridgeInfo.php
kabassanov: Useless you mean?

src/Html.php
kabassanov: Hi, I'm back to this PR... I'm really not sure about the place Alerts should be put into... Printers cartridges alerts are not related to GLPI administration, they are representing much more a supervision tool like Tools/Reports... Do I miss something?

front/printercartridgelevelalert.php
kabassanov: Feature moved from Administration to a Tools sub-menu. So this suggestion is outdated.

install/migrations/update_10.0.7_to_10.0.8/cartridgelevelalert.php
kabassanov: Done.

install/migrations/update_10.0.7_to_10.0.8/cartridgelevelalert.php
kabassanov: Done.

stonebuzz (changes requested):
Before going any further in the review, can you rebase your work on the ```main``` branch?

All new features are now integrated into the ```main``` branch, while the ```bugfixes``` branch is used only for fixing.

cedric-anne:
I close this PR as it is stale and I guess noone on our side will have time to finish it.
We should have spent some time on it when it was first opened to help you finish it. Sorry we did not.

If you want to work on this for a future GLPI version, please contact us before starting your development.

PR #14875, opened 2023-06-07
feat(search): check field pattern

Rom1-B:
Checks the format entered with the "contains" search entry.

This informs the user that their input is incorrect while they are typing and does not execute the search query (which can be a significant resource saving in some cases).

Suggested warning, probably to be improved:

image

| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !26543

### DIFF
--- a/css/includes/components/_search_input.scss
+++ b/css/includes/components/_search_input.scss
@@ -97,3 +97,17 @@ div.search-input {
         }
     }
 }
+
+input:invalid {
+    background-color: $warning !important;
+}
+
+input + span.info {
+    display: none;
+}
+
+input:invalid + span.info {
+    display: inline;
+    color: $warning;
+    font-size: 0.8rem;
+}
--- a/src/Search/Input/QueryBuilder.php
+++ b/src/Search/Input/QueryBuilder.php
@@ -313,8 +313,13 @@ final class QueryBuilder implements SearchInputInterface
 
         // Default case : text field
         if (!$display) {
+            $fieldpattern = self::getInputValidationPattern($searchopt['datatype'] ?? '', false);
+            $pattern = $fieldpattern['pattern'];
+            $message = $fieldpattern['validation_message'];
+
             echo "<input type='text' class='form-control' size='13' name='$inputname' value=\"" .
-                htmlspecialchars($request['value']) . "\">";
+                htmlspecialchars($request['value']) . "\" pattern=\"" . htmlspecialchars($pattern) . "\">" .
+                "<span class='info'>" . htmlspecialchars($message) . "</span>";
         }
     }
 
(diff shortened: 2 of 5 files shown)


src/Search.php
```diff
@@ -3427,8 +3427,10 @@ public static function displaySearchoptionValue($request = [])
 
        // Default case : text field
         if (!$display) {
+            global $DB;
+            $pattern = $DB->getFieldPattern($searchopt['table'], $searchopt['field'], false);
```
cconard96, line 3431: $searchopt can be an empty array here at least when there is an Items Seen criteria since it isn't actually a search option.

cconard96:
The input seems to think operators in the contains field are invalid when they are not.
`<2023-01-01` is a valid contains search for dates but marked invalid.
Similar searches can be done for other data types.

trasher:
I'm not sure this should target bugfixes branch

Rom1-B:
@cconard96 
> The input seems to think operators in the contains field are invalid when they are not. `<2023-01-01` is a valid contains search for dates but marked invalid. Similar searches can be done for other data types.

I didn't know about this feature.
According to the documentation, I have to manage `NULL`, `>`, `<`, `^`, `$`, I forget some?

https://glpi-user-documentation.readthedocs.io/fr/latest/first-steps/search.html?highlight=contains#perform-a-multi-criteria-search



@trasher 
> I'm not sure this should target bugfixes branch

I don't know, I targeted BF thinking it was a feature that could avoid problems, but I can change the target.

orthagh:
I confirm this should target main branch

cconard96:
> I didn't know about this feature. According to the documentation, I have to manage `NULL`, `>`, `<`, `^`, `$`, I forget some?

<= and >= too I think.
That is all I am aware of then.

Rom1-B:
Visual corresponding to  https://github.com/glpi-project/glpi/pull/14875/commits/5b4b1fe2a906df3e57d0f8db098299cc3f15f269

image

src/DBmysql.php
```diff
@@ -950,6 +950,54 @@ public function getField(string $table, string $field, $usecache = true): ?array
...
 
+    /**
+     * Get field pattern of a table field based on field type
+     *
+     * @param string $table
+     * @param string $field
+     *
+     * @return string Field pattern
+     */
+    public function getFieldPattern($table, $field, $with_delimiter = true)
```
trasher, line 961: This method must be tested

trasher (dismissed):
See comment. Also, a test should be added to check the whole feature is working.

tests/units/DB.php
```diff
@@ -856,4 +856,151 @@ public function testDecodeFetchResult(string $method, mixed $row, mixed $expecte
...
+                ]
+            ],
+            [
+                'params' => [
+                    'table' => 'glpi_tickets',
+                    'field' => 'solvedate', //timestamp
+                    'with_delimiter' => true
+                ],
+                'expected' => [
+                    '/^(NULL|null|(\^|>|<|>=|<=|%)?(\d{0,4}-?\d{0,2}-?\d{0,2} ?\d{0,2}:?\d{0,2}:?\d{0,2})(%|\$)?)$/',
```
cedric-anne, line 925: This will allow `20230701140005` format for `2023-07-01 14:00:05` datetime. Should we allow this?

Also, it does not seems to allow searching on partial date string. For instance, it is currently possible to search for  `2023`, `2023-07`, `2023-06-2` (from day 20 to 29), ... We should probably keep this feature.

tests/units/DB.php
```diff
@@ -856,4 +856,151 @@ public function testDecodeFetchResult(string $method, mixed $row, mixed $expecte
...
+                ]
+            ],
+            [
+                'params' => [
+                    'table' => 'glpi_calendarsegments',
+                    'field' => 'begin', //time
+                    'with_delimiter' => true
+                ],
+                'expected' => [
+                    '/^(NULL|null|(\^|>|<|>=|<=|%)?(\d{0,2}:?\d{0,2}:?\d{0,2})(%|\$)?)$/',
```
cedric-anne, line 947: Same remark here.

tests/units/DB.php
```diff
@@ -856,4 +856,151 @@ public function testDecodeFetchResult(string $method, mixed $row, mixed $expecte
...
+                ]
+            ],
+            [
+                'params' => [
+                    'table' => 'glpi_computers',
+                    'field' => '',
+                    'with_delimiter' => true
+                ],
+                'expected' => [
+                    '/^(NULL|null|(\^|>|<|>=|<=|%)?.*(%|\$)?)$/',
```
cedric-anne, line 969: It should probably return `'/^(.*)$/',`.

cedric-anne (changes requested):

tests/units/DB.php
Rom1-B: > Also, it does not seems to allow searching on partial date string. For instance, it is currently possible to search for `2023`, `2023-07`, `2023-06-2` (from day 20 to 29), ... We should probably keep this feature.

It's still possible

image

src/DBmysql.php
```diff
@@ -973,6 +973,56 @@ public function getField(string $table, string $field, $usecache = true): ?array
...
+     * @return string Field pattern
+     */
+    final public function getFieldPattern(string $table, string $field, bool $with_delimiter = true): array
+    {
+        global $DB;
+        $field_pattern = '.*';
+        $field_message = '';
+        if ($DB->tableExists($table) && ($field_attr = $this->getField($table, $field)) !== null) {
+            // Keep only main column type by removing anything that is preceded by a parenthesis (e.g. "(255)").
+            $field_type = preg_replace('/\s*\(.+$/', '', $field_attr['Type']);
```
cedric-anne, line 991:
```suggestion
            // Keep only main type by removing length parameters
            // e.g. `varchar(255)` -> `varchar`
            // e.g. `int(11) unsigned` -> `int unsigned`
            $field_type = trim(preg_replace('/^(\w+)\s*(?:\([^)]+\))?\s*(\w+)?$/', '$1 $2', $field_attr['Type']));
```

cedric-anne (commented):
A new test should be added, with a call to `tests\units\Search::doSearch()`, based on values provided by a data provider method, in order to validate that a given pattern for a given field produces the expected query.
The data provider should contain all the supported cases, in order to validate that we will not break existing saved searches.

tests/units/DB.php
```diff
@@ -913,4 +913,319 @@ function () {
...
+            'ENDYEAR',
+
+            // intervals
+            '2DAY',
+            '-2DAY',
+            '4WEEK',
+            '-1WEEK',
+            '3MONTH',
+            '-10MONTH',
+            '6YEAR',
```
cedric-anne, line 1041: @orthagh 

Should we support this kind of values in a `contains` search? It could potentially permit to do more precise search than with the limited list proposed with the `is` operator, but this is a bit more complex to define a pattern for it.

tests/units/DB.php
```diff
@@ -913,4 +913,323 @@ function () {
...
+            ),
+            'invalid_values' => array_merge(
+                $text_values,
+                $date_values,
+                $float_values,
+                $datetime_values,
+                $relative_datetime_values,
+            ),
+        ];
+        */
```
cedric-anne, line 1205: Tests were failing here due to some unexpected values allowed by the pattern, but I figured out that there is no `time` field that can be used in search, so I think we should just drop this case.

cedric-anne:
I rebased to be able to get result of newly added tests on `Search` class. They will probably fail, as you may have fix some issues.

cedric-anne (approved):

PR #14896, opened 2023-06-12
Ensure upload failure messages are always displayed

btry:
In Formcreator, we may have textarea fields with uploads enabled **and** without showing an area to drop files. 

With this context, if someone uploads a too big image (usually 2MB), the upload is rejected, but no feedback is provided to the user. The file seems uploaded successfully because it appears in the text area, and its name shows below it.

image

(test picture credit: a wallpaper picked [here](https://www.pexels.com/photo/reflection-of-mountain-on-lake-braies-1525041/) by Fancesco Ungaro) size : 5 MB.

With the patch, the upload failure becomes visible

image

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | internal ref 28203

### DIFF
--- a/src/Html.php
+++ b/src/Html.php
@@ -5547,21 +5547,18 @@ HTML;
                       . ($p['multiple'] ? " multiple='multiple'" : "")
                       . ($p['onlyimages'] ? " accept='.gif,.png,.jpg,.jpeg'" : "") . ">";
 
-        $progressall_js = '';
-        if (!$p['only_uploaded_files']) {
-            $display .= "<div id='progress{$p['rand']}' style='display:none'>" .
-                 "<div class='uploadbar' style='width: 0%;'></div></div>";
-            $progressall_js = "
-            progressall: function(event, data) {
-               var progress = parseInt(data.loaded / data.total * 100, 10);
-               $('#progress{$p['rand']}').show();
-               $('#progress{$p['rand']} .uploadbar')
-                  .text(progress + '%')
-                  .css('width', progress + '%')
-                  .show();
-            },
-         ";
-        }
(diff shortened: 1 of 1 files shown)


src/Html.php
```diff
@@ -5612,7 +5609,8 @@ function(index, file) {
                         $('#progress{$p['rand']}').show();
                         $('#progress{$p['rand']} .uploadbar')
                            .text(file.error)
-                           .css('width', '100%');
+                           .css('width', '100%')
+                           .show();
```
btry, line 5613: when a user uploads an image with acceptable size (< 2MB) this element shows a success message then fades out and display is disabled with style.

This line ensures it becomes visible to show the error when a 2nd image (too big) is dropped in the textarea)

src/Html.php
```diff
@@ -5547,21 +5547,18 @@ class='form-control'
                       . ($p['multiple'] ? " multiple='multiple'" : "")
                       . ($p['onlyimages'] ? " accept='.gif,.png,.jpg,.jpeg'" : "") . ">";
 
-        $progressall_js = '';
-        if (!$p['only_uploaded_files']) {
```
btry, line 5551: from docblock 
```
only_uploaded_files boolean  show only the uploaded files block, i.e. no title, no dropzone
     *                                   (should be false when upload has to be enable only from rich text editor)
```

then we can safely consider that the success / error feedback area should be rendered no matter the drop zone is rendered or not.

cedric-anne (approved):

trasher (approved):

PR #14908, opened 2023-06-13
Preventing php memory saturation from user.php page

kabassanov:
If additional columns are added for display through the front/user.php page, in particular Number of tickets as requester and Entities (Profile) or Profiles (Entity), displaying users with a lot of tickets increases significantly the amount of memory used in php and eventually goes over the max allowed memory limit (significantly = +/- 60Mio or more per user).

As additional bug, AJAX code fails and does not allow to increase the number of entries to display on the page. 

The reason is a per ticket concatenation of identical values  related to entities and profiles in the sql result.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

### DIFF
--- a/src/Search.php
+++ b/src/Search.php
@@ -1642,7 +1642,19 @@ class Search
                             $handled = false;
                             if ($fieldname != 'content' && is_string($val) && strpos($val, self::SHORTSEP) !== false) {
                                 $split2                    = self::explodeWithID(self::SHORTSEP, $val);
-                                if (is_numeric($split2[1])) {
+                                if ($j == "User_80") {
+                                    $newrow[$j][0][$fieldname] = $split2[0];
+                                    $newrow[$j][0]["profiles_id"] = $split2[1];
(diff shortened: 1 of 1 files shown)


orthagh (approved):

orthagh (commented):
Somehow, this changes the sub order of "Entity - Profiles" column, but it doesn't seems to be an issue

kabassanov:
> Somehow, this changes the sub order of "Entity - Profiles" column, but it doesn't seems to be an issue

Like if before the modification, GLPI was displaying in the order of last appearence of the Entity from the SQL result?

kabassanov:
> Somehow, this changes the sub order of "Entity - Profiles" column, but it doesn't seems to be an issue

If you think that the order matters, GROUP_CONCAT allows to define the order...

cconard96:
Functionally, this seems OK. My GLPI installation may not be complex enough to notice the performance difference.

cconard96 (approved):

PR #14402, opened 2023-03-28
[WIP] feat(core): webhooks implementation

stonebuzz:
Webhook implementation

- [x] Object ```Webhook```
- [x] ```CRUD``` operations for ```Webhook```
- [x] Check if ```secret``` is already used by another  ```Webhook```
- [x] Notice about ```HTTPS``` usage
- [x] ```CRA``` (Challenge–response authentication) implementation
- [x] ```twig``` template for  ```Webhook```
- [ ] Object ```WebhookQueued```
- [x] ```CRUD``` operations for ```Webhook```
- [ ] ```twig``` template for  ```WebhookQueued```
- [ ] Method to trigger and send data from API layer ([High-level API](https://github.com/glpi-project/glpi/pull/12221))
- [ ] TU

image

| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #number

PR #14916, opened 2023-06-14
Outgoing Webhooks

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !26398

Replaces #14402

### DIFF
--- a/front/webhook.form.php
+++ b/front/webhook.form.php
@@ -1,0 +1,118 @@
+<?php
+
+/**
+ * ---------------------------------------------------------------------
+ *
+ * GLPI - Gestionnaire Libre de Parc Informatique
+ *
+ * http://glpi-project.org
+ *
+ * @copyright 2015-2023 Teclib' and contributors.
+ * @copyright 2003-2014 by the INDEPNET Development Team.
+ * @licence   https://www.gnu.org/licenses/gpl-3.0.html
+ *
+ * ---------------------------------------------------------------------
+ *
+ * LICENSE
+ *
+ * This file is part of GLPI.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ * ---------------------------------------------------------------------
+ */
+
+use Glpi\Event;
+
+include('../inc/includes.php');
+
+Session::checkRight("config", READ);
+
+if (empty($_GET["id"])) {
+    $_GET["id"] = "";
+}
+
+$webhook = new Webhook();
+
+if (isset($_POST["add"])) {
+    $webhook->check(-1, CREATE);
+    if ($newID = $webhook->add($_POST)) {
+        Event::log(
+            $newID,
+            "webhook",
+            4,
+            "setup",
+            sprintf(__('%1$s adds the item %2$s'), $_SESSION["glpiname"], $_POST["name"])
+        );
+        if ($_SESSION['glpibackcreated']) {
+            Html::redirect($webhook->getLinkURL());
+        }
+    }
+    Html::back();
+} else if (isset($_POST["delete"])) {
+    $webhook->check($_POST["id"], DELETE);
+    if ($webhook->delete($_POST)) {
+        Event::log(
+            $_POST["id"],
+            "webhook",
+            4,
+            "setup",
+            //TRANS: %s is the user login
+            sprintf(__('%s deletes an item'), $_SESSION["glpiname"])
+        );
+    }
+    $webhook->redirectToList();
+} else if (isset($_POST["restore"])) {
+    $webhook->check($_POST["id"], DELETE);
+    if ($webhook->restore($_POST)) {
+        Event::log(
+            $_POST["id"],
+            "webhook",
+            4,
+            "webhook",
+            //TRANS: %s is the user login
+            sprintf(__('%s restores an item'), $_SESSION["glpiname"])
+        );
+    }
+    $webhook->redirectToList();
+} else if (isset($_POST["purge"])) {
+    $webhook->check($_POST["id"], PURGE);
+    if ($webhook->delete($_POST, 1)) {
+        Event::log(
+            $_POST["id"],
+            "webhook",
+            4,
+            "setup",
+            //TRANS: %s is the user login
+            sprintf(__('%s purges an item'), $_SESSION["glpiname"])
+        );
+    }
+    $webhook->redirectToList();
+} else if (isset($_POST["update"])) {
+    $webhook->check($_POST["id"], UPDATE);
+    if ($webhook->update($_POST)) {
+        Event::log(
+            $_POST["id"],
+            "webhook",
+            4,
+            "setup",
+            //TRANS: %s is the user login
+            sprintf(__('%s updates an item'), $_SESSION["glpiname"])
+        );
+    }
+    Html::back();
+} else {
+    $menus = ["config", Webhook::class];
+    Webhook::displayFullPageForItem($_GET["id"], $menus);
+}
--- a/inc/relation.constant.php
+++ b/inc/relation.constant.php
@@ -667,6 +667,8 @@ $RELATION = [
         'glpi_users'                       => 'entities_id',
         'glpi_vlans'                       => 'entities_id',
         'glpi_wifinetworks'                => 'entities_id',
+        'glpi_webhooks'                    => 'entities_id',
+        'glpi_queuedwebhooks'              => 'entities_id',
     ],
 
     'glpi_filesystems' => [
@@ -1638,6 +1640,9 @@ $RELATION = [
 
     'glpi_wifinetworks' => [
         'glpi_networkportwifis' => 'wifinetworks_id',
+    ],
+    'glpi_webhooks' => [
+        '_glpi_queuedwebhooks' => 'webhooks_id',
     ],
 
     'glpi_itilvalidationtemplates' => [
--- a/js/modules/Monaco/MonacoEditor.js
+++ b/js/modules/Monaco/MonacoEditor.js
@@ -1,0 +1,125 @@
+/**
+ * ---------------------------------------------------------------------
+ *
+ * GLPI - Gestionnaire Libre de Parc Informatique
+ *
+ * http://glpi-project.org
+ *
+ * @copyright 2015-2023 Teclib' and contributors.
+ * @copyright 2003-2014 by the INDEPNET Development Team.
+ * @licence   https://www.gnu.org/licenses/gpl-3.0.html
+ *
+ * ---------------------------------------------------------------------
+ *
+ * LICENSE
+ *
+ * This file is part of GLPI.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ * ---------------------------------------------------------------------
+ */
+
+window.GLPI = window.GLPI || {};
+
+/**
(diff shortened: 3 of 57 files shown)


orthagh:
Just to note as I checked, X-GLPI-signature should be verified against body . X-GLPI-timestamp, like explained in original rfe.

cconard96:
> Just to note as I checked, X-GLPI-signature should be verified against body . X-GLPI-timestamp, like explained in original rfe.

Is it not already the case that the headers are set properly and sent?
https://github.com/glpi-project/glpi/pull/14916/files#diff-7ed7e08a51abcf0217ec71739e11d15194e6d54108571c46c0efb7a6512276ceR820-R828
https://github.com/glpi-project/glpi/pull/14916/files#diff-98a543faffbe214c919517134ca5fa9149f738f537b6ba22166e9659d6e519fbR167-R176

orthagh:
Not exactly, the signature header should contains an hmac of the concatenation of body and timestamp, currently only the body is sent

orthagh (approved):
We're good from a functional pov. UX is great !
@cedric-anne, if you could a last round on this ?
After, we'll manage a new backport release to the customer.

src/Webhook.php
```diff
@@ -0,0 +1,1044 @@
...
+        ];
+        Search::showList(QueuedWebhook::class, $params);
+        return true;
+    }
+
+    /**
+     * Check if secret is already use dby another webhook
+     * @return array of webhook using same secret
+     */
+    public function getWebhookWithSameSecret()
```
cedric-anne, line 748:
```suggestion
    private function getWebhookWithSameSecret(): array
```

src/Webhook.php
```diff
@@ -0,0 +1,1044 @@
...
+
+    public static function signWebhookRequest($body, $secret)
+    {
+        $body = json_encode($body);
+        $timestamp = time();
+        $signature = self::getSignature($body . $timestamp, $secret);
+        header("X-GLPI-signature: " . $signature);
+        header("X-GLPI-timestamp: " . $timestamp);
+        echo $body;
+    }
```
cedric-anne, line 782: Method is not used.
```suggestion
```

src/Webhook.php
```diff
@@ -0,0 +1,1044 @@
...
+
+    public static function getSignature($data, $secret)
```
cedric-anne, line 784:
```suggestion
    public static function getSignature($data, $secret): string
```

src/Webhook.php
```diff
@@ -0,0 +1,1044 @@
...
+            self::send($data);
+        }
+    }
+
+    /**
+     * Send a webhook to the queue
+     * @param array $data The data for the webhook
+     * @return void
+     */
+    public static function send(array $data)
```
cedric-anne, line 935:
```suggestion
    public static function send(array $data): void
```

src/Webhook.php
```diff
@@ -0,0 +1,1044 @@
...
+            'webhooks_id' => $data['id'],
+            'url' => $data['url'],
+            'body' => $data['body'],
+            'event' => $data['event'],
+            'headers' => $data['headers'],
+        ]);
+        // Try sending the webhook immediately
+        if ($queued_webhooks_id !== false) {
+            QueuedWebhook::sendById($queued_webhooks_id);
+        }
```
cedric-anne, line 951: IMHO, this should not be done synchronously.

In the past, we had issues with some `QueuedNotification` that were sent synchronously and were sometine degrading performances when a ticket was created/updated, or were triggering an error that prevented, for instance, the mail collector to work properly.

templates/pages/setup/webhook/webhook_security.html.twig
```diff
@@ -0,0 +1,199 @@
...
+      ) }}
+
+      {% if secret_already_used is not empty %}
+         {% set warning_secret_already_used %}
+            <div class="alert alert-warning align-items-center flex-column" role="alert">
+               <i class='ti ti-alert-triangle-filled'></i>
+               <span class='ms-2'> {{ __('Secret is already use by another webhooks') }}</span>
+               <span>
+                  {% for webhook in secret_already_used %}
+                     <li  class='ms-4'>{{ webhook.link|raw }}</li>
```
cedric-anne, line 60: Here we could prevent usage of `|raw`, either by passing an array of `Webhook` in `secret_already_used` to be able to use `webhook.getLink()`, either by passing and array of ids to be able to use `{{ get_item_link('Webhook', webhook_id) }}`.

cedric-anne (changes requested):

src/QueuedWebhook.php
```diff
@@ -0,0 +1,495 @@
...
+
+        $client = new \GuzzleHttp\Client($options);
+        $headers = json_decode($queued_webhook->fields['headers'], true);
+        try {
+            $response = $client->request($webhook->fields['http_method'], $queued_webhook->fields['url'], [
+                \GuzzleHttp\RequestOptions::HEADERS => $headers,
+                \GuzzleHttp\RequestOptions::BODY => $queued_webhook->fields['body'],
+            ]);
+        } catch (\GuzzleHttp\Exception\GuzzleException $e) {
+            //TODO log the error
```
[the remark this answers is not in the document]
orthagh: Agree

js/modules/Monaco/MonacoEditor.js
```diff
@@ -0,0 +1,125 @@
...
+     * @param {string} language The code language
+     * @param {string} value The default value for the editor
+     * @param {CompletionItemDefinition[]} completions List of completion items
+     */
+    constructor(element_id, language, value = '', completions = []) {
+        const el = document.getElementById(element_id);
+        const trigger_characters = {
+            twig: ['{', ' '],
+        };
+        window.monaco.languages.registerCompletionItemProvider(language, {
```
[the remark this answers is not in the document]
cconard96: I'm not sure it is worth it at this time.
Completion data is globally managed in the `window.monaco` object which affects any editor created from it.
If plugins really want to modify this, they have access to the window global. Having it done as methods on the MonacoEditor class could confuse them into thinking they are only changing the behavior of that one editor.
Also, you can have multiple providers at a time. So, if you really want to replace the provider, you need to track all the Disposable items returned by registerCompletionItemProvider.

I don't see a need for a plugin to control the behavior of the trigger characters or provider. They do have the option of passing custom completion data (variables, functions, etc) per-editor though.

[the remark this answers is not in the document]
cconard96: > I don't see a need for a plugin to control the behavior of the trigger characters or provider.

To clarify, we control the available languages on the GLPI side because it is in the webpack config. So, if a language is allowed from the core, we can provide and maintain a set of good providers for them too (if the default isn't sufficient).

templates/pages/setup/webhook/webhook_security.html.twig
cconard96: This part seems odd in general. I'm not sure it is a good idea to tell anyone which webhooks have the same secret.
It doesn't make any difference from a security standpoint to just say "Secret is already use by another webhooks" without the list.

src/Webhook.php
```diff
@@ -0,0 +1,1031 @@
...
+    public function handleInput($input)
+    {
+        //empty choice (0) update to empty ('')
+        if (isset($input["itemtype"]) && !$input["itemtype"]) {
+            Session::addMessageAfterRedirect(__('An item type is required'), false, ERROR);
+        }
+
+        if (isset($input["event"]) && !$input["event"]) {
+            Session::addMessageAfterRedirect(__('An event is required'), false, ERROR);
+        }
```
cedric-anne, line 965:
```suggestion
        $valid_input = true;

        if (isset($input["itemtype"]) && !$input["itemtype"]) {
            Session::addMessageAfterRedirect(__('An item type is required'), false, ERROR);
            $valid_input = false;
        }

        if (isset($input["event"]) && !$input["event"]) {
            Session::addMessageAfterRedirect(__('An event is required'), false, ERROR);
            $valid_input = false;
        }

        if (!$valid_input) {
            return false;
        }
```

templates/pages/setup/webhook/webhook_security.html.twig
cedric-anne: Is it really a problem to have webhooks that share the same secret? If you need to send multiple events for a single object (e.g. Ticket::add and Ticket::update), I guess the target application admin will probably ask you if it is possible to use a unique secret, because it will consider your two webhooks as the same source.

Also, since this is an auto-generated field, it is intentional to have the same secret for multiple webhooks. Therefore, we could consider that the webhooks administrator is aware of what he is doing and that he considers that in this specific case there is no risk in sharing the same secret.

inc/relation.constant.php
```diff
@@ -1639,6 +1641,9 @@
     'glpi_wifinetworks' => [
         'glpi_networkportwifis' => 'wifinetworks_id',
     ],
+    'glpi_webhooks' => [
+        '_glpi_queuedwebhooks' => 'webhooks_id',
```
cedric-anne, line 1645: You should add the following method to `Webhook`, to automatically delete all queued webhooks when a webhook is deleted from DB.

```php
    public function cleanDBonPurge()
    {

        $this->deleteChildrenAndRelationsFromDb(
            [
                QueudWebhook::class,
            ]
        );
    }
```

front/webhook.form.php
```diff
@@ -0,0 +1,118 @@
...
+            $newID,
+            "webhook",
+            4,
+            "setup",
+            sprintf(__('%1$s adds the item %2$s'), $_SESSION["glpiname"], $_POST["name"])
+        );
+    }
+    if ($_SESSION['glpibackcreated']) {
+        Html::redirect($webhook->getLinkURL());
+    }
```
cedric-anne, line 61: Redirection to created item can only be made if item is actually created.
```suggestion
        if ($_SESSION['glpibackcreated']) {
            Html::redirect($webhook->getLinkURL());
        }
    }
```

templates/pages/setup/webhook/webhook_security.html.twig
cconard96: I don't think so, but this alert message was defined in the specs for this feature.

orthagh: I don't remember precisely why we wrote about the duplication alert.
But I think we can remove it. As said, normal usage will generate different secret for each webhooks. If someone really want to have one shared secret, I guess it doesn't really matter.

src/WebhookTest.php
```diff
@@ -0,0 +1,111 @@
...
+use Glpi\Application\View\TemplateRenderer;
+
+class WebhookTest extends CommonGLPI
+{
+    public static $rightname         = 'config';
+
+
+    public static function getTypeName($nb = 0)
+    {
+        return _n('Test', 'Test', $nb);
```
cedric-anne, line 45: To fix lint.
```suggestion
        return __('Test');
```

Or maybe, as @orthagh suggests it to me IRL, change it to `Preview`, that will may be more appropriate as the request is not actually sent.

cedric-anne (commented):
Seems technically OK for me. I did not tested the full chain, but I trust @ortagh and the customer to detect any potential issue.

flegastelois:
Queries log tab refresh is incorrect, on successfull manual send, all others queries logs are displayed (and not only those related to the current webhook).

queries log OK, click to "Send":
image

on refresh, others weebhook queries log are displayed:
image

flegastelois:
Is it possible to add "Event" itemtype?

cconard96:
> Is it possible to add "Event" itemtype? or more generally, all compatible itemtypes rather than a hard-coded list (getAPIItemtypeData)? To avoid having a new version to release if someone wants an extra itemtype (while it is functionally supported).
> 
> Maybe a new "CFG_GLPI['webhook_types']"?

It should be possible in the future to support everything that the new API provides data for. The biggest challenge is finding the correct endpoints for a specific itemtype and the path attribute names. Mainly, the relation item types. It was scaled back to be able to be able to deliver this feature quickly to the client. Events are not supported in the new API yet.

cconard96:
> Queries log tab refresh is incorrect, on successfull manual send, all others queries logs are displayed (and not only those related to the current webhook).
> 
> queries log OK, click to "Send": image
> 
> on refresh, others weebhook queries log are displayed: image

Should be fixed with the last commit.

cedric-anne:
@cconard96

Could you fix conflicts ?

cedric-anne (changes requested):
Tests are failing.

cedric-anne (approved):
Seems technically OK.

cedric-anne:
Validated IRL by @orthagh .

oneillsun:
where/when we'll see this new feature?

cconard96:
> where/when we'll see this new feature?

GLPI 10.1, the next major release. I don't have a release timeframe.

HydrelioxGitHub:
Any news on the release timeframe. I would like to use this feature to integrate notification in our internal messaging app. My boss keep me asking when it should be available ....

orthagh:
Currently 10.1 is in dev progress, so no date yet.
Please contact us on https://glpi-project.org/contact_us/ if you want a backport for your version

ldgmaia:
We can already create webhooks on Microsoft Teams, and I was wondering if it will be possible to customize the JSON data sent by this new webhook feature. Sending messages to Teams requires a specific JSON structure, such as when creating a card. I recall that there wasn’t a release timeframe mentioned last October, but I was curious if there’s any update on that now. Additionally, if there’s anything I can do to help or contribute, I’d be more than happy to assist!

PR #14966, opened 2023-06-20
Remove "Personal token" display

ccailly:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

This field no longer seems to be used and is a source of confusion.

image

### DIFF
--- a/src/User.php
+++ b/src/User.php
@@ -2866,29 +2866,6 @@ HTML;
                 echo "<tr class='tab_bg_1'><th colspan='4'>" . __('Remote access keys') . "</th></tr>";
 
                 echo "<tr class='tab_bg_1'><td>";
-                echo __("Personal token");
-                echo "</td><td colspan='2'>";
-
-                if (!empty($this->fields["personal_token"])) {
-                    echo "<div class='copy_to_clipboard_wrapper'>";
-                    echo Html::input('_personal_token', [
-                        'value'    => $this->fields["personal_token"],
-                        'style'    => 'width:90%'
-                    ]);
-                    echo "</div>";
-                    echo "(" . sprintf(
-                        __('generated on %s'),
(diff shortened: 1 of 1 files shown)


cconard96:
Seems to be used in some way with the planning feature.
https://github.com/glpi-project/glpi/blob/ec088dbb0b62f656b089b809874cd111b9573bbd/front/planning.php#L67-L75
https://github.com/glpi-project/glpi/blob/ec088dbb0b62f656b089b809874cd111b9573bbd/src/Planning.php#L1062-L1066

`User::getAuthToken` has a parameter for a field name which defaults to 'personal_token'.

AdrienClairembault:
> Seems to be used in some way with the planning feature.
> 
> https://github.com/glpi-project/glpi/blob/ec088dbb0b62f656b089b809874cd111b9573bbd/front/planning.php#L67-L75
> 
> https://github.com/glpi-project/glpi/blob/ec088dbb0b62f656b089b809874cd111b9573bbd/src/Planning.php#L1062-L1066
> 
> `User::getAuthToken` has a parameter for a field name which defaults to 'personal_token'.

I think @orthagh said that the planning part does not require the user to fetch the token from its preferences, there is a button in the planning to get a formatted url containing the token.

AdrienClairembault (approved):

orthagh (approved):

orthagh:
> I think @orthagh said that the planning part does not require the user to fetch the token from its preferences, there is a button in the planning to get a formatted url containing the token.

Exactly, the remaining concern should be the regenerate checkbox, but in the case of webcal planning and as the token does not service anymore features (so no security issue anymore), I guess the it's not usefull to keep it also.

cedric-anne (approved):

PR #14978, opened 2023-06-21
Fix "read" right for reservations

ccailly:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !28443

Added the ability to view reservations with the `READ` right.
Until now, the `ReservationItem::RESERVEANITEM` right was required to display reservations.
The `READ` right was already present but had no effect.

### DIFF
--- a/js/reservations.js
+++ b/js/reservations.js
@@ -43,6 +43,7 @@ var Reservations = function() {
     this.license_key = null;
     this.currentv    = null;
     this.defaultDate = null;
+    this.can_reserve = true;
 
     var my = this;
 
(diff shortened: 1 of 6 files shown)


js/reservations.js
```diff
@@ -227,17 +231,19 @@ var Reservations = function() {
             // ADD EVENTS
             selectable: true,
```
AdrienClairembault, line 232: I think we can also disable `selectable` if the user can't add items.

```suggestion
            selectable: my.can_reserve,
```

AdrienClairembault (changes requested):

AdrienClairembault (approved):

orthagh (approved):

cedric-anne (approved):

PR #14983, opened 2023-06-22
Add redirection if user lacks rights to view the page upon profile change

ccailly:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

When a user changes their profile in GLPI, the current page is reloaded. However, the new active profile does not guarantee that the user has the necessary rights to view the page, which can result in a permissions error. This pull request aims to address this issue by redirecting the user to the homepage if they lack the required rights. If the user has the necessary rights, no changes are made.

### DIFF
--- a/src/Html.php
+++ b/src/Html.php
@@ -626,6 +626,7 @@ class Html
      **/
     public static function displayRightError(string $additional_info = '')
     {
+        Toolbox::handleProfileChangeRedirect();
         $requested_url = (isset($_SERVER['REQUEST_URI']) ? $_SERVER['REQUEST_URI'] : 'Unknown');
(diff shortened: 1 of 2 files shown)


AdrienClairembault (changes requested):
I think you should put this call directly in `Html::displayRightError()`, to cover all possibles entry points (search for `displayRightError` and you will see that it can be called from other places that `Session::checkRight`).

AdrienClairembault (approved):

orthagh (approved):

cedric-anne (approved):

PR #14992, opened 2023-06-22
Debug bar UX

AdrienClairembault:
Some UI and UX suggestions for the debug bar

Before:

image

After;

image

I have a few other changes in mind but I'll split them in another PR to keep things readable.

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |

### DIFF
--- a/js/modules/Debug/Debug.js
+++ b/js/modules/Debug/Debug.js
@@ -392,11 +392,11 @@ window.GLPI.Debug = new class Debug {
 
         // Database performance
         const sql_data = this.getCombinedSQLData();
-        const database_button_label = `${sql_data.total_requests} <span class="text-muted"> in </span> ${sql_data.total_duration} <span class="text-muted"> ms </span>`;
+        const database_button_label = `${sql_data.total_requests} <span class="text-muted"> queries </span>`;
         this.getWidgetButton('sql').find('.debug-text').html(database_button_label);
 
         // Requests
-        this.getWidgetButton('requests').find('.debug-text').text(this.ajax_requests.length);
+        this.getWidgetButton('requests').find('.debug-text').html(`${this.ajax_requests.length} <span class="text-muted"> requests </span>`);
 
         // Client performances
         const dom_timing = window.performance.getEntriesByType('navigation')[0].domComplete;
--- a/js/modules/Debug/Debug.js
+++ b/js/modules/Debug/Debug.js
@@ -348,11 +348,13 @@ window.GLPI.Debug = new class Debug {
     }
 
     showDebugToolbar() {
+        $('.debug-logo').prop('disabled', true);
         $('.debug-toolbar-content').removeClass('d-none');
         $('#debug-toolbar').addClass('w-100').css('width', null);
     }
 
     hideDebugToolbar() {
+        $('.debug-logo').prop('disabled', false);
         $('.debug-toolbar-content').addClass('d-none');
         $('#debug-toolbar-expanded-content').addClass('d-none');
         $('#debug-toolbar').removeClass('w-100').css('width', 'fit-content');
--- a/js/modules/Debug/Debug.js
+++ b/js/modules/Debug/Debug.js
@@ -382,15 +382,35 @@ window.GLPI.Debug = new class Debug {
     }
 
     refreshWidgetButtons() {
-        const server_performance_button = this.getWidgetButton('server_performance');
+        // Server performance
         const server_perf = this.initial_request.server_performance;
         const memory_usage_mio = (server_perf.memory_usage / 1024 / 1024).toFixed(2);
-        server_performance_button.find('.debug-text').text(`${server_perf.execution_time}ms | ${memory_usage_mio}mio`);
+        const server_performance_button_label = __("$1s ms using $2s mio")
+            .replace("$1s", `${server_perf.execution_time} <span class="text-muted">`)
+            .replace("$2s", `</span> ${memory_usage_mio} <span class="text-muted">`)
+            + "</span>"
+        ;
+        this.getWidgetButton('server_performance').find('.debug-text').html(server_performance_button_label);
 
-        const ajax_requests_button = this.getWidgetButton('ajax_requests');
-        ajax_requests_button.find('.debug-text').text(this.ajax_requests.length);
+        // Database performance
+        const sql_data = this.getCombinedSQLData();
+        const database_button_label = __("$1s in $2s ms")
+            .replace("$1s", `${sql_data.total_requests} <span class="text-muted">`)
+            .replace("$2s", `</span> ${sql_data.total_duration} <span class="text-muted">`)
+            + "</span>"
+        ;
+        this.getWidgetButton('sql').find('.debug-text').html(database_button_label);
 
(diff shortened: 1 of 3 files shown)


cconard96 (approved):

templates/components/debug/debug_toolbar.html.twig
```diff
@@ -34,35 +34,35 @@
...
          </button>
       </div>
-      <div class="debug-toolbar-content p-1 d-flex w-100 justify-content-between">
+      <div class="debug-toolbar-content d-flex w-100 justify-content-between align-items-center">
          <ul class="debug-toolbar-widgets nav nav-tabs align-items-center border-0" data-bs-toggle="tabs">
-            {{ _self.widget_button('server_performance', _x('debug', 'Server performance')) }}
+            {{ _self.widget_button('server_performance', _x('debug', 'Server performance'), 'ti ti-clock-play') }}
             {{ _self.widget_button('sql', _x('debug', 'SQL Requests'), 'ti ti-database') }}
-            {{ _self.widget_button('requests', _x('debug', 'Requests'), 'ti ti-refresh') }}
+            {{ _self.widget_button('requests', _x('debug', 'HTTP Requests'), 'ti ti-refresh') }}
```
trasher, line 59: Translations are frozen for 10.0.8.

Also, as discussed today in our channel, I wonder if debug bar strings should really be translated; it sounds like a waste of time for translators. Keeping debug bar in English only could make sense.

So, either we remove those translations, either you put the string back (and change it after 10.0.8 has been released).

js/modules/Debug/Debug.js
```diff
@@ -382,15 +384,35 @@ window.GLPI.Debug = new class Debug {
...
+        // Server performance
         const server_perf = this.initial_request.server_performance;
         const memory_usage_mio = (server_perf.memory_usage / 1024 / 1024).toFixed(2);
-        server_performance_button.find('.debug-text').text(`${server_perf.execution_time}ms | ${memory_usage_mio}mio`);
-
-        const ajax_requests_button = this.getWidgetButton('ajax_requests');
-        ajax_requests_button.find('.debug-text').text(this.ajax_requests.length);
-
-        this.getWidgetButton('requests').find('.debug-text').text('Requests');
+        const server_performance_button_label = __("$1s ms using $2s mio")
```
trasher, line 390: See my other comment on translations. Also, if we keep translations on debug bar, this one should use the same domain as all other.

trasher (changes requested):
See commeents

templates/components/debug/debug_toolbar.html.twig
cedric-anne: > I wonder if debug bar strings should really be translated

When we have to do support, we may sometimes ask what information is displayed in the debug bar. When I made translations on this part yesterday, I realized that some terms are difficult to translate and that it might be better to leave everything in English to avoid misunderstandings due to translation errors.

js/modules/Debug/Debug.js
```diff
@@ -382,15 +384,35 @@ window.GLPI.Debug = new class Debug {
...
+            + "</span>"
+        ;
+        this.getWidgetButton('server_performance').find('.debug-text').html(server_performance_button_label);
+
+        // Database performance
+        const sql_data = this.getCombinedSQLData();
+        const database_button_label = __("$1s in $2s ms")
+            .replace("$1s", `${sql_data.total_requests} <span class="text-muted">`)
+            .replace("$2s", `</span> ${sql_data.total_duration} <span class="text-muted">`)
+            + "</span>"
```
cedric-anne, line 402: As said in #14594, this is weird to display at the same place the time/weight of initial request and the sum of all SQL queries for both the initial request and all ajax requests.
I think we should hide SQL queries count. It is confusing to have an increment on every ajax request (especially when ajax notifications are active and increase the counter very often).

cedric-anne: Maybe the `Server performance` button/tab could be removed.
I do not think it is necessary to highlight the initial request time and memory consumption, especially since in GLPI, the majority of the execution time is often spent loading the current tab, which is done in a AJAX query.

AdrienClairembault: > I think we should hide SQL queries count. It is confusing to have an increment on every ajax request (especially when ajax notifications are active and increase the counter very often).

I kinda like seeing the request count tbh.

Maybe some endpoints like ajax notifications could simply be blacklisted ?

templates/components/debug/debug_toolbar.html.twig
AdrienClairembault: Ok to remove translations for the debug bar if everyone agrees

AdrienClairembault:
Also as a side note, I've seen that the clients side metrics use deprecated browser functions like https://developer.mozilla.org/en-US/docs/Web/API/PerformanceTiming/domComplete.

I don't know if this was discussed in the previous pull request.

js/modules/Debug/Debug.js
cedric-anne: Problem is not on the count but on the total time. Having both initial time in "Server performance" tab title and total SQL time in "SQL" tab title may be confusing.

cedric-anne: I did not said that information is useless, but as it can be found easilly in the `HTTP requests` tab, I am not sure having it highlighted here is mandatory.

templates/components/debug/debug_toolbar.html.twig
cedric-anne: @orthagh @cconard96 do you think that translations could be removed here?

cconard96:
> Also as a side note, I've seen that the clients side metrics use deprecated browser functions like https://developer.mozilla.org/en-US/docs/Web/API/PerformanceTiming/domComplete.
> 
> I don't know if this was already discussed in the previous pull request.

`window.performance.timing` is the deprecated part along with a few other properties. For the timing property, it is a PerformanceTiming object which MDN says is deprecated. This feature uses `window.performance.getEntriesByType` to get `PerformanceNavigationTiming` objects instead which seem to be the correct replacement.

As far as I know, just the performance metrics directly in `window.performance` are deprecated and we should use the `getEntries*` functions to get them instead.

templates/components/debug/debug_toolbar.html.twig
orthagh: I agree

cconard96: > @orthagh @cconard96 do you think that translations could be removed here?

Yes.

cedric-anne: Done in #15007 . You will have to rebase.

js/modules/Debug/Debug.js
AdrienClairembault: I agree with @cconard96 that it's an important metric that should stay on the bar (and in the future, GLPI may not load the maj tab through AJAX).

AdrienClairembault: Latest changes after our discussion @cedric-anne 

image

cedric-anne (approved):

orthagh (approved):

PR #14994, opened 2023-06-22
refs #14500

crasyhorse:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #14500

### DIFF
--- a/src/Config.php
+++ b/src/Config.php
@@ -2965,6 +2965,10 @@ HTML;
             $CFG_GLPI['planning_work_days'] = importArrayFromDB($CFG_GLPI['planning_work_days']);
         }
 
+        if (isset($CFG_GLPI['mailing_signature'])) {
+            $CFG_GLPI['mailing_signature'] = Sanitizer::unsanitize($CFG_GLPI['mailing_signature']);
+        }
+
         return true;
     }
 


cedric-anne (changes requested):
Fo security reasons, call to `Sanitizer::unsanitize()` should be done only when variable is used in a "secure" context.

For instance:
```diff
- $signature_html = RichText::getSafeHtml($this->signature);
+ $signature_html = RichText::getSafeHtml(Sanitizer::unsanitize($this->signature));
```

It is a pain, but it will change in GLPI 10.1 thanks to #14302.

Also, if we officially support HTML tags in signatures, following changes should be done:
 - corresponding search option should probably be updated, if any,
 - in both notification settings and entity config forms, the rich text editor should be enabled,
 - the `RichText::getTextFromHtml()` method should be used everywhere the signature is used to generate `plain-text` messages.

trasher:
I wa snot entirely sure, tha'ts why I aksed you @cedric-anne a review... But this is a new feature; therefore it should target main branch, not 10.0/bugfixes (as it seems it will require more changes that the one proposed).

crasyhorse:
Thanks for the review. Please keep me posted if you know how to proceed with this issue.

PR #14998, opened 2023-06-23
fix(inventory): fix ID keep in memory (and use from SoftwareVersion)

stonebuzz:
```operatingystems_id``` used by ```SoftwareVersion``` comes from ```\Glpi\Inventory\Asset\OperatingSystem->getId()```

But from ```OperatingSystem``` when an ```OS``` is handled, ```getId()```  return ID of link between ```Computer``` and ```OS```

```php
$this->operatingsystems_id =  $ios->fields['id'];
``` 

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !28498


trasher:
Could you please add a test?

trasher (approved):

cedric-anne (approved):

PR #15003, opened 2023-06-24
feat(user) Adding last activity record for users.

miguelanruiz:
Hi guys, 

I was thinking about helping to build the last access mechanism for users, it really seems to me a powerful utility to be exploited by plugins in the future, but for now I think that instead of using a socket it would be possible to handle a timeout script to manage the time update from the browser.

An ajax method, could be a solution to transfer the state of the user to GLPI, to store the data would have to extend the schema and through this file (Session:#L972) that in each successful check this last hour data is updated.

For now I am working on the proof of concept, clearly I know that much of the code is missing, but I wanted to have a clear opinion from your side. The main idea is that with time the forms will allow to consult and present this last access as an indicator that the user is online or offline, maybe with a colored light bulb or similar.

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 0


miguelanruiz:
Im not sure yet how to include last_access to db (Manually for now) from install empty or update, but for now i can show you how as i expect it will approximately look:

Logged and browser opened:
image
image

Recently logged:
image
image

Account never used:
image
image

As i mentioned i included the last_access update precisely in each session check, since it seems to me opportune and simple. I think timediff threshold can be a const value for User class

miguelanruiz:
It is complete and functional

miguelanruiz:
Without comments, i assume that this does not enough

cconard96:
> Without comments, i assume that this does not enough

Not necessarily. With bug fixes and other work, there may not have been time to review/discuss this feature proposal yet.

miguelanruiz:
its okay, i will reopen even leave it open to proposals for changes that I consider necessary and possible improvements.

cedric-anne (changes requested):
Sorry for the really late reply.

I know that this feature was requested multiple times in the past and I guess it would be appreciated by the community.

Some changes should be made to make this PR acceptable.

First, polling the server every 5 seconds to update the last time a user was seen may generate a lot of request on a GLPI instance that is massively used by a lot of users. The delay should be increased to reduce the polling rate. I do not know how this information will be used, but having this information refreshed every 30 seconds seems enough. Also, the polling should stop when the page is not displayed (when the tab is inactive for instance), maybe using the `document.visibilityState` property.

Second, adding a DB update query each time the session is checked does not seems to be a good idea. This check can be triggered multiple times on the same request, so it may add some unexpected load. Updating the database only from the AJAX polling logic seems sufficient.

Also, as this is a new feature, it should target the `main` branch.

cedric-anne:
I am closing this PR because it seems stale. Feel free to reopen it if you wish to continue working on it.

PR #15043, opened 2023-06-29
fix(knowbase): search article with multi categories

Rom1-B:
When an article belongs to several categories and you filter on the category in the search, in the tree you have the counter on the categories excluded by the filter, but no result when you open the category.

Before:
image

After:
image

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

### DIFF
--- a/src/Features/TreeBrowse.php
+++ b/src/Features/TreeBrowse.php
@@ -203,13 +203,25 @@ JAVASCRIPT;
         $cat_fk    = $cat_itemtype::getForeignKeyField();
         $cat_join = $itemtype . '_' . $cat_itemtype;
 
+        // This query is used to get the IDs of all results matching the search criteria
+        $sql = $data['sql']['search'];
+        // We can remove all the SELECT fields and replace it with just the ID field
+        $raw_select = $data['sql']['raw']['SELECT'];
+        $replacement_select = "SELECT DISTINCT " . $cat_join::getTableField($cat_fk);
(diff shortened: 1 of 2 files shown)


src/Features/TreeBrowse.php
```diff
@@ -203,13 +203,25 @@ public static function getTreeCategoryList(string $itemtype, array $params): arr
         $cat_fk    = $cat_itemtype::getForeignKeyField();
         $cat_join = $itemtype . '_' . $cat_itemtype;
 
+        // This query is used to get the IDs of all results matching the search criteria
+        $sql = $data['sql']['search'];
+        // We can remove all the SELECT fields and replace it with just the ID field
+        $raw_select = $data['sql']['raw']['SELECT'];
```
trasher, line 209:
```suggestion
```
Since this is already done just before

trasher (approved):
I'm not at all fan of replacement solution, but since this is based on Search; I think we currently have no other solution. ALso, this is already done just like this just before; so I guess this is OK.

trasher (approved):

stonebuzz:
When a search does not contains criteria about "category" an error SQL occurs

image

AdrienClairembault:
FYI, this break searches on documents:

image

cedric-anne (approved):
Seems OK, but it is a real pain to read and would probably a pain to maintain.

trasher (approved):
LGTM, but I'm very unsure :D

AdrienClairembault (approved):

PR #15052, opened 2023-06-30
Debug bar formating + Use datagrid instead of html tables

AdrienClairembault:
Somes changes suggested by @cedric-anne regarding formatting :
* Round X.00 values
* Always use spaces before units
* mio -> MB

I've also replaced the HTML tables by [tabler's datagrids](https://tabler.io/docs/components/datagrid) + added some margins/padding + reduced titles to h2.

Server perf before:

image

Server perf after:

image

Client perf before:

image

Client perf after:

image

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |

### DIFF
--- a/js/modules/Debug/Debug.js
+++ b/js/modules/Debug/Debug.js
@@ -729,15 +729,22 @@ window.GLPI.Debug = new class Debug {
             const total_heap = perf.memory.totalJSHeapSize / 1024 / 1024;
 
             // Non-standard feature supported by Chrome
-            content_area.find('tbody').append(`
-                <tr><th colspan="4">Memory</th></tr>
-                <tr>
-                    <th>Used JS Heap</th><td>${used_heap.toFixed(2)}mio</td>
-                    <th>Total JS Heap</th><td>${total_heap.toFixed(2)}mio</td>
-                </tr>
-                <tr>
-                    <th>JS Heap Limit</th><td>${heap_limit.toFixed(2)}mio</td>
-                </tr>
+            content_area.find('.datagrid:last').append(`
+                <<h3 class="mt-3 mb-2">Memory</h3>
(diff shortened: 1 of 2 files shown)


orthagh (approved):

cconard96 (approved):

js/modules/Debug/Debug.js
```diff
@@ -386,7 +386,7 @@ window.GLPI.Debug = new class Debug {
     refreshWidgetButtons() {
         // Server performance
         const server_perf = this.initial_request.server_performance;
-        const memory_usage_mio = (server_perf.memory_usage / 1024 / 1024).toFixed(2);
+        const memory_usage_mio = +(server_perf.memory_usage / 1024 / 1024).toFixed(2);
```
cedric-anne, line 389: Maybe we should replace `/ 1024 / 1024` by `/ 1000 / 1000` if we use the `MB` unit.

cedric-anne, line 389:
```suggestion
        const memory_usage_mb = +(server_perf.memory_usage / 1000 / 1000).toFixed(2);
```

trasher:
Please fix CS

js/modules/Debug/Debug.js
cedric-anne: I put back MiB values, because this is the unit used by PHP for the memory limit configuration.

With a `memory_limit=128M` value, if we expose MB, the max value will seems weird.
image

If we expose MiB, it looks more correct:
image

cedric-anne (approved):

cconard96 (approved):

orthagh (approved):

PR #14776, opened 2023-05-25
fix(search): &nbsp; visible in dropdown values

Rom1-B:
In the ticket search (I didn't find any cases elsewhere) in Portuguese, on some values of the dropdown criteria `&nbsp;` was not interpreted, but was on others.

Here, the 2 values under "SLAs" and "OLA" had `&nbsp;` :

image

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !28078

PR #15093, opened 2023-07-03
fix(html): use sprintf instead of concat space

stonebuzz:
Prevent ```&nbsp;```

image

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !28636

### DIFF
--- a/src/Ticket.php
+++ b/src/Ticket.php
@@ -3260,7 +3260,7 @@ JAVASCRIPT;
             'id'                 => '32',
             'table'              => 'glpi_slalevels',
             'field'              => 'name',
-            'name'               => __('SLA') . ' ' . _n('Escalation level', 'Escalation levels', 1),
+            'name'               => sprintf('%s %s', __('SLA'), _n('Escalation level', 'Escalation levels', 1)),
             'massiveaction'      => false,
             'datatype'           => 'dropdown',
             'joinparams'         => [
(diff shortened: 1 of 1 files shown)


src/Ticket.php
```diff
@@ -3260,7 +3260,7 @@ public function rawSearchOptions()
             'id'                 => '32',
             'table'              => 'glpi_slalevels',
             'field'              => 'name',
-            'name'               => __('SLA') . ' ' . _n('Escalation level', 'Escalation levels', 1),
+            'name'               => sprintf('%s %s', __('SLA'), _n('Escalation level', 'Escalation levels', 1)),
```
trasher, line 3263: Hum... "SLA escalation level" seems good in english, but "SLA niveau d'escalade" sounds weird in French.

Pattern should be translated in order to fix that, but I guess this is a terrible idea for translators; it should be better to create both "SAL escalation level" and "OLA escalation level".

BTW, current fix proposal is OK for 10.0.8, new strings won't be introduced before release.

trasher: Alternatively, since this is displayed as a sub element in the drodown, we can just keep "Escalation level"; should'nt we? This would fix all issues.

cedric-anne (commented):
I think it has been fixed by #14776.

stonebuzz:
> I think it has been fixed by https://github.com/glpi-project/glpi/pull/14776.

Yep

we can close this one

PR #15123, opened 2023-07-05
Fix computation of joins inside nested objects

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

While working on https://github.com/pluginsGLPI/fields/pull/670 I found that joins were not working for the custom fields.
Until now, joins where only needed for top-level properties like the `location` property. The nested properties like `location.id` were always scalar properties.
For the fields plugin, the complex nature of the DB schema where each field is in its own table, the joins were nested. There is a `custom_fields` object property and then for each field, there is a child object which has the join info.

This is still a draft until I fix remaining issues with the fields plugin integration.


cconard96:
Replaced by #15788

PR #15217, opened 2023-07-13
Prevent invalid predefined associated item in template

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #15197

### DIFF
--- a/src/ITILTemplatePredefinedField.php
+++ b/src/ITILTemplatePredefinedField.php
@@ -74,6 +74,17 @@ abstract class ITILTemplatePredefinedField extends ITILTemplateField
             }
         }
 
+        if ((int) $input['num'] === 13) {
+            if ((string) $input['value'] === '0') {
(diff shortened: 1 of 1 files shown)


src/ITILTemplatePredefinedField.php
```diff
@@ -74,6 +74,17 @@ public function prepareInputForAdd($input)
             }
         }
 
+        if ((int) $input['num'] === 13) {
```
trasher, line 77: Where is this `13` from?

trasher: OK, maybe add a comment then, I'm pretty sure I'll forgot that in 5 minutes :D

cedric-anne (approved):

PR #15218, opened 2023-07-13
Try fixing battery wear level display from device model form

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #15099

When looking at batteries from the DeviceBattery form on the Items tab, it was trying to get the capacity from the asset rather than the device so nothing was showing and a warning was thrown. From the asset form, it was displaying correctly.

This PR fixes it by always using the device rather than the peer (the opposite of whichever type you are looking at). It also makes the available information visible when the individual battery has a capacity set but the battery model does not. Finally, it improves the display of progress bars in general by allowing the text to overflow since it is set on the inner part of the progress bar. This resulted in text being cut off or invisible when the percent is too low.

### DIFF
--- a/src/Html.php
+++ b/src/Html.php
@@ -922,7 +922,7 @@ class Html
             $out = <<<HTML
             <div class="progress" style="height: 16px" id="{$id}">
                <div class="progress-bar progress-bar-striped bg-info" role="progressbar"
-                     style="width: 0%;"
+                     style="width: 0%; overflow: visible"
                      aria-valuenow="0"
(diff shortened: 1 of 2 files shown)


trasher (approved):

cedric-anne (changes requested):
Tests are failing.

cedric-anne (approved):

PR #15229, opened 2023-07-17
Graphical slot definition for NetworkEquipmentModels

ccailly:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

### Defining ports on the model
image

### Add/modify port interface
image

### Once the ports have been defined, each device in the model will see the ports and their status graphically.
image

This mechanism is intended to be reused, so I made sure that the development was generic.

### DIFF
--- a/src/Features/ZonableModelPicture.php
+++ b/src/Features/ZonableModelPicture.php
@@ -33,26 +33,21 @@
  * ---------------------------------------------------------------------
  */
 
-/// Class NetworkEquipmentModel
-class NetworkEquipmentModel extends CommonDCModelDropdown
+namespace Glpi\Features;
+
+interface ZonableModelPicture
 {
-    public static function getTypeName($nb = 0)
-    {
-        return _n('Networking equipment model', 'Networking equipment models', $nb);
-    }
+    /**
+     * Display the stencil with the zones
+     *
+     * @return void
+     */
+    public function displayStencil(): void;
 
-    public function defineTabs($options = [])
-    {
-        $ong = parent::defineTabs($options);
-
-        // Add stencil tab if there is at least one picture field defined
-        foreach ((new NetworkEquipmentModelStencil())->getPicturesFields() as $picture_field) {
-            if (!empty($this->getItemtypeOrModelPicture($picture_field))) {
-                $this->addStandardTab('NetworkEquipmentModelStencil', $ong, $options);
-                break;
-            }
-        }
-
-        return $ong;
-    }
+    /**
+     * Display the stencil editor
+     *
+     * @return void
+     */
+    public function displayStencilEditor(): void;
 }
--- a/src/NetworkEquipmentModelStencil.php
+++ b/src/NetworkEquipmentModelStencil.php
@@ -1,0 +1,123 @@
+<?php
+
+/**
+ * ---------------------------------------------------------------------
+ *
+ * GLPI - Gestionnaire Libre de Parc Informatique
+ *
+ * http://glpi-project.org
+ *
+ * @copyright 2015-2023 Teclib' and contributors.
+ * @copyright 2003-2014 by the INDEPNET Development Team.
+ * @licence   https://www.gnu.org/licenses/gpl-3.0.html
+ *
+ * ---------------------------------------------------------------------
+ *
+ * LICENSE
+ *
+ * This file is part of GLPI.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ * ---------------------------------------------------------------------
+ */
+
+use Glpi\Application\View\TemplateRenderer;
+
+class NetworkEquipmentModelStencil extends Stencil
+{
+    public static function getTypeName($nb = 0): string
+    {
+        return __('Graphical slot definition');
+    }
+
+    public function getPicturesFields(): array
+    {
+        return ['picture_front', 'picture_rear'];
+    }
+
+    public function getParams(bool $editor): array
+    {
+        if ($editor) {
+            return [
+                'nb_zones_label' => __('Set number of ports'),
+                'define_zones_label' => __('Define port data in image'),
+                'zone_label' => __('Port Label'),
+                'zone_number_label' => __('Port Number'),
+                'save_zone_data_label' => __('Save port data'),
+                'add_zone_label' => __('Add a new port'),
+                'remove_zone_label' => __('Remove last port'),
+            ];
+        } else {
+            return [
+                'anchor_id' => 'port_number_',
+            ];
+        }
+    }
+
+    public static function displayTabContentForItem(CommonGLPI $item, $tabnum = 1, $withtemplate = 0)
+    {
+        $stencil = Stencil::getStencilFromItem($item);
+        if ($stencil != null) {
+            $stencil->displayStencilEditor();
+            return true;
+        }
+
+        return false;
+    }
+
+    public function getZoneLabel(bool $editor, array $zone): string
+    {
+        $zoneLabel = parent::getZoneLabel($editor, $zone);
+        if (!$editor) {
+            $portInformation = self::getPortInformation($zone);
+            $statusHtml = TemplateRenderer::getInstance()->render('stencil/parts/port/status.html.twig', [
+                'port' => $portInformation,
+            ]);
+            $zoneLabel .= $statusHtml;
+        }
+        return $zoneLabel;
+    }
+
+    public function getZonePopover(bool $editor, array $zone): string
+    {
+        $zonePopover = parent::getZonePopover($editor, $zone);
+        if (!$editor) {
+            $portInformation = self::getPortInformation($zone);
+            $popoverHtml = TemplateRenderer::getInstance()->render('stencil/parts/port/popover.html.twig', [
+                'port' => $portInformation,
+            ]);
+            $zonePopover .= $popoverHtml;
+        }
+        return $zonePopover;
+    }
+
+    private function getPortInformation(array $port): array
+    {
+        $networkPort = new NetworkPort();
+
+        if (
+            $networkPort->getFromDBByCrit([
+                'logical_number' => $port['number'],
+                'items_id'  => $this->getStencilItem()->getID(),
+                'itemtype'  => $this->getStencilItem()->getType(),
+                'is_deleted' => 0,
+            ]) && $networkPort->fields['ifstatus']
+        ) {
+            $port['ifstatus'] = $networkPort->fields['ifstatus'];
+        }
+
+        return $port;
+    }
+}
--- a/src/Stencil.php
+++ b/src/Stencil.php
@@ -1,0 +1,334 @@
+<?php
+
+/**
+ * ---------------------------------------------------------------------
+ *
+ * GLPI - Gestionnaire Libre de Parc Informatique
+ *
+ * http://glpi-project.org
+ *
+ * @copyright 2015-2023 Teclib' and contributors.
+ * @copyright 2003-2014 by the INDEPNET Development Team.
+ * @licence   https://www.gnu.org/licenses/gpl-3.0.html
+ *
+ * ---------------------------------------------------------------------
+ *
+ * LICENSE
+ *
+ * This file is part of GLPI.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ * ---------------------------------------------------------------------
+ */
+
+use Glpi\Application\View\TemplateRenderer;
+use Glpi\Features\ZonableModelPicture;
+
+class Stencil extends CommonDBChild implements ZonableModelPicture
+{
+    public static $itemtype = "itemtype";
+    public static $items_id = "items_id";
+
+    private CommonDBTM $item;
+
+    public static function getTypeName($nb = 0): string
+    {
+        return __('Stencil');
+    }
+
+    public static function getIcon(): string
+    {
+        return 'ti ti-shape';
+    }
+
+    public static function getTable($classname = null): string
+    {
+        return 'glpi_stencils';
+    }
+
+    /**
+     * Returns the stencil associated with the given item
+     *
+     * @param CommonDBTM $item
+     * @return Stencil|null
+     */
+    public static function getStencilFromItem(CommonDBTM $item): ?Stencil
+    {
+        $itemtype = $item::getType();
+        $items_id = $item->getID();
+
+        switch ($item->getType()) {
+            case NetworkEquipment::getType():
+            case NetworkEquipmentModel::getType():
+                $stencil = new NetworkEquipmentModelStencil();
+                break;
+            default:
+                return null;
+        }
+
+        $stencil->getFromDBByCrit([
+            'itemtype' => $itemtype,
+            'items_id' => $items_id,
+        ]);
+        $stencil->item = $item;
+
+        return $stencil;
+    }
+
+    /**
+     * Returns the stencil associated with the given ID
+     *
+     * @param int $id
+     * @return Stencil|null
+     */
+    public static function getStencilFromID(int $id): ?Stencil
+    {
+        global $DB;
+
+        $iterator = $DB->request([
+            'SELECT' => [
+                'glpi_stencils.id',
+                'glpi_stencils.itemtype',
+                'glpi_stencils.items_id',
+            ],
+            'FROM'   => self::getTable(),
+            'WHERE'  => [
+                'glpi_stencils.id' => $id,
+            ],
+        ]);
+
+        if ($iterator->count() == 1) {
+            $itemtype = $iterator->current()['itemtype'];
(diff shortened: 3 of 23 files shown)


AdrienClairembault (changes requested):

src/NetworkEquipmentModelStencil.php
```diff
@@ -0,0 +1,113 @@
...
+    public static function displayTabContentForItem(CommonGLPI $item, $tabnum = 1, $withtemplate = 0)
+    {
+        self::displayStencilEditor($item);
+
+        return true;
+    }
+
+    public static function displayStencil(CommonDBTM $item, array $picture_fields = [], array $params = [])
+    {
+        parent::displayStencil(
```
cedric-anne, line 54: `$picture_fields` and `$params` are not supposed to be defined, unless by the concrete class, so you should not expose them.

This remark also applies to `displayStencilEditor()` method.

src/NetworkEquipmentModelStencil.php
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    public static function displayTabContentForItem(CommonGLPI $item, $tabnum = 1, $withtemplate = 0)
+    {
+        self::displayStencilEditor($item);
+
+        return true;
+    }
+
+    public static function displayStencil(CommonDBTM $item, array $picture_fields = [], array $params = [])
```
cedric-anne, line 52: IMHO, all display methods could be moved into an interface (with a corresponding trait, if possible), and it would be easier to use.
Also, please avoid using static methods.

```php
namespace Glpi\Features;

interface ZonableModelPicture
{
    public function displayStencil(): void;

    public function displayStencilEditor(): void;
}
```

To use it in PHP:
```php
// to display the stencil
if ($item instanceof ZonableModelPicture) {
    $item->displayStencil();
}

// to display editor
if ($item instanceof ZonableModelPicture) {
    $item->displayStencilEditor();
}
```

src/System/Diagnostic/DatabaseSchemaIntegrityChecker.php
```diff
@@ -468,6 +468,8 @@ function ($matches) {
             '/(DEFAULT) \'([-|+]?\d+(\.\d+)?)\'/i' => '$1 $2',
             // Remove surrounding quotes on collate values (quotes are optional)
             '/(COLLATE) \'([-|+]?\w+(\.\d+)?)\'/i' => '$1 $2',
+            // MariaDB does not have a JSON type and instead uses an alias
+            '/longtext (CHARACTER SET utf8mb4 COLLATE utf8mb4_bin )?((\S\s?)*)CHECK \(json_valid\(`(\w+)`\)\)/' => 'json $2',
```
cedric-anne, line 472: Please update existing tests to test this case.

templates/stencil/editor.html.twig
```diff
@@ -0,0 +1,162 @@
...
+        <input type="hidden" name="itemtype" value="{{ itemtype }}">
+        <input type="hidden" name="items_id" value="{{ items_id }}">
+    {% endif %}
+
+    {% if id <= 0 %}
+        {% if params.nb_zones_label is defined %}
+            <label for="nb_zones" class="form-label">{{ params['nb_zones_label'] }}</label>
+        {% else %}
+            <label for="nb_zones" class="form-label">{{ __("Set number of zones") }}</label>
+        {% endif %}
```
cedric-anne, line 50: You can simplify this.

```suggestion
        <label for="nb_zones" class="form-label">{{ params['nb_zones_label']|default(__("Set number of zones")) }}</label>
```

This remark also applies on other labels.

templates/stencil/editor.html.twig
```diff
@@ -0,0 +1,162 @@
...
+            'stencil': stencil,
+            'zones': zones,
+            'current_side': loop.index - 1,
+            'params': params
+        } only %}
+    </div>
+{% endfor %}
+
+<script type="text/javascript">
+    let editor = new StencilEditor(JSON.parse('{{ zones_json|raw }}'));
```
cedric-anne, line 161:
```suggestion
    let editor = new StencilEditor({{ zones_json|json_encode|raw }});
```

templates/stencil/parts/zones.html.twig
```diff
@@ -0,0 +1,61 @@
...
+ #
+ # ---------------------------------------------------------------------
+ #}
+
+{% for zone_number, zone in zones %}
+    {% if (zone['side'] ?? front) == current_side %}
+        {% if params.with_popover is defined and params.with_popover %}
+            {% set popover_content %}
+                {% apply escape('html_attr') %}
+                    {{ call(stencil ~ '::showZonePopover', [item, zone]) }}
```
cedric-anne, line 39: Would be better if it was a non static method on an instanciated item.
```twig
{{ stencil.getZonePopover(zone) }}
```

templates/stencil/parts/zones.html.twig
```diff
@@ -0,0 +1,61 @@
...
+ #
+ # You should have received a copy of the GNU General Public License
+ # along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ #
+ # ---------------------------------------------------------------------
+ #}
+
+{% for zone_number, zone in zones %}
+    {% if (zone['side'] ?? front) == current_side %}
+        {% if params.with_popover is defined and params.with_popover %}
```
cedric-anne, line 36: The `with_popover` parameter could probably be removed. Indeed, you can choose whether a popover has to be displayed or not by checking the emptyness of the generated popover content.

templates/stencil/parts/zones.html.twig
```diff
@@ -0,0 +1,61 @@
...
+            {% set popover_content %}
+                {% apply escape('html_attr') %}
```
cedric-anne, line 38: The `showZonePopover` method should be changed to a `getZonePopover` method that returns the generated content, instead of displaying it directly.
It would permit to replace the `apply` block by a simple `|escape('html_attr')` inline filter.

cedric-anne (changes requested):

templates/stencil/editor.html.twig
ccailly: This modification is not correct. I use `JSON.parse()` to convert a JSON string into a JavaScript object, while `json_encode` generates a JSON string from a PHP array.
I haven't found a cleaner alternative to the original code.

cedric-anne: My bad, you indeed store a JSON string in db.

orthagh (changes requested):
Some backtraces when viewing panel in ports tab of an equipment:

```
[2023-09-01 14:49:41] glpiphplog.WARNING:   *** PHP Warning (2): Undefined array key 0 in /var/www/html/glpi/10.1.git/src/Stencil.php at line 201
  Backtrace :
  src/NetworkPort.php:789                            Stencil->displayStencil()
  src/NetworkPort.php:1837                           NetworkPort::showForItem()
  src/CommonGLPI.php:699                             NetworkPort::displayTabContentForItem()
  ajax/common.tabs.php:112                           CommonGLPI::displayStandardTab()
  
[2023-09-01 14:49:41] glpiphplog.WARNING:   *** PHP Warning (2): Trying to access array offset on value of type null in /var/www/html/glpi/10.1.git/src/Stencil.php at line 201
  Backtrace :
  src/NetworkPort.php:789                            Stencil->displayStencil()
  src/NetworkPort.php:1837                           NetworkPort::showForItem()
  src/CommonGLPI.php:699                             NetworkPort::displayTabContentForItem()
  ajax/common.tabs.php:112                           CommonGLPI::displayStandardTab()
  
[2023-09-01 14:49:41] glpiphplog.WARNING:   *** PHP Warning (2): Undefined array key 0 in /var/www/html/glpi/10.1.git/src/Stencil.php at line 201
  Backtrace :
  src/NetworkPort.php:789                            Stencil->displayStencil()
  src/NetworkPort.php:1837                           NetworkPort::showForItem()
  src/CommonGLPI.php:699                             NetworkPort::displayTabContentForItem()
  ajax/common.tabs.php:112                           CommonGLPI::displayStandardTab()
  
[2023-09-01 14:49:41] glpiphplog.WARNING:   *** PHP Warning (2): Trying to access array offset on value of type null in /var/www/html/glpi/10.1.git/src/Stencil.php at line 201
  Backtrace :
  src/NetworkPort.php:789                            Stencil->displayStencil()
  src/NetworkPort.php:1837                           NetworkPort::showForItem()
  src/CommonGLPI.php:699                             NetworkPort::displayTabContentForItem()
  ajax/common.tabs.php:112                           CommonGLPI::displayStandardTab()
  
```

orthagh:
"Graphical slot definition" tab should not be displayed when no front/rear pictures exist in model

orthagh (changes requested):
This data in DB :
```json
{"1": {"side": 0, "image": [5.731012918906948, 0, 0, 5.731012918906948, 1443.6249938964852, 77.70001373291016], "label": "\tGi1/0/1", "number": "1", "selection": {"x": 81, "y": 45, "width": 103, "height": 70}, "x_percent": 6.1552271038322885, "y_percent": 32.49519564210454, "width_percent": 2.6547076623750083, "height_percent": 18.79110574863193}, "2": {"side": 0, "image": [8.390767026167008, 0, 0, 8.390767026167008, 2096.2025177001974, 102.00253143310552], "label": "2", "number": "2", "selection": {"x": 126, "y": 45, "width": 148, "height": 90}, "x_percent": 9.357662267317393, "y_percent": 33.590029201527244, "width_percent": 2.6053816107355092, "height_percent": 16.501649426343416}, "3": {"side": 0, "image": [5.7310129189069485, 0, 0, 5.7310129189069485, 1252.7941557450736, 43.663500837846215], "label": "3", "number": "3", "selection": {"x": 138, "y": 14, "width": 102, "height": 62}, "x_percent": 12.54284920089076, "y_percent": 33.31395178323642, "width_percent": 2.628933801575251, "height_percent": 16.643550805931138}}
```

fails with javascript error:

```
Uncaught SyntaxError: JSON.parse: bad control character in string literal at line 1 column 123 of the JSON data
    <anonymous> networkequipmentmodel.form.php:2
```

orthagh:
UX review:

- [ ] when a model doen't have yet a number of port set, don't display the pictures/editor, just the form
- [ ] when a model has a number of ports set, there is a [Add port] button with a primary style. I suggest to add the button at the end of the list port buttons, with sthe same style (outline-secondary) and just an [+] label (with a `bs-toggle-tooltip` and a title "Add a port")
- [ ] [Clear Data] button should have a confirmation before doing the deletion
- [ ] sending [Enter] key when editing a label submit a global form and doesn't save anything. I suggest intercepting the form submit event and redirect it correclty 
- [ ] There is some overflow of labels when displaying ports both in model edition and in network equipment panel
    image
    image
    Side of net equipment, i suggest to recall the label in the tooltip. And display only the start of the label (with no elipsis). Maybe a 0.5rem font size could help.

- [ ] I suggest adding an horizontal margin between the button
   image

orthagh:
One last thing

> when a model has a number of ports set, there is a [Add port] button with a primary style. I suggest to add the button at the end of the list port buttons, with sthe same style (outline-secondary) and just an [+] label (with a bs-toggle-tooltip and a title "Add a port")

Good job for the addition, but there's no way to remove a port.
BUT, we want to be able to only remove the last port (to keep the index)

orthagh (approved):

src/Features/ZonableModelPicture.php
```diff
@@ -0,0 +1,43 @@
...
+ * ---------------------------------------------------------------------
+ */
+
+namespace Glpi\Features;
+
+interface ZonableModelPicture
+{
+    public function displayStencil(): void;
+
+    public function displayStencilEditor(): void;
```
AdrienClairembault, line 42: PHPDoc to explain these method would be nice :)

src/NetworkEquipmentModelStencil.php
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    private function getPortInformation(array $port): array
+    {
+        $networkPort = new NetworkPort();
+
+        if (
+            $networkPort->getFromDBByCrit([
+                'logical_number' => $port['number'],
+                'items_id'  => $this->getStencilItem()->getID(),
```
AdrienClairembault, line 101: `getStencilItem` return type is `CommonGLPI`, which does not have the `getID()` method.
Maybe the return type should be `CommondDBTM` instead ?

src/NetworkEquipmentModelStencil.php
```diff
@@ -0,0 +1,111 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ * ---------------------------------------------------------------------
+ */
+
+use Glpi\Application\View\TemplateRenderer;
+
+class NetworkEquipmentModelStencil extends Stencil
+{
```
AdrienClairembault, line 39: PHPDoc on any non inherited method would be nice :)

src/NetworkEquipmentModelStencil.php
```diff
@@ -0,0 +1,111 @@
...
+            'port' => self::getPortInformation($port),
+        ]) : '');
+    }
+
+    public function getZonePopover(bool $editor, array $port)
+    {
+        return parent::getZonePopover($editor, $port) . (!$editor ? TemplateRenderer::getInstance()->render('stencil/parts/port/popover.html.twig', [
+            'port' => self::getPortInformation($port),
+        ]) : '');
+    }
```
AdrienClairembault, line 92: The super long ternary conditions spanning into multiple lines are a bit hard to read.
I think splitting the code into separate instructions would increase readability.

src/Stencil.php
```diff
@@ -0,0 +1,294 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ * ---------------------------------------------------------------------
+ */
+
+use Glpi\Application\View\TemplateRenderer;
+use Glpi\Features\ZonableModelPicture;
+
+class Stencil extends CommonDBChild implements ZonableModelPicture
```
AdrienClairembault, line 39: Same remark on PHPdoc for this class

src/Stencil.php
```diff
@@ -0,0 +1,294 @@
...
+                'itemtype' => $itemtype,
+                'items_id' => $items_id,
+            ]);
+            $stencil->item = $item;
+
+            return $stencil;
+        }
+
+        return null;
+    }
```
AdrienClairembault, line 104: Return type is `Stencil` but the method may return `null`.

src/Stencil.php
```diff
@@ -0,0 +1,294 @@
...
+
+    public function canPurgeStencil(): bool
+    {
+        return $this->getStencilItem()->canPurge();
+    }
+
+    public static function getStencilFromItem(CommonGLPI $item): Stencil
+    {
+        $itemtype = $item::getType();
+        $items_id = $item->getID();
```
AdrienClairembault, line 84: As noted before, `getID` is not available on `CommonGLPI`.
Parameter type should probably be `CommonDBTM` instead.

src/Stencil.php
```diff
@@ -0,0 +1,294 @@
...
+            $stencil->getFromDBByCrit([
+                'itemtype' => $itemtype,
+                'items_id' => $items_id,
+            ]);
+            $stencil->item = $item;
+
+            return $stencil;
+        }
+
+        return null;
```
AdrienClairembault, line 103: Maybe `return null;` should be the default case of the switch ? 
Then you no longer need to check if `isset($stencil)`.

src/Stencil.php
```diff
@@ -0,0 +1,294 @@
...
+            $items_id = $iterator->current()['items_id'];
+
+            $item = getItemForItemtype($itemtype);
+            if ($item->getFromDB($items_id)) {
+                return self::getStencilFromItem($item);
+            }
+        }
+
+        return null;
+    }
```
AdrienClairembault, line 133: Same return type issue here with a possible `null` value.

src/Stencil.php
```diff
@@ -0,0 +1,294 @@
...
+        }
+
+        if (!isset($this->fields['itemtype']) || !isset($this->fields['items_id'])) {
+            return null;
+        }
+
+        $itemtype = $this->fields['itemtype'];
+        $items_id = $this->fields['items_id'];
+
+        $item = getItemForItemtype($itemtype);
```
AdrienClairembault, line 148: A bit of an edge case but `getItemForItemtype` can return `false`, which mean `$item` will be false and we will get an exception when calling `getFromDB()` on it.

Granted that this is not very likely to happen ;)

src/Stencil.php
```diff
@@ -0,0 +1,294 @@
...
+            'item'              => $item,
+            'stencil'           => $this,
+            'zones'             => json_decode($model_stencil->fields['zones'] ?? '{}', true),
+            'pictures'          => $pictures,
+            'params'            => array_merge(
+                $this->getParams(false),
+                ['is_editor_view' => false]
+            ),
+        ]);
+    }
```
AdrienClairembault, line 246: Lots of warning because of `$item` being a `CommonGLPI`:

image

Shouldn't `getStencilItem` return `CommonDBTM` items instead ?

src/Stencil.php
```diff
@@ -0,0 +1,294 @@
...
+    public function getZoneLabel(bool $editor, array $zone)
+    {
+        return TemplateRenderer::getInstance()->render('stencil/parts/label.html.twig', [
+            'label' => $zone['label'],
+        ]);
+    }
+
+    public function getZonePopover(bool $editor, array $port)
+    {
+    }
```
AdrienClairembault, line 293: Is it normal to have an empty function here ?

AdrienClairembault (changes requested):

AdrienClairembault (approved):

cedric-anne:
I forced-push to rebase on current state of main branch. I am currently reviewing the code, I have, at least, some small changes to push.

AdrienClairembault (commented):
Feature and code looks good.

Still some UX suggestion:

1) When adding or editing a zone, allow to validate the area by pressing the "Enter" key (avoid having to move the mouse and click on "save port data". I think when setting a lot of ports this could speed up the process by a lot).

image

The "Esc" key could also be mapped to the "Cancel" action.

2) Use AJAX instead to avoid page reload for these actions:

image

3) Set a reasonable max value for this input:

image

If I put something stupid like 10000, my browser crash and I have to edit the value in the database to recover my network equipment model.

Google tells me the max number of ports for a switch is 52.
Maybe a hard limit to something like 64 or 128 to be large ?

----

Do you agree with these suggestions @cedric-anne @orthagh ?
They could be done in another PR to avoid re-validating this one once again (which has a lot of code).

cedric-anne:
> Do you agree with these suggestions @cedric-anne @orthagh ?
> They could be done in another PR to avoid re-validating this one once again (which has a lot of code).

I agree, and yes, they could be made on another PR. I consider that current state is already production ready, and can be merged.

orthagh:
> could be made on another PR

Seems a good compromise.
Make @ccailly working on a new pr at his returnal

PR #15242, opened 2023-07-18
Fix software version endpoints

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -


cconard96:
Changed Software versions endpoints to always be available under a specific Software rather than under a `/SoftwareVersion` endpoint.

PR #15250, opened 2023-07-19
Form editor

AdrienClairembault:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |


AdrienClairembault:
@cedric-anne Let's merge to main ?

cedric-anne (approved):

orthagh:
bulk list of things (not everything need to be addressed right now, but add them to backlog):

- permits to rename sections (maybe we can call them steps, no ?). (I wrote along I discover) In fact, I just found you can rename section, but I saw the ribbon before. There something here that need be addressed. We have 2 distinct zones related to the same thing. I don't have suggestion (yet).
- an hover effect on editable content (questions, descriptions, etc) to let the user know it's editable. I saw the "text" cursor but I think it's not sufficient (it may represent also selection)
- We talk about a floating sidebar for adding parts, this seems mandatory to lighten the page.
- I think we already talk about this, but long answer example content may hide its tinymce controls when it's not hovered.
- There is an annoying scrolling when you are initially on an active rich text field and you select a field anywhere below. nothing much, but I'm curious to see if it exists also in google forms.
- Default value feature could be indicated with placeholder. Like maybe "Enter a default value"
- Action for delete a section may display a confirm dialog when containing questions. A miss-click can happen. Not sure it's required when deleting a question for the moment, but if there any sub content related (like in a future, conditions), it will be.
- 1st section doesn't exist until you create the second, it's both interesting and disturbing. An inspiration from google ?
- when a form has been set as deleted, buttons [Restore] and [Purge] should display their text. The current state of the form make these actions primary (without requiring a style change)
- There's something odd in preview and the visual distinction between questions. I suggest to increase spacing between main description and section. And maybe reduce the spacing between questions. I may suggest to question in the same card (at least I'm curious of the render)
- mandatory fields (in preview) are not managed before section changing
- When adding a new question or a section, focus the label field.

orthagh:
To be clear, the current state is obviously mergeable.

AdrienClairembault:
> 1. permits to rename sections (maybe we can call them steps, no ?). (I wrote along I discover) In fact, I just found you can rename section, but I saw the ribbon before. There something here that need be addressed. We have 2 distinct zones related to the same thing. I don't have suggestion (yet).

Ok for "Steps" (I started calling them that way in the preview already)
Ribon is not the name of the section but its number.

For example on google forms:

image

IMO, the goal a the ribbon is to give sections their own distinct visual "signature".
Without it, it become very difficult to differentiate sections and questions.

> 2. an hover effect on editable content (questions, descriptions, etc) to let the user know it's editable. I saw the "text" cursor but I think it's not sufficient (it may represent also selection)

Earlier versions had the hover, I can add it again.
Google does not do it but I agree with you that it is good UX to include it.

> 3. We talk about a floating sidebar for adding parts, this seems mandatory to lighten the page.

Agreed, it's planned -> https://github.com/orgs/glpi-project/projects/7/views/3?pane=issue&itemId=54016731

> 4. I think we already talk about this, but long answer example content may hide its tinymce controls when it's not hovered.

Same, I confirm its planned ->https://github.com/orgs/glpi-project/projects/7/views/3?pane=issue&itemId=54016854

> 5.  There is an annoying scrolling when you are initially on an active rich text field and you select a field anywhere below. nothing much, but I'm curious to see if it exists also in google forms.

Same things happen on google forms.
Not sure what we can do about it but it doesn't bother me too much.

> 6.  Default value feature could be indicated with placeholder. Like maybe "Enter a default value"
Ok but maybe only when the question is active OR when the questions has a default value ?

Or recall the type in the placeholder like `Enter a default value (text)` ?

-> https://github.com/orgs/glpi-project/projects/7/views/3?pane=issue&itemId=54017328

> 7. Action for delete a section may display a confirm dialog when containing questions. A miss-click can happen. Not sure it's required when deleting a question for the moment, but if there any sub content related (like in a future, conditions), it will be.

Added -> https://github.com/orgs/glpi-project/projects/7/views/3?pane=issue&itemId=54017628
We also need a "Merge with previous section" action, added too.

> 8. 1st section doesn't exist until you create the second, it's both interesting and disturbing. An inspiration from google ?

Yes, its what google forms kinda do.
No opinion on it yet, we could show the first section from the start if you like ?

> 9. When a form has been set as deleted, buttons [Restore] and [Purge] should display their text. The current state of the form make these actions primary (without requiring a style change)

Noted -> https://github.com/orgs/glpi-project/projects/7/views/3?pane=issue&itemId=54017628

> 10. There's something odd in preview and the visual distinction between questions. I suggest to increase spacing between main description and section. And maybe reduce the spacing between questions. I may suggest to question in the same card (at least I'm curious of the render)

> 11.  mandatory fields (in preview) are not managed before section changing

Yeah, planning on working on it later when condition are added.

> 12.  When adding a new question or a section, focus the label field.

Good idea, added -> https://github.com/orgs/glpi-project/projects/7/views/3?pane=issue&itemId=54018573

PR #15274, opened 2023-07-25
Debug bar Vue

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

POC for using Vue.js for the new debug bar feature


.vue.webpack.config.js
```diff
@@ -0,0 +1,14 @@
+const config = require('./.webpack.config.js');
```
cconard96, line 1: Separate webpack config to allow building just the Vue components.

templates/components/debug/debug_toolbar.html.twig
```diff
@@ -31,51 +31,20 @@
...
-      </div>
-   </div>
-   <div id="debug-toolbar-expanded-content" class="w-100 card d-none pe-2">
-   </div>
-</div>
-<script>
-    $(document).on('ready', () => {
-        window.GLPI.Debug.init({{ debug_info|json_encode|raw }});
+<script type="module">
+    const debug_toolbar_applet = window.Vue.createApp(window.Vue.components['DebugToolbar'].component, {
```
cconard96, line 37: Creating a Vue "app" just for the toolbar element. If you try creating an app and mounting it to the body element currently, the entire page will be blank. This seems to be caused by invalid HTML somewhere. One of those issues was solved recently where an `=` was missing between an attribute name and value.

There is no limitation keeping us from having multiple apps, so it is probably best to have one for each feature container element.

js/src/Debug/Widget/HTTPRequests.vue
```diff
@@ -0,0 +1,298 @@
...
+    function switchSubwidget(widget_id) {
+        active_subwidget.value = widget_id;
+    }
+</script>
+
+<template>
+    <div ref="content_area" @mousemove="onMouseMove($event)" @mouseup="onMouseUp()">
+        <div class="request-timeline" v-if="show_timeline"></div>
+        <div class="d-flex flex-row h-100 split-panel-h">
+            <div class="left-panel" :style="`flex-basis: ${left_panel_flexbasis}`">
```
cconard96, line 195: Vue should allow us to use variables in the scoped style tag like `v-bind(left_panel_flexbasis)`. However, I've not been able to get this to work completely. It would replace the `v-bind` with a a reference to a CSS property, but it never initialized that property on the component element.

That is why there is an inline style here.

js/src/Debug/Toolbar.vue
```diff
@@ -0,0 +1,445 @@
...
+            </div>
+        </div>
+        <div id="debug-toolbar-expanded-content" class="w-100 card d-none pe-2">
+            <component v-if="active_widget" :is="active_widget_component" :initial_request="props.initial_request"
+               :ajax_requests="ajax_requests" @switchWidget="switchWidget" :widgets="widgets"></component>
+        </div>
+    </div>
+</template>
+
+<style scoped>
```
cconard96, line 401: These styles will be automatically prefixed with the generated component ID. Unless the selectors are wrapped inside a `:deep()`, they won't apply to child components. No need to worry about generic selectors leaking outside the scope of the feature.

js/src/Debug/Toolbar.vue
```diff
@@ -0,0 +1,445 @@
...
+        if (widget === undefined) {
+            return null;
+        }
+        return widget.component_registered_name;
+    });
+
+    const ajax_requests = ref([]);
+
+    const current_request = ref(props.initial_request.id);
+    const current_profile = computed(() => {
```
cconard96, line 50: This is a computed value. Vue will track all the refs used inside the function and should only update the value returned if a ref changes.

js/src/Debug/Toolbar.vue
```diff
@@ -0,0 +1,445 @@
...
+    import {computed, ref} from "vue";
+
+    const props = defineProps({
+        initial_request: {
+            type: Object,
+            default: null,
+        },
+    });
+
+    const active_widget = ref(null);
```
cconard96, line 35: This is a reference being initialized as null. We can get/set the value within the script with `.value` or use the name of the ref alone within the template. If the value is updated, the value in the template is also updated. We could watch changes to this variable using `watch(active_widget, (new_value, old_value) => {})`.

js/src/Debug/Toolbar.vue
```diff
@@ -0,0 +1,445 @@
...
+     * @property {function(button: jQuery)} refreshButton
+     */
+    /**
+     * @typedef SubWidget
+     * @extends Widget
+     * @property {boolean} main_widget=false
+     */
+    import {computed, ref} from "vue";
+
+    const props = defineProps({
```
cconard96, line 28: These are the properties that can be passed into this component.

js/src/Debug/Widget/HTTPRequests.vue
```diff
@@ -0,0 +1,298 @@
+<script setup>
+import {computed, ref, watch} from "vue";
+
+    const props = defineProps({
+        initial_request: {
+            type: Object,
+            required: true
+        },
+        ajax_requests: {
```
cconard96, line 9: Props are passed by ref so when a new AJAX request is added to the array by the main toolbar component, this child component will be aware of the change and it will be able to trigger adding the new request to the table.

cconard96:
Vue components are built automatically with `composer build` and `npm run build`. They can be built separately if you make changes to them with `npm run build-vue` or start a watcher to automatically rebuild them when you change a Vue component with `npm run watch-vue`.

templates/components/debug/debug_toolbar.html.twig
AdrienClairembault: As a side note, it would be very nice if we get the single app mounted to the body to work as it would allow us to use any external component anywhere without hassle.

For example, [tabler icons components](https://tabler.io/docs/icons/vue) or the [gitlab filter UX](https://gitlab-org.gitlab.io/gitlab-ui/?path=/story/base-filtered-search--default) (that we had to recreate from scratch for our kanban as we couldn't use their component without Vue).

Maybe not in this first pull request but something to keep in mind for later.

js/src/Debug/Widget/HTTPRequests.vue
cconard96: `v-bind` is working properly now.

AdrienClairembault (approved):

orthagh (approved):
Seems good
Go for merge, another pr requires this one

PR #15280, opened 2023-07-25
Use search engine to display consumables

AdrienClairembault:
Rewrite the "Consumable" tab to use the search engine rendering.

The main goals with this change are to :
* Allow users to add custom columns to the results through display preferences
* Allow users to export data (e.g. CSV)

Before:

image

image

After:

image

image

The original layout of the page is conserved with two separate tables for unused and used items.

These changes required a few tweaks of the search engine to work properly:
* A new `no_search` parameter for `SearchEngine::show()` in order to display only search results (the search criteria section is hidden but still rendered as we need the criteria inputs defined in the page for results pagination to work properly)
* Add a rand to the `massform$itemtype` form to avoid conflicts when displaying multiple search results on the same page.
* Support a new `htmlname` option for searchoptions to display raw html (such as a an icon, which is very useful for the infocom column as it will only contain a single button so keeping a short icon as its header instead of a very long label save a lot of space).
* A new `init_session_data` parameter for `SearchEngine::show()` needed due to technical limitations of our search engine code. Explained in details in comments directly in the code.
* Added a new data property `data-searchopt-content-id` in the search engine results `<td>`, which allow to make some easy css selectors targetting a full column of results. This is useful in this case as the original page does not display the same columns for unused and used items so we can hide them accordingly using css).

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !28730

### DIFF
--- a/css/includes/pages/_search.scss
+++ b/css/includes/pages/_search.scss
@@ -41,7 +41,16 @@
     background-color: $danger;
 }
 
+// Prevent search results from filling all available vertical space
+// Useful when displaying multipler search results on the same page
+.search-no-forced-height {
+    .search-container {
+        height: unset !important;
+    }
+}
+
 .search_page {
+
     @include media-breakpoint-up(lg) {
         display: flex;
         align-items: stretch;
@@ -53,6 +62,7 @@
 
         @include media-breakpoint-up(lg) {
             overflow: auto;
+            height: calc(100vh - #{$contextbar-height} - #{$content-margin});
             padding: 0;
             margin-left: calc(var(--tblr-gutter-x) / 2);
             margin-right: calc(var(--tblr-gutter-x) / 2);
--- a/css/includes/pages/_search.scss
+++ b/css/includes/pages/_search.scss
@@ -50,9 +50,9 @@
     .search-container {
         @include media-breakpoint-up(lg) {
             overflow: auto;
-            height: calc(100vh - #{$contextbar-height} - #{$content-margin});
             padding: 0;
             margin-left: calc(var(--tblr-gutter-x) / 2);
+            margin-right: calc(var(--tblr-gutter-x) / 2);
             width: 100%;
 
             .dashboard-card {
--- a/css/includes/pages/_search.scss
+++ b/css/includes/pages/_search.scss
@@ -50,7 +50,6 @@
 }
 
 .search_page {
-
     @include media-breakpoint-up(lg) {
         display: flex;
         align-items: stretch;
--- a/css/includes/pages/_search.scss
+++ b/css/includes/pages/_search.scss
@@ -48,6 +48,9 @@
     }
 
     .search-container {
+        border: var(--tblr-card-border-width) solid var(--tblr-card-border-color);
+        border-radius: var(--tblr-border-radius);
+
         @include media-breakpoint-up(lg) {
             overflow: auto;
             padding: 0;
@@ -90,6 +93,14 @@
         }
 
         .search-card {
+            &.card {
+                // Borders are unreliable here as they will scroll out of sight
+                // due to the sticky header behavior
+                // Its better to move it to a parent item that wont scroll like
+                // .search-container
+                border: unset;
+            }
+
             @include media-breakpoint-up(lg) {
                 min-width: 100%;
                 width: fit-content;
--- a/src/Consumable.php
+++ b/src/Consumable.php
@@ -543,6 +543,7 @@ class Consumable extends CommonDBChild
         Toolbox::deprecated('Replaced by Consumable::displayConsumableList()');
 
         $ID = $consitem->getField('id');
+
         if (!$consitem->can($ID, UPDATE)) {
             return;
         }
@@ -565,6 +566,7 @@ class Consumable extends CommonDBChild
             echo "</div>";
         }
     }
+
 
     /**
      * Print out the consumables of a defined type
--- a/src/Consumable.php
+++ b/src/Consumable.php
@@ -1164,7 +1164,7 @@ class Consumable extends CommonDBChild
             'field'              => 'id',
             'name'               => $infocom_label,
             // Content will be very short so using an icon here save space
-            'htmlname'           => "<i class=\"fa fa-coins\" title=\"$infocom_label\"></i>",
+            'htmlname'           => "<i class=\"ti ti-coins\" title=\"$infocom_label\"></i>",
             'massiveaction'      => false,
             'nosearch'           => true,
             'datatype'           => 'specific',
--- a/src/Consumable.php
+++ b/src/Consumable.php
@@ -1163,8 +1163,6 @@ class Consumable extends CommonDBChild
             'table'              => $this->getTable(),
             'field'              => 'id',
             'name'               => $infocom_label,
-            // Content will be very short so using an icon here save space
-            'htmlname'           => "<i class=\"ti ti-coins\" title=\"$infocom_label\"></i>",
             'massiveaction'      => false,
             'nosearch'           => true,
             'datatype'           => 'specific',
--- a/src/Consumable.php
+++ b/src/Consumable.php
@@ -1197,10 +1197,11 @@ class Consumable extends CommonDBChild
             case '6': // Given to
                 $itemtype = $values['itemtype'];
                 $items_id = $values['items_id'];
-                if (!is_null($itemtype)) {
+                if (is_a($itemtype, CommonDBTM::class, true)) {
                     $item = new $itemtype();
-                    $item->getFromDB($items_id);
-                    return $item->getLink();
+                    if ($item->getFromDB($items_id)) {
+                        return $item->getLink();
+                    }
                 }
 
                 // Must not be empty
--- a/src/Consumable.php
+++ b/src/Consumable.php
@@ -36,6 +36,7 @@
 use Glpi\Application\View\TemplateRenderer;
 use Glpi\DBAL\QuerySubQuery;
 use Glpi\Event;
+use Glpi\Search\SearchEngine;
 
 //!  Consumable Class
 /**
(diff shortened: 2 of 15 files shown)


cconard96:
A similar change was made in the Outgoing Webhook PR to allow displaying the list of sent webhooks using the search engine.
There may be some duplicated/conflicting changes.

Some related sections from the PR (may not be all the code related to the changes):
https://github.com/glpi-project/glpi/pull/14916/files#diff-7ed7e08a51abcf0217ec71739e11d15194e6d54108571c46c0efb7a6512276ceR851-R856

https://github.com/glpi-project/glpi/pull/14916/files#diff-c120910901cd3b4a25f617dc7605298247a92b7c3a0b2c42b8e8393f01dc1269L374-R391

https://github.com/glpi-project/glpi/pull/14916/files#diff-4d56b440afc480f30a7c2e9057086ce021e5ad22adfcb93c77e3459485f9fee5L40-R105

cedric-anne:
@AdrienClairembault, could you please rebase your PR to solve conflicts and adapt your code according to previous comment?

templates/components/search/table.html.twig
```diff
@@ -82,7 +83,7 @@
                <th data-searchopt-id="{{ col['id'] }}" {% if not can_sort %}data-nosort="true"{% endif %} data-sort-order="{{ sort_order }}"
                   {% if sort_num is not empty %}data-sort-num="{{ sort_num - 1 }}"{% endif %}>
                   {% set sort_icon = sort_order == 'ASC' ? 'fas fa-sort-up' : (sort_order == 'DESC' ? 'fas fa-sort-down' : '') %}
-                  {{ col_name }}
+                  {{ col_name|raw }}
```
cedric-anne, line 86: We should avoid using `raw`.
```suggestion
                  {{ col_name }}
```

cedric-anne (changes requested):

src/Search/Input/QueryBuilder.php
```diff
@@ -80,6 +80,7 @@ public static function showGenericSearch($itemtype, array $params)
         $p['hide_controls']                 = false;
         $p['showmassiveactions']            = true;
         $p['extra_actions_templates']       = [];
+        $p['no_search']                     = $params['no_search'] ?? false;
```
cconard96, line 83: There is already a `no_search` parameter used within the search engine, so this naming could add to confusion. `hide_criteria` would probably be less confusing and match with the `hide_controls` parameter.

src/Consumable.php
```diff
@@ -1014,4 +1104,116 @@ public static function convertFiltersValuesToSqlCriteria(array $filters = []): a
...
+        ];
+
+        $infocom_label = Infocom::getTypeName();
+        $options[] = [
+            'id'                 => '7',
+            'table'              => $this->getTable(),
+            'field'              => 'id',
+            'name'               => $infocom_label,
+            // Content will be very short so using an icon here save space
+            'htmlname'           => "<i class=\"fa fa-coins\" title=\"$infocom_label\"></i>",
```
cconard96, line 1167:
```suggestion
            'htmlname'           => "<i class=\"ti ti-coins\" title=\"$infocom_label\"></i>",
```

cconard96:
Closing the infocom modal could trigger a "fluid search" results refresh to update the icon.

AdrienClairembault:
Latest commit fix an issue with border rendering when the content is scrolled down:

image

css/includes/pages/_search.scss
```diff
@@ -50,9 +50,9 @@
     .search-container {
         @include media-breakpoint-up(lg) {
             overflow: auto;
-            height: calc(100vh - #{$contextbar-height} - #{$content-margin});
```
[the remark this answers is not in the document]
AdrienClairembault: Which screen size do you use ?
I don't have the issue on my side (horizontal layout).

src/Consumable.php
```diff
@@ -191,6 +202,21 @@ public function out($ID, $itemtype = '', $items_id = 0)
...
+        CommonDBTM $checkitem = null
+    ) {
+        // Special actions only for self
+        if ($itemtype != self::getType()) {
+            return;
+        }
+
+        $action_prefix = self::getType() . MassiveAction::CLASS_ACTION_SEPARATOR;
+        $actions[$action_prefix . 'backtostock'] = __('Back to stock');
+        $actions[$action_prefix . 'give'] = _x('button', 'Give');
```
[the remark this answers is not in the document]
AdrienClairembault: I don't think its possible to filter actions here.
Functionally it has no impact:
* Moving back to stock an unused item do noting
* Giving an already set item is the same as moving back to stock + give item so it avoid an useless step

src/Search/Provider/SQLProvider.php
```diff
@@ -4309,6 +4309,7 @@ public static function constructData(array &$data, $onlycount = false)
                     'itemtype'  => $data['itemtype'],
                     'id'        => $opt_id,
                     'name'      => $searchopt[$opt_id]["name"],
+                    'htmlname'  => $searchopt[$opt_id]["htmlname"] ?? false,
```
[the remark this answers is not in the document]
AdrienClairembault: What CSS solution do you have in mind ?
The goal is to use an icon as the column name.

templates/components/search/table.html.twig
AdrienClairembault: Needed if we keep the htmlname option

src/Consumable.php
AdrienClairembault: Done cdcab4f2e6.

src/Search/Input/QueryBuilder.php
AdrienClairembault: Good call, done e6fce4698e.

AdrienClairembault:
> Closing the infocom modal could trigger a "fluid search" results refresh to update the icon.

Would be nice to have but it doesn't seem easy to implement as the infocom modal is handled through the `Ajax::createIframeModalWindow` function.

I'll accept contributions if anyone is brave enough to make it work ;)

css/includes/pages/_search.scss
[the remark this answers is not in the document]
cedric-anne: 1920x1200

src/Search/Provider/SQLProvider.php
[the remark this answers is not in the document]
cedric-anne: Something like
```css
.used-consumables td[data-searchopt-content-id="XXX"] {
    font-size: 0;
}
.used-consumables td[data-searchopt-content-id="XXX"]::before {
    content: "icon";
    font-size: 0.625rem;
}
```

cconard96:
> > Closing the infocom modal could trigger a "fluid search" results refresh to update the icon.
> 
> Would be nice to have but it doesn't seem easy to implement as the infocom modal is handled through the `Ajax::createIframeModalWindow` function.
> 
> I'll accept contributions if anyone is brave enough to make it work ;)

Well, the modal generation is something that was identified as a performance issue previously (all modals created on page load instead of as needed), so it should probably be redone at some point anyways. For now though, I think adding this in a script table in the consumable_list template would be sufficient to at least keep the icon updated.
```js
$('div.modal[id^="infocomConsumable"]').on('hide.bs.modal', () => {
    // Try finding a fluid search instance
    const search_display = $('.ajax-container.search-display-data');
    try {
        if (search_display.length === 1 && search_display.data('js_class') !== undefined) {
            // Trigger a reload of just the results
            search_display.data('js_class').view.refreshResults();
        } else {
            // There is no (or multiple) search results, reload the page
            window.location.reload();
        }
    } catch (err) {
        window.location.reload();
    }
});
```

src/Search/Provider/SQLProvider.php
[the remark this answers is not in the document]
AdrienClairembault: Done d81786ac39.

templates/components/search/table.html.twig
AdrienClairembault: Deleted in d81786ac39.

AdrienClairembault:
> > > Closing the infocom modal could trigger a "fluid search" results refresh to update the icon.
> > 
> > 
> > Would be nice to have but it doesn't seem easy to implement as the infocom modal is handled through the `Ajax::createIframeModalWindow` function.
> > I'll accept contributions if anyone is brave enough to make it work ;)
> 
> Well, the modal generation is something that was identified as a performance issue previously (all modals created on page load instead of as needed), so it should probably be redone at some point anyways. For now though, I think adding this in a script tag in the consumable_list template would be sufficient to at least keep the icon updated.
> 
> ```js
> $('div.modal[id^="infocomConsumable"]').on('hide.bs.modal', () => {
>     // Try finding a fluid search instance
>     const search_display = $('.ajax-container.search-display-data');
>     try {
>         if (search_display.length === 1 && search_display.data('js_class') !== undefined) {
>             // Trigger a reload of just the results
>             search_display.data('js_class').view.refreshResults();
>         } else {
>             // There is no (or multiple) search results, reload the page
>             window.location.reload();
>         }
>     } catch (err) {
>         window.location.reload();
>     }
> });
> ```

Thanks, works great 👍

src/Search/Input/QueryBuilder.php
cconard96, line 83:
```suggestion
        $p['hide_criteria']                     = $params['hide_criteria'] ?? false;
```
Missed replacement

cedric-anne (approved):

AdrienClairembault:
Good for merge ?

templates/pages/assets/consumable_list.html.twig
```diff
@@ -0,0 +1,155 @@
...
+    .unused-consumables td[data-searchopt-content-id="8"],
+    .used-consumables th[data-searchopt-id="8"],
+    .used-consumables td[data-searchopt-content-id="8"]
+    {
+        display: none;
+    }
+
+    /* Hack to display an icon as a column header */
+    .unused-consumables th[data-searchopt-id="7"],
+    .used-consumables td[data-searchopt-id="7"]
```
AdrienClairembault, line 121:
```suggestion
    .used-consumables th[data-searchopt-id="7"]
```

AdrienClairembault:
Internal ref reported an issue with the export dropdown display, which I was partly able to fix with the lastest commit.

However there is still a conflict with the sticky header below that I don't know how to fix (changing z-index doesn't change anything):

image

Help wanted :)

AdrienClairembault:
Got it in [d97c19d](https://github.com/glpi-project/glpi/pull/15280/commits/d97c19db9e1a7f55bc828785c37bfe4ee1b54bd9) but its quite ugly.

orthagh (approved):

trasher (approved):

trasher:
Remaining comments concerns part of this (quite huge) PR. I'll merge, fix -if any- can be done later since we target main branch.

PR #15290, opened 2023-07-27
Notification to users before of maintenance

dxdemetriou:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #number

Hello,

I started here for showing what I mean and learning some of twig coding in conjunction with config's php & html form, but if I should make a feature request, then please close this pull request and I'll follow the process.

The idea is to use the same way as the "GLPI MAINTENANCE MODE" to show a prior warning to users that there will following a maintenance, or for any reason that should be informed the GLPI users.

Thanks


trasher:
Please use [suggestion website](http://suggest.glpi-project.org) for RFE.

PR #15296, opened 2023-07-31
fix(notification): private timeline items visible

Rom1-B:
In the case of a notification template using "FOREACHtimelineitems".

In a ticket, if a technician adds a private follow-up, other actors who don't have the right to see private follow-ups don't receive a notification (this is normal). If the same technician then adds a public follow-up, the other actors receive a notification containing the new follow-up (normal), but also the previous private follow-ups written by the same technician, even though they don't have access to them.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !28943

### DIFF
--- a/src/NotificationTemplate.php
+++ b/src/NotificationTemplate.php
@@ -268,6 +268,10 @@ class NotificationTemplate extends CommonDBTM
             $bak_language = $_SESSION["glpilanguage"];
             $_SESSION["glpilanguage"] = $language;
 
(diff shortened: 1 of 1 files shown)


Rom1-B:
Replaced by https://github.com/glpi-project/glpi/pull/15297

PR #15301, opened 2023-07-31
Display preferences UX improvements

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

Notable changes/improvements:
- AJAX actions now use the itemtype, user, and num instead of the ID for ordering and removal.
- Closing the modal will now attempt to trigger the `refreshResults` method for the fluid search view and only reload the full page when a fluid search view cannot be found (or finds multiple) or an error occurs.
- De-duplicated code between global and personal config forms
- Options are now orderable by drag and drop instead of one... at... a... time... up and down controls. Adding and removing options now doesn't trigger a reload of the iframe either. Activating and deactivating personal preferences are still "synchronous" and require reloading the iframe in the modal.
- Added ability to view all global display preferences from the Config form in a new "Search result display" tab.
- Added ability to edit display preferences directly from the global/personal lists. Previously, it was only possible to do so from the search result page which could be problematic if someone added too many options or an option that otherwise breaks the search page.
- Added ability to reset global display preferences to the defaults specified in the `empty_data.php` script or provided by plugins via a new `default_display_prefs` hook.
- Added informational message to the global preference config form to make sure users are aware they would be changing the display for everyone that doesn't have a personal display configured, and not just changing their own display.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -102,6 +102,7 @@ The present file will list all changes made to the project; according to the
 - `PrinterLog::getMetrics()` method has been made final.
 - `SavedSearch::showSaveButton()` replaced with `pages/tools/savedsearch/save_button.html.twig` template.
 - `showSystemInformations` method for `$CFG_GLPI['systeminformations_types']` types renamed to `getSystemInformation` and should return an array with a label and content.
+- `DisplayPreference` config form POST handling moved to `ajax/displaypreference.php` script. The front file is for displaying the tabs only.
 
 #### Deprecated
 - Usage of `GLPI_USE_CSRF_CHECK` constant.
@@ -225,7 +226,13 @@ The present file will list all changes made to the project; according to the
 - Javascript file upload functions `dataURItoBlob`, `extractSrcFromImgTag`, `insertImgFromFile()`, `insertImageInTinyMCE`, `isImageBlobFromPaste`, `isImageFromPaste`.
 - `CommonDBTM::$fkfield` property.
 - `getHTML` action for `ajax/fuzzysearch.php` endpoint.
+<<<<<<< HEAD
 - `Config::showLibrariesInformation()`
+=======
+- `DisplayPreference::showFormGlobal` `target` parameter.
+- `DisplayPreference::showFormPerso` `target_id` parameter.
+
+>>>>>>> 9f421e80d2 (migrate display pref config to twig)
 
 ## [10.0.10] unreleased
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -226,9 +226,7 @@ The present file will list all changes made to the project; according to the
 - Javascript file upload functions `dataURItoBlob`, `extractSrcFromImgTag`, `insertImgFromFile()`, `insertImageInTinyMCE`, `isImageBlobFromPaste`, `isImageFromPaste`.
 - `CommonDBTM::$fkfield` property.
 - `getHTML` action for `ajax/fuzzysearch.php` endpoint.
-<<<<<<< HEAD
 - `Config::showLibrariesInformation()`
-=======
 - `DisplayPreference::showFormGlobal` `target` parameter.
 - `DisplayPreference::showFormPerso` `target_id` parameter.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -230,7 +230,6 @@ The present file will list all changes made to the project; according to the
 - `DisplayPreference::showFormGlobal` `target` parameter.
 - `DisplayPreference::showFormPerso` `target_id` parameter.
 
->>>>>>> 9f421e80d2 (migrate display pref config to twig)
 
 ## [10.0.10] unreleased
 
--- a/src/DisplayPreference.php
+++ b/src/DisplayPreference.php
@@ -373,6 +373,9 @@ class DisplayPreference extends CommonDBTM
         }
         $entries = [];
         foreach ($fixed_columns as $key => $val) {
+            if (!isset($searchopt[$val])) {
+                continue;
+            }
             $entries[] = [
                 'id'   => $val,
                 'name' => $searchopt[$val]['name'],
@@ -381,6 +384,9 @@ class DisplayPreference extends CommonDBTM
             ];
         }
         foreach ($already_added as $key => $val) {
+            if (!isset($searchopt[$val])) {
+                continue;
+            }
             $entries[] = [
                 'id'   => $val,
                 'name' => $searchopt[$val]['name'],
--- a/src/DisplayPreference.php
+++ b/src/DisplayPreference.php
@@ -74,6 +74,19 @@ class DisplayPreference extends CommonDBTM
         return $input;
     }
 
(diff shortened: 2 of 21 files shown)


src/DisplayPreference.php
```diff
@@ -329,146 +337,78 @@ public function showFormPerso($target, $itemtype)
...
+                'id'   => $val,
+                'name' => $searchopt[$val]['name'],
+                'group' => $this->nameOfGroupForItemInSearchopt($searchopt, $val),
+                'fixed' => true
+            ];
+        }
+        foreach ($already_added as $key => $val) {
+            $entries[] = [
+                'id'   => $val,
+                'name' => $searchopt[$val]['name'],
```
orthagh, line 386: Probably due to wrong data in database (removed plugin for example), but I guess you need to manage the case

```
[2023-08-23 14:11:27] glpiphplog.WARNING:   *** PHP Warning (2): Undefined array key 47812 in /var/www/html/glpi/10.1.git/src/DisplayPreference.php at line 386
  Backtrace :
 ```
and
```
[2023-08-23 14:11:27] glpiphplog.WARNING:   *** PHP Warning (2): Trying to access array offset on value of type null in /var/www/html/glpi/10.1.git/src/DisplayPreference.php at line 386
```

**Review by orthagh — CHANGES_REQUESTED** [review]
Also, not sure it's required to do it in this pr but maybe we can move (lack of time previously) to a drag and drop design.

**Review by cedric-anne — COMMENTED** [review]
When opening the display pref modal, the dropdown contains a field that is already shown:
image

I can select it:
image

**Comment by cedric-anne** [comment]
Also, when display preferences are updated, it seems that columns added by default filters are added, even when default filters are supposed to be disabled..

**Comment by cconard96** [comment]
> When opening the display pref modal, the dropdown contains a field that is already shown

Odd behavior of select2. When using a result template to hide options, it still has no problem selecting filtered out options by default or programmatically.

**Comment by cconard96** [comment]
> Also, when display preferences are updated, it seems that columns added by default filters are added, even when default filters are supposed to be disabled..

Do you have an example?

**Comment by cedric-anne** [comment]
> > Also, when display preferences are updated, it seems that columns added by default filters are added, even when default filters are supposed to be disabled..
> 
> Do you have an example?

Issue is already present on main branch, so it is not related to current PR.

**Review by cedric-anne — APPROVED** [review]

File: templates/components/search/displaypreference_config.html.twig [hunk 2]
```diff
@@ -0,0 +1,227 @@
...
+            {{ __('These preferences are used by everyone that does not have a personal view configured.') }}
+         </div>
+      {% endif %}
+      {% if not is_global and not has_personal %}
+         <div class="alert alert-info">
+            {{ __('No personal criteria. Create personal parameters?') }}
+         </div>
+         <button type="button" class="btn btn-primary" name="activate" value="1">{{ __('Create') }}</button>
+      {% else %}
+         <table class="table table-striped card-table m-n2">
```
orthagh, line 55: are a table mandatory here ? If yes, at least remove `.m-n2` and move header stuff outside the table
if no, could we move to a ul > li design (use flex to make delete button at the right) ?

templates/components/search/displaypreference_list.html.twig
```diff
@@ -0,0 +1,108 @@
...
+   <div class="mb-3">
+      {% if massiveactionparams['specific_actions']|length > 0 %}
+         <form id="massDisplayPreference{{ rand }}" method="get" action="{{ path('front/massiveaction.php') }}"
+            data-search-itemtype="DisplayPreference" data-submit-once>
+         {% do call('Html::showMassiveActions', [massiveactionparams]) %}
+      {% endif %}
+      {{ inputs.hidden('users_id', users_id, {
+         'data-glpicore-ma-tags': 'common'
+      }) }}
+      <table class="table table-striped display-prefs-list">
```
orthagh, line 46: is `.display-prefs-list` class still used ?
I suggest also to a margin top and a border to the table here

templates/components/search/displaypreference_list.html.twig
```diff
@@ -0,0 +1,108 @@
...
+         <thead>
+            <tr>
+               <th style="width: 10px">
+                  {% if massiveactionparams['specific_actions']|length > 0 %}
+                     {{ call('Html::getCheckAllAsCheckbox', ['massDisplayPreference' ~ rand])|raw }}
+                  {% endif %}
+               </th>
+               <th>{{ _n('Type', 'Types', 1) }}</th>
+               <th class="text-end">{{ __('Number of modifiable columns') }}</th>
```
orthagh, line 55: Not caused by this pr, I'm not sure about the purpose of this column. Seems to be the number of columns set by the user for an itemtype.
So the label seems wrong but whatever, I think we should drop completly this th, the feature is useless. 

If we do this (the removal), move to ul>li design, as having a table for one columns is useless

orthagh (changes requested):
Also

Make form inline : image

About drag and drop UX:
- when dragging, add a placeholder with a yellow color (and darker yellow border) to let user identify where he will drop the item
- identify visually draggable items by appending `ti ti-grip-vertical` icon

templates/components/search/displaypreference_list.html.twig
cconard96: I don't know. The column always existed, but without a label. It currently shows the number of columns in the saved preferences. That would include any added by default or by a user, but not fixed columns which are always shown.

I don't know if the original purpose was just to show how "heavy" each list was or some other reason.

orthagh:
Two very last comment about ui:

- add a [+] icon to "Add" buton
- if possible, remove the card (or at least the border)  from the `.tab-content` surrounding the modal content

image

Apart this, good for me.

cconard96:
> if possible, remove the card (or at least the border) from the `.tab-content` surrounding the modal content

It would affect all tabbed forms in both cases..

orthagh:
I pushed a commit to quickly polish last things:
image

Things done:
- move drag icon to left side
- remove border (I added a ` in_modal` option to manage that)
- indicate visually non-movable list items
- some minor adjustements

orthagh:
And tried to condensate a little more
image

cconard96:
I can't approve my own PR, but these changes are OK for me.
:heavy_check_mark:

cedric-anne (approved):

CHANGELOG.md
```diff
@@ -226,6 +227,10 @@ The present file will list all changes made to the project; according to the
 - `CommonDBTM::$fkfield` property.
 - `getHTML` action for `ajax/fuzzysearch.php` endpoint.
 - `Config::showLibrariesInformation()`
+- `DisplayPreference::showFormGlobal` `target` parameter.
+- `DisplayPreference::showFormPerso` `target_id` parameter.
+
+>>>>>>> 9f421e80d2 (migrate display pref config to twig)
```
cedric-anne, line 233:
```suggestion
```

trasher (approved):

PR #15342, opened 2023-08-01
fix(reservation): itemtype list empty

Rom1-B:
When we searched for a free item for a specific period, we found items of different types, but the itemtype dropdown was empty, which made it impossible to filter by this field.

After:
image

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !28083

### DIFF
--- a/src/ReservationItem.php
+++ b/src/ReservationItem.php
@@ -512,7 +512,7 @@ class ReservationItem extends CommonDBChild
         ]);
 
         foreach ($iterator as $data) {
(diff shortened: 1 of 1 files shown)


cedric-anne (approved):
OK for me.
A user should be able to reserve an item, even if he does not have the right to view the full information of this item (a self-service user cannot see the computer form, but is still able to reserve it).

PR #15345, opened 2023-08-01
Files integrity check

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

- `bin/console diagnostic:source_code_integrity --baseline` - (Re-)Generate the baseline file manifest in the `version/X.Y.Z` file based on the current files. This is what is used when building the release archive.
- `bin/console diagnostic:source_code_integrity` - Display a summary of differences between the baseline file manifest and the current files. This is just a table of relative paths and the status (missing, added, or altered). Specific differences are not displayed.
-  `bin/console diagnostic:source_code_integrity --diff` - Outputs a diff of changes between the release archive and the current files. Does not work with development versions. The summary is added to the start of the diff as a comment.

When generating the baseline, you can use the `-a`/`--algorithm` option to specify the hashing function. Could be used to change the algorithm in the future if needed.

### DIFF
--- a/src/System/Diagnostic/SourceCodeIntegrityChecker.php
+++ b/src/System/Diagnostic/SourceCodeIntegrityChecker.php
@@ -43,7 +43,7 @@ use GuzzleHttp\Exception\GuzzleException;
 use RecursiveDirectoryIterator;
 use RecursiveIteratorIterator;
 use SebastianBergmann\Diff\Differ;
-use SebastianBergmann\Diff\Output\UnifiedDiffOutputBuilder;
+use SebastianBergmann\Diff\Output\StrictUnifiedDiffOutputBuilder;
 
 /**
  * @since 10.1.0
@@ -255,14 +255,15 @@ class SourceCodeIntegrityChecker
         $diff = '';
 
         foreach ($summary as $file => $status) {
-            // print diff --git line for the file
-            $diff .= 'diff --git a/' . $file . ' b/' . $file . "\n";
             $original_file = 'a/' . $file;
             $current_file = 'b/' . $file;
             $original_content = '';
             $current_content = '';
+            $extra_header = '';
             if ($status === self::STATUS_ADDED || !file_exists('phar://' . $release_path . '/glpi/' . $file)) {
                 $original_file = '/dev/null';
+                $file_perms   = @substr(sprintf('%o', fileperms($this->getCheckRootDir() . '/' . $file)), -4) ?: '0644';
+                $extra_header = 'new file mode 10' . $file_perms;
             } else {
                 $original_content = file_get_contents('phar://' . $release_path . '/glpi/' . $file);
                 if ($original_content === false) {
@@ -272,6 +273,8 @@ class SourceCodeIntegrityChecker
             }
             if ($status === self::STATUS_MISSING || !file_exists($this->getCheckRootDir() . '/' . $file)) {
                 $current_file = '/dev/null';
+                $file_perms   = @substr(sprintf('%o', fileperms('phar://' . $release_path . '/glpi/' . $file)), -4) ?: '0644';
+                $extra_header = 'deleted file mode 10' . $file_perms;
             } else {
                 $current_content = file_get_contents($this->getCheckRootDir() . '/' . $file);
                 if ($current_content === false) {
@@ -280,12 +283,23 @@ class SourceCodeIntegrityChecker
                 }
             }
             try {
-                $differ = new Differ(new UnifiedDiffOutputBuilder("--- $original_file\n+++ $current_file\n", true));
-                $diff .= $differ->diff(
+                $differ = new Differ(
+                    new StrictUnifiedDiffOutputBuilder(
+                        [
+                            'fromFile' => $original_file,
+                            'toFile'   => $current_file,
+                        ]
+                    )
+                );
+                $file_diff = $differ->diff(
                     $original_content,
                     $current_content
                 );
-                $diff .= "\n";
+                $diff .= 'diff --git a/' . $file . ' b/' . $file . "\n";
+                if ($extra_header !== '') {
+                    $diff .= "$extra_header\n";
+                }
+                $diff .= "$file_diff\n";
             } catch (\Exception $e) {
                 $errors[] = $e->getMessage();
             }
--- a/src/System/Diagnostic/SourceCodeIntegrityChecker.php
+++ b/src/System/Diagnostic/SourceCodeIntegrityChecker.php
@@ -63,9 +63,11 @@ class SourceCodeIntegrityChecker
     }
 
     /**
+     * Get list of files/directories to check.
+     *
      * @return array
      */
-    public function getPathsToCheck(): array
+    private function getPathsToCheck(): array
     {
         return [
             'ajax', 'css', 'front', 'inc', 'install', 'js', 'lib', 'locales', 'pics',
@@ -75,6 +77,8 @@ class SourceCodeIntegrityChecker
     }
 
     /**
+     * Returns the source code manifest contents.
+     *
      * @return array|null
      * @phpstan-return array{algorithm: string, files: array<string, string>}|null
      * @throws \JsonException
@@ -86,15 +90,17 @@ class SourceCodeIntegrityChecker
         try {
             $content = json_decode($manifest, true, 512, JSON_THROW_ON_ERROR);
         } catch (\Throwable $e) {
-            throw new \RuntimeException('The source code file manifest is invalid', previous: $e);
+            throw new \RuntimeException('The source code file manifest is invalid.', previous: $e);
         }
         if (!isset($content['algorithm'], $content['files']) || !is_string($content['algorithm']) || !is_array($content['files'])) {
-            throw new \RuntimeException('The source code file manifest is invalid');
+            throw new \RuntimeException('The source code file manifest is invalid.');
         }
         return $content;
     }
 
     /**
+     * Generates the source code manifest contents.
+     *
      * @param string $algorithm
      * @return array {algorithm: string, files: array<string, string>}
      */
@@ -130,6 +136,11 @@ class SourceCodeIntegrityChecker
         ];
     }
 
+    /**
+     * Get a summary of differences between current source code and expected source code.
+     *
+     * @return array
+     */
     public function getSummary(): array
     {
         $baseline = $this->getBaselineManifest();
@@ -161,7 +172,13 @@ class SourceCodeIntegrityChecker
         return $summary;
     }
 
-    private function getGLPIRelease(&$errors = []): string|null
+    /**
+     * Download the GLPI release.
+     *
+     * @param array $errors
+     * @return string|null Release file path.
+     */
+    private function getGLPIRelease(array &$errors = []): string|null
     {
         $version_to_get = VersionParser::getNormalizedVersion(GLPI_VERSION);
         $gh_releases_endpoint = 'https://api.github.com/repos/glpi-project/glpi/releases/tags/' . $version_to_get;
@@ -188,7 +205,14 @@ class SourceCodeIntegrityChecker
         return $dest;
     }
 
-    public function getDiff(array &$errors = [], bool $allow_download = false): string
+    /**
+     * Get the diff (unified diff format) between current source code and expected source code.
+     *
+     * @param bool $allow_download  Whether the release download is allowed.
+     * @param array $errors         Array on which any errors will be added.
+     * @return string|null
+     */
+    public function getDiff(bool $allow_download = false, array &$errors = []): ?string
     {
         $summary = $this->getSummary();
         // ignore OK files in case they are present
@@ -199,12 +223,12 @@ class SourceCodeIntegrityChecker
             if ($allow_download) {
                 $release_path = $this->getGLPIRelease($errors);
                 if ($release_path === null) {
-                    $errors[] = 'An error occurred while downloading the release';
-                    return '';
+                    $errors[] = 'An error occurred while downloading the release.';
+                    return null;
                 }
             } else {
-                $errors[] = 'The release is not downloaded and downloading it was not allowed';
-                return '';
+                $errors[] = 'The release is not downloaded and downloading it was not allowed.';
+                return null;
             }
         }
 
@@ -221,11 +245,19 @@ class SourceCodeIntegrityChecker
                 $original_file = '/dev/null';
             } else {
                 $original_content = file_get_contents('phar://' . $release_path . '/glpi/' . $file);
+                if ($original_content === false) {
+                    $errors[] = sprintf('Failed to get original contents of file `%s`.', $file);
+                    continue;
+                }
             }
             if ($status === self::STATUS_MISSING || !file_exists($this->getCheckRootDir() . '/' . $file)) {
                 $current_file = '/dev/null';
             } else {
                 $current_content = file_get_contents($this->getCheckRootDir() . '/' . $file);
+                if ($current_content === false) {
+                    $errors[] = sprintf('Fails to get current contents of file `%s`.', $file);
+                    continue;
+                }
             }
             try {
                 $differ = new Differ(new UnifiedDiffOutputBuilder("--- $original_file\n+++ $current_file\n", true));
--- a/src/System/Diagnostic/SourceCodeIntegrityChecker.php
+++ b/src/System/Diagnostic/SourceCodeIntegrityChecker.php
@@ -37,6 +37,7 @@ namespace Glpi\System\Diagnostic;
 
 use FilesystemIterator;
 use Glpi\Toolbox\VersionParser;
+use GLPIKey;
 use GuzzleHttp\Client;
 use GuzzleHttp\Exception\GuzzleException;
 use RecursiveDirectoryIterator;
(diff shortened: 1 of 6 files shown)


src/System/Diagnostic/SourceCodeIntegrityChecker.php
```diff
@@ -0,0 +1,225 @@
...
+            // print diff --git line for the file
+            $diff .= 'diff --git a/' . $file . ' b/' . $file . "\n";
+            $original_file = 'a/' . $file;
+            $current_file = 'b/' . $file;
+            $original_content = '';
+            $current_content = '';
+            // If status is missing, just add the info to the diff string
+            if ($status === self::STATUS_MISSING) {
+                $current_file = '/dev/null';
+            } else if ($status === self::STATUS_ADDED || !file_exists('phar://' . $release_path . '/' . $file)) {
```
cconard96, line 205: The original spec was to use UnifiedArchive for reading files directly in the archive without needing to extract them. This was not a functional solution as UnifiedArchive scans the file info for every single file in the archive as soon as you create the object instance, and it was failing to read info about at least one `vendor` file even though it seemed OK. None of the functionality of the class seemed to be needed and it is really just a wrapper for a `phar://` Uri file stream anyways, so that is what I ended up using.

src/Console/Diagnostic/SourceCodeIntegrityCommand.php
```diff
@@ -0,0 +1,196 @@
...
+        parent::configure();
+
+        $this->setName('diagnostic:source_code_integrity');
+        $this->setDescription(__('Check GLPI source code file integrity. This command runs in validation mode by default unless --baseline is specified.'));
+        $this->addOption(
+            'baseline',
+            'b',
+            InputOption::VALUE_NONE,
+            __('Generate a baseline manifest of the file hashes. Does not check the files against the official manifest.'),
+        );
```
cedric-anne, line 64: IMHO, this should be moved into a different command (e.g. `build:generate_code_baseline`) and make this command hidden (unless in dev environment). Indeed, we should not expose the baseline generation feature to end users.

trasher:
Does not work on my side. I've ran `php bin/console build:generate_code_baseline -a crc32c`, then edited a `src` file:
```diff
diff --git a/src/DeviceBatteryType.php b/src/DeviceBatteryType.php
index 613689c13e..b6ccacdbcd 100644
--- a/src/DeviceBatteryType.php
+++ b/src/DeviceBatteryType.php
@@ -35,8 +35,4 @@
 
 class DeviceBatteryType extends CommonDeviceType
 {
-    public static function getTypeName($nb = 0)
-    {
-        return _n('Battery type', 'Battery types', $nb);
-    }
 }
```

And then:
```
 % php bin/console diagnostic:check_documents_integrity
    0 [->--------------------------]
```

trasher:
(also, the command requires db to be updated, this does not seems really required at this point)

trasher (approved):

src/Console/Build/SourceCodeIntegrityBaselineCommand.php
```diff
@@ -0,0 +1,87 @@
...
+
+use GLPI;
+use Glpi\Console\AbstractCommand;
+use Glpi\System\Diagnostic\SourceCodeIntegrityChecker;
+use Glpi\Toolbox\VersionParser;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+
+class SourceCodeIntegrityBaselineCommand extends AbstractCommand
```
cedric-anne, line 46: The `tools/make_release.sh` script is failing because the requirements checks are not met. Indeed, the empty `config` directory is not exported by `git checkout-index`, and its writable state cannot be validated (I will do a dedicated PR to put it back in the release archive).

Anyway, I think build commands should not extend the `Glpi\Console\AbstractCommand`, as checks added by this abstract command are not appliable to the build context. 

```suggestion
class SourceCodeIntegrityBaselineCommand extends AbstractCommand
```

src/Console/Diagnostic/SourceCodeIntegrityCommand.php
```diff
@@ -0,0 +1,164 @@
...
+        return 0;
+    }
+
+    protected function interact(InputInterface $input, OutputInterface $output)
+    {
+        // If --allow-download missing, ask if it is OK
+        if ($input->getOption('diff') && !$input->getOption('allow-download')) {
+            $helper = $this->getHelper('question');
+            $question = new ConfirmationQuestion(
+                __('The source code diff cannot be generated if the GLPI release is missing. Do you want to download it?'),
```
cedric-anne, line 156:
```suggestion
                __('Generating the source code diff could require to download the GLPI release archive. Do you allow this operation?'),
```

src/System/Diagnostic/SourceCodeIntegrityChecker.php
```diff
@@ -0,0 +1,248 @@
...
+            'public', 'resources', 'sound', 'src', 'templates', 'vendor',
+            'api.php', 'apirest.md', 'apirest.php', 'caldav.php', 'index.php', 'status.php'
+        ];
+    }
+
+    /**
+     * @return array
+     * @phpstan-return array{algorithm: string, files: array<string, string>}|null
+     */
+    private function getBaselineManifest(): array
```
cedric-anne, line 73: To prevent the following error:
```
Uncaught Exception TypeError: Glpi\System\Diagnostic\SourceCodeIntegrityChecker::getBaselineManifest(): Return value must be of type array, null returned in /var/www/glpi/src/System/Diagnostic/SourceCodeIntegrityChecker.php at line 80
```

```suggestion
    private function getBaselineManifest(): ?array
```

src/System/Diagnostic/SourceCodeIntegrityChecker.php
```diff
@@ -0,0 +1,248 @@
...
+        // Ensure the release is downloaded
+        if ($allow_download) {
+            $release_path = $this->getGLPIRelease($errors);
+        } else {
+            $release_path = GLPI_TMP_DIR . '/glpi-' . VersionParser::getNormalizedVersion(GLPI_VERSION) . '.tgz';
+            if (!file_exists($release_path)) {
+                $errors[] = 'The release is not downloaded and downloading it was not allowed';
+                return '';
+            }
+        }
```
cedric-anne, line 202: When using `bin/console diagnostic:source_code_integrity --diff`, the `interact()` method will set `$allow_download ` to `true` is user answers `yes` (this will probably be the case).

On first execution, release will be downloaded to be able to generate the diff.
On next executions, even if the release is already present on disk, it will be downloaded again. I think this part could be optimized.

src/System/Diagnostic/SourceCodeIntegrityChecker.php
```diff
@@ -0,0 +1,248 @@
...
+        if ($release_path === null) {
+            return 'An error occurred while downloading the release';
```
cedric-anne, line 204: When this occurs, the message should be displayed in red (surrounded by `<error></error>`), and command result should not be `0` (0=success).

image

src/System/Diagnostic/SourceCodeIntegrityChecker.php
```diff
@@ -0,0 +1,248 @@
...
+            // print diff --git line for the file
+            $diff .= 'diff --git a/' . $file . ' b/' . $file . "\n";
+            $original_file = 'a/' . $file;
+            $current_file = 'b/' . $file;
+            $original_content = '';
+            $current_content = '';
+            // If status is missing, just add the info to the diff string
+            if ($status === self::STATUS_MISSING) {
+                $current_file = '/dev/null';
+            } else if ($status === self::STATUS_ADDED || !file_exists('phar://' . $release_path . '/' . $file)) {
```
cedric-anne, line 218: In the official release archive, all files are under a `glpi` directory.
```suggestion
            } else if ($status === self::STATUS_ADDED || !file_exists('phar://' . $release_path . '/glpi/' . $file)) {
```

src/System/Diagnostic/SourceCodeIntegrityChecker.php
```diff
@@ -0,0 +1,248 @@
...
+                $original_file = '/dev/null';
+                $current_content = file_get_contents(GLPI_ROOT . '/' . $file);
+            } else {
+                $original_content = file_get_contents('phar://' . $release_path . '/' . $file);
```
cedric-anne, line 222:
```suggestion
                $original_content = file_get_contents('phar://' . $release_path . '/glpi/' . $file);
```

src/System/Diagnostic/SourceCodeIntegrityChecker.php
```diff
@@ -0,0 +1,248 @@
...
+                $current_content = file_get_contents(GLPI_ROOT . '/' . $file);
+            }
+            try {
+                $differ = new Differ(new UnifiedDiffOutputBuilder("--- $original_file\n+++ $current_file\n"));
```
cedric-anne, line 226: The `$addLineNumbers` param default value is `false`. It need to be passed to `true` to get line numbers.

```suggestion
                $differ = new Differ(new UnifiedDiffOutputBuilder("--- $original_file\n+++ $current_file\n", true));
```

src/System/Diagnostic/SourceCodeIntegrityChecker.php
```diff
@@ -0,0 +1,248 @@
...
+    /**
+     * @return array
+     * @phpstan-return array{algorithm: string, files: array<string, string>}|null
+     */
+    private function getBaselineManifest(): array
+    {
+        $version = VersionParser::getNormalizedVersion(GLPI_VERSION, false);
+        $manifest = file_get_contents(GLPI_ROOT . '/version/' . $version);
+        try {
+            return json_decode($manifest, true, 512, JSON_THROW_ON_ERROR);
```
cedric-anne, line 78: Manifest format should be validated, to ensure there is an `algorithm` key that corresponds to a string, and a `files` key that contains an array of strings. It would prevent the code to fail later and help to detect corruption of this file.

src/Console/Diagnostic/SourceCodeIntegrityCommand.php
```diff
@@ -0,0 +1,164 @@
...
+                    $diff .= '# ' . fgets($output_buffer);
+                }
+                $diff .= "\n" . $checker->getDiff($errors, $allow_download);
+                if (count($errors) > 0) {
+                    $output->writeln('<error>' . __('Errors:') . '</error>');
+                    foreach ($errors as $error) {
+                        $output->writeln('<error>' . $error . '</error>');
+                    }
+                    return 1;
+                }
```
cedric-anne, line 140: If there is an error on a unique file, we should still display the diff we were able to compute.
Moving these lines after `$output->write(OutputFormatter::escape($diff));` would permit to do this, and to ensure that errors are still visible in the result.

src/Console/Diagnostic/SourceCodeIntegrityCommand.php
```diff
@@ -0,0 +1,164 @@
...
+
+class SourceCodeIntegrityCommand extends AbstractCommand
+{
+    protected $requires_db = false;
+
+    protected function configure()
+    {
+        parent::configure();
+
+        $this->setName('diagnostic:source_code_integrity');
```
cedric-anne, line 57:
```suggestion
        $this->setName('diagnostic:check_source_code_integrity');
```

src/Console/Build/SourceCodeIntegrityBaselineCommand.php
cedric-anne, line 46: The class name should reflect the command name.
```suggestion
class GenerateCodeBaselineCommand extends AbstractCommand
```

src/System/Diagnostic/SourceCodeIntegrityChecker.php
```diff
@@ -0,0 +1,248 @@
...
+        } catch (\JsonException $e) {
+            return null;
```
cedric-anne, line 80: Throwing the exception would probably make the code simplier. It would just have to be catched in the command itself, instead of doing
```php

        if ($baseline === null) {
            return [
                'error' => 'No baseline manifest found.'
            ];
        }
```
then
```php
        if (isset($summary['error'])) {
            $output->writeln('<error>' . $summary['error'] . '</error>');
            return 1;
        }
```

cedric-anne (changes requested):
I do not know if it can be optimized, but memory usage is huge: `Memory usage: 395.7 MiB.`.

cedric-anne:
Maybe some tests could be added to the `SourceCodeIntegrityChecker` class, using `vfsStream` to mock files contents and  some mock to simulate release archive files contents.

src/System/Diagnostic/SourceCodeIntegrityChecker.php
cconard96: That shouldn't be the case. `getGLPIRelease` is called every time a download is allowed, but that method doesn't always re-download the archive. Instead, it:
1. Makes a single request to the GitHub API to get a list of releases
2. Checks the releases for the name of the archive related to the current version (in case extensions change at some point)
3. Checks if the correct archive is already downloaded. If yes, the path to the downloaded file is returned without re-downloading. If no, it attempts to download it before returning the expected path.

cconard96:
> I do not know if it can be optimized, but memory usage is huge: `Memory usage: 395.7 MiB.`.

Which command? Both are around 25mb for me, at least after making the other requested changes.

src/System/Diagnostic/SourceCodeIntegrityChecker.php
```diff
@@ -0,0 +1,242 @@
...
+    /**
+     * @return array|null
+     * @phpstan-return array{algorithm: string, files: array<string, string>}|null
+     * @throws \JsonException
+     */
+    private function getBaselineManifest(): ?array
+    {
+        $version = VersionParser::getNormalizedVersion(GLPI_VERSION, false);
+        $manifest = file_get_contents($this->getCheckRootDir() . '/version/' . $version);
+        $content = json_decode($manifest, true, 512, JSON_THROW_ON_ERROR);
```
cedric-anne, line 86: If the manifest file is empty, the command fails with an unclear `Syntax error` message. It would be usefull to have a clearer message.

```suggestion
        try {
            $content = json_decode($manifest, true, 512, JSON_THROW_ON_ERROR);
        } catch (\Throwable $e) {
            throw new \RuntimeException('The source code file manifest is invalid', previous: $e);
        } 
```

cedric-anne:
> > I do not know if it can be optimized, but memory usage is huge: `Memory usage: 395.7 MiB.`.
> 
> Which command? Both are around 25mb for me, at least after making the other requested changes.

As long as I use the `--diff` option, it uses a lot of memory.

src/System/Diagnostic/SourceCodeIntegrityChecker.php
```diff
@@ -0,0 +1,242 @@
...
+            // If status is missing, just add the info to the diff string
+            if ($status === self::STATUS_MISSING) {
+                $current_file = '/dev/null';
+            } else if ($status === self::STATUS_ADDED || !file_exists('phar://' . $release_path . '/glpi/' . $file)) {
+                $original_file = '/dev/null';
+                $current_content = file_get_contents($this->getCheckRootDir() . '/' . $file);
+            } else {
+                $original_content = file_get_contents('phar://' . $release_path . '/glpi/' . $file);
+                $current_content = file_get_contents($this->getCheckRootDir() . '/' . $file);
+            }
```
cedric-anne, line 228: This improves the diff of deleted files.

```suggestion
            if ($status === self::STATUS_ADDED || !file_exists('phar://' . $release_path . '/glpi/' . $file)) {
                $original_file = '/dev/null';
            } else {
                $original_content = file_get_contents('phar://' . $release_path . '/glpi/' . $file);
            }
            if ($status === self::STATUS_MISSING || !file_exists($this->getCheckRootDir() . '/' . $file)) {
                $current_file = '/dev/null';
            } else {
                $current_content = file_get_contents($this->getCheckRootDir() . '/' . $file);
            }
```

tests/functional/Glpi/System/Diagnostic/SourceCodeIntegrityChecker.php
```diff
@@ -0,0 +1,197 @@
...
+        // Why not isEmpty? Because then atoum will not tell you what the contents are when this fails.
+        $this->array($errors)->isEqualTo([]);
+        $this->string(trim($diff))->isEqualTo(<<<EOL
+diff --git a/src/test2.php b/src/test2.php
+--- /dev/null
++++ b/src/test2.php
+@@ -1,0 +1,3 @@
++line 1
++line b
++line 3
```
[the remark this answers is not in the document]
cconard96: The fake archive in the tests wasn't putting files into a 'glpi' subfolder. That was fixed.

cedric-anne:
I pushed 2 commits.
The first one to make small adjustments.
The second one as I figured out that there is no need to prefix comments by a `#` in the diff file. I figured out that Github patches are containing comment not prefixed by a `#`. Applying the patch using `bin/console diagnostic:check_source_code_integrity --diff -n | patch -p1 -R` is actually working to revert changes made to the GLPI source code.

I will made some more test ASAP in order to validate this PR.

cedric-anne (approved):

PR #15360, opened 2023-08-03
move some rules ui to twig

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

### DIFF
--- a/templates/pages/admin/rules/engine_preview_criteria.html.twig
+++ b/templates/pages/admin/rules/engine_preview_criteria.html.twig
@@ -1,0 +1,70 @@
+{#
+ # ---------------------------------------------------------------------
+ #
+ # GLPI - Gestionnaire Libre de Parc Informatique
+ #
+ # http://glpi-project.org
+ #
+ # @copyright 2015-2023 Teclib' and contributors.
+ # @copyright 2003-2014 by the INDEPNET Development Team.
+ # @licence   https://www.gnu.org/licenses/gpl-3.0.html
+ #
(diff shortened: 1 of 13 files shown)


orthagh (changes requested):
About preview modal

Before:
image

After:
image

We may do better, please.

templates/pages/admin/rules/engine_preview_criteria.html.twig
```diff
@@ -0,0 +1,74 @@
...
+         <h3 class="card-title">{{ _n('Criterion', 'Criteria', get_plural_number()) }}</h3>
+      </div>
+      <div class="card-body d-flex flex-wrap">
+         <form name="testrule_form" id="testrulesengine_form" method="post" action="{{ params.target }}">
+            <div class="col-12 col-xxl12 flex-column">
+               <div class="d-flex flex-row flex-wrap flex-xl-nowrap">
+                  <div class="row flex-row align-items-start flex-grow-1">
+                     <div class="row flex-row">
+                        {% for criterion in input %}
+                           {% set field_content %}
```
orthagh, line 49: One last point, as the html here is generated by `Rule::displayCriteriaSelectPattern`, the default case (text input at l2645) will call an `Html::input` with size = 70.
We should completly remove the size attribute to avoid breaking preview UI.

Apart this, all ok 👍🏽

trasher (approved):

PR #15361, opened 2023-08-04
show global kanban task creation issues

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #15353

- Prevent creating project task with an ID of 0 (only checked missing value before)
- Fetch session messages after trying to add an item via Kanban and display them
- Prevent bulk-creating project tasks on the global Kanban


cedric-anne (approved):
Seems OK, but tests are failing.

cedric-anne (approved):

orthagh (commented):
There exists a file similar to the new added `ajax/getSessionMessages.php`
-> `ajax/displayMessageAfterRedirect.php`

They may be different in usage (raw text vs json), but maybe we can update the former to support json returning (and avoid proliferation of ajax files)

PR #15371, opened 2023-08-05
Move user preferences (and defaults) form to Twig (part 2 of config ui)

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -


orthagh (commented):
Not so mandatory as #15369, but you may priority in term of space usage the label rather the field

Apart this, seems ok

orthagh (approved):

PR #15374, opened 2023-08-07
convert management setup to twig (last of config ui)

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

### DIFF
--- a/templates/pages/setup/general/management_setup.html.twig
+++ b/templates/pages/setup/general/management_setup.html.twig
@@ -1,0 +1,45 @@
+{#
+# ---------------------------------------------------------------------
+#
+# GLPI - Gestionnaire Libre de Parc Informatique
+#
+# http://glpi-project.org
+#
+# @copyright 2015-2023 Teclib' and contributors.
+# @copyright 2003-2014 by the INDEPNET Development Team.
+# @licence   https://www.gnu.org/licenses/gpl-3.0.html
+#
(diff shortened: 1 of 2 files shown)


templates/pages/setup/general/management_setup.html.twig
```diff
@@ -0,0 +1,45 @@
...
+ #}
+
+{% extends "pages/setup/general/base_form.html.twig" %}
+{% import 'components/form/fields_macros.html.twig' as fields %}
+
+{% block config_fields %}
+   {{ fields.largeTitle(__('Documents setup'), 'ti ti-files') }}
+   <div class="alert alert-info">
+      {{ __('The maximum upload size is primarily determined by the "upload_max_filesize" and "post_max_size" PHP settings. The setting below is to further restrict the uploads for just GLPI.') }}
+   </div>
```
orthagh, line 41: I suggest to encapsulate the alert into a "container" div (to avoid parent row css styles) and add some x margins for a better look

```suggestion
<div>
        <div class="alert alert-info px-2">
            {{ __('The maximum upload size is primarily determined by the "upload_max_filesize" and "post_max_size" PHP settings. The setting below is to further restrict the uploads for just GLPI.') }}
        </div>
    <div>
```

PR #15377, opened 2023-08-07
Migrate Auth, MailCollector and DomainRelation to twig

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

### DIFF
--- a/js/common.js
+++ b/js/common.js
@@ -1137,7 +1137,11 @@ $(document).ready(function() {
         }
     });
 
-    $(document).on('submit', 'form', function() {
+    $(document).on('submit', 'form', (e) => {
+        // if the submitter has a data-block-on-unsaved attribute, do not clear the unsaved changes flag
+        if ($(e.originalEvent.submitter).attr('data-block-on-unsaved') === 'true') {
+            return;
+        }
         window.glpiUnsavedFormChanges = false;
     });
 });
@@ -1438,6 +1442,25 @@ $(() => {
             e.preventDefault();
             return false;
         } else {
+            let submitter = null;
+            if (e.originalEvent && e.originalEvent.submitter) {
+                submitter = $(e.originalEvent.submitter);
+            }
+            if (submitter !== null && submitter.is('button') && submitter.attr('data-block-on-unsaved') === 'true' && window.glpiUnsavedFormChanges) {
+                // This submit may be cancelled by the unsaved changes warning so we cannot permanently block it
+                // We fall back to a timed block
+                const block = function(e) {
+                    e.preventDefault();
+                };
+                submitter.on('click', block);
+                submitter.data('original_html', submitter.html());
+                submitter.html(`<span class="spinner-border spinner-border-sm" role="status" aria-hidden="true"></span>`);
+                window.setTimeout(() => {
+                    submitter.off('click', block);
+                    submitter.html(submitter.data('original_html'));
+                }, 100);
+                return;
+            }
             blockFormSubmit(form, e);
         }
     });
--- a/templates/pages/setup/mailcollector/results_title.html.twig
+++ b/templates/pages/setup/mailcollector/results_title.html.twig
@@ -31,29 +31,18 @@
  # ---------------------------------------------------------------------
  #}
 
-{% set messages = pull_messages() %}
-{% if messages|length %}
-
-   {% for type, message in messages %}
-      {% set message = message|join('<br />') %}
-      {% set class = '' %}
-      {% set title = '' %}
-      {% if type == constant('ERROR') %}
-         {% set class = 'alert-danger' %}
-         {% set title = _n('Error', 'Errors', 1) %}
-      {% elseif type == constant('WARNING') %}
-         {% set class = 'alert-warning' %}
-         {% set title = __('Warning') %}
-      {% else %}
-         {% set class = 'alert-info' %}
-         {% set title = _n('Information', 'Informations', 1) %}
-      {% endif %}
-
-      <div class="alert alert-important {{ class }}" role="alert">
-         <h3>{{ title }}</h3>
-         <p>
-            {{ message|raw }}
-         </p>
-      </div>
-   {% endfor %}
-{% endif %}
+<div class="btn-group flex-wrap mb-3">
+   {% if servers|length > 0 %}
+      <span class="border-danger text-danger border-1 border p-0 ps-2 rounded-start">
+         <i class="ti ti-alert-triangle me-2"></i>
+         {% set server_links = servers|map((v) => '<a class="btn btn-ghost-danger align-baseline" href="' ~ v['link'] ~ '">' ~ v['name']|e ~ '</a>')|join(' ') %}
+         <span>{{ __('Receivers in error: %s')|format(server_links)|raw }}</span>
+      </span>
+   {% endif %}
+   {% if has_collector %}
+      <a class="btn btn-outline-warning" href="{{ path('/front/notimportedemail.php') }}">
+         <i class="ti ti-list me-2"></i>
+         <span>{{ __('List of not imported emails') }}</span>
+      </a>
+   {% endif %}
+</div>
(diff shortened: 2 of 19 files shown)


templates/pages/setup/mailcollector/results_title.html.twig
```diff
@@ -0,0 +1,48 @@
...
+ # along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ #
+ # ---------------------------------------------------------------------
+ #}
+
+<div class="btn-group flex-wrap mb-3">
+   {% if servers|length > 0 %}
+      <span class="border-danger text-danger border-1 border p-0 ps-2 rounded-start">
+         <i class="ti ti-alert-triangle me-2"></i>
+         {% set server_links = servers|map((v) => '<a class="btn btn-ghost-danger" href="' ~ v['link'] ~ '">' ~ v['name'] ~ '</a>')|join(' ') %}
```
cedric-anne, line 38: Name should be escaped to prevent XSS.

```suggestion
         {% set server_links = servers|map((v) => '<a class="btn btn-ghost-danger" href="' ~ v['link'] ~ '">' ~ v['name']|e ~ '</a>')|join(' ') %}
```

templates/pages/setup/mailcollector/server_config_fields.html.twig
```diff
@@ -0,0 +1,123 @@
...
+
+{% set connection_options_fields %}
+   <div class="d-flex">
+      {% set type_value = connect_opts['type'] is not empty ? ('/' ~ connect_opts['type']) : '' %}
+      {{ fields.dropdownArrayField('server_type', type_value, protocols, '', {
+         no_label: true,
+         display_emptychoice: true,
+         field_class: 'col-5'
+      }) }}
+      {% set ssl_value = connect_opts['ssl'] is not empty ? ('/' ~ connect_opts['ssl']) : '' %}
```
cedric-anne, line 54: `connect_opts['ssl']` is a boolean.

```suggestion
      {% set ssl_value = connect_opts['ssl'] ? '/ssl' : '' %}
```

templates/pages/setup/mailcollector/server_config_fields.html.twig
```diff
@@ -0,0 +1,123 @@
...
+      {{ fields.dropdownArrayField('server_ssl', ssl_value, {
+         '/ssl': __('SSL')
+      }, '', {
+         no_label: true,
+         display_emptychoice: true,
+         field_class: 'col-5'
+      }) }}
+      {% set tls_value = connect_opts['tls'] == true ? '/tls' : (connect_opts['tls'] == false ? '/notls' : '') %}
```
cedric-anne, line 62: Comparison have to be strict to not consider empty value as false.
```suggestion
      {% set tls_value = connect_opts['tls'] is same as(true) ? '/tls' : (connect_opts['tls'] is same as(false) ? '/notls' : '') %}
```

templates/pages/setup/mailcollector/server_config_fields.html.twig
```diff
@@ -0,0 +1,123 @@
...
+      {{ fields.dropdownArrayField('server_tls', tls_value, {
+         '/tls': __('TLS'),
+         '/notls': __('NO-TLS')
+      }, '', {
+         no_label: true,
+         display_emptychoice: true,
+         field_class: 'col-5'
+      }) }}
+      {% set cert_value = connect_opts['validate-cert'] == true ? '/validate-cert' : (connect_opts['validate-cert'] == false ? '/novalidate-cert' : '') %}
```
cedric-anne, line 71:
```suggestion
      {% set cert_value = connect_opts['validate-cert'] is same as(true) ? '/validate-cert' : (connect_opts['validate-cert'] is same as(false) ? '/novalidate-cert' : '') %}
```

cedric-anne (changes requested):

cedric-anne (approved):

templates/pages/setup/mailcollector/collect_now.html.twig
```diff
@@ -0,0 +1,42 @@
...
+ # ---------------------------------------------------------------------
+ #}
+
+{% import 'components/form/basic_inputs_macros.html.twig' as inputs %}
+
+<div class="center mt-4">
+   <form name="form" method="post" action="{{ path('/front/mailcollector.form.php') }}">
+      {{ inputs.hidden('_glpi_csrf_token', csrf_token()) }}
+      {{ inputs.hidden('id', id) }}
+      {{ inputs.submit('get_mails', _x('button', 'Get email tickets now'), 1) }}
```
orthagh, line 40: Not sure how easy it is to move this to main form. 
We have a tab named "Actions" (plural) with only one action and nothing displayed in this tab apart the button.

templates/pages/setup/mailcollector/setup_form.html.twig
```diff
@@ -0,0 +1,139 @@
...
+{% extends "generic_show_form.html.twig" %}
+{% import 'components/form/fields_macros.html.twig' as fields %}
+{% set params  = params ?? [] %}
+
+{% block form_fields %}
+   {{ fields.textField('name', item.fields['name'], __('Name'), {
+      helper: __('If the name is a valid email address, it will be automatically added to blacklisted senders.')
+   }) }}
+   {{ fields.nullField() }}
+   {% if item.fields['errors'] %}
```
orthagh, line 43: I think this should be displayed above the name.
When errors exists, resulting UI is strange with the current way

templates/pages/setup/mailcollector/results_title.html.twig
```diff
@@ -0,0 +1,48 @@
...
+ #
+ # ---------------------------------------------------------------------
+ #}
+
+<div class="btn-group flex-wrap mb-3">
+   {% if servers|length > 0 %}
+      <span class="border-danger text-danger border-1 border p-0 ps-2 rounded-start">
+         <i class="ti ti-alert-triangle me-2"></i>
+         {% set server_links = servers|map((v) => '<a class="btn btn-ghost-danger" href="' ~ v['link'] ~ '">' ~ v['name']|e ~ '</a>')|join(' ') %}
+         <span>{{ __('Receivers in error: %s')|format(server_links)|raw }}</span>
```
orthagh, line 39: image

There is strange vertical alignement and (due to tabler upgrade) missing padding in server button(s)

orthagh (dismissed):
Sorry to be late to the party.
I just saw I started a review weeks ago, but missed to Submit it (It stayed as Pending and only visible by me)

js/common.js
```diff
@@ -1137,7 +1137,11 @@ $(document).ready(function() {
...
 
-    $(document).on('submit', 'form', function() {
+    $(document).on('submit', 'form', (e) => {
+        // if the submitter has a data-block-on-unsaved attribute, do not clear the unsaved changes flag
+        if ($(e.originalEvent.submitter).attr('data-block-on-unsaved') === 'true') {
+            return;
+        }
         window.glpiUnsavedFormChanges = false;
     });
 });
```
cedric-anne, line 1147: When clicking on the `Get email tickets now` button, if the "Cancel" button is used when the "Are you sure ...." message is displayed, then the form buttons are not usable anymore:

image

Also, I am not sure it is the best place for this kind of button. There are already some items that have action buttons in the form header, like users. Maybe we should put the button at this place. @orthagh What is your opinion?
image

orthagh: For one action, I prefer the current iteration (Main form I guess).
Actions in user's header is, at least for me, an aberation : exception on ui, only user has this kind of button. But now, it's difficult to move to another design.

Another design I don't like, is the global action button still in header where it should be placed side of Save button (like in tickets)

cedric-anne: > When clicking on the `Get email tickets now` button, if the "Cancel" button is used when the "Are you sure ...." message is displayed, then the form buttons are not usable anymore:
> 
> image

This issue is still present.

cedric-anne (approved):

PR #15379, opened 2023-08-09
rewrite menu fuzzy search as vue component

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

Also add:
- add/fix linting for Vue component files
- add jest setup to allow testing. Unfortunately, I can only get this to work with babel (we just got rid of it too...). I split the regular JS tests and Vue tests into separate projects with different configs. The regular unit tests shouldn't be using babel.


AdrienClairembault (changes requested):
Functional review: 

Works great but there is a vue warning:

image

It seems we are loading the javascript code twice (the twig template is probably included multiple times by mistake).

image


image

cconard96:
There are 3 calls to the goto button template in the page_header template. When the menu is in vertical mode at least, it ends up being included twice. I'm not sure what one of those calls is used for, so I'm not convinced it can be removed without any issues.

I've at least fixed the check done to ensure the component is only attempted to be mounted once though. The component should activate the modal when any of the `.trigger-fuzzy` buttons are clicked.

AdrienClairembault (approved):

orthagh (changes requested):
Overall, seems very great.
I think this component fits very well with a vue rewrite.

One functional issue I found : when navigating with up/down arrows and the scroll is at the end, we cannot go the the last (5?) items. You want to reduce the list by searching a term (Computer for example) to see the issue.
Also, when calling again the modal while already open, the issue is worse (it stops middle of the list)

cconard96:
> One functional issue I found : when navigating with up/down arrows and the scroll is at the end, we cannot go the the last (5?) items. You want to reduce the list by searching a term (Computer for example) to see the issue. Also, when calling again the modal while already open, the issue is worse (it stops middle of the list)

The results shown were not being updated properly so they didn't reflect that actual results value. When you press the down arrow, you were actually at the end of the results (by array length) but old results were still in the HTML. Changing the key used for the list seems to have fixed the issue.

orthagh (approved):

cedric-anne (changes requested):
On 10.0 branch, matching terms are highlighted. Now, the whole entries are using bold font.

On 10.0 branch:
image

On your branch:
image

cedric-anne (approved):

PR #15382, opened 2023-08-09
Fix default dashboard computation

AdrienClairembault:
Prevent calling `loadAllDashboards` if the default dashboard can be found in GLPI's config (which will be the case 99.9% of the time).

This save around ~70 SQL queries on every page for any user that didn't override the default dashboards in their preferences.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |

### DIFF
--- a/src/Dashboard/Grid.php
+++ b/src/Dashboard/Grid.php
@@ -35,6 +35,7 @@
 
 namespace Glpi\Dashboard;
 
+use Config;
 use DateInterval;
 use Dropdown;
 use Glpi\Application\View\TemplateRenderer;
@@ -1442,6 +1443,8 @@ HTML;
      */
     public static function getDefaultDashboardForMenu(string $menu = "", bool $strict = false): string
     {
+        global $CFG_GLPI;
+
         $grid = new self();
 
         if (!$strict) {
@@ -1451,6 +1454,7 @@ HTML;
             }
         }
 
+        // Try loading default from user preferences
         $config_key = 'default_dashboard_' . $menu;
         $default    = $_SESSION["glpi$config_key"] ?? "";
         if (strlen($default)) {
@@ -1461,7 +1465,17 @@ HTML;
             }
         }
 
-       // if default not found, return first dashboards
+        // Try loading default from config
+        $default = $CFG_GLPI[$config_key] ?? "";
+        if (strlen($default)) {
+            $dasboard = new Dashboard($default);
+
+            if ($dasboard->load() && $dasboard->canViewCurrent()) {
+                return $default;
+            }
+        }
+
+        // if default not found, return first dashboard
         if (!$strict) {
             self::loadAllDashboards();
             $first_dashboard = array_shift(self::$all_dashboards);


trasher (approved):

cedric-anne (approved):

froberts0n:
Hi,

I think there is a problem with the fix, now, in 10.0.10, from the previous behavior.

Previously, if the user wanted to remove the default dashboard, it selected "-----" (`$_SESSION["glpidefault_dashboard_mini_ticket"]="0"`, for example), instead of the inherited value (which would be `$_SESSION["glpidefault_dashboard_mini_ticket"]=null` in this case).

I understand that there was no way to return to the inherited value from the config other than selecting it (by chance), and that therefore, if that base config was changed, it would probably come to a surprise for the user that his dashboard changed as well. Though this might not happen often. Hence the suggested change, I guess.

The problem, I think,  is that now, the user cannot disable a dashboard if there is a config's default dashboard because in `getDefaultDashboardForMenu()` from `src/Dashboard/Grid.php`, when `$_SESSION["glpi$config_key"]` returns `"0"`, `(new Dashboard("0"))->load()` returns `false` and the code proceeds to the new code which defaults from `$CFG_GLPI` instead of skipping to the end and returning `""` and hiding the dashboard as the user intended.

I think this comes from the fact that the inherited value is not explicitly shown as such in the user's preferences. My understanding of the previous behavior was that: "_in the user's preferences dropdown, use `null` as the current's config value and use the dashboard's id, or `"0"` for none, **otherwise**_".

One solution could be to allow resetting the dropdown to "_inherit from general config_" (or something like that) so that any choice explicitly made by the user would overwrite the general config.

What do you think?

AdrienClairembault:
Hi @froberts0n,

From what I understand, you used to set the default dashboard to  "empty" and that would disable the dashboard ?

image

If I save this value in my personal preferences and revert the changes from this Pull Request, the dashboard is still enabled (and the first dashboard is selected by default) :

image

Unless I'm missing something, this does not disable the dashboard ?

froberts0n:
Hi @AdrienClairembault ,

Sorry for my complicated explantation. I'll try to clarify.

Say the "Mini Tickets Dashboard" is set as the default Tickets dashboard in the general preferences (for everyone). Previously, if I went to my personal preferences and set that dashboard to "-----", this would disable the dashboard for me and I would be happy. But now, what this does is use the general default preference instead of my choice (as now, I guess, ----- is not considered a choice anymore).

Also, previously, if I hadn't set a personal preference before, and the general preferences' default was "Mini Tickets Dashboard", that dashboard (Mini Tickets Dashboard) it would be set (by default) as my choice in my personal preference screen (although, in the DB, the value would be null as it is the same as the default). Then, if the general preference was modified and I went to see my personal preference screen, the new default choice would be selected.

Of course, this behavior might have been a bit confusing for some, but the new behavior makes removing the dashboard impossible for individual users if default dashboards have been set for all. Therefore, in my previous comment, I was trying to suggest other behaviors which would allow default values for all to be overwritten by user preferences only if they want to.

Does that make sense and clarifies a my comment a bit?

AdrienClairembault:
I see what you mean, its only the "Mini Tickets Dashboard" that could be disabled this way.
Simply adding a new "Disabled" choice should be enough to allow user to disable it again from their preferences.

We will look into fixing this for the next version, I'll come back to you when its done.

PR #15388, opened 2023-08-09
Lazy load actors

AdrienClairembault:
Implement lazy loading for actors.

This shouldn't cause BC break as actors are stored in protected properties, the only impact could be if some plugins have classes extending Ticket/Changes/Problems/CommonITIL (which seems unlikely).

These change save another 30% performance on my test subject (ticket with 19 various items in its timeline).
Request count:
| e1185c8684 (a week ago; before all recent optimisations) | Before these changes | After these changes |
|--------|--------|--------|
| 1786 requests| 1052 requests| 766 requests|

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |

### DIFF
--- a/src/Change.php
+++ b/src/Change.php
@@ -1277,31 +1277,21 @@ class Change extends CommonITILObject
                         ];
 
                         $requesters = [];
-                        if (
-                            isset($change->users[CommonITILActor::REQUESTER])
-                            && count($change->users[CommonITILActor::REQUESTER])
-                        ) {
-                            foreach ($change->users[CommonITILActor::REQUESTER] as $d) {
-                                if ($d["users_id"] > 0) {
-                                    $userdata = getUserName($d["users_id"], 2);
-                                    $name = '<i class="fas fa-sm fa-fw fa-user text-muted me-1"></i>' .
-                                        $userdata['name'];
-                                    $requesters[] = $name;
-                                } else {
-                                    $requesters[] = '<i class="fas fa-sm fa-fw fa-envelope text-muted me-1"></i>' .
-                                        $d['alternative_email'];
-                                }
+                        foreach ($change->getUsers(CommonITILActor::REQUESTER) as $d) {
+                            if ($d["users_id"] > 0) {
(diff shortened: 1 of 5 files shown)


cedric-anne (changes requested):
If you want to be sure that there will be no BC-break, you have to remove the existing `$users`, `$groups` and `$suppliers` properties and handle access to them using magic `__get` method. Also, you should probably implement the `__isset` method to ensure that usage of `empty($this->users)` will return expected result. You could then store loaded data into new private property (e.g. `_loaded_actors`) and revert all changes made elsewhere in the code.

AdrienClairembault:
> If you want to be sure that there will be no BC-break, you have to remove the existing `$users`, `$groups` and `$suppliers` properties and handle access to them using magic `__get` method. Also, you should probably implement the `__isset` method to ensure that usage of `empty($this->users)` will return expected result. You could then store loaded data into new private property (e.g. `_loaded_actors`) and revert all changes made elsewhere in the code.

I tried to use __get but didn't think of deleting the properties to get it to work, seems like a nice solution to get a smaller diff and avoid mistakes.

I'll revert the current change (and move out the fix for actor count) and try your suggested approach 👍

AdrienClairembault:
Logging of unexpected properties is disabled (commented line) as it would trigger a lot of warnings due to mistakes in our code base.
I'll look into into fixing theses code mistake and enabling the warnings in another pull request to not complicate the diff here.

cedric-anne (commented):
Seems OK, but it should not be merged without reactivating warnings on unknown properties.
Please open another PR to fix them, so we will be able to see what are the problems you are mentionning.

trasher:
This would probably require tests

AdrienClairembault:
Rebased to incorporate #15401, I'll now enable warnings and add some units tests then it will be ready for reviews.

cedric-anne (approved):

trasher (approved):

PR #15397, opened 2023-08-11
Actualization of the rus translation

krote5k:


trasher:
Translations are done on transifex: https://app.transifex.com/glpi/GLPI/dashboard/

PR #15401, opened 2023-08-11
Fix undefined properties

AdrienClairembault:
Fix call to undefined properties to avoid triggering the `__get` magic method that will be added in #15388.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |

### DIFF
--- a/src/NotificationTemplate.php
+++ b/src/NotificationTemplate.php
@@ -613,7 +613,7 @@ class NotificationTemplate extends CommonDBTM
         $mailing_options['items_id']     = method_exists($target->obj, "getField")
          ? $target->obj->getField('id')
          : 0;
-        if (isset($target->obj->documents)) {
+        if (property_exists($target, 'documents') && isset($target->obj->documents)) {
             $mailing_options['documents'] = $target->obj->documents;
         }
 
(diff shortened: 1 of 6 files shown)


cedric-anne (approved):

trasher (approved):

src/NotificationTemplate.php
```diff
@@ -613,7 +613,7 @@ public function getDataToSend(NotificationTarget $target, $tid, $to, array $user
         $mailing_options['items_id']     = method_exists($target->obj, "getField")
          ? $target->obj->getField('id')
          : 0;
-        if (isset($target->obj->documents)) {
+        if (property_exists($target, 'documents') && isset($target->obj->documents)) {
```
orthagh, line 616: Pr closed, but I don't understand this change

Should the part added be `property_exists($target, 'obj')`, no ? @trasher

orthagh: @AdrienClairembault sorry (missping)

AdrienClairembault: I think there is indeed a mistake and it should be `if (property_exists($target->obj, 'documents')`

PR #15405, opened 2023-08-14
Implement Kanban as Vue components

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

Improvements:
- Use of async and promises instead of callbacks.
- Cards not added to DOM when column is collapsed.
- Multiple changes to use reactive states. For example, cards automatically get shown/hid when the filters object is updated.
- Use of native Bootstrap dropdowns instead of custom event handlers and manually changing CSS styles.

New:
- Debug mode actions added to the main dropdown (same menu as the "Reset view" option available for regular users). Maybe not permanent? They are helpful as this is being migrated at least.
  - Load state (Debug) - Force-get the user state from the server
  - Save state (Debug) - Force save the user state
  - Refresh (Debug) - Force a refresh action

To look at later?:
- Only add the cards that would be visible into the DOM, plus some extra for smooth scrolling, to improve performance. Cards could be removed and added as needed as the user scrolls. More useful for ITIL Kanban where there could be thousands of items even when excluding closed ones.


cconard96:
Need to add tests (Kanban had no JS tests before), but the feature itself should be OK for review.

cedric-anne (changes requested):
I have multiple issues:

1. The global Kanban for tickets does not work anymore. I have some kind of blank page, and I cannot add any column.
image
```
Uncaught (in promise) TypeError: Cannot convert undefined or null to object
    at Function.values (<anonymous>)
    at Proxy.isColumnVisible (Kanban.vue:1152:28)
    at Kanban.vue:1317:50
    at renderList (runtime-core.esm-bundler.js:2776:1)
    at Proxy.render (Kanban.vue:1324:38)
    at renderComponentRoot (runtime-core.esm-bundler.js:816:1)
    at ReactiveEffect.componentUpdateFn [as fn] (runtime-core.esm-bundler.js:5764:1)
    at ReactiveEffect.run (reactivity.esm-bundler.js:178:1)
    at instance.update (runtime-core.esm-bundler.js:5814:1)
    at callWithErrorHandling (runtime-core.esm-bundler.js:158:1)
```

2. On project global Kanban (after a view reset), I have an error when I try to add a column.
```
Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'folded')
    at Object.<anonymous> (Kanban.vue:506:82)
    at mightThrow (jquery.js:3489:1)
    at process (jquery.js:3557:1)
```
The column is still added, and I am able to saw it when I refresh the view.

3. When trying to unfold a column, I have a JS error.
```
Uncaught TypeError: Cannot set properties of undefined (setting 'folded')
    at Proxy.updateFoldColumn (Kanban.vue:584:112)
    at onKanban:column_fold._cache.<computed>._cache.<computed> (Kanban.vue:1380:46)
    at callWithErrorHandling (runtime-core.esm-bundler.js:158:1)
    at callWithAsyncErrorHandling (runtime-core.esm-bundler.js:166:1)
    at emit (runtime-core.esm-bundler.js:664:1)
    at runtime-core.esm-bundler.js:7422:1
    at Proxy.toggleFolded (Column.vue:84:9)
    at onClick._cache.<computed>._cache.<computed> (Column.vue:171:32)
    at callWithErrorHandling (runtime-core.esm-bundler.js:158:1)
    at callWithAsyncErrorHandling (runtime-core.esm-bundler.js:166:1)
```

4. When I create a new project in the global kanban, even if I create it from the `+` on the `new` status column, it is created inside the first column (with no state). If I try to move it, I have JS errors.
```
Uncaught TypeError: target.getAttribute is not a function
    at HTMLUListElement.<anonymous> (html5sortable.amd.js:977:1)
(anonyme) @ html5sortable.amd.js:977
```
```
Uncaught TypeError: Cannot read properties of null (reading 'parentElement')
    at HTMLUListElement.<anonymous> (html5sortable.amd.js:1104:1)
```

cedric-anne:
1) After last commit, on a "blank" kanban, the "Add column" menu is not displayed entirely:
image

Also, the "There are no columns added to this Kanban yet" message disapears when this menu is open, even if no column is added yet.

2) There is an encoding issue inside the message displayed in the "Closed" column.
image

3) When there is too many closed tickets, clicking on creation menu is not working. Same for mass creation menu.
image

4) When trying to add a ticket, from any column, it does not work. Also, it then tries to load the following URL: `/front/undefineddisplayMessageAfterRedirect.php?get_raw=true`

5) Reordering columns does not seems to work.

6) The "There are no columns added to this Kanban yet" message  is displayed, even when there are already existing columns in the Kanban.
image

7) Moving a ticket from a column to another seems to work as long as the target column is empty. Once it contains another ticket, it does not work anymore.

cedric-anne:
There are some style/ux issues.

1. The item details panel overlaps the page header.
image

2. The "closed" column background was in yellow hen number of items was too important to be shown. Now it has no specific background:
Before:
image

After:
image

3. The drag and drop placeholder, that is supposed to show where the item will be dropped, is not prevent anymore.

cedric-anne (approved):

AdrienClairembault (changes requested):
Few points I've found after a quick functional review:

Some are bugs, others are differences I've noticed with the 10.0 version (maybe some of them are done on purpose and/or unavoidable ?).

1\) Missing padding in the "Add column" form:

image

GLPI 10.0 for comparison:

image

2\) Missing label and color for the "No status" column

image

GLPI 10.0 for comparison:

image

3\) Drag and drop is less clear than in GLPI 10

kanban_drag_main

GLPI 10:

kanban_drag

4\) Weird height changes when editing filters:

kanban_height

5\) Filters break when using multiple words:

kanban_filter_space

6\) Create status dropdown no longer hide the previous dropdown:

image

GLPI 10:

image

7\) Error when trying to delete or restore an item:

image

Does not happens everytime, I need to do some random actions before being able to reproduce the anomaly.

8\) Sometime I get this error in my logs when moving a card

image

It does not seems to impact the feature (card is moved successfully).

cconard96:
> 1) Missing padding in the "Add column" form:

I see nothing in the 10.0 code that indicates the empty space was intentional. It looks better now anyways except the missing spacing between the checkbox and color preview.

> 2) Missing label and color for the "No status" column

Default column color is now transparent instead of whatever the current theme designates for badge styles. Only issue is the text color was still defaulting to the one used for badges which is white. Looked fine in dark themes.

> 3) Drag and drop is less clear than in GLPI 10

Fixed.

> 4) Weird height changes when editing filters:

The tokenizer/search input code wasn't touched by this PR. I haven't figured out yet why it suddenly forgot how to show a popover properly (but only sometimes).

> 5) Filters break when using multiple words:

Same comment as above. Search input and the tokenizer for it wasn't touched. Use quotes around a phrase containing spaces to force it to stay part of the same filter.

> 6) Create status dropdown no longer hide the previous dropdown:

Side effect of using bootstrap's native dropdown functionality instead of custom CSS hacks. new behavior seems OK to me.

> 7) Error when trying to delete or restore an item:
> Does not happens everytime, I need to do some random actions before being able to reproduce the anomaly.

I haven't been able to recreate this error yet. I also don't know how a column could suddenly contain an undefined value in the array of cards.

> 8) Sometime I get this error in my logs when moving a card
> 
> It does not seems to impact the feature (card is moved successfully).

The saved state seems messed up if, as the error indicates, there is a column saved in the state with a "cards" property set to null instead of at least an empty array. Forcing a re-save of the user state should resolve the issue. In debug mode, you can use the "Save state" option in the "3 dot" menu next to the "Add column" button.

AdrienClairembault:
I confirm points 1, 2 and 3 are fixed 👍

Ok for points 4 and 5, they are indeed already broken in main.
Maybe we should create an issue to be sure we don't forget to fix it ?

Ok for point 6, I think the UX could be slightly improved but doesn't really need to be done in this pull request.

I've found how you can reproduce point 7: you need to delete a projet from the kanban which have some sub tasks.
The sub tasks are still displayed and will throw errors when trying to delete them (and some 403 when moving them).
Like points 4 and 5, its probably not related to your changes.
Lets add it to the know issues ?

AdrienClairembault (approved):

cedric-anne (changes requested):
Deleting a column from the view does not work. When page is refreshed, column is still shown.

Also, could you rebase ? I am not sure that change made in d1794ed0749b60bea33fd74be5bb418bc98e2df0 should be preserved.

cedric-anne (changes requested):
In the ticket global Kanaban, moving column seems to trigger the `move_column` request twice, and it produces unexpected results.

cconard96:
> In the ticket global Kanaban, moving column seems to trigger the `move_column` request twice, and it produces unexpected results.

The library used for sorting fires the sortstop event twice in the case of sorting cards too. This isn't a new behavior with this PR. I assume there is an issue with the library itself since my code removes all sortstop listeners during a refresh of sortables prior to adding one that sends the AJAX call.

cedric-anne (approved):

PR #15414, opened 2023-08-17
Fix deprecation notice for phpmarkdown lib

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #15412

There is already a bug report and a proposed PR in the library GitHub, but no activity from the maintainer for a while.


trasher:
Is that lib replaced in main branch? If it's no longer maintained, we should no longer use it :/

trasher (approved):

cconard96:
It is still listed as a dependency in the main branch.

orthagh:
I add in the [roadmap project](https://github.com/orgs/glpi-project/projects/2) we need to find a replacement for this lib

orthagh (approved):

PR #15415, opened 2023-08-17
Clear old escalation levels on SLA/OLA change

AdrienClairembault:
When a ticket is created, it's SLA/OLA escalation's date is computed properly:

image

However, if the SLA and/or OLA are changed through a rule, the escalation date does not change and keep the previous value:

image
_Escalation date should be 11h33, not 13h33_

This happens because we do not clear the data from the `glpi_slalevels_tickets` / `glpi_olalevels_tickets` tables when the SLA/OLA are updated.
This mean the old escalation level is kept in the database and shown in the tickets.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !28778

### DIFF
--- a/src/LevelAgreement.php
+++ b/src/LevelAgreement.php
@@ -1164,4 +1164,39 @@ abstract class LevelAgreement extends CommonDBChild
     {
         return static::$levelticketclass;
     }
+
+    /**
+     * Remove level of previously assigned level agreements
+     *
+     * @return void
+     */
+    public function clearInvalidLevels(): void
+    {
+        // CLear levels of others LA of the same type
+        // e.g. if a new LA TTR was assigned, clear levels from others (= previous) LA TTR
+        $level_ticket_class = $this->getLevelTicketClass();
+        $level_class = $this->getLevelClass();
+        $levels = (new $level_ticket_class())->find([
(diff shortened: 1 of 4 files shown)


src/LevelAgreement.php
```diff
@@ -1164,4 +1164,39 @@ public function getLevelTicketClass(): string
...
+     *
+     * @return void
+     */
+    public function clearInvalidLevels(): void
+    {
+        // CLear levels of others LA of the same type
+        // e.g. if a new LA TTR was assigned, clear levels from others (= previous) LA TTR
+        $level_ticket_class = $this->getLevelTicketClass();
+        $level_class = $this->getLevelClass();
+        $levels = (new $level_ticket_class())->find([
```
cconard96, line 1179: Is there a criteria missing to ensure only levels are cleared from a specific ticket?

AdrienClairembault: Indeed, good catch.
Looking into it.

cconard96 (approved):

orthagh (approved):

trasher (approved):

PR #15416, opened 2023-08-17
Wrong status when adding a planned task

ccailly:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #15350 #15355 

There are several inconsistencies in the ticket status transition.
- First, when a planned task is performed on a ticket in "Pending" status and the "Pending" status is unchecked in the add task menu, the ticket status changes to "Processing (assigned)" instead of the expected "Processing (planned)" status.
- Second, if a planned task is added, and the ticket is set to "pending" at the same time, the ticket status will be "Processing (planned)" instead of "Pending".

### DIFF
--- a/src/Features/ParentStatus.php
+++ b/src/Features/ParentStatus.php
@@ -135,7 +135,12 @@ trait ParentStatus
                     || $parentitem::isAllowedStatus($parentitem->fields["status"], CommonITILObject::ASSIGNED)
                 ) {
                     $needupdateparent = true;
-                    $update['status'] = CommonITILObject::ASSIGNED;
+                    // If begin date is defined, the status must be planned if it exists, rather than assigned.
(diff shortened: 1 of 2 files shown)


AdrienClairembault (changes requested):
Looks good overall but the second change (line 162) does not seem to be tested properly, 

Indeed, if I revert this specific change on line 162, the `testUpdateParentStatus` case pass without issue, which shouldn't happen.

AdrienClairembault (approved):

cconard96 (approved):

trasher (approved):

orthagh (approved):

PR #15417, opened 2023-08-17
Compare printers statistics on printer graph

ccailly:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !28627

Adds the ability to compare multiple printers.

image
image

### DIFF
--- a/front/printerlogcsv.php
+++ b/front/printerlogcsv.php
@@ -35,23 +35,34 @@
 
 use Glpi\Csv\CsvResponse;
 use Glpi\Csv\PrinterLogCsvExport;
+use Glpi\Csv\PrinterLogCsvExportComparison;
 
 include('../inc/includes.php');
 
 Session::checkRight("printer", READ);
 
 if (isset($_GET["id"])) {
-    $printer = Printer::getById($_GET["id"]);
+    $printers = array_map(fn ($id) => Printer::getById($id), $_GET["id"]);
+    $interval = $_GET['interval'] ?? 'P1Y';
+    $start = empty($_GET['start']) ? null : new DateTime($_GET['start']);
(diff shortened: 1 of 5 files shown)


AdrienClairembault:
I have some unexpected renderings:

image

It seems the missing dots happens when we don't have access to values for all printers on certains dates:

image

In this case I guess the value should be set to the previous known page counter to avoid this issue.

Also a PHP warning when one of the printers has 0 pages:

image

ccailly:
I solved the problem of rendering problems due to missing data by adding the `connectNulls` option, which connects points containing a null value to each other without adding any points.

However, when hovering over the graph, the tooltip doesn't necessarily display all values. (As illustrated below)
Is this appropriate, or would it be better to replace the `null` values with the previous value?

image

AdrienClairembault:
> Is this appropriate, or would it be better to replace the null values with the previous value?

I think it's better to fill the values with the previous known value, as it makes the data easier to read (especially for CSV exports).

AdrienClairembault:
Looks good functionally 👍

AdrienClairembault (approved):

AdrienClairembault:
@orthagh we didn't get a change to review together the specs for this, lets us know if the current implementation is correct for you.

I've already warned @ccailly that you may require some changes as I seem to recall that you weren't really convinced with the addition of this feature directly into the printer page.

AdrienClairembault (changes requested):
Export looks good 👍 

I've just noticed that there are still some empty values at the start of the graph:

image

AdrienClairembault:
After second thoughts that's OK, we don't need to show data for dates before where we have any informations on the page counters values.

orthagh (approved):

PR #15448, opened 2023-08-23
Equipment collision in rack due to model modification

ccailly:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !29207

Modifying an asset's model (by increasing its height, width or depth) can lead to overlapping with other assets in the rack.

image
For example, if I increase the height of my network equipment model to 5, the display becomes inconsistent (we can no longer interact with the computer).

### DIFF
--- a/src/CommonDCModelDropdown.php
+++ b/src/CommonDCModelDropdown.php
@@ -212,6 +212,36 @@ abstract class CommonDCModelDropdown extends CommonDropdown
         return parent::getSpecificValueToDisplay($field, $values, $options);
     }
 
+    public function getItemRackForModel(): array
+    {
+        $itemtype = str_replace('Model', '', get_called_class());
+        return array_filter(
+            (new Item_Rack())->find([
+                'itemtype' => $itemtype,
+            ]),
+            // Filter out items that are not using the model
+            fn ($item_rack) => $itemtype::getById($item_rack['items_id'])->fields[strtolower(get_called_class()) . 's_id'] == $this->fields['id']
+        );
+    }
+
+    private function checkCellFilled(array $cell, int $orientation, int $hpos, float $depth): bool
+    {
+        if ($hpos == Rack::POS_NONE) {
+            return $this->checkCellFilled($cell, $orientation, Rack::POS_LEFT, $depth)
+                || $this->checkCellFilled($cell, $orientation, Rack::POS_RIGHT, $depth);
+        }
+
+        if (isset($cell[$hpos])) {
+            $accurateCell = array_slice($orientation ? array_reverse($cell[$hpos]) : $cell[$hpos], 0, $depth * 4);
+
+            if (in_array(1, $accurateCell)) {
+                return true;
+            }
+        }
+
+        return false;
+    }
+
     public function prepareInputForAdd($input)
     {
         return $this->managePictures($input);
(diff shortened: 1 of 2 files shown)


src/CommonDCModelDropdown.php
```diff
@@ -212,14 +212,130 @@ public static function getSpecificValueToDisplay($field, $values, array $options
...
-        return $this->managePictures($input);
+        $input = $this->managePictures($input);
+
+        // Check if the model is used by an asset in a rack
+        // If so, check whether the new units required fit into the rack without modifying the positions
+        if (
+            isset($input['required_units']) && $input['required_units'] > $this->fields['required_units']
+            || isset($input['is_half_rack']) && $input['is_half_rack'] != $this->fields['is_half_rack']
+            || isset($input['depth']) && $input['depth'] != $this->fields['depth']
+        ) {
```
AdrienClairembault, line 260: The code use a bit too much nested loops and condition which we want to avoid if possible (https://en.wikibooks.org/wiki/Computer_Programming/Coding_Style/Minimize_nesting)

I would suggest:
* Isolate the code into a function (to keep `prepareInputForUpdate` light)
* Invert the first condition (if the input is OK -> return xxx; example here: https://www.jetbrains.com/help/resharper/InvertIf.html)
* Split the iterating on position + error reporting into two distinct parts (first part: build an array of position that need to be checked, second part: check each position and add error) -> this avoid duplicating the error logging code and reduce nesting a bit when handling the errors

src/CommonDCModelDropdown.php
```diff
@@ -212,14 +212,130 @@ public static function getSpecificValueToDisplay($field, $values, array $options
...
+                if ($item['hpos'] == Rack::POS_NONE) {
+                    continue;
+                }
+
+                $item_rack->update([
+                    'id' => $item['id'],
+                    'hpos' => Rack::POS_NONE,
+                ]);
+            }
+        }
```
AdrienClairembault, line 338: Could you add a small code comment to indicate that we are doing this because a full rack model can't use an half space ?

Also maybe the opposite check is also needed ? An half-rack model should always use an half space (move it to `POS_NONE` -> `POS_LEFT` ?) 

Unless I'm mistaken this code does not seem to be tested, could you add some tests ?

tests/functional/Item_Rack.php
```diff
@@ -490,4 +491,295 @@ public function testAdd()
...
+        // Create a ComputerModel
+        $model1 = new \ComputerModel();
+        $this->integer(
+            (int)$model1->add([
+                'required_units'  => 1,
+                'depth'           => 1,
+                'is_half_rack'    => 0,
+                'entities_id'     => 0,
+            ])
+        )->isGreaterThan(0);
```
AdrienClairembault, line 509: Its not mandatory to use it but in case you don't know we have a `createItem` method that does the `isGreaterThan(0)` + some others checks for you.

We also have `createItems` and `updateItem`.

src/CommonDCModelDropdown.php
AdrienClairembault: Also could be isolated into a method with a fitting name :)

AdrienClairembault (changes requested):

AdrienClairembault (approved):

trasher (approved):

orthagh (approved):

src/CommonDCModelDropdown.php
```diff
@@ -212,14 +212,202 @@ public static function getSpecificValueToDisplay($field, $values, array $options
...
+
+        if (isset($cell[$hpos])) {
+            // Get the first $depth * 4 units of the cell to check if they are filled
+            $accurateCell = array_slice(
+                $orientation ?
+                    array_reverse($cell[$hpos]) // If orientation is rear, reverse the array
+                    : $cell[$hpos],
+                0,
+                $depth * 4
```
AdrienClairembault, line 274: @ccailly 

Depth can be a float, did you test for this scenario ?

I don't remember why we do `*4` here but isn't there a risk of issue if depth is a float since array_slice only accept int values ?

PR #15457, opened 2023-08-25
login can be an email (+ char accepted)

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !28929

### DIFF
--- a/tests/FrontBaseClass.php
+++ b/tests/FrontBaseClass.php
@@ -59,14 +59,26 @@ class FrontBaseClass extends \GLPITestCase
         parent::afterTestMethod($method);
     }
 
-    protected function logIn()
+    protected function loginProvider()
+    {
+        return [
+            [TU_USER, TU_PASS],
+            ['jsmith123', TU_PASS],
+            ['john+doe@doe.com', TU_PASS],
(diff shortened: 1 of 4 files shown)


trasher:
This should be tested

trasher (approved):

tests/FrontBaseClass.php
```diff
@@ -59,14 +59,26 @@ public function afterTestMethod($method)
...
+    protected function loginProvider()
+    {
+        return [
+            [TU_USER, TU_PASS],
+            ['jsmith123', TU_PASS],
+            ['john+doe@doe.com', TU_PASS],
+        ];
+    }
+
+    /**
```
trasher, line 71: I'm not sure this is a good idea. This login method is meant to be called from subclasses several times, and we need a real login on a well known user in this case.

I guess a better solution would be to create a new method with the provider which will just call the existing one

trasher (changes requested):

trasher (approved):

PR #15478, opened 2023-08-31
Fix : Duplicate mail notification

Lainow:
### <samp>🤖 Generated by Copilot at 9c3ab02</samp>

Fix notification issue for satisfaction surveys. Prevent sending a notification when only the `date_answered` field of a `TicketSatisfaction` object is updated by the system.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #number

### DIFF
--- a/src/TicketSatisfaction.php
+++ b/src/TicketSatisfaction.php
@@ -201,7 +201,7 @@ class TicketSatisfaction extends CommonDBTM
 
         if (!isset($this->input['_disablenotif']) && $CFG_GLPI["use_notifications"]) {
             $ticket = new Ticket();
-            // date_answer is always set when updating
+            // date_answer is always updated even if the comment or rate does not change
             if (count($this->updates) > 1 && $ticket->getFromDB($this->fields['tickets_id'])) {
                 NotificationEvent::raiseEvent("replysatisfaction", $ticket);
             }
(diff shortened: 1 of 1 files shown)


trasher (approved):

src/TicketSatisfaction.php
```diff
@@ -201,7 +201,8 @@ public function post_UpdateItem($history = 1)
 
         if (!isset($this->input['_disablenotif']) && $CFG_GLPI["use_notifications"]) {
             $ticket = new Ticket();
-            if ($ticket->getFromDB($this->fields['tickets_id'])) {
+            // date_answer is always set when updating
```
stonebuzz, line 204:
```suggestion
            // date_answer is always updated even if the comment or rate does not change
```

stonebuzz (changes requested):

stonebuzz (approved):

trasher:
As said in our internal discussion channel, I wonder if the date should have been updated if nothing else has been.

orthagh:
> if nothing else has been

I agree.
- If only date updated -> discard changes
- if something else updated -> send notification

stonebuzz:
As discussed IRL with @orthagh 

Loading the data from the ```DB``` into the ```prepareInputForUpdate``` to compare with the data sent (```$input```) by the user doesn't seem like a good idea.


```pre_updateInDB``` doesn't condition the update action.
Even if I return ```false``` / ```empty array``` subsequent processing will be executed

Current PR is easier to manage

PR #15536, opened 2023-09-05
Show only the rack item form itself in modal

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

Currently, the entire rack item form, including tabs, are being displayed in the modal when adding an item to the rack. This causes an issue as the new tabs share the same ID as the panel on the page. Before the "All" tab was reworked, this didn't seem to be an issue, but now the tabs on the page are hidden when the modal is opened (and an error trying to redefine a function).

Since the modal will only ever be used for a new item, it makes sense to just show the form itself without the extra tab stuff. Currently, the only way to have multiple tabbed forms loaded at once would be to load the others in an iframe. Randomizing the IDs and function names isn't a clean solution as many places simple call `reloadTab` and we wouldn't know which set of tabs it is referring to. The calling script wouldn't know if it is on the main page or a modal either.

### DIFF
--- a/front/item_rack.form.php
+++ b/front/item_rack.form.php
@@ -81,7 +81,13 @@ if (isset($_GET['id'])) {
 $ajax = isset($_REQUEST['ajax']) ? true : false;
 
 if ($ajax) {
-    $ira->display($params);
+    $item = new Item_Rack();
+    $id = $params['id'] ?? 0;
+    if ($id > 0 && !$item->getFromDB($params['id'])) {
+        Html::displayNotFoundError();
+        return;
+    }
+    $item->showForm($id, $params);
 } else {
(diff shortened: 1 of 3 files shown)


orthagh (approved):
In fact, I wonder if we could edit an item_rack with a modal (with the pencil icon on the line)
but same, there is no tab currently.

Can you check if it's quick to move the edit control to modal ?

orthagh (approved):

trasher (approved):

cedric-anne (approved):

PR #15537, opened 2023-09-05
Fix dashboard browser cache hash

AdrienClairembault:
The cache key does not take current entities into account due to a typo that's probably here since the original implementation in GLPI 9.5.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #15431


AdrienClairembault:
Fix #15431.

orthagh (approved):

trasher (approved):

Issue #15435, opened 2023-08-22
SIM card reset


R0ot16:
### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Version

10.0.7

### Bug description

Hi,

when i create a new sim card, this card is reseted.

The glpi agent for android is installed but the phone is not used for the moment.

Sorry for my english.

Field has been reseted :
![image](https://github.com/glpi-project/glpi/assets/26124413/28f8bcdc-7dfe-49fa-99ce-fbcb8ca6959f)

### Relevant log output

```shell
PHP ERROR : 
[2023-06-05 16:11:11] glpiphplog.WARNING: Test logger  
[2023-06-05 16:12:58] glpiphplog.WARNING: Test logger  
[2023-06-05 16:13:12] glpiphplog.WARNING: Test logger  

[2023-08-12 14:20:22] glpisqllog.WARNING: DBmysql::query() in /var/www/html/glpi/src/DBmysql.php line 404
  *** MySQL query warnings:
  SQL: SELECT * FROM `glpi_lockedfields` WHERE `itemtype` = 'Item_DeviceSimcard' AND ((`items_id` = '' OR `is_global` = '1'))
  Warnings: 
1292: Truncated incorrect DECIMAL value: ''
  Backtrace :
  src/DBmysqlIterator.php:112                        
  src/DBmysql.php:1056                               DBmysqlIterator->execute()
  src/Lockedfield.php:180                            DBmysql->request()
  src/Lockedfield.php:153                            Lockedfield->getLocks()
  src/Inventory/Asset/InventoryAsset.php:433         Lockedfield->getLockedNames()
  src/Inventory/Asset/Device.php:175                 Glpi\Inventory\Asset\InventoryAsset->handleInput()
  src/Inventory/Asset/MainAsset.php:921              Glpi\Inventory\Asset\Device->handle()
  src/Inventory/Asset/MainAsset.php:841              Glpi\Inventory\Asset\MainAsset->handleAssets()
  src/RuleImportAsset.php:969                        Glpi\Inventory\Asset\MainAsset->rulepassed()
  src/Rule.php:1525                                  RuleImportAsset->executeActions()
  src/RuleCollection.php:1625                        Rule->process()
  src/Inventory/Asset/MainAsset.php:575              RuleCollection->processAllRules()
[...]

PR #15541, opened 2023-09-06
fix(inventory): fix inventory / GLPI mapping

stonebuzz:
Try to fix https://github.com/glpi-project/glpi/issues/15435

The reconciliation (to know whether to add or update the SIM card) is done on the ```serial``` / ```msin``` field which (in the latest versions of Android due to Google restrictions) are empty.

So GLPI deletes the SIM card and adds a new one at each inventory.

This first draft consists of basing the reconciliation on the link (```imtes_id```, ```itemtype```, ```is_dynamic```) to the asset

This work if asset have only one SIM card.

But  could cause problems if the asset has several simcards.

It may be necessary to generate a ```UUID``` to be used as the serial number (from ```android-inventory-library```).

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #15435

### DIFF
--- a/src/Inventory/Asset/Simcard.php
+++ b/src/Inventory/Asset/Simcard.php
@@ -43,8 +43,9 @@ class Simcard extends Device
     public function prepare(): array
     {
         $mapping = [
-            'imsi' => 'serial'
+            'subscriber_id' => 'msin'
         ];
+
         foreach ($this->data as $k => &$val) {
             foreach ($mapping as $origin => $dest) {
                 if (property_exists($val, $origin)) {
--- a/src/Inventory/Asset/Simcard.php
+++ b/src/Inventory/Asset/Simcard.php
@@ -37,13 +37,14 @@ namespace Glpi\Inventory\Asset;
 
 use CommonDBTM;
 use Glpi\Inventory\Conf;
+use Toolbox;
 
 class Simcard extends Device
 {
     public function prepare(): array
     {
         $mapping = [
-            'subscriber_id' => 'msin'
+            'subscriber_id' => 'msin',
         ];
 
         foreach ($this->data as $k => &$val) {
@@ -53,6 +54,7 @@ class Simcard extends Device
                 }
             }
         }
+
         return $this->data;
     }
 
--- a/src/Item_DeviceSimcard.php
+++ b/src/Item_DeviceSimcard.php
@@ -139,8 +139,9 @@ class Item_DeviceSimcard extends Item_Devices
     public function getImportCriteria(): array
     {
         return [
-            'serial' => 'equal',
-            'msin' => 'equal',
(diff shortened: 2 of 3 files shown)


trasher:
I'm really not fan of proposed solution; asset link should not be a "dedup" key. The PR opened on inventory format seems to be another solution, am I wrong? If so, we'll wait for it to be merged and adapt current PR.

stonebuzz:
So, 
with https://github.com/glpi-project/inventory_format/pull/120

We now have the ```subscriber_id``` field, which is incremental and tracks the number of SIM cards.

(I was able to pick up an inventory with two sim cards)

```xml
    <SIMCARDS>
      <COUNTRY>fr</COUNTRY>
      <OPERATOR_CODE>2081</OPERATOR_CODE>
      <OPERATOR_NAME>Orange F</OPERATOR_NAME>
      <SERIAL>89330126162002971850</SERIAL>
      <STATE>SIM_STATE_READY</STATE>
      <LINE_NUMBER></LINE_NUMBER>
      <SUBSCRIBER_ID>1</SUBSCRIBER_ID>
    </SIMCARDS>
    <SIMCARDS>
      <COUNTRY>fr</COUNTRY>
      <OPERATOR_CODE>20815</OPERATOR_CODE>
      <OPERATOR_NAME>Free</OPERATOR_NAME>
      <SERIAL>8933150418115179588</SERIAL>
      <STATE>SIM_STATE_READY</STATE>
      <LINE_NUMBER>+33xxxxxxxx</LINE_NUMBER>
      <SUBSCRIBER_ID>2</SUBSCRIBER_ID>
    </SIMCARDS>
```

The reconciliation method from ```Item_DeviceSimcard``` can be kept as follows

```php
    public function getImportCriteria(): array
    {
        return [
            'serial' => 'equal',
            'msin' => 'equal',
        ];
    }
```

the mapping between the fields from inventory file and those in GLPI has been fixed

```php
        $mapping = [
            'subscriber_id' => 'msin'
        ];
```
Because old version, never work (```imsi``` never exist from original file)

```php
        $mapping = [
            'imsi' => 'serial'
        ];
```

I'm waiting for with https://github.com/glpi-project/inventory_format/pull/120 to be merged so I can set up the unit tests

stonebuzz:
No problem with the old inventories because the ```imsi``` field never exist from phone inventory and the ```serial``` field was removed ```inventory_format```.

So the old inventories never had ```serial``` or ```imsi``` (subscriber ID).

cedric-anne (approved):

trasher:
Looks good to me, but PR title has to be changed, it no longer suits the change

PR #15564, opened 2023-09-12
feat(replica): review config/status tab

flegastelois:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | N/A

- ~~TODO: test if current connect database user have corrects grants to use "show master status" or "show slave status"~~

image

image


flegastelois:
image

Is this sufficient as an error message in case of missing "grant"?

flegastelois:
image

After the last commit (note that this is a MySQL error message).

templates/pages/setup/general/dbreplica_setup.html.twig
```diff
@@ -52,10 +58,113 @@
...
+                  <div id="collapse-primary" class="accordion-collapse collapse" aria-labelledby="heading-primary" data-bs-parent="#accordion-primary">
+                     <div class="accordion-body">
+                        {% if primary_unknow %}
+                           <div class="alert alert-danger" role="alert">
+                              <span class="text-secondary">{{ __('MySQL server returned an error') }}:</span>
+                              <p class="mt-2">{{ replication_status['primary']['error'] }}</p>
+                           </div>
+                        {% else %}
+                           <ul class="p-0 list-unstyled space-y-3">
+                              <li><span class="text-secondary">Server ID:</span> {{ replication_status['primary']['server_id'] }}</li>
```
trasher, line 102: Is that intended those strings are not translatable?

trasher (commented):
Seems good to me appart comment on translations

templates/pages/setup/general/dbreplica_setup.html.twig
flegastelois: I told myself that as these are technical terms, linked to MySQL variables which are in English, the translation does not make sense.

flegastelois:
Switch to draft again, I have to find a good solution to display the current "Difference between main and replica", indicating that it is not the value of the MySQL status, but a comparison of the date_mod of the glpi_logs table...

Issue #403, opened 2016-01-27
erreur dans log php


Sismic:
2016-01-27 17:32:14 [26@PLAGLPI001]
  *** PHP Notice(8): Undefined index: lock_use_lock_item
  Backtrace :
  inc/objectlock.class.php:379
  inc/commonglpi.class.php:1117                      ObjectLock::manageObjectLock()
  front/ticket.form.php:181                          CommonGLPI->display()

tomolimo:
Hello
How do reproduce the errors?
Thank you
Tomolimo


-------- Original message --------
From: Sismic
Date:27/01/2016 17:33 (GMT+01:00)
To: glpi-project/glpi
Subject: [glpi] erreur dans log php (#403)


2016-01-27 17:32:14 [26@PLAGLPI001]
*** PHP Notice(8): Undefined index: lock_use_lock_item
Backtrace :
inc/objectlockclassphp:379
inc/commonglpiclassphp:1117 ObjectLock::manageObjectLock()
front/ticketformphp:181 CommonGLPI->display()

orthagh:
imo, update from 0.91 to 0.91 (reproduced one hour ago)

So, @Sismic , change you glpi version in glpi_configs table to 0.90.1 and redo update

Sismic:
PHP Notice: Undefined index: names_format in /var/www/glpi/inc/db.function.php at line 1186
PHP Notice: Undefined index: is_ids_visible in /var/www/glpi/inc/db.function.php at line 1191
PHP Notice: Undefined index: names_format in /var/www/glpi/inc/db.function.php at line 1186
PHP Notice: Undefined index: is_ids_visible in /var/www/glpi/inc/db.function.php at line 1191
PHP Notice: Undefined index: names_format in /var/www/glpi/inc/db.function.php at line 1186
PHP Notice: Undefined index: is_ids_visible in /var/www/glpi/inc/db.function.php at line 1191
PHP Notice: Undefined index: names_format in /var/www/glpi/inc/db.function.php at line 1186
PHP Notice: Undefined index: is_ids_visible in /var/www/glpi/inc/db.function.php at line 1191
PHP Notice: Undefined index: names_format in /var/www/glpi/inc/db.function.php at line 1186
PHP Notice: Undefined index: is_ids_visible in /var/www/glpi/inc/db.function.php at line 1191
PHP Notice: Undefined index: names_format in /var/www/glpi/inc/db.function.php at line 1186
[...]

PR #15583, opened 2023-09-15
fix(inventory): prevent agent deletion from unmanaged process

stonebuzz:
Try to avoids agent deletion, when an ```Unmanaged device``` matches (by MAC) an asset managed by another agent

Use case

1. Do inventory of a computer with ```Agent``` ```A``` and MAC 00:1B:44:11:3A:B7
2. Do Network Discovery with ```Agent``` ```B``` on the computer IP (I know it's not meant to be, but on an IP range it can happen without meaning to.)
3. When GLPI handle ```Unmanaged device```,   it check if an asset already exist with mac if found (this is the case here)
4. GLPI delete old agent related to ```itemtype``` / ```items_id``` and different ```Agent```  (this is the case too)

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #403


tests/functional/Glpi/Inventory/Assets/Unmanaged.php
```diff
@@ -35,6 +35,7 @@
 
 namespace tests\units\Glpi\Inventory\Asset;
 
+use Agent;
```
trasher, line 38:
```suggestion
```

tests/functional/Glpi/Inventory/Assets/Unmanaged.php
```diff
@@ -457,4 +458,86 @@ public function testInventory()
...
+
+        //redo inventory
+        $this->doInventory($xml, true);
+
+        //check Unmanaged not exist (since an asset with the same MAC exists)
+        $unmanaged = new \Unmanaged();
+        $this->boolean($unmanaged->getFromDbByCrit(['name' => 'glpixps']))->isFalse();
+
+        //reload agent
+        $agent_reload = new Agent();
```
trasher, line 536:
```suggestion
        $agent_reload = new \Agent();
```

tests/functional/Glpi/Inventory/Assets/Unmanaged.php
stonebuzz: remove useless use statement

trasher (approved):

cedric-anne (approved):

orthagh (approved):

PR #15592, opened 2023-09-16
HLAPI Expand endpoint docs

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

Changes should affect Swagger UI documentation only.

- Added valid parameters for search endpoints (RSQL query, start, and limit)
- When a component schema has no description but has an itemtype associated, the fallback description is the itemtype's name
- GLPI-Entity, GLPI-Entity-Recursive and GLPI-Profile headers available in Swagger UI as parameters
- Added missing parameters and responses to route docs
- Added Accept-Langauge header parameter for every endpoint along with some examples for quick testing and to show the user the required format. Examples can be selected to pre-fill the value, but can also be entered manually.
- Added placeholder support for cases where generic endpoints are expanded in the documentation. For example, `/Assistance/{itemtype}` expands to endpoints specific for Ticket, Change and Problem as `itemtype` is specified that it can only ever be Ticket, Change or Problem. Placeholders can now be used in the schema property for the parameters and responses for the documentation on the `/Assistance/{itemtype}` endpoint. A schema of `{itemtype}` is resolved to the expanded value. `{itemtype}[]` is similarly resolved, but as an array type.

### DIFF
--- a/src/Api/HL/Controller/AbstractController.php
+++ b/src/Api/HL/Controller/AbstractController.php
@@ -81,6 +81,29 @@ abstract class AbstractController
         ]
     ];
 
+    protected const PARAMETER_START = [
+        'name' => 'start',
+        'description' => 'The first item to return',
+        'location' => 'query',
+        'schema' => [
+            'type' => Doc\Schema::TYPE_INTEGER,
+            'format' => Doc\Schema::FORMAT_INTEGER_INT64,
+            'minimum' => 0,
+            'default' => 0,
+        ]
+    ];
+
+    protected const PARAMETER_LIMIT = [
+        'name' => 'limit',
+        'description' => 'The maximum number of items to return',
+        'location' => 'query',
+        'schema' => [
+            'type' => Doc\Schema::TYPE_INTEGER,
+            'format' => Doc\Schema::FORMAT_INTEGER_INT64,
+            'minimum' => 0,
+        ]
+    ];
+
     /**
      * @return array<string, Doc\Schema>
      */
--- a/src/Api/HL/Controller/AbstractController.php
+++ b/src/Api/HL/Controller/AbstractController.php
@@ -72,6 +72,15 @@ abstract class AbstractController
     public const CRUD_ACTION_RESTORE = 'restore';
     public const CRUD_ACTION_LIST = 'list';
 
+    protected const PARAMETER_RSQL_FILTER = [
+        'name' => 'filter',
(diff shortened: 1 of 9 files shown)


orthagh:
I can't understand the last point, do you mind explain a bit ?

cconard96:
> I can't understand the last point, do you mind explain a bit ?

For example, there is a route defined in the ITILController for searching tickets, changes and problems.
```php
#[Route(path: '/{itemtype}/', methods: ['POST'])]
#[Doc\Route(
        description: 'List or search Tickets, Changes or Problems'
        description: 'List or search Tickets, Changes or Problems',
        parameters: [self::PARAMETER_RSQL_FILTER, self::PARAMETER_START, self::PARAMETER_LIMIT],
        responses: [
            ['schema' => '{itemtype}[]']
        ]
    )]
```
It has a placeholder in the path for the specific itemtype. On the controller itself, this placeholder has a pattern defined as a static set of values 'Ticket|Change|Problem'.

When the OpenAPI schema is generated, instead of the route being added as-is, it sees that the placeholder in the path has specific values it can be so it goes through those values and adds a route for each possibility.

That was done previously. The change, is that when it expands the routes into the separate ones, it can look at the parameters and responses defined in the documentation attribute and apply a similar replacement for the schema properties.

In the example above, the response is defined as `{itemtype}[]` which means that the endpoint will return an array of whatever the current itemtype is. When the documentation is generated and it goes through the possible values of `itemtype` in the path, it does the same here. `{itemtype}[]` becomes `Ticket[]` for example which then is resolved to the final representation needed for OpenAPI:

```php
[
    'type' => 'array',
    'items' => [
        '$ref' => '#/components/schemas/Ticket'
    ]
]
```

An example of the placeholder as a parameter can be seen for the route to create tickets, changes and problems:
```php
#[Doc\Route(
        description: 'Create a new Ticket, Change or Problem',
        parameters: [
            [
                'name' => '_',
                'location' => Doc\Parameter::LOCATION_BODY,
                'schema' => '{itemtype}',
            ]
        ]
    )]
```

In that case, it isn't an array so it gets resolved simply as:
```php
[
    '$ref' => '#/components/schemas/Ticket'
]
```

orthagh (approved):

cedric-anne (approved):

PR #15611, opened 2023-09-20
Fix relative path for themes stored outside of GLPI_ROOT + cache on last update

AdrienClairembault:
`str_replace(GLPI_ROOT, '', $path);` won't work if `GLPI_THEMES` is outside `GLPI_ROOT`.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !28015

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -69,6 +69,7 @@
         "symfony/console": "^6.3",
         "symfony/css-selector": "^6.3",
         "symfony/dom-crawler": "^6.3",
+        "symfony/filesystem": "^6.3",
         "symfony/html-sanitizer": "^6.3",
         "symfony/mailer": "^6.3",
         "symfony/mime": "^6.3",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "8eed5bf22715616da1009259f556ff64",
+    "content-hash": "7b6f83fbddf268da10657e07bf04956d",
     "packages": [
         {
             "name": "apereo/phpcas",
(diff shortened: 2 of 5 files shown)


trasher:
Why not just replace `GLPI_ROOT` with `GLPI_THEMES`?

AdrienClairembault:
> Why not just replace `GLPI_ROOT` with `GLPI_THEMES`?

The path will be exposed on the client side with an http request to `css.php`, thus we need to avoid disclosing any absolute path that would expose the server directory tree.

trasher:
I do not understand. This is a str_replace, no absolute path will be exposed; GLPI_ROOT was not.

AdrienClairembault:
Note: changes to `front/css.php` could probably be done to remove the need for a relative URL by adding an extra `&theme=true` param which will set the base dir to `GLPI_THEMES`

Not sure why it wasn't done that way originally, maybe some constraint I'm not aware of ?

AdrienClairembault:
> I do not understand. This is a str_replace, no absolute path will be exposed; GLPI_ROOT was not.

`GLPI_ROOT` was not exposed because it was replaced by an empty string, leaving only the relative path to the theme directory (`/files/_themes/my_theme.css`).
However, this only work if `GLPI_THEMES` is a subfolder of `GLPI_ROOT` (= `GLPI_THEMES` contains `GLPI_ROOT`).

If you replace `GLPI_THEMES` completely you end up with `/my_theme.css`, which `front/css.php` wont load because it doesn't know you are requesting a theme (it could be a css file from a plugin; you need an exact path to the file).

Thus my alternate proposal of adding an additional query parameter to `front/css.php` to indicate we want to load a theme.
Then we can only send the theme name and `front/css.php` will know to look for it in `GLPI_THEMES` + the core theme folder, without the need for a full qualified path.

AdrienClairembault:
After trying for a while the alternate solution seems possible but difficult to integrate properly without rewriting a lot of code, I think this one this better for now.

AdrienClairembault:
Second commit handle some cache issue when custom themes are modified.

Our usual URL key based on GLPI version is not enough here since the theme files may be modified in between versions by the system administrator or some plugins.

trasher (approved):

cedric-anne (approved):

PR #15631, opened 2023-09-26
Fix category predefined field in ITIL templates

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #15363

### DIFF
--- a/src/CommonITILObject.php
+++ b/src/CommonITILObject.php
@@ -7646,7 +7646,7 @@ abstract class CommonITILObject extends CommonDBTM
 
        // Check if profile / entity set type and category and try to load template for these values
         if ($template_loaded) { // template loaded for profile or entity
-            $newtype              = $type;
+            $newtype              = null;
             $newitilcategories_id = $itilcategories_id;
            // Get predefined values for ticket template
             if (isset($tt->predefined['itilcategories_id']) && $tt->predefined['itilcategories_id']) {


trasher:
All the logic of applying predefined fields is UI side; I've not been able to add a test on that. 
It's working on my instance.

cedric-anne (commented):
I do not really understand the method itself, it is a pain to understand (`$type` may be an integer `INCIDENT_TYPE/DEMAND_TYPE`, `null`, or `true`), but the previous affectation seems OK to me.
As far as I understand, we just want to check if the template loaded previously assigns a new type/category, and, in this case, try to see if there is a better template to use. You change just seems to disable this behaviour (unless the template itself defines a type), and I do not know if it is correct.
Maybe an alternative solution could be:
```diff
diff --git a/src/CommonITILObject.php b/src/CommonITILObject.php
index 579b388b75..8885e1dc27 100644
--- a/src/CommonITILObject.php
+++ b/src/CommonITILObject.php
@@ -7656,8 +7656,8 @@ abstract class CommonITILObject extends CommonDBTM
                 $newtype = $tt->predefined['type'];
             }
             if (
-                $newtype
-                && $newitilcategories_id
+                $newtype != $type
+                && $newitilcategories_id != $itilcategories_id
             ) {
                 $categ = new ITILCategory();
                 if ($categ->getFromDB($newitilcategories_id)) {
```

trasher:
Alternative solution may indeed be correct; but I do not really understand the whole code :/

trasher:
@orthagh an opinion?

trasher:
> @orthagh an opinion?

ping?

PR #15656, opened 2023-09-29
feat(search): use keyboard arrows for pagination

flegastelois:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | N/A

### <samp>🤖 Generated by Copilot at 8004de3</samp>

Add keyboard navigation shortcuts for search pagination. The change modifies the `templates/components/search/controls.html.twig` file to include a JavaScript code block that listens for keypress events and triggers the appropriate page link.

### DIFF
--- a/src/AuthLDAP.php
+++ b/src/AuthLDAP.php
@@ -2091,6 +2091,8 @@ class AuthLDAP extends CommonDBTM
                 $filter = $values['ldap_filter'];
             }
 
+            d("LDAP filter: $filter");
+
             if ($values['script'] && !empty($values['begin_date'])) {
                 $filter_timestamp = self::addTimestampRestrictions(
                     $values['begin_date'],
--- a/src/AuthLDAP.php
+++ b/src/AuthLDAP.php
@@ -2091,8 +2091,6 @@ class AuthLDAP extends CommonDBTM
                 $filter = $values['ldap_filter'];
             }
 
-            d("LDAP filter: $filter");
-
             if ($values['script'] && !empty($values['begin_date'])) {
                 $filter_timestamp = self::addTimestampRestrictions(
                     $values['begin_date'],
--- a/templates/components/search/controls.html.twig
+++ b/templates/components/search/controls.html.twig
@@ -243,5 +243,46 @@ $(document).ready(function() {
          navigator.clipboard.writeText(data).then(copy_success, copy_error);
       }).fail(copy_error);
    });
+
+   // Keyboard navigation shortcuts (left/right arrows)
+   var page_link_start = $(".page-link-start");
+   var page_link_prev  = $(".page-link-prev");
+   var page_link_next  = $(".page-link-next");
+   var page_link_last  = $(".page-link-last");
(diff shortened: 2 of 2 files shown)


cconard96:
These shortcuts interfere with regular behavior like moving the caret in a text box (like a text input for search criteria). Ctrl + Left/Right arrow also is used for moving to the previous or next word in a text input, at least in Chrome.
A check needs added to ensure there is no focus on an input element.

flegastelois:
I added the exception for the focus, and I removed the "Ctrl+..." to use the "Home" and "End" keys instead.

cconard96:
Home and End are also used by Chrome 😛. Used to go to the top or bottom of a page. Ctrl + Home and Ctrl + End seem unused in Chrome, but Firefox uses them to go to the start or end of text. Alt + Home is used to go to your homepage.

https://support.google.com/chrome/answer/157179?
https://support.mozilla.org/en-US/kb/keyboard-shortcuts-perform-firefox-tasks-quickly

flegastelois:
> Home and End are also used by Chrome 😛. Used to go to the top or bottom of a page. Ctrl + Home and Ctrl + End seem unused in Chrome, but Firefox uses them to go to the start or end of text. Alt + Home is used to go to your homepage.
> 
> https://support.google.com/chrome/answer/157179 https://support.mozilla.org/en-US/kb/keyboard-shortcuts-perform-firefox-tasks-quickly

😞 so only ArrowLeft/ArrowRight are simple and seems not to be used alone outside an input?

I don't know if we can find simple combination for start/last links? If not, I remove this piece of code.
(by that I mean, we should not offer users to use "Ctrl+Alt+Left Elbow+nose on Shift+End" 😆)

cconard96:
I think the original shortcuts are usable since they only seem to have a default use when inside of a form input (or element with the contenteditable attribute set to true). As long as one of those types of elements is not focused, it should be OK.

cconard96:
The other thing I see from searching Google, is a jQuery plugin that handled quick double-presses of the left and right arrow keys to go to the first/last page and regular presses to navigate to previous/next pages.

flegastelois:
> I think the original shortcuts are usable since they only seem to have a default use when inside of a form input (or element with the contenteditable attribute set to true). As long as one of those types of elements is not focused, it should be OK.

Done 🔝

flegastelois:
My last commit add "j" and "k" as suggested by nerds like @orthagh 😆 

(https://www.skipser.com/p/2/p/why-google-uses-j-and-k-keys.html)

cedric-anne (commented):
Seems OK, but I wonder if it may break some specific controls provided by accessibility tools/devices.

cconard96 (approved):

flegastelois:
Although bringing an improvement for keyboard navigation, we must not be satisfied with that alone, we must go further for accessibility (https://www.w3.org/WAI/WCAG21/Understanding/keyboard -accessible). I close this PR.

Issue #15657, opened 2023-09-29
Task date in the list


evertondemoura:
### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

10.0.9 

### Bug description

When creating a task from a ticket, the list shows the date of creation or editing and does **not include the date of the task**.
To view the task date, I need to edit the task and if the ticket is already finished, I need to reopen the ticket and change the task to view the task date.

In GLPI 9 this function existed, but in GLPI 10 I can't configure it.
Does it still exist or do I now need a plugin?
If so, what would it be?

exemple GLPI 10
![image](https://github.com/glpi-project/glpi/assets/15331384/47cec8dc-0086-4836-9df9-28f9b6b31fee)

![image](https://github.com/glpi-project/glpi/assets/15331384/34aca14d-061b-455c-94cd-ad527a4ba268)

exemple GLPI 9 
![image](https://github.com/glpi-project/glpi/assets/15331384/171aba03-327b-4cb5-8f38-c20ccb9442ca)

stonebuzz:
Hi @evertondemoura 

can you try this 

https://github.com/glpi-project/glpi/issues/15657

Best regards

evertondemoura:
Correct, in glpi 9 it was in this shared format.
However, in 10 I couldn't implement it.
If there is a plugin for this, can you share?

stonebuzz:
Sorry for wrong link 

There is #15659

PR #15659, opened 2023-09-29
feat(Timeline): add task date

stonebuzz:
Add missing task date in the timeline display

 image

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #15657

### DIFF
--- a/templates/components/itilobject/timeline/form_task.html.twig
+++ b/templates/components/itilobject/timeline/form_task.html.twig
@@ -86,6 +86,13 @@
                </span>
             {% endif %}
 
+            {% if entry_i['date'] %}
+               <span class="planification badge bg-orange-lt">
+                  <i class="fa-regular fa-calendar me-1"></i>
+                  {{ entry_i['date']|formatted_datetime }}
+               </span>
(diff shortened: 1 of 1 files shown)


cedric-anne (commented):
Honestly, I still do not understand what this date is. Indeed, this is neither the creation date (that is displayed in task header), neither the planned date. So what is it?

stonebuzz:
Some use it this way

- A supervisor creates a task for a tech for a date, the tech schedules it according to his calendar (so not necessarily the same date)

- Creation of a task with this date (for info) without scheduling
- and certainly another use case

orthagh:
After irl discussion:
- I'm not in favor of adding this in view (already too much overloaded)
- there is a misunderstanding of the purpose of the field, rename it, or better see if can be merged with other date field.

pedroalves5770:
Is it possible to add this now by Enable CSS customization? 
It would be very helpful to be able to see task date in the display

PR #15660, opened 2023-09-29
[Entity] Permit mail_domain as a semicolum separated list of domains

Berthe01:
Updated getEntityIDByDomain function to be abale to retrive the first Entity wich includes "domain" in mail_domain attribute as a semicolum separated list of domains.

If multiple Entities found, return the first one of SQL query. Could eventually be modified to return -1 if needed to be more strict about domain unicity.

I think this update can address many use cases, because this was ofen requested on forums, but I don't know if this function behavior modification implies other functionnality than MailCollector Rules.

Hope this can help :)

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes

### DIFF
--- a/src/Entity.php
+++ b/src/Entity.php
@@ -2702,10 +2702,39 @@ class Entity extends CommonTreeDropdown
      * @since 0.84 (before in entitydata.class)
      *
      * @param $value
+     * Berthe01-Permit mail_domain as list formatted as 'dom1.com;dom2.com;dom3.org'
      **/
     public static function getEntityIDByDomain($value)
     {
-        return self::getEntityIDByField("mail_domain", $value);
+        global $DB;
+
+        $iterator = $DB->request([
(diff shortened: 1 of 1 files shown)


Berthe01:
Hi,

Any news about this PR ?

The only potential problem I see is about 'mail_domain' type which is VARCHAR(255) type in the entites table, and may be should be switched to LONGTEXT() type to handle much more domains than 2 or 3 depending on the length of domain name.

trasher (changes requested):
Storing comma separated values seems not a very good idea most of the time; and current implementation seems wrong on some cases at first view.
Anyway, this would require adding unit tests; and I'm almost sure permitting this would require changes elsewhere...

Also, features PR must target main branch, not a bugfixes one.

trasher:
no feedback from a while, closing.

PR #15695, opened 2023-10-03
Handle modified values from business rules during inventory

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !29764 #14609

### DIFF
--- a/src/Inventory/Asset/MainAsset.php
+++ b/src/Inventory/Asset/MainAsset.php
@@ -707,9 +707,6 @@ abstract class MainAsset extends InventoryAsset
             $this->setNew();
 
             //update val according to commonDBTM process (rule)
-            $lockeds = new \Lockedfield();
-            $locks = $lockeds->getLockedValues($this->item->getType(), $this->item->fields['id']);
-
             foreach ($this->item->fields as $onadd_key => $onadd_value) {
                 // handle after
                 // we need to manage states_id on both cases (add / update)
@@ -720,8 +717,8 @@ abstract class MainAsset extends InventoryAsset
 
                 // set value computed on add by GLPI (and rules) to $val
                 // to prevent data loss during updates
-                // only for known keys (from $val) and for keys without global lock
-                if (property_exists($val, $onadd_key) && !array_key_exists($onadd_key, $locks)) {
+                // only for known keys (from $val)
+                if (property_exists($val, $onadd_key)) {
                     $val->$onadd_key = $onadd_value;
                     //update known_list
                     $known_key = md5($onadd_key . $val->$onadd_key);
--- a/src/Inventory/Asset/MainAsset.php
+++ b/src/Inventory/Asset/MainAsset.php
@@ -705,6 +705,39 @@ abstract class MainAsset extends InventoryAsset
                 throw new \Exception('Unable to create item.');
             }
             $this->setNew();
+
+            //update val according to commonDBTM process (rule)
+            $lockeds = new \Lockedfield();
+            $locks = $lockeds->getLockedValues($this->item->getType(), $this->item->fields['id']);
+
+            foreach ($this->item->fields as $onadd_key => $onadd_value) {
+                // handle after
+                // we need to manage states_id on both cases (add / update)
+                // because states_id can be configured globally by inventory configuration
+                if ($onadd_key == 'states_id') {
+                    continue;
+                }
+
+                // set value computed on add by GLPI (and rules) to $val
+                // to prevent data loss during updates
+                // only for known keys (from $val) and for keys without global lock
+                if (property_exists($val, $onadd_key) && !array_key_exists($onadd_key, $locks)) {
+                    $val->$onadd_key = $onadd_value;
+                    //update known_list
+                    $known_key = md5($onadd_key . $val->$onadd_key);
+                    $this->known_links[$known_key] = $onadd_value;
+                    $this->raw_links[$known_key] = $onadd_value;
+                }
+            }
+        }
+
+        // special case for states_id (no need for entites_id because RuleAsset has no action for entities_id)
+        // when default states is set (> 0) from inventory configuration
+        // force it to be set on the item
+        if ($default_states_id && $default_states_id != '-1') {
+            $val->states_id = $default_states_id;
+        } else {
+            $val->states_id =  $this->item->fields['states_id'];
         }
 
         $val->id = $this->item->fields['id'];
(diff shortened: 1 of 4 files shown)


stonebuzz:
I don't know if unit tests are really necessary here, 

we test everywhere (for the inventory) the values on creation and update

trasher:
We're trying to address a specific issue here with business rules; so we have to add tests. I have actually no idea if that will work as expected, and on all possible cases

src/Inventory/Asset/MainAsset.php
```diff
@@ -702,6 +702,26 @@ public function rulepassed($items_id, $itemtype, $rules_id, $ports_id = 0)
             $this->setNew();
         }
 
+        //update val according to commonDBTM process (rule)
+        if ($itemtype != NetworkEquipment::getType() && $itemtype != Printer::getType()) {
+            foreach ($this->item->fields as $onadd_key => $onadd_value) {
+                if ($this->isNew() && $default_states_id != '-1' && $onadd_key == 'states_id') {
```
stonebuzz, line 708:
```suggestion
                if ($default_states_id != '-1' && $onadd_key == 'states_id') {
```

Both ```add``` and ```update``` must be ignored  ```states_id``` if needed

stonebuzz (changes requested):

src/Inventory/Asset/MainAsset.php
```diff
@@ -705,6 +705,39 @@ public function rulepassed($items_id, $itemtype, $rules_id, $ports_id = 0)
...
             }
             $this->setNew();
+
+            //update val according to commonDBTM process (rule)
+            $lockeds = new \Lockedfield();
+            $locks = $lockeds->getLockedValues($this->item->getType(), $this->item->fields['id']);
+
+            foreach ($this->item->fields as $onadd_key => $onadd_value) {
+                // handle after
+                // we need to manage states_id on both cases (add / after)
```
stonebuzz, line 715:
```suggestion
                // we need to manage states_id on both cases (add / update)
```

stonebuzz (changes requested):

src/Inventory/Asset/MainAsset.php
```diff
@@ -705,6 +705,39 @@ public function rulepassed($items_id, $itemtype, $rules_id, $ports_id = 0)
...
+        }
+
+        // special case for states_id (no need for entites_id because RuleAsset has no action for entities_id)
+        // when default states is set (> 0) from inventory configuration
+        // force it to be set on the item
+        if ($default_states_id && $default_states_id != '-1') {
+            $val->states_id = $default_states_id;
+        } else {
+            $val->states_id =  $this->item->fields['states_id'];
         }
```
stonebuzz, line 742:
```suggestion
                    $known_key = md5('states_id' . $val->states_id);
                    $this->known_links[$known_key] = $val->states_id;
                    $this->raw_links[$known_key] = $val->states_id;
```
Maybe we need to save value in ```raw_links``` and ```known_links``` 
Unit tests are OK with this modification

stonebuzz (changes requested):

src/Inventory/Asset/MainAsset.php
```diff
@@ -705,8 +705,43 @@ public function rulepassed($items_id, $itemtype, $rules_id, $ports_id = 0)
...
+        // when default states is set (> 0) from inventory configuration
+        // force it to be set on the item
+        if ($default_states_id && $default_states_id != '-1') {
+            $val->states_id = $default_states_id;
+        } else {
+            $val->states_id =  $this->item->fields['states_id'];
+        }
+                    $known_key = md5('states_id' . $val->states_id);
+                    $this->known_links[$known_key] = $val->states_id;
+                    $this->raw_links[$known_key] = $val->states_id;
```
stonebuzz, line 744:
```suggestion
        $known_key = md5('states_id' . $val->states_id);
        $this->known_links[$known_key] = $val->states_id;
        $this->raw_links[$known_key] = $val->states_id;
```

stonebuzz (changes requested):

src/Inventory/Asset/MainAsset.php
```diff
@@ -705,8 +705,43 @@ public function rulepassed($items_id, $itemtype, $rules_id, $ports_id = 0)
...
+        // when default states is set (> 0) from inventory configuration
+        // force it to be set on the item
+        if ($default_states_id && $default_states_id != '-1') {
+            $val->states_id = $default_states_id;
+        } else {
+            $val->states_id =  $this->item->fields['states_id'];
+        }
+        $known_key = md5('states_id' . $val->states_id);
+        $this->known_links[$known_key] = $val->states_id;
+        $this->raw_links[$known_key] = $val->states_id;
```
stonebuzz, line 744:
```suggestion
        $lockeds = new \Lockedfield();
        $locks = $lockeds->getLockedValues($this->item->getType(), $this->item->fields['id']);

        if (!array_key_exists('states_id', $locks)) {
            if ($default_states_id && $default_states_id != '-1') {
                $val->states_id = $default_states_id;
            } else {
                $val->states_id =  $this->item->fields['states_id'];
            }
            $known_key = md5('states_id' . $val->states_id);
            $this->known_links[$known_key] = $val->states_id;
            $this->raw_links[$known_key] = $val->states_id;
        }
```

We need to check if the ```states_id``` is locked

stonebuzz (changes requested):

src/Inventory/Asset/MainAsset.php
trasher: OK. I have no idea if values from business rules should be blocked from LockedFields. Initial goal of this stuff is to prevent values to be modified from inventory data.
If we prevent business rules to change status here, we'lle have inconsistencies because that will work on "is_dynamic=" asset, and not on "is_dynamic=1" ones, even if "is inventory" is not set in rule.

trasher: Also, what is the value of "is_dynamic" fields, the rule will apply if modified by hand.

To me, even we must block changes from locked fileds for all rules results from all changes methods, or we should allow it everywhere (I do not know if I'm clear)

trasher: After tests and discussions, all seems OK as is.

cedric-anne (approved):

stonebuzz (approved):

PR #15715, opened 2023-10-04
Allow pending reason edition from new fup/task form

AdrienClairembault:
Similarly to #15696, this was a feature from the 9.5 backport that was not integrated properly into GLPI 10.
I think @cedric-anne also suggested that we fix that a while ago but it didn't go anywhere at the time.

When adding a new followup to a ticket, it is possible to remove the `pending` status but we can't interact with the pending reasons: 

image

After changes:
image

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !29772

### DIFF
--- a/src/CommonITILTask.php
+++ b/src/CommonITILTask.php
@@ -634,6 +634,8 @@ abstract class CommonITILTask extends CommonDBTM implements CalDAVCompatibleItem
             NotificationEvent::raiseEvent('add_task', $this->input["_job"], $options);
         }
 
+        PendingReason_Item::handlePendingReasonUpdateFromNewTimelineItem($this);
+
        // Add log entry in the ITIL object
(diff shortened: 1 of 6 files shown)


orthagh (commented):
Just to be sure, this doesn't interfere with the auto reopening of the ticket by the requester ?

AdrienClairembault:
> Just to be sure, this doesn't interfere with the auto reopening of the ticket by the requester ?

Seems good:

image

I think the relevant code is here, which keep disable pending action when the "add_reopen" flag is active: 

image

cedric-anne (approved):

orthagh (approved):

PR #15716, opened 2023-10-04
fix: imap folders globalname

flegastelois:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | I hope
| Fixed tickets | N/A

Mailbox hierarchy delimiter can be "/" or "." (it depend on server NAMESPACE configuration).

So I propose to use "globalName" (which already contains the full path of folder constructed by Laminas lib with the correct server delimiter), instead of using "localName" and construct the path on GLPI side.

image

This only affect "preselection" feature in collector UI configuration form.

### DIFF
--- a/src/MailCollector.php
+++ b/src/MailCollector.php
@@ -360,14 +360,7 @@ class MailCollector extends CommonDBTM
                var input_id = li.data('input-id');
                var folder   = li.children('.folder-name').html();
 
-               var _label = '';
-               var _parents = li.parents('li').children('.folder-name');
-               for (i = _parents.length -1 ; i >= 0; i--) {
-                  _label += $(_parents[i]).html() + '/';
-               }
-               _label += folder;
-
-               $('#'+input_id).val(_label);
+               $('#'+input_id).val(folder);
 
                var modalEl = $('#'+input_id+'_modal')[0];
                var modal = bootstrap.Modal.getInstance(modalEl);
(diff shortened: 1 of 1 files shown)


trasher (approved):

cedric-anne (commented):
I think it will makes reading more complicated for sub-folder aficionados.

Before:
image

After:
image

cedric-anne:
> I think it will makes reading more complicated for sub-folder aficionados.
> 
> Before: image
> 
> After: image

You can probably find a way to remove the name of the parent folder from the name of the current folder. It will permit to keep both the correct separator and short folders names.

flegastelois:
> > I think it will makes reading more complicated for sub-folder aficionados.
> > Before: image
> > After: image
> 
> You can probably find a way to remove the name of the parent folder from the name of the current folder. It will permit to keep both the correct separator and short folders names.

Something like the last commit?

src/MailCollector.php
```diff
@@ -427,10 +420,11 @@ public function displayFoldersList($input_id = "")
      */
     private function displayFolder($folder, $input_id)
     {
-        $fname = mb_convert_encoding($folder->getLocalName(), "UTF-8", "UTF7-IMAP");
+        $fglobalname = mb_convert_encoding($folder->getGlobalName(), "UTF-8", "UTF7-IMAP");
+        $flocalname  = mb_convert_encoding($folder->getLocalName(), "UTF-8", "UTF7-IMAP");
```
cedric-anne, line 424: To prevent issues with quotes in folder names.
```suggestion
        $fglobalname = htmlspecialchars(mb_convert_encoding($folder->getGlobalName(), "UTF-8", "UTF7-IMAP"), ENT_QUOTES);
        $flocalname  = htmlspecialchars(mb_convert_encoding($folder->getLocalName(), "UTF-8", "UTF7-IMAP"), ENT_QUOTES);
```

cedric-anne (approved):

PR #15725, opened 2023-10-06
fix: disable submit via enter key

Rom1-B:
For example: On a ticket, if you add a text input via the Fields plugin. When you modify this field and press enter, the ticket is deleted, because the 1st submit button on the form is delete.

_I don't think this case is specific to the plugin._

This PR disables submit via enter key from an input field (itilobject only).

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

### DIFF
--- a/templates/components/itilobject/footer.html.twig
+++ b/templates/components/itilobject/footer.html.twig
@@ -128,36 +128,7 @@
             </button>
          {% else %}
 
-            <div class="btn-group" role="group" id="right-actions">
-               {% if item.canDeleteItem() %}
-                  {% if item.isDeleted() %}
-                     <button class="btn btn-outline-secondary" type="submit" name="restore" form="itil-form"
-                           title="{{ _x('button', 'Restore') }}">
-                        <i class="fas fa-trash-restore-alt"></i>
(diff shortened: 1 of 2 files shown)


orthagh (commented):
I'm not sure I would have taken this road.
Alternative solution:

- reverse in html and use flex to reverse also the display
- make an hidden submit button in first position (acting like the save one)

cedric-anne:
> * reverse in html and use flex to reverse also the display

This is what we do in other forms, and it is the best solution IMHO.

Rom1-B:
Bootstrap doesn't handle reverse with btn-group, so border-radius is wrong.

image

I made 2 commits for the 2 proposed roads

orthagh:
One comment because I didn't check when suggestion the solution, is the second works on Enter key ?

Rom1-B:
> One comment because I didn't check when suggestion the solution, is the second works on Enter key ?

Yes, but the tests fail. So there's still work to be done on the 2 solutions. Which one should I keep working on?

orthagh (approved):

trasher (approved):

PR #15757, opened 2023-10-11
fix: Adds a "disabled" option to the dashboard preferences for tickets

ccailly:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 

Restores the ability for users to disable the mini dashboard at the top of the ticket page. Previously, users could leave the value empty to hide the mini dashboard. However, PR #15382 changed this mechanism, using the empty value to default to the general configuration. This PR introduces a new 'Disabled' option for the 'Default for tickets (mini dashboard)' setting in user preferences, reinstating the ability for users to hide the mini dashboard if desired.

image

### DIFF
--- a/src/Dashboard/Grid.php
+++ b/src/Dashboard/Grid.php
@@ -1493,7 +1493,7 @@ HTML;
     }
 
 
-    public static function dropdownDashboard(string $name = "", array $params = [], bool $disabledOption = false): string
+    public static function dropdownDashboard(string $name = "", array $params = [], bool $disabled_option = false): string
     {
         $to_show = Dashboard::getAll(false, true, $params['context'] ?? 'core');
         $can_view_all = $params['can_view_all'] ?? false;
@@ -1505,7 +1505,7 @@ HTML;
             }
         }
 
-        if ($disabledOption) {
+        if ($disabled_option) {
             $options_dashboards = ['disabled' => __('Disabled')] + $options_dashboards;
         }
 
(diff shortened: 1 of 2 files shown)


AdrienClairembault:
@froberts0n, could you confirm that it fixes your issues relative to changes done in https://github.com/glpi-project/glpi/pull/15382 by adding back the possibility of disabling the tickets dashboard ?

cedric-anne (approved):

AdrienClairembault:
Small non blocking note: `$disabledOption` could be named `$disabled_option` or `$add_disabled_option` instead.

Its not a hard rule but GLPI historically use snake case for variable names and we try to keep in consistent.

orthagh (approved):

src/Dashboard/Grid.php
```diff
@@ -1488,7 +1493,7 @@ public static function getDefaultDashboardForMenu(string $menu = "", bool $stric
     }
 
 
-    public static function dropdownDashboard(string $name = "", array $params = []): string
+    public static function dropdownDashboard(string $name = "", array $params = [], bool $disabledOption = false): string
```
trasher, line 1496:
```suggestion
    public static function dropdownDashboard(string $name = "", array $params = [], bool $disabled_option = false): string
```

src/Dashboard/Grid.php
```diff
@@ -1500,6 +1505,10 @@ public static function dropdownDashboard(string $name = "", array $params = []):
             }
         }
 
+        if ($disabledOption) {
```
trasher, line 1508:
```suggestion
        if ($disabled_option) {
```

trasher (commented):
Agree with Adrien's comment about variable name; OK otherwise

trasher (approved):

PR #15771, opened 2023-10-12
Fix invalid PHPdoc

AdrienClairembault:
Avoid seeing these kind of warnings everywhere in my code because of invalid PHPdoc ;)

image

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |

### DIFF
--- a/src/Html.php
+++ b/src/Html.php
@@ -5342,7 +5342,7 @@ HTML;
      * @param array   $options Array of HTML attributes
      * @param boolean $minify  Try to load minified file (defaults to true)
      *
-     * @return string Script tags
+     * @return string
      **/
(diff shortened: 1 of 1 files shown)


src/Html.php
```diff
@@ -5342,7 +5342,7 @@ public static function scriptBlock($script)
      * @param array   $options Array of HTML attributes
      * @param boolean $minify  Try to load minified file (defaults to true)
      *
-     * @return String of script tags
+     * @return string Script tags
```
trasher, line 5345:
```suggestion
     * @return string
```

trasher (approved):

PR #15782, opened 2023-10-13
keep filename when updating document

btry:
When a document is being uploaded but already exists, the existing document is updated, but the prefix is not passed to Document::update. If the file original file contains a prefix, it is not removed and the filename is updated with the full filename (with prefix).

This occurs in Formcreator when the plugin uploads a file from the form. When the plugin creates the a ticket the plugin simulates an upload. In this case, the hash of the newly uploaded file matches the already uploaded one, and the filename contains a prefix which must be removed.

Tested 

- upload of a document from front/document/.form.php (creating item and updating item)
- upload of a file when creating ticket
- upload of a pasted image when creating a ticket
- upload of a file when answering a form (from Formcreator), generating a ticket 
- upload of a pasted image when answering a form (from Formcreator), generating a ticket 

| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | internal ref 29645


cedric-anne (approved):
As this is a reuse of an identical piece of code existing a few lines above, it is OK for me.

However, it seems weird to me to update the initial document name, but it is already the case with previous code. The main difference is that now, for the tested case, document name is updated to `bar.txt` instead of `65292dc32d6a87.22222222bar.txt`.

orthagh (approved):

trasher (approved):

PR #15789, opened 2023-10-16
Add change followup

xacobofg:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #number

When responding to a notification of a change add the response as a followup

### DIFF
--- a/src/MailCollector.php
+++ b/src/MailCollector.php
@@ -650,7 +650,15 @@ class MailCollector extends CommonDBTM
                             $tkt['_users_id_requester'] = $rejected[$head['message_id']]['users_id'];
                             $tkt['entities_id']         = $entity;
 
-                            if (!isset($tkt['tickets_id'])) {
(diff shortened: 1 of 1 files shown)


ftoledo:
hi @xacobofg Is it possible that you update the PR so that it complies with lint's checks?

cedric-anne (commented):
Sorry for the late reply.

@xacobofg Do you think it is possible to replace hardcoded references to `Ticket`/`tickets_id`/... by the usage of the items from the `$CFG_GLPI['itil_types']` list that has been uintroduced in GLPI 11.0 ?

IMHO, it could prevent having to duplicate the code as you do and can even permit to handle Problem followups.

orthagh:
Stale PR

PR #15800, opened 2023-10-17
Search engine: disable sort on default assets requests

AdrienClairembault:
Significant performances gain for huge databases by removing the `ORDER BY` clause on the default assets searches.

The functional impact is very low as, most often than not, users go to assets search pages with the goal of researching with some specifics criteria and don't really care about the default results.

Event then, users can still explicitly request a sort by clicking a column or re-enable the default sort by using it in their default saved searches.

On my benchmark with ~300k total assets, the default search on the "All assets" page go down from 15s to 0.2s with these changes.

Same goes for individual itemtypes, with some requests that previously took between 1s and 3s going down to 0.001s.

This is very impactful as these "default searches" will be executed each time you go on an assets page with a clean session.

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !29967


orthagh (approved):

AdrienClairembault:
Will move changes to main instead as discussed.

AdrienClairembault:
Solution in latest commit should fix the mariadb issues.

This is done by optimizing the request even more by adding a LIMIT to each subquery.
This can only be done on the first page of an empty search as it won't work with pagination but it should be enough here since we are fixing the "default search" which will obviously be displayed using its first page.

AdrienClairembault:
Regarding the PR in itself, I'm still planning to move it to `main` and implement the changes suggested by @orthagh but didn't get the time do it yet.

AdrienClairembault:
Rebased to main, should be good for review and merge.

The following changes were discussed with @orthagh last month and will be implemented in another PR (to keep review short and simples).

> If the user doesn't explicitly requested a sort key and the itemtype doesn't define a default search:
> * Do not sort the results at all (even if some search criteria are defined)
> * Display the ID column as the first column and "fake" its use as a search key

orthagh (approved):

trasher (approved):
LGTM. Please rebase to get CI running

trasher:
Tests are failing

cedric-anne (approved):

PR #15805, opened 2023-10-17
debug: allow to copy sql queries to clipboard

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

image


cconard96 (approved):

js/modules/Debug/Debug.js
```diff
@@ -536,13 +536,22 @@ window.GLPI.Debug = new class Debug {
                 (filtered_request_id !== undefined && filtered_request_id !== request_id)) {
                 return;
             }
+            const clipboard_title = __("Copy query to clipboard");
```
trasher, line 539: Since this is debug, this should not be translated (nothing else seems translated in this file).
```suggestion
            const clipboard_title = "Copy query to clipboard";
```

js/modules/Debug/Debug.js
```diff
@@ -553,6 +562,19 @@ window.GLPI.Debug = new class Debug {
         });
         sql_table_body.append(rows_to_append);
 
+        $(".copy-code").on('click', function () {
+            // copy content of code block in clipboard
+            const code = $(this).parent().find('code');
+            copyTextToClipboard(code.text());
```
AdrienClairembault, line 568: Whats is the logic behind using a custom function to write to the clipboard instead of `navigator.clipboard.writeText()` ?

From my point of view, it looks like `copyTextToClipboard` is some old code that was used a while ago before `navigator.clipboard.writeText` was implemented.
It call deprecated methods such as `document.execCommand`, maybe we are better off not using it in new code ?

https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Interact_with_the_clipboard
https://developer.mozilla.org/en-US/docs/Web/API/Document/execCommand

AdrienClairembault (changes requested):

orthagh:
Because navigator api doesn't work on non https env. So for convenience I
use something that already works. If it bothers you, we can lately
reworking the dedicated functions

AdrienClairembault:
I see, Ok then

cedric-anne (approved):

trasher (approved):

PR #15811, opened 2023-10-18
fix(inventory): restoring fields

Rom1-B:
Restores fields: `hostid`, `company`, `owner` that were brought up by FusionInventory

| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | https://github.com/glpi-project/glpi-inventory-plugin/issues/429


stonebuzz (changes requested):
In both case ```json``` and ```xml```

image

https://github.com/glpi-project/inventory_format should be adapt

See ```inventory.schema.json```  (```hardware``` key)

stonebuzz:
> In both case `json` and `xml`
> 
> image
> 
> https://github.com/glpi-project/inventory_format should be adapt
> 
> See `inventory.schema.json` (`hardware` key)

Disregard this follow-up,

my inventory files were poorly done

trasher (approved):
I'm not absolutely sure this is really usefull; but I'm not really against neither.

cedric-anne (approved):

cedric-anne:
Maybe a test could be added, to ensure that the value is correctly retrieved from the inventory file.

src/Inventory/Asset/OperatingSystem.php
```diff
@@ -76,6 +76,14 @@ public function prepare(): array
...
                 $val->license_number = $this->extra_data['hardware']->winprodkey;
             }
+
+            if (property_exists($this->extra_data['hardware'], 'wincompany')) {
+                $val->company = $this->extra_data['hardware']->wincompany;
+            }
+
+            if (property_exists($this->extra_data['hardware'], 'winowner')) {
+                $val->owner = $this->extra_data['hardware']->winowner;
+            }
```
g-bougard, line 86: What about `hostid` ? I don't see it imported from the inventory.

trasher: Indeed, seems it's not handled... Maybe tests are indeed needed

stonebuzz: > What about `hostid` ? I don't see it imported from the inventory.

Because inventory ```key``` and database ```column name``` are same, so we d'ont need to add ```mapping```

Rom1-B: HostID does not require conversion, unlike the other 2. The 3 fields are tested in my latest commit.

trasher (approved):
Seems OK, waiting for tests results

stonebuzz (approved):

PR #15817, opened 2023-10-18
fix id attribute in macro field files

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

Since #14852

mailing_settings seems to be fixed


orthagh:
move to draft, I need to think to some edge case scenarios

PR #15831, opened 2023-10-19
scroll to anchor after tab loading

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

Anchors are present from a long time in the timeline and worked if you call the same url when the page is open.
But as timeline (like the rest of the ticket) is loaded by ajax, on opening a new url, anchor scrolling didn't worked.

I set the behavior in the global code because it also fix the KB (also have anchors)
Atm, I don't see where it can cause issue (non wanted scroll)

### DIFF
--- a/src/Ajax.php
+++ b/src/Ajax.php
@@ -372,7 +372,14 @@ JAVASCRIPT;
                      tab_key: href_url_params.get('_glpi_tab'),
                      withtemplate: " . (int)($_GET['withtemplate'] ?? 0) . "
                   }
-               );
+               ).done(function() {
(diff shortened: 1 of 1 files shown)


trasher (approved):

cedric-anne (approved):

AdrienClairembault (approved):
Code look good, nitpicking a bit but var name could be improved (its not logical to have something named `initial_XXXXX` which is then modified).

AdrienClairembault (approved):

PR #15836, opened 2023-10-19
feat(Approval): use anchor to point directly to the form in the timeline

stonebuzz:
To resume the following PR https://github.com/glpi-project/glpi/pull/14428
which continued the work of  https://github.com/glpi-project/glpi/pull/14256
and since https://github.com/glpi-project/glpi/pull/15831

The idea would be (via HTML anchors) to position the user directly on the validation form (which concerns him) in the time line.

you can use this URL (must be adapted to your GLPI url) to test

jsut  replace ```#ID_TICKET#``` and ```#ID_VALIDATION#```

```
/index.php?redirect=ticket_#ID_TICKET#_Ticket%24main%23TicketValidation_#ID_VALIDATION#
```

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 

Need : https://github.com/glpi-project/glpi/pull/15853 (to be merged into ```main```)

### DIFF
--- a/front/css.php
+++ b/front/css.php
@@ -40,6 +40,7 @@ if (!defined('GLPI_ROOT')) {
 }
 
 use Glpi\Application\ErrorHandler;
+use Glpi\UI\ThemeManager;
 
 $_GET["donotcheckversion"]   = true;
 $dont_check_maintenance_mode = true;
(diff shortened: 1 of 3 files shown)


src/Toolbox.php
```diff
@@ -1645,7 +1645,9 @@ public static function manageRedirect($where)
                     Html::redirect($redirect_to);
                 }
 
-                $data = explode("_", $where);
+                // explode with limit 3 to preserve the last part of the url
+                // /index.php?redirect=ticket_2_Ticket$main#TicketValidation_1 (preserve anchor)
+                $data = explode("_", $where, 3);
```
trasher, line 1650: Are we sure 3 will always be enough? What about exploding on `#` then append it back when/where needed?

stonebuzz: Yes, explode ```_``` is only used to find the ticket ID, 

and originally we didn't go any further than index ```2``` (so ```3``` part)

orthagh (approved):
OK, could you rebase to get the last changes from previous pr

Issue #15801, opened 2023-10-17
The translation in the ticket categories does not work


FreeMacs:
### Code of Conduct

- [ ] I agree to follow this project's Code of Conduct

### Is there an existing issue for this?

- [ ] I have searched the existing issues

### Version

10.0.10

### Bug description

I have created a category under Setup > Dropdowns > ITIL categories with a description of the category in the comment field. There was also an English translation added under Translations but it doesn't work. if I set GLPI to English the text in the comments stays on the German version.

Here are a few example images

![image](https://github.com/glpi-project/glpi/assets/115316656/026141fb-5d04-4e21-b0d8-77c921539acd)

English translation created

![image](https://github.com/glpi-project/glpi/assets/115316656/3acc3fc6-fa87-42ef-b70b-7e2b1084c661)

Glpi changed to English

![image](https://github.com/glpi-project/glpi/assets/115316656/cdcce283-16b5-422e-a4ff-0f61e3eaf542)

Comments are still in German

As described in ticket #15180, the subcategories also disappear in the dropdown lists where a translation has been inserted.

![image](https://github.com/glpi-project/glpi/assets/115316656/1a80a89a-1c88-4cab-92f6-73b3349f883c)
![image](https://github.com/glpi-project/glpi/assets/115316656/0bc130e4-1193-4378-8c24-1d9f0241e16e)

To do this, simply set GLPI to the language in which the translation was created.

cconard96:
I cannot recreate the issue. Do you have the translation of dropdowns enabled in Setup > General on the General setup tab?

cconard96:
I confirm. It seems that if I add a name translation the complete name is generated properly along with it. If I then add a translation for the comments, the complete name translation in the DB is suddenly empty.

PR #15837, opened 2023-10-19
fix dropdown complete name translation

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #15801

Steps I found to recreate the issue:
1. While using en_GB language, add a name translation for an ITIL Category for another language.
2. Load the other language to verify the complete name translation was generated.
3. Switch back to en_GB and add a comments translation for the same category.
4. Switch back to the other language and see that the complete name translation is missing. The value in the DB for the translation was replaced with an empty string (the empty string is stored in the DB).

### DIFF
--- a/src/DropdownTranslation.php
+++ b/src/DropdownTranslation.php
@@ -677,10 +677,7 @@ JAVASCRIPT
        //If dropdown translation is globally off, or if this itemtype cannot be translated,
        //then original value should be returned
         $item = new $itemtype();
-        if (
-            !$ID
-            || !Session::haveTranslations($itemtype, $field)
-        ) {
+        if (!self::canBeTranslated($item)) {
             return $value;
         }
        //ID > 0 : dropdown item might be translated !
--- a/tests/functional/DropdownTranslation.php
+++ b/tests/functional/DropdownTranslation.php
@@ -1,0 +1,105 @@
+<?php
+
+/**
+ * ---------------------------------------------------------------------
+ *
+ * GLPI - Gestionnaire Libre de Parc Informatique
+ *
+ * http://glpi-project.org
+ *
+ * @copyright 2015-2023 Teclib' and contributors.
+ * @copyright 2003-2014 by the INDEPNET Development Team.
+ * @licence   https://www.gnu.org/licenses/gpl-3.0.html
+ *
+ * ---------------------------------------------------------------------
+ *
+ * LICENSE
+ *
+ * This file is part of GLPI.
+ *
+ * This program is free software: you can redistribute it and/or modify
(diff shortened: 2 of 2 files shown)


trasher:
This seems to cause side effects on existing tests :/

cedric-anne (commented):
It seems that the `DropdownTranslation::getTranslatedValue()` has a bad design and does not fallback to the "non translated" value if there is not at least one item that has this field translated. This can probably be changed, as you proposed, but it would have an impact on performances and on method results (the default value that is passed in most of this method call will not be returned as expected, even if in most cases it is probably the same as the non translated value finally returned).

On 10.0/bugfixes branch, we should try to limit potential side effects, so I propose to just fix the way the `getTranslatedValue()` method is call. It would be something like this (not tested):
```diff
diff --git a/src/DropdownTranslation.php b/src/DropdownTranslation.php
index 14f50bdbff..e6750acc80 100644
--- a/src/DropdownTranslation.php
+++ b/src/DropdownTranslation.php
@@ -279,8 +279,6 @@ class DropdownTranslation extends CommonDBChild
     {
         /** @var \DBmysql $DB */
         global $DB;
-       // Force completename translated : used for the first translation
-        $_SESSION['glpi_dropdowntranslations'][$input['itemtype']]['completename'] = 'completename';
 
        //If there's already a completename for this language, get it's ID, otherwise 0
         $completenames_id = self::getTranslationID(
@@ -304,11 +302,14 @@ class DropdownTranslation extends CommonDBChild
            //Get ancestors as an array
 
             if ($item->fields[$foreignKey] != 0) {
+                $parent = new $input['itemtype']();
+                $parent->getFromDB($item->fields[$foreignKey]);
                 $completename = self::getTranslatedValue(
-                    $item->fields[$foreignKey],
+                    $parent->getID(),
                     $input['itemtype'],
                     'completename',
-                    $input['language']
+                    $input['language'],
+                    $parent->fields['completename']
                 );
             }
 
@@ -319,7 +320,8 @@ class DropdownTranslation extends CommonDBChild
                 $item->getID(),
                 $input['itemtype'],
                 'name',
-                $input['language']
+                $input['language'],
+                $item->fields['name']
             );
 
            //Add or update completename for this language
```

On main branch, we should try to find a way to simplify this, in order to be able to remove usage of this `$_SESSION['glpi_dropdowntranslations']` entry.

cconard96:
This doesn't resolve the issue. Instead of blanking the 'completename' translation in the DB, now it completely removes it which causes the English/default completename to be shown.

PR #15839, opened 2023-10-20
Explicitely define valid elements for tinymce

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #15636 

According to documentation:
«Use * to include all elements and all attributes. This can be very useful when used with the invalid_elements option.»

Maybe that fix related issue.

### DIFF
--- a/src/Html.php
+++ b/src/Html.php
@@ -3926,6 +3926,7 @@ JS;
 
                // Content settings
                entity_encoding: 'raw',
+               valid_elements: '*',
                invalid_elements: '{$invalid_elements}',
                readonly: {$readonlyjs},
                relative_urls: false,


orthagh (approved):

cedric-anne (changes requested):
It does not fix the issue. Worse, it removes the ability to add styles on tables.
To add styles on a table:
 - create a table in the editor
 - right click on it
 - click "table properties"
 - go in "advanced" tab
 - select a background color.

Without the patch, the background color is defined and visible. With proposed patch, it is not applied.

PR #15857, opened 2023-10-25
fix(search): deprecated htmlspecialchars

Rom1-B:
Fix warning message on ticket search

`Deprecated: htmlspecialchars(): Passing null to parameter (Sstring) of type string is deprecated in /mnt/data/glpi/src/Planning.php on line 263`

6538b5b65d4dd8 53298316image_paste

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !30152

### DIFF
--- a/src/Planning.php
+++ b/src/Planning.php
@@ -262,9 +262,12 @@ class Planning extends CommonGLPI
      */
     public static function getStatusIcon($status): string
     {
+        $label = htmlspecialchars(Planning::getState($status), ENT_QUOTES);
+        if (empty($label)) {
+            return '';
+        }
         $class = Planning::getStatusClass($status);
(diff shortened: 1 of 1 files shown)


trasher:
I guess entire icon should return "nothing" rather than html without information

trasher:
(I talk about `getStatusIcon()`)

trasher:
And... Is that expected we receive an unknown status in first place?

trasher (approved):

Rom1-B:
For rows without tasks, `status = ''`, so the value is legitimate.

cedric-anne (approved):

PR #15861, opened 2023-10-25
Feat/search ux

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

image

### DIFF
--- a/js/modules/Search/Table.js
+++ b/js/modules/Search/Table.js
@@ -278,6 +278,12 @@ window.GLPI.Search.Table = class Table extends GenericView {
             this.refreshResults();
         });
         $(ajax_container).on('click', '.sort-reset', (e) => {
+             // force removal of tooltip
+             const tooltip = bootstrap.Tooltip.getInstance($(e.currentTarget)[0]);
+             if (tooltip !== null) {
+                 tooltip.dispose();
+             }
+
             e.preventDefault();
             this.resetSortState();
             this.refreshResults();
--- a/js/modules/Search/Table.js
+++ b/js/modules/Search/Table.js
@@ -278,11 +278,11 @@ window.GLPI.Search.Table = class Table extends GenericView {
             this.refreshResults();
         });
         $(ajax_container).on('click', '.sort-reset', (e) => {
-             // force removal of tooltip
-             const tooltip = bootstrap.Tooltip.getInstance($(e.currentTarget)[0]);
-             if (tooltip !== null) {
-                 tooltip.dispose();
-             }
+            // force removal of tooltip
+            const tooltip = bootstrap.Tooltip.getInstance($(e.currentTarget)[0]);
+            if (tooltip !== null) {
+                tooltip.dispose();
+            }
 
             e.preventDefault();
             this.resetSortState();
--- a/js/modules/Search/Table.js
+++ b/js/modules/Search/Table.js
@@ -31,6 +31,8 @@
  * ---------------------------------------------------------------------
  */
 
+/* global bootstrap */
+
 import GenericView from './GenericView.js';
 
 // Explicitly bind to window so Jest tests work properly
@@ -38,9 +40,14 @@ window.GLPI = window.GLPI || {};
 window.GLPI.Search = window.GLPI.Search || {};
 
 window.GLPI.Search.Table = class Table extends GenericView {
+    // track sort/order changes globally
+    sort_state = {
+        sort: [],
+        order: []
+    };
 
     constructor(result_view_element_id, push_history = true, forced_params = {}) {
-        const element_id = $('#'+result_view_element_id).find('table.search-results').attr('id');
+        const element_id = $('#'+result_view_element_id).parent().find('table.search-results').attr('id');
         super(element_id);
 
         this.push_history = push_history;
@@ -131,6 +138,7 @@ window.GLPI.Search.Table = class Table extends GenericView {
             target_column.attr('data-sort-num', undefined);
         }
 
+        this.setSortStateFromColumns();
         this.refreshResults();
     }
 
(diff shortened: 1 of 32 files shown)


cconard96:
Here is a commit that fixes some JS conflicts after the search results are refreshed via AJAX and also fixes an issue where the criteria disappears after a search.
https://github.com/glpi-project/glpi/commit/69ab7d5b877ee9b27be9993adac549948688d9bd

cconard96:
I made a few fixes on a fork of this PR:
https://github.com/cconard96/glpi/commits/feat/search_ux

The only remaining issue I noticed is when refreshing the results, the `ajax/search.php` endpoint occasionally responds with a JS script tag that would trigger a redirect to the saved search I selected last instead of the results data. This is considered an unexpected response by the Fluid Search JS, so it triggers a full page reload. After the reload, the previously selected saved search is showing in the button label even though I have custom criteria selected.

orthagh:
TODO:

- [ ] savedsearches split
- [ ] responsive (hide some labels)

AdrienClairembault (changes requested):
1) The thrash bin toggle is lacking some padding and probably the same rounded borders that are used by the display type toggle:

image

2) You can also see in the previous image that the map toggle is missing when inside the thrash bin.

3) Didn't you also wanted to integrate the kanban here ? Maybe delayed to another PR ?

image

4) The "filter" dropdown is not highlighted on my side compared to your screen shot.
Maybe it should have some hint of the primary color as it is an important action (more important than sorting IMO) ?

image

5) When using the map mode, the filter indicate that items are restricted on latitude and longitude but we can't see the associated criteria.

image

Maybe it will confuse some users ? Maybe the lat and long criteria should be shown as readonly ?

6) After using massive actions the items remained checked when the page is reloaded but the "actions" button is not displayed.

image

----

Otherwise I think it overall looks great, very clean 👍

orthagh:
- 1, 4 & 6 fixed
- 2 cannot replicate on my side
- 3 later

orthagh:
2 fixed

AdrienClairembault (approved):

src/Search.php
```diff
@@ -378,6 +378,22 @@ public static function showGenericSearch($itemtype, array $params)
...
+     * Print generic sort form
+     *
+     * Params need to parsed before using Search::manageParams function
+     *
+     * @param class-string<CommonDBTM> $itemtype  Type to display the form
+     * @param array  $params    Array of parameters may include sort, order
+     *
+     * @return void
+     **/
+    public static function showGenericSort($itemtype, array $params)
```
cconard96, line 392: Not sure these two methods need added here. `Search` only exists still for backwards-compatibility.

src/Search.php
```diff
@@ -461,6 +477,24 @@ public static function displaySearchoptionValue($request = [])
     }
 
 
+    /**
+     * Display a criteria field set, this function should be called by ajax/search.php
+     *
+     * @since 9.4
```
cconard96, line 483: Wrong version

cconard96 (changes requested):

src/Search.php
cconard96: I think the warnings on `Glpi\Search\SearchEngine` and `Search` can be removed/ignored now too.

trasher:
Search seems OK, I had an issue with sorting. I've selected:
image

And once I've applied sort, this is shown:
image


I can see following in logs:
```
[2023-11-29 07:51:49] glpiphplog.WARNING:   *** PHP Warning (2): Undefined array key "" in /var/www/webapps/glpi/src/Search/Output/HTMLSearchOutput.php at line 167
  Backtrace :
  src/Search.php:350                                 Glpi\Search\Output\HTMLSearchOutput::displayData()
  ajax/search.php:90                                 Search::displayData()
  
[2023-11-29 07:51:49] glpiphplog.WARNING:   *** PHP Warning (2): Trying to access array offset on value of type null in /var/www/webapps/glpi/src/Search/Output/HTMLSearchOutput.php at line 167
  Backtrace :
  src/Search.php:350                                 Glpi\Search\Output\HTMLSearchOutput::displayData()
  ajax/search.php:90                                 Search::displayData()
```

trasher:
I guess this happens when I sort on a non displayed column

AdrienClairembault:
I guess it doesn't make sense to be able to sort multiple time on the same column: 

image

js/modules/Search/Table.js
```diff
@@ -151,6 +153,35 @@ window.GLPI.Search.Table = class Table extends GenericView {
...
+        $(sort_container).find('select[name^="sort"]').each(function() {
+            sorts.push($(this).val());
+        });
+
+        // Remove data-sort-num from all headers and change all data-sort-order to nosort to reset the sort state
+        this.getElement().find('thead th').each((i, c) => {
+            const col = $(c);
+            col.attr('data-sort-num', null);
+            col.attr('data-sort-order', 'nosort');
+        });
```
AdrienClairembault, line 172: FYI jQuery's `attr` method work on the whole collection, there is no need to iterate on each items.

Something like that would work for example:

```suggestion
        this.getElement().find('thead th').attr({
            'data-sort-num': null,
            'data-sort-order': 'nosort',
        });
```

AdrienClairembault:
> > I guess it doesn't make sense to be able to sort multiple time on the same column:
> 
> Right, but there's more in fact, I just think about it : but as we defer the sort from the selects to th, any options not present in table header will not be sorted.
> 
> @cconard96 I guess we need to track sorts/orders in a class property and both sources of sorting should alter them rather than forwarding to headers.

A temporary fix would be to allow only the displayed headers to be selectable in the sort list.

orthagh:
Sort/order issues should be fixed.
More, I added "used" management, adding a new sort row (from card panel), should now take into account previous sorted columns and not presented them

cconard96 (approved):

trasher (approved):

cedric-anne (approved):

PR #15876, opened 2023-10-27
Fix getReplyTo when not allow response

tsmr:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #number

When you not allow response on glpi notifications, the sender is replaced by getNoReplyEmailSender() but not the address used for response


cedric-anne (approved):

trasher:
Maybe it is possible to add a test?

orthagh (approved):

PR #15880, opened 2023-10-27
Add Waiting On Send Ticket Validation

tsmr:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #number

Add waiting icon on ticket validation from

image

### DIFF
--- a/src/CommonITILValidation.php
+++ b/src/CommonITILValidation.php
@@ -311,9 +311,11 @@ abstract class CommonITILValidation extends CommonDBChild
         // Handle uploaded documents
         $this->input = $this->addFiles($this->input);
 
-        if (isset($this->input['itemtype'])
+        if (
+            isset($this->input['itemtype'])
             && $this->input['itemtype'] == "Ticket"
-            && $this->input['pending'] == 1) {
+            && $this->input['pending'] == 1
+        ) {
             $object     = new Ticket();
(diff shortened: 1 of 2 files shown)


src/CommonITILValidation.php
```diff
@@ -311,6 +311,21 @@ public function post_addItem()
...
+            $object     = new Ticket();
+            $object->getFromDB($this->input['tickets_id']);
+            $update['status'] = CommonITILObject::WAITING;
+            $update['id']     = $object->fields['id'];
+            // don't notify on ITILObject - update event
+            $update['_disablenotif'] = true;
+            // Use update method for history
+            $object->update($update);
+        }
+
```
AdrienClairembault, line 328: I think you should be able to reuse what we are already doing for followup and tasks here in this method:

image

(Don't forget to add the `use Glpi\Features\ParentStatus;` statement).

And you will also need to handle pending reason edition in the `post_updateItem` function, like this:

image

AdrienClairembault (changes requested):

cedric-anne:
@tsmr I changed target branch to main. We are trying to limit as much as functionnal changes to GLPI 10.0.

orthagh:
staled I guess. If you have time to work again on it, please let us know.

PR #15886, opened 2023-10-30
Fix Locales for Software

tsmr:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #number

Fix Locales for Software __('Technician in charge of the software') / __('Group in charge of the software')


trasher:
Won't `__('Technician in charge')` be a better solution? (not only for this case).

PR #15900, opened 2023-10-31
fix incorrect decimal sql warning on new ticket form

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

A new install of GLPI 10.0.10 generates a "Truncated incorrect DECIMAL value: ''" SQL warning when viewing the new ticket form as it tries calling `CommonITILActor::getActors` with an empty string as the ID.

### DIFF
--- a/src/CommonITILActor.php
+++ b/src/CommonITILActor.php
@@ -126,7 +126,7 @@ abstract class CommonITILActor extends CommonDBRelation
         global $DB;
 
         $iterator = $DB->request([
-            'FROM'   => $this->getTable(),
+            'FROM'   => static::getTable(),
             'WHERE'  => [
                 static::getItilObjectForeignKey()   => $items_id,
                 'alternative_email'                 => $email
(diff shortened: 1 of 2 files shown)


cedric-anne (changes requested):
The condition in `loadUsers()`, `loadGroups()` and `loadSuppliers()` should also be changed from `isset($this->fields['id'])` to `$this->isNewItem().

cedric-anne (approved):

cconard96:
> The changes themselves looks good to me as they ensure that we can handle these unexpected values without errors.
> 
> However it would also be nice to know why we are sending an empty string here ? This way we could also fix the origin of this invalid parameter

Its just a matter of blindly passing item IDs into SQL requests when the item may be new. When you create a new CommonDBTM instance, there is no ID set at all. When you call getEmpty, all fields are set as an empty string. Maybe that part can be changed, but I don't know if it will break something else or handle every case.

trasher (approved):

PR #15926, opened 2023-11-01
Strict types for getActors method

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

More correct fix of #15900 for main branch to avoid bad SQL requests.

### DIFF
--- a/src/CommonITILActor.php
+++ b/src/CommonITILActor.php
@@ -101,7 +101,7 @@ abstract class CommonITILActor extends CommonDBRelation
         /** @var \DBmysql $DB */
         global $DB;
 
-        if (empty($items_id)) {
+        if ($items_id <= 0) {
             return [];
         }
 
--- a/src/CommonITILActor.php
+++ b/src/CommonITILActor.php
@@ -92,9 +92,11 @@ abstract class CommonITILActor extends CommonDBRelation
 
 
     /**
-     * @param $items_id
+     * @param int $items_id
+     * @phpstan-param positive-int $items_id
+     * @return array Array of actors
      **/
-    public function getActors($items_id)
+    public function getActors(int $items_id): array
     {
         /** @var \DBmysql $DB */
         global $DB;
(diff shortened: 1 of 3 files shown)


src/CommonITILActor.php
```diff
@@ -92,16 +92,18 @@ public function isAttach2Valid(array &$input)
...
-    public function getActors($items_id)
+    public function getActors(int $items_id): array
     {
         /** @var \DBmysql $DB */
         global $DB;
 
         $users = [];
         $iterator = $DB->request([
-            'FROM'   => $this->getTable(),
+            'FROM'   => static::getTable(),
```
trasher, line 106: This should be changed on 10.0/bf branch

src/CommonITILObject.php
```diff
@@ -123,7 +123,7 @@ public function loadUsers(): void
     {
         if (!empty($this->userlinkclass) && isset($this->fields['id'])) {
             $class = new $this->userlinkclass();
-            $this->lazy_loaded_users = $class->getActors($this->fields['id']);
+            $this->lazy_loaded_users = $class->getActors((int) $this->fields['id']);
```
trasher, line 126: This (and other similar changes) should be changed on 10.0/bf branch

trasher (commented):
There are at least some changes that really should be applied on bugfixes branch. I wonder if the whole PR should not target bugfixes branch, this does not seems to be a BC break.

src/CommonITILObject.php
cconard96: No longer needed. 10.0 fix changes the `if` condition so `getActors` isn't called at all for new items.

trasher (approved):

cedric-anne (approved):

cedric-anne (commented):
PHPStan issue has to be fixed.

cedric-anne (approved):

PR #15965, opened 2023-11-08
feat(Ruleticket): add '$this->fields[$key]' value if key is used by rule criteria and not update

stonebuzz:
The idea of this PR is to inject into the ```$input```  (used by the rules) the values that would not be modified during an update.

Currently, only the ```keys``` updated **and** used by the rules are sent.

With this PR, we can now send ```keys``` that have been : 
- updated (with ```$obj->update(['key' => $val]);``` and used by the rules and not already added by rules (see ```$post_added```)
- not updated but used by the rules and not already added by rules (see ```$post_added```)

**Case study** 

A rule checks the status of a ticket when a validation request is approved.

The status is never sent as it is not updated in the process (```CommonITILValidation```)

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #number

### DIFF
--- a/src/Ticket.php
+++ b/src/Ticket.php
@@ -1248,6 +1248,13 @@ class Ticket extends CommonITILObject
                 ) {
                     $changes[] = $key;
                 }
+            } elseif (
+                array_key_exists($key, $this->fields) //existing in original field
+                && !array_key_exists($key, $post_added) //not already added by rules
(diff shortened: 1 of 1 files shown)


stonebuzz:
Require lot of work ATM

PR #15970, opened 2023-11-09
Add helper function to read glpi_currenttime

AdrienClairembault:
Allow us to benefit from IDE auto completion when trying to get this value rather than trying to guess the correct key name from memory.

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  |no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |

### DIFF
--- a/src/Session.php
+++ b/src/Session.php
@@ -2025,4 +2025,14 @@ class Session
     {
         return $_SESSION['glpilanguage'] ?? null;
     }
+
+    /**
+     * Helper function to get the date stored in $_SESSION['glpi_currenttime']
+     *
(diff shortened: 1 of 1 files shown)


trasher:
Maybe alos replace usage, to be consistent - at least in core.

AdrienClairembault:
> Maybe also replace usage, to be consistent - at least in core.

I agree with the thought but I'd like to avoid any refactoring on the bugfixes branches (can be done in `main` later tho).

cedric-anne (approved):

PR #15971, opened 2023-11-09
[HLAPI] Rules

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

Adds multiple endpoints for managing rules via the new API.
There are the basic CRUD endpoints for rules, all rule criteria, all rule actions, specific criteria and specific actions.

A notable improvement over the legacy API when it comes to rules is the ability to discover the available action types (assign, fromitem, etc), action fields, criteria conditions (How fields are being checks like 'is' or 'regex match') and criteria 'criteria' (The fields that are being checked) AND says in which cases they can be used such as which action types apply to a specific action field.


src/RuleCollection.php
```diff
@@ -862,21 +862,28 @@ public function moveRule($ID, $ref_ID, $type = self::MOVE_AFTER)
...
-            $rank = $ruleDescription->fields["ranking"];
-        } else if ($type == self::MOVE_AFTER) {
-           // Move after all
-            $result = $DB->request([
-                'SELECT' => ['MAX' => 'ranking AS maxi'],
-                'FROM'   => 'glpi_rules',
-                'WHERE'  => ['sub_type' => $this->getRuleClassName()]
-            ])->current();
-            $rank   = $result['maxi'];
+        $max_ranking_criteria = [
```
orthagh, line 865: I can't understand the changes below.
And you don't seems to use them in the current PR.
Do I miss something ?

cconard96: It has been a while since I made this PR, but I seem to remember making this change to make it possible to specify the new rank directly via the API without having duplicates. Having to make the rank an object during create/update actions so it can take a reference ID and a type (before/after) would have been complicated and potentially confusing.

Indeed, I don't see anything calling this method, so I will test it later.

orthagh (approved):

cedric-anne (approved):

PR #15995, opened 2023-11-15
feat(Profile): add right scope for each right

stonebuzz:
Add a tooltip to each right to see its scope

image

We still need : 

- [x] Do not display a tootlip if is scope is limited to a single use, e.g.

image

- [x] optimization to avoid searching / reloading / checking classes for each right

| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #number

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -62,6 +62,7 @@
         "symfony/console": "^5.4",
         "symfony/css-selector": "^5.4",
         "symfony/dom-crawler": "^5.4",
+        "symfony/finder": "^5.4",
         "symfony/polyfill-ctype": "^1.26",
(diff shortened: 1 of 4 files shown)


trasher:
please fix tests ;)

cedric-anne (changes requested):
I am not sure that introducing a new dependency in a bugfixes release, for a so small change, is a good idea.

stonebuzz:
Replaced by 
https://github.com/glpi-project/glpi/pull/16007

PR #15999, opened 2023-11-15
Fix "Operating system name" search option

AdrienClairembault:
This search option may return multiple results and should thus use `GROUP CONCAT` to prevent duplicated results.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |

### DIFF
--- a/src/Item_OperatingSystem.php
+++ b/src/Item_OperatingSystem.php
@@ -392,6 +392,7 @@ class Item_OperatingSystem extends CommonDBRelation
             'field'              => 'name',
             'name'               => _n('Version', 'Versions', 1),
             'datatype'           => 'dropdown',
+            'forcegroupby'       => true,
             'massiveaction'      => false,
             'joinparams'         => [
                 'beforejoin'         => [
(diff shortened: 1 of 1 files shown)


trasher:
Some other SO may suffer of the same issue (61, 64, 48, 63), reported by @cedric-anne on local chat

AdrienClairembault:
The others SO do not recreate the issue on my side.
I guess that while they could technically causes issues (since this is a 1-N relation) they don't end up with more than a 1-1 relationship in real world cases.

@stonebuzz was also telling me that even for the original case it should no longer happens with GLPI's native inventory.

Since `GROUP CONCAT` is probably not great performances wise, maybe its not worth it to add it the others options ?

trasher:
If original issue should not happen; maybe could we try to fix what's wrongly stored in DB instead of trying to fix the search? I guess it's the better option for performances.

Several OS on a same asset never has been supported in facts, it should not have been done that way in the code (my bad at the origin).

This is an issue that keeps rearing its ugly head; I wonder if we should not change the relation to a real 1-1 to definitely fix that (not before 10.1 anyway).

orthagh:
Please extends the `forcegroupby` key to all searchoptions like said by Cedric.
We'll make a decision about relation nature of OS later.
In the case we move to 1-1, the forcegroupby will probably not interfer from a UI pov

trasher (approved):

cedric-anne (approved):

PR #16004, opened 2023-11-16
fix(central): warning displayed on all profiles

Rom1-B:
The warning message about root directory configuration was displayed on all profiles, whereas other such messages are filtered so that they are only visible to admins.

image

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !30455


cedric-anne (changes requested):
This is not the right place to check rights.

Change have to be made on `Central::getMessages()`.

cedric-anne (approved):

orthagh (approved):

PR #16011, opened 2023-11-17
Fix(Inventory): fix RuleImportAsset for Phone

stonebuzz:
I don't have a history of the changes made to this part, 
but the rules for importing and linking material that concerning ```Phones``` (which are deactivated by default) don't work at all.
As it stands, it's compulsory to adapt them, otherwise the devices are refused for import.

The default criteria do not seem consistent.

image

This PR corrects the XML resources for creating/resetting XML rules

Because we don't update the rules directly (despite the presence of the ```uuid``` now) -> too complicated

I simply duplicated the rules on the PCs and adapted them to the Phones.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !30284


trasher (approved):

Rom1-B:
"Phone import denied" should be last?

image

In the "All" tab, the first "Phones" rules are mixed with those for "Computers".

image

stonebuzz:
Last commit fix rule rank and bad copy past

Rom1-B (approved):

cedric-anne (approved):

cedric-anne:
Please rebase to fix conflicts.

cedric-anne (changes requested):
It seems that you have many duplicated rules in the XML file.

resources/Rules/RuleImportAsset.xml
```diff
@@ -1933,4 +1817,426 @@
             <value>1</value>
         </ruleaction>
     </rule>
+    <rule>
+        <entities_id>Root entity</entities_id>
+        <sub_type>RuleImportAsset</sub_type>
+        <ranking>38</ranking>
```
cedric-anne, line 1823: ranking 38 is after 62, it does not seems logical

Rom1-B (approved):

cedric-anne (approved):

PR #16023, opened 2023-11-20
Do not log agent last contact change

AdrienClairembault:
Agent last contact information create a log of logs that do not seems very interesting to me.

Would it be OK to blacklist this field for the history or do you think it is useful to keep track of past inventories date ?

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |

### DIFF
--- a/src/Agent.php
+++ b/src/Agent.php
@@ -69,6 +69,8 @@ class Agent extends CommonDBTM
 
     private static $found_address = false;
 
+    public $history_blacklist = ['last_contact'];
+
     public static function getTypeName($nb = 0)
     {
         return _n('Agent', 'Agents', $nb);


trasher:
No preference for me.

Last contact is not the same as last rule matched as far as I know; but I have no idea if that can be useful or not.

orthagh (approved):
Agree on removing from history this changes

trasher (approved):

AdrienClairembault:
Not sure why last commit is needed, it seems there is an entry log entry being found:
```
array(10) {
  ["id"]=>
  int(67120)
  ["itemtype"]=>
  string(5) "Agent"
  ["items_id"]=>
  int(7)
  ["itemtype_link"]=>
  string(0) ""
  ["linked_action"]=>
  int(0)
  ["user_name"]=>
  string(9) "inventory"
  ["date_mod"]=>
  string(19) "2023-11-20 15:29:33"
  ["id_search_option"]=>
  int(8)
  ["old_value"]=>
  string(6) "2.3.19"
  ["new_value"]=>
  string(6) "2.3.20"
}
```

cedric-anne (approved):

PR #16026, opened 2023-11-20
Fix notification entity restriction

AdrienClairembault:
The notification system does not behave correctly with recursive items.

Lets assume we have a contract on the root entity with children entities enabled.
A user with an authorization on a sub entity can see the contract but won't receive any notifications.

You can see this example in details in the added test case, which is failing without the other changes.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !29820

### DIFF
--- a/src/NotificationTarget.php
+++ b/src/NotificationTarget.php
@@ -586,6 +586,17 @@ class NotificationTarget extends CommonDBChild
                 $this->getEntity(),
                 true
             );
+
+            // If one of the item is recursive, then we should also check the child entities
+            if ($this->mustCheckChildrenEntities()) {
+                $filt = [
+                    'OR' => [
+                        $filt,
+                        ['entities_id' => getSonsOf(Entity::getTable(), $this->getEntity())]
+                    ]
+                ];
+            }
+
             $prof = Profile_User::getUserProfiles($data['users_id'], $filt);
             if (!count($prof)) {
                 // No right on the entity of the object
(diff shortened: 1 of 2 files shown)


AdrienClairembault:
For now I only have a failing test case to prove the issue exist.

I think the incorrect checks are mainly done here and there :

image

image

This code only look for entities above the current one.
It seems we should also look for any sub-entities if the current item is recursive.

src/NotificationTarget.php
```diff
@@ -1710,4 +1734,22 @@ public function setAllowResponse($allow_response)
 
         return $this;
     }
+
+    /**
+     * Check if at least one target item is recursive
+     *
+     * @return bool
+     */
+    protected function mustCheckChildrenEntities(): bool
```
cedric-anne, line 1743: IMHO, if there are multiple target objects and one of them is NOT recursive, we should not include recipients that have only access to child entities.

Also, maybe the method should be named `canRecipientsFromAddChildEntities` or something like that.

cedric-anne (changes requested):

src/NotificationTarget.php
AdrienClairembault: But wouldn't some valid users still don't get notifications in this case ?
I don't see how to get a perfect solution without huge breaking changes, the fact that an event can accept multiple items is a huge problem to handle entities correctly

AdrienClairembault: Maybe I could target `$this->obj` instead, they its the responsibility of each notification to be triggered correctly ?

AdrienClairembault:
I can add a TODO so we don't forgot to sort this out on main ?

cedric-anne (approved):

cedric-anne:
> I can add a TODO so we don't forgot to sort this out on main ?

I discussed about it with @orthagh and we concluded that it will not be an issue.

PR #16033, opened 2023-11-21
Association of a technician with a ticket if he is assigned a ticket task

Lainow:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !29551

This request associates a user or group with a ticket if they are assigned to a task linked to that ticket.
image
image
image
This is handled before calling commonitiltask.form.php in tickettask.form.php.

### DIFF
--- a/front/tickettask.form.php
+++ b/front/tickettask.form.php
@@ -36,5 +36,4 @@
 include('../inc/includes.php');
 
 $task = new TicketTask();
-
 include(GLPI_ROOT . "/front/commonitiltask.form.php");
--- a/front/tickettask.form.php
+++ b/front/tickettask.form.php
@@ -36,4 +36,19 @@
 include('../inc/includes.php');
 
 $task = new TicketTask();
+
+//if user or group assigned to ticket task, add it to the ticket
+$entities = [
+    'users_id_tech' => new Ticket_User(),
+    'groups_id_tech' => new Group_Ticket(),
+];
+
+foreach ($entities as $key => $entity) {
+    if (isset($_POST[$key])) {
+        $entityId = str_replace('_tech', '', $key);
+        if ($entity->getFromDBByCrit(['tickets_id' => $_POST['tickets_id'], $entityId => $_POST[$key], 'type' => 2]) == false) {
+            $entity->add(['tickets_id' => $_POST['tickets_id'], $entityId => $_POST[$key], 'type' => 2]);
+        }
+    }
+}
 include(GLPI_ROOT . "/front/commonitiltask.form.php");
--- a/front/tickettask.form.php
+++ b/front/tickettask.form.php
@@ -37,18 +37,4 @@ include('../inc/includes.php');
 
 $task = new TicketTask();
 
-//if user or group assigned to ticket task, add it to the ticket
-$entities = [
-    'users_id_tech' => new Ticket_User(),
-    'groups_id_tech' => new Group_Ticket(),
-];
-
-foreach ($entities as $key => $entity) {
-    if (isset($_POST[$key])) {
-        $entityId = str_replace('_tech', '', $key);
-        if ($entity->getFromDBByCrit(['tickets_id' => $_POST['tickets_id'], $entityId => $_POST[$key], 'type' => 2]) == false) {
-            $entity->add(['tickets_id' => $_POST['tickets_id'], $entityId => $_POST[$key], 'type' => 2]);
-        }
-    }
-}
 include(GLPI_ROOT . "/front/commonitiltask.form.php");
--- a/src/CommonITILTask.php
+++ b/src/CommonITILTask.php
@@ -293,7 +293,7 @@ abstract class CommonITILTask extends CommonDBTM implements CalDAVCompatibleItem
 
     public function assignTechFromtask(array $input): void
     {
-        //if user or group assigned to ticket task, add it to the ticket
+        //if user or group assigned to CommonITIL task, add it to the main item
         $itemtype = $this->getItilObjectItemType();
         $item = new $itemtype();
         $itemData = [
--- a/src/CommonITILTask.php
+++ b/src/CommonITILTask.php
@@ -297,20 +297,20 @@ abstract class CommonITILTask extends CommonDBTM implements CalDAVCompatibleItem
         $entities = [
             'Ticket' => [
                 'users_id_tech' => new Ticket_User(),
-                'groups_id_tech' => new Group_Ticket(),
+                'groups_id_tech' => new Group_Ticket()
             ],
             'Change' => [
                 'users_id_tech' => new Change_User(),
-                'groups_id_tech' => new Change_Group(),
+                'groups_id_tech' => new Change_Group()
             ],
             'Problem' => [
                 'users_id_tech' => new Problem_User(),
-                'groups_id_tech' => new Group_Problem(),
+                'groups_id_tech' => new Group_Problem()
             ],
         ];
         foreach ($entities as $entityType => $entityData) {
-            if ($entityType == isset($input['itemtype'])) {
-                $entityType = strtolower($entityType);
+            $entityType = strtolower($entityType);
+            if (isset($input[$entityType . 's_id'])) {
                 foreach ($entityData as $key => $entity) {
                     if (isset($input[$key])) {
                         $entityId = str_replace('_tech', '', $key);
@@ -428,8 +428,10 @@ abstract class CommonITILTask extends CommonDBTM implements CalDAVCompatibleItem
                 }
             }
         }
-        // Assign technician to ticket from ticket task
-        self::assignTechFromtask($input);
+        if (isset($input["changes_id"]) || isset($input["problems_id"]) || isset($input["tickets_id"])) {
+            // Assign technician to ticket from ticket task
+            self::assignTechFromtask($input);
+        }
 
         return $input;
     }
@@ -638,8 +640,10 @@ abstract class CommonITILTask extends CommonDBTM implements CalDAVCompatibleItem
         if (isset($input["users_id"])) {
             $input['timeline_position'] = $itemtype::getTimelinePosition($input["_job"]->getID(), $this->getType(), $input["users_id"]);
         }
-        // Assign technician to ticket from ticket task
-        self::assignTechFromtask($input);
+        if (isset($input["changes_id"]) || isset($input["problems_id"]) || isset($input["tickets_id"])) {
+            // Assign technician to ticket from ticket task
+            self::assignTechFromtask($input);
+        }
 
         return $input;
     }
--- a/src/CommonITILTask.php
+++ b/src/CommonITILTask.php
@@ -312,7 +312,7 @@ abstract class CommonITILTask extends CommonDBTM implements CalDAVCompatibleItem
             $entityType = strtolower($entityType);
             if (isset($input[$entityType . 's_id'])) {
                 foreach ($entityData as $key => $entity) {
-                    if (isset($input[$key])) {
+                    if (isset($input[$key]) && !empty($input[$key])) {
                         $entityId = str_replace('_tech', '', $key);
                         if (
                             $entity->getFromDBByCrit(
--- a/src/CommonITILTask.php
+++ b/src/CommonITILTask.php
@@ -721,7 +721,7 @@ abstract class CommonITILTask extends CommonDBTM implements CalDAVCompatibleItem
             Log::HISTORY_ADD_SUBITEM
         );
 
-        // Assign technician to ticket from ticket task
+        // Assign technician to main item  from task
         self::assignTechFromtask($this->input);
 
         parent::post_addItem();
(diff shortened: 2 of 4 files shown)


stonebuzz:
As seen IRL,

check whether the behavior you're trying to correct is observable on ```Change``` and ```Problem```.

If so, we can implement the feature directly in ```CommonITILTask``` (instead of using ```tickettask.form.php```.

Lainow:
> As seen IRL,
> 
> check whether the behavior you're trying to correct is observable on `Change` and `Problem`.
> 
> If so, we can implement the feature directly in `CommonITILTask` (instead of using `tickettask.form.php`).

After trying, we have the same problem for Change and Problem. I've moved this task to CommonITILTask to solve the problem.

src/CommonITILTask.php
```diff
@@ -309,7 +309,7 @@ public function assignTechFromtask($input)
             ],
         ];
         foreach ($entities as $entityType => $entityData) {
-            if ($entityType == $input['itemtype']) {
+            if ($entityType == isset($input['itemtype'])) {
```
stonebuzz, line 312: Seems wrong here

You affected the result of ```isset``` function instead of affected ```$input['itemtype']``` value

stonebuzz (changes requested):

src/CommonITILTask.php
trasher: TO be 100% sure, add parenthesis:
```suggestion
            if ($entityType == ($input['itemtype'] ?? '')) {
```

Lainow: As the item type is not given every time, I removed this check and replaced it with if (isset($input[$entityType . 's_id'])) which matches problems_id, changes_id and tickets_id.

src/CommonITILTask.php
```diff
@@ -291,10 +291,55 @@ private function handleTaskDuration(array &$input, int $timestart, int $timeend)
         }
     }
 
+    public function assignTechFromtask($input)
+    {
+        //if user or group assigned to ticket task, add it to the ticket
+        $entities = [
```
AdrienClairembault, line 297: "entities" is a bit ambiguous here because we often use it to refer to the entity feature of GLPI.
Maybe something like "itemtype" would be clearer ?

src/CommonITILTask.php
```diff
@@ -291,10 +291,55 @@ private function handleTaskDuration(array &$input, int $timestart, int $timeend)
...
+                'users_id_tech' => new Change_User(),
+                'groups_id_tech' => new Change_Group()
+            ],
+            'Problem' => [
+                'users_id_tech' => new Problem_User(),
+                'groups_id_tech' => new Group_Problem()
+            ],
+        ];
+        foreach ($entities as $entityType => $entityData) {
+            $entityType = strtolower($entityType);
```
AdrienClairembault, line 312: I suppose you are iterating on every possible parent type here because there is no method to get the final ITIL class for a given task class.

I think it would be clearer to add a abstract `getTargetITILClass` method to the `CommonITILTask` class.
Each child class would then define its correct target (e.g. `return Ticket::class` for `TicketTask`).

With this new method you don't need to loop on types anymore and can just call `$this->getTargetITILClass()`.

src/CommonITILTask.php
```diff
@@ -291,10 +291,55 @@ private function handleTaskDuration(array &$input, int $timestart, int $timeend)
...
+            if (isset($input[$entityType . 's_id'])) {
+                foreach ($entityData as $key => $entity) {
+                    if (isset($input[$key]) && !empty($input[$key])) {
+                        $entityId = str_replace('_tech', '', $key);
+                        if (
+                            $entity->getFromDBByCrit(
+                                [
+                                    $entityType . 's_id' => $input[$entityType . 's_id'],
```
AdrienClairembault, line 320: You have a few method that allow you to get the foreign key name for any itemtype (tickets_id for ticket) :
* `CommonDBTM->getForeignKeyField()`
* `getForeignKeyFieldForItemType`
* `getForeignKeyFieldForTable`

The rule are a bit more complicated than just adding "s_id" so you must always use these methods :)

src/CommonITILTask.php
```diff
@@ -291,10 +291,55 @@ private function handleTaskDuration(array &$input, int $timestart, int $timeend)
...
+                                    $entityId => $input[$key],
+                                    'type' => 2
```
AdrienClairembault, line 322: The constant for the actor types are found in the `CommonITILActor` class.

For example you can use `CommonITILActor::ASSIGN` here instead of 2 :)

src/CommonITILTask.php
```diff
@@ -291,10 +291,55 @@ private function handleTaskDuration(array &$input, int $timestart, int $timeend)
         }
     }
 
+    public function assignTechFromtask($input)
+    {
+        //if user or group assigned to ticket task, add it to the ticket
+        $entities = [
+            'Ticket' => [
+                'users_id_tech' => new Ticket_User(),
+                'groups_id_tech' => new Group_Ticket()
```
AdrienClairembault, line 300: To get the correct xxx_user and group_xxx classes for an ITIL object, you have access to these properties:

image

tests/functional/CommonITILTask.php
```diff
@@ -0,0 +1,138 @@
...
+namespace tests\units;
+
+use DbTestCase;
+
+class CommonITILTask extends DbTestCase
+{
+    public function testAddTechToTicketFromTask()
+    {
+        $ticket = new \Ticket();
+        $ticket_id = 101;
```
AdrienClairembault, line 45: If this test is the exact same as `testAddTechToProblemFromTask` but for tickets, you should probably use a dataprovider instead.

Data provider allow you to run the same tests with different inputs.
Look for some example using the `@dataProvider` keywork in ours existing tests

AdrienClairembault (changes requested):

AdrienClairembault (approved):

front/tickettask.form.php
```diff
@@ -36,4 +36,5 @@
 include('../inc/includes.php');
 
 $task = new TicketTask();
+
```
stonebuzz, line 39: useless =)

tests/functional/Ticket.php
```diff
@@ -1860,12 +1861,15 @@ public function testClone()
     {
         $this->login();
         $this->setEntity('Root entity', true);
-        $ticket = getItemByTypeName('Ticket', '_ticket01');
-
+        $ticket = new GlobalTicket();
```
stonebuzz, line 1864:
```suggestion
        // this test originally used the ticket created in the dataset 
        // this ticket contained a task, which on creation added the tech to the assigned actors 
        // and add  additional log entry relating to this action 
        // so $this->integer((int)$clonedTicket->getTimelineItems())->isEqualTo(0); return false 
        // because getTimelineItems also returns the history of the actors
        
        //$ticket = getItemByTypeName('Ticket', '_ticket01');
        
        $ticket = new GlobalTicket();
```

stonebuzz (changes requested):

src/CommonITILTask.php
```diff
@@ -291,10 +291,45 @@ private function handleTaskDuration(array &$input, int $timestart, int $timeend)
         }
     }
 
+    public function assignTechFromtask($input, string $type)
```
cedric-anne, line 294: 1. `$type` is not used
2. Please use strict types.
```suggestion
    public function assignTechFromtask(array $input): void
```

src/CommonITILTask.php
```diff
@@ -291,10 +291,45 @@ private function handleTaskDuration(array &$input, int $timestart, int $timeend)
...
+    {
+        //if user or group assigned to ticket task, add it to the ticket
+        $itemtype = $this->getItilObjectItemType();
+        $item = new $itemtype();
+        $itemData = [
+            'users_id_tech' => new $item->userlinkclass(),
+            'groups_id_tech' => new $item->grouplinkclass(),
+        ];
+        $foreignkey = getForeignKeyFieldForItemType($itemtype);
+        if (isset($input[$foreignkey])) {
```
cedric-anne, line 304: When it is possible, please try to do something like this:
```php
if (!isset($input[$foreignkey])) {
    return;
}
```

It will remove some indentation levels and make the code easier to read.

src/CommonITILTask.php
```diff
@@ -291,10 +291,45 @@ private function handleTaskDuration(array &$input, int $timestart, int $timeend)
...
+            foreach ($itemData as $key => $value) {
+                if (isset($input[$key]) && !empty($input[$key])) {
```
cedric-anne, line 306:
```php
if (empty($input[$key])) {
    continue;
}
```

cedric-anne (changes requested):

tests/functional/Ticket.php
```diff
@@ -1860,12 +1861,21 @@ public function testClone()
...
         $this->login();
         $this->setEntity('Root entity', true);
-        $ticket = getItemByTypeName('Ticket', '_ticket01');
-
+        // this test originally used the ticket created in the dataset
+        // this ticket contained a task, which on creation added the tech to the assigned actors
+        // and add  additional log entry relating to this action
+        // so $this->integer((int)$clonedTicket->getTimelineItems())->isEqualTo(0); return false
+        // because getTimelineItems also returns the history of the actors
+        //$ticket = getItemByTypeName('Ticket', '_ticket01');
```
cedric-anne, line 1869:
```suggestion
```

tests/functional/Ticket.php
```diff
@@ -1860,12 +1861,21 @@ public function testClone()
...
+        $ticket = new GlobalTicket();
+        $ticket_id = $ticket->add([
+            'name'    => 'Ticket to check cloning',
+            'content' => 'Ticket to check cloning',
+        ]);
```
cedric-anne, line 1874:
```suggestion
        ]);
        $this->integer($ticket_id)->isGreaterThan(0);
```

tests/functional/CommonITILTask.php
```diff
@@ -0,0 +1,119 @@
...
+        $problem = new \Problem();
+        $problem_id = $problem->add([
+            'name'        => "Test tech task",
+            'content'     => "Test tech task",
+            'entities_id' => getItemByTypeName('Entity', '_test_root_entity', true),
+            'date'        => $_SESSION['glpi_currenttime'],
+        ]);
+        return [
+            [
+                'id' => 101,
```
cedric-anne, line 60: You should not use an hardcoded ID here.

src/CommonITILTask.php
```diff
@@ -291,10 +291,47 @@ private function handleTaskDuration(array &$input, int $timestart, int $timeend)
...
+        foreach ($itemData as $key => $value) {
+            if (empty($input[$key])) {
+                continue;
+            }
+            $actorFkey = str_replace('_tech', '', $key);
+            if (
+                $value->getFromDBByCrit(
+                    [
+                        $foreignkey => $input[$foreignkey],
+                        $entityId => $input[$key],
```
cedric-anne, line 316:
```suggestion
                        $actorFkey => $input[$key],
```

src/CommonITILTask.php
```diff
@@ -291,10 +291,47 @@ private function handleTaskDuration(array &$input, int $timestart, int $timeend)
...
+                        'type' => CommonITILActor::ASSIGN
+                    ]
+                ) == false
+            ) {
+                $value->add(
+                    [
+                        $foreignkey => $input[$foreignkey],
+                        $entityId => $input[$key],
```
cedric-anne, line 324:
```suggestion
                        $actorFkey => $input[$key],
```

cedric-anne (approved):

src/CommonITILTask.php
```diff
@@ -494,6 +531,9 @@ public function post_updateItem($history = true)
             );
         }
 
+        // Assign technician to ticket from ticket task
+        self::assignTechFromtask($this->input, get_class($this->input["_job"]));
```
cedric-anne, line 535:
```suggestion
        self::assignTechFromtask($this->input);
```

src/CommonITILTask.php
```diff
@@ -681,6 +721,9 @@ public function post_addItem()
             Log::HISTORY_ADD_SUBITEM
         );
 
+        // Assign technician to ticket from ticket task
+        self::assignTechFromtask($this->input, get_class($this->input["_job"]));
```
cedric-anne, line 725:
```suggestion
        self::assignTechFromtask($this->input);
```

tests/functional/CommonITILTask.php
```diff
@@ -0,0 +1,126 @@
...
+                'type' => \CommonITILActor::ASSIGN
+            ]
+        )))->isEqualTo(1);
+
+        $task->update([
+            'id' => $task_id,
+            $foreignkey => $id,
+            'users_id' => 3,
+            'users_id_tech' => 3,
+        ]);
```
trasher, line 116:
```suggestion
        $this->boolean(
            $task->update([
                'id' => $task_id,
                $foreignkey => $id,
                'users_id' => 3,
                'users_id_tech' => 3,
            ])
        )->isTrue();
```

PR #16041, opened 2023-11-23
Fix PHPDoc

AdrienClairembault:
This method may also return an array:

image

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |

### DIFF
--- a/src/AuthLDAP.php
+++ b/src/AuthLDAP.php
@@ -2236,7 +2236,7 @@ class AuthLDAP extends CommonDBTM
      * @param array  $ldap_infos ldap user search result
      * @param string $user_dn    user dn to look for
      *
-     * @return bool|array false if the user dn doesn't exist, user ldap infos otherwise
+     * @return false|array false if the user dn doesn't exist, user ldap infos otherwise
      */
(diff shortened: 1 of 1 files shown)


src/AuthLDAP.php
```diff
@@ -2236,7 +2236,7 @@ public static function getAllUsers(array $options, &$results, &$limitexceeded)
      * @param array  $ldap_infos ldap user search result
      * @param string $user_dn    user dn to look for
      *
-     * @return boolean false if the user dn doesn't exist, user ldap infos otherwise
+     * @return bool|array false if the user dn doesn't exist, user ldap infos otherwise
```
trasher, line 2239:
```suggestion
     * @return false|array false if the user dn doesn't exist, user ldap infos otherwise
```

trasher (approved):
Next time please do not open a PR for only one change...

There are several hundreds of phpdocs to fix to reach the next phpstan level we target; you can pick a few other ones for a PR.

Thanks :)

PR #16042, opened 2023-11-23
Fix: carriage return in log files

Rom1-B:
In some log files the line break was missing

Example here before the date:
image

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

### DIFF
--- a/src/Migration.php
+++ b/src/Migration.php
@@ -198,7 +198,7 @@ class Migration
        // Do not log if more than 3 log error
         if (
             $this->log_errors < 3
-            && !Toolbox::logInFile($log_file_name, $message . ' @ ', true)
+            && !Toolbox::logInFile($log_file_name, $message . ' @\n', true)
         ) {
             $this->log_errors++;
         }
--- a/src/MailCollector.php
+++ b/src/MailCollector.php
@@ -1082,7 +1082,7 @@ class MailCollector extends CommonDBTM
                 $tkt['_tag']      = $this->tags;
             } else {
                //TRANS: %s is a directory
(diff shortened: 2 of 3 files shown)


trasher:
Seem correct. Maybe one missed:
```
diff --git a/src/GLPIMailer.php b/src/GLPIMailer.php
index 0933932898..633ef01771 100644
--- a/src/GLPIMailer.php
+++ b/src/GLPIMailer.php
@@ -122,7 +122,7 @@ class GLPIMailer extends PHPMailer
             $this->Debugoutput = function ($message, $level) {
                 Toolbox::logInFile(
                     'mail-debug',
-                    "$level - $message"
+                    "$level - $message\n"
                 );
             };
         }
```

cedric-anne (approved):

src/Migration.php
```diff
@@ -198,7 +198,7 @@ public function log($message, $warning)
        // Do not log if more than 3 log error
         if (
             $this->log_errors < 3
-            && !Toolbox::logInFile($log_file_name, $message . ' @ ', true)
+            && !Toolbox::logInFile($log_file_name, $message . ' @\n', true)
```
cedric-anne, line 201:
```suggestion
            && !Toolbox::logInFile($log_file_name, $message . ' @ ', true)
```

cedric-anne: This one is normal:

image

Rom1-B: Is it time for the next line event?

cedric-anne: Indeed. I reverted my commit.

PR #16070, opened 2023-11-27
Anonymous ticket actor notifications cannot be enabled/disabled

RomainLvr:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !30499

## Problem
In the current version 10.0.10 of the application, when the requester of a ticket is an (anonymous) email address, it is currently impossible to disable email tracking for this address. This issue results in automatic reactivation of email tracking after the ticket has been saved, making the deactivation feature ineffective.

## Impact
This limitation can be particularly troublesome in cases where users wish to specifically disable e-mail tracking for requestor e-mail addresses. Currently, this action does not have the desired effect and the tracking bell remains active, resulting in an unnecessary overload of notifications for these users.

### DIFF
--- a/src/CommonITILObject.php
+++ b/src/CommonITILObject.php
@@ -8649,11 +8649,14 @@ abstract class CommonITILObject extends CommonDBTM
                     ) {
                         // "email" actor found
                         if ($actor['alternative_email'] == $existing['alternative_email']) {
+                            // The anonymous actor matches an existing one, update it
+                            $updated[] = $actor + ['id' => $existing['id']];
                             $found = true;
                             break;
+                        } else {
+                            // Do not check for modifications on "email" actors (they should be deleted then re-added on email change)
+                            continue;
                         }
-                        // Do not check for modifications on "email" actors (they should be deleted then re-added on email change)
-                        continue;
                     }
 
                     if ($actor['itemtype'] != $existing['itemtype'] || $actor['items_id'] != $existing['items_id']) {
--- a/src/User.php
+++ b/src/User.php
@@ -652,7 +652,7 @@ class User extends CommonDBTM
     public function getDefaultEmail()
     {
 
-        if (empty($this->fields['id'])) {
+        if ($this->isNewItem()) {
             return '';
         }
 
--- a/src/User.php
+++ b/src/User.php
@@ -652,7 +652,7 @@ class User extends CommonDBTM
     public function getDefaultEmail()
     {
 
-        if (!isset($this->fields['id'])) {
+        if (empty($this->fields['id'])) {
             return '';
         }
 
--- a/tests/functional/Ticket.php
+++ b/tests/functional/Ticket.php
@@ -58,7 +58,7 @@ use User;
 
 class Ticket extends DbTestCase
 {
-    protected function actorsProvider(): iterable
+    protected function addActorsProvider(): iterable
     {
         $default_use_notifications = 1;
 
@@ -519,7 +519,7 @@ class Ticket extends DbTestCase
     }
 
     /**
-     * @dataProvider actorsProvider
+     * @dataProvider addActorsProvider
      */
     public function testCreateTicketWithActors(array $actors_input, array $expected_actors): void
     {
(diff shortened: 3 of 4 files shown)


trasher:
Please add a test case

tests/functional/Ticket.php
```diff
@@ -560,6 +560,144 @@ public function testUpdateTicketWithActors(array $actors_input, array $expected_
...
+
+        // Create ticket
+        $ticket_id = $this->createItem(
+            \Ticket::class,
+            [
+                'name'        => 'ticket title',
+                'content'     => 'a description',
+                'entities_id' => getItemByTypeName('Entity', '_test_root_entity', true),
+            ]
+        )->fields['id'];
```
Rom1-B, line 588: As explained IRL, you have to check each step of the test (afterwards too):
```suggestion
        // Create ticket
        $ticket_id = $this->createItem(
            \Ticket::class,
            [
                'name'        => 'ticket title',
                'content'     => 'a description',
                'entities_id' => getItemByTypeName('Entity', '_test_root_entity', true),
            ]
        )->fields['id'];
        $this->integer($ticket_id)->isGreaterThan(0);
```

Rom1-B (changes requested):

tests/functional/Ticket.php
RomainLvr: Fixed in commit https://github.com/glpi-project/glpi/pull/16070/commits/f474aae7213b52af6e5b7e9c66b3de582ba36cd0
Thank :+1:

cedric-anne: `$this->createItem()` already checks that item is correctly created.

src/CommonITILObject.php
```diff
@@ -8620,20 +8620,6 @@ protected function updateActors(bool $disable_notifications = false)
...
-                    ) {
-                        // "email" actor found
-                        if ($actor['alternative_email'] == $existing['alternative_email']) {
-                            $found = true;
-                            break;
-                        }
-                        // Do not check for modifications on "email" actors (they should be deleted then re-added on email change)
-                        continue;
-                    }
-
```
cedric-anne, line 8636: Removing this whole piece of code result in unexpected behaviour when modifiying actors.

For instance:
1. Create a ticket.
2. Add `test1@domain.tld` in requester field and save the form.
3. Add `test2@domain.tld` in requester field and save the form.
Expected result: `test1@domain.tld` and `test2@domain.tld` are both present in requester fields.
Current result: only `test2@domain.tld` is present in requester fields.

AdrienClairembault: Good catch, we should probably add a test on this specific action to ensure it isn't broken in a future update.

AdrienClairembault (changes requested):

tests/functional/Ticket.php
RomainLvr: Thanks :+1: useless code has been removed https://github.com/glpi-project/glpi/pull/16070/commits/4e27a0d7970f917bef330706c616f0f495c05622

tests/functional/Ticket.php
```diff
@@ -560,6 +560,383 @@ public function testUpdateTicketWithActors(array $actors_input, array $expected_
...
+     *
+     * This method tests the functionality of updating ticket notifications for anonymous users.
+     * It creates a ticket with different actors (requester, observer, assign) and enables notifications.
+     * It then adds a follow-up to the ticket to trigger notifications and checks the notification queue.
+     * It updates the anonymous users to disable their notifications and checks the notification queue again.
+     * It then adds a second anonymous requester to the ticket and checks the notification queue again.
+     *
+     * @return void
+     */
+    public function testUpdateActorsNotification(): void
```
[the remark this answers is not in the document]
RomainLvr: suggestion applied

src/CommonITILObject.php
RomainLvr: The deleted code has been reinstated with a slight modification that should solve the problem : https://github.com/glpi-project/glpi/pull/16070/commits/4e27a0d7970f917bef330706c616f0f495c05622

src/CommonITILObject.php
```diff
@@ -8607,33 +8607,24 @@ protected function updateActors(bool $disable_notifications = false)
...
-                        && $actor['items_id'] == 0
-                        && $actor['itemtype'] == $existing['itemtype']
-                    ) {
-                        // "email" actor found
-                        if ($actor['alternative_email'] == $existing['alternative_email']) {
-                            $found = true;
-                            break;
+                    if ($actor['itemtype'] === User::class && $actor['items_id'] == 0) {
+                        // This is an anonymous actor
+                        if ($existing['itemtype'] === User::class && $existing['items_id'] == 0) {
```
cedric-anne, line 8614: Both if could be combined.

RomainLvr: Fixed in https://github.com/glpi-project/glpi/pull/16070/commits/436d0b4db1c83391ff6faaa54da6daefd7bda1d7

tests/functional/Ticket.php
```diff
@@ -560,6 +560,402 @@ public function testUpdateTicketWithActors(array $actors_input, array $expected_
...
+            }
+        }
+
+        $this->updateItem(
+            \Ticket::class,
+            $ticket_id,
+            [
+                '_actors' => $actors
+            ]
+        );
```
Rom1-B, line 654: You can link the actors during the creation of the ticket, which could simplify the reading. This will impact your `$this->checkActors(...` which follow

```suggestion
        // Create ticket
        $ticket_id = $this->createItem(
            \Ticket::class,
            [
                'name'        => 'ticket title',
                'content'     => 'a description',
                'entities_id' => getItemByTypeName('Entity', '_test_root_entity', true),
                '_users_id_requester' => 0,
                '_users_id_requester_notif' => [
                    'use_notification' => 1,
                    'alternative_email' => 'test@user.com'
                ],
                '_users_id_observer' => 0,
                '_users_id_observer_notif' => [
                    'use_notification' => 1,
                    'alternative_email' => 'test2@user.com'
                ],
                '_users_id_assign' => $user_db_id,
                '_users_id_assign_notif' => [
                    'use_notification' => 1,
                    'default_email' => $user_db_email
                ]
            ]
        )->fields['id'];
```

src/CommonITILObject.php
```diff
@@ -8617,23 +8617,26 @@ protected function updateActors(bool $disable_notifications = false)
...
+                        } else {
+                            // "email" actor found but emails do not match
+                            // Do not check for modifications on "email" actors (they should be deleted then re-added on email change)
+                            continue;
                         }
-                        // Do not check for modifications on "email" actors (they should be deleted then re-added on email change)
-                        continue;
                     }
-
                     if ($actor['itemtype'] != $existing['itemtype'] || $actor['items_id'] != $existing['items_id']) {
```
Rom1-B, line 8640: I'm not sure it was necessary to modify so many lines, only the addition of the line `$updated[] = ...` impact your test.

src/User.php
```diff
@@ -642,7 +642,7 @@ public function getFromDBbyEmail($email, $condition = [])
     public function getDefaultEmail()
     {
 
-        if (!isset($this->fields['id'])) {
```
Rom1-B, line 645: This change has no impact on your test, so it may not be necessary.

tests/functional/Ticket.php
```diff
@@ -560,6 +560,402 @@ public function testUpdateTicketWithActors(array $actors_input, array $expected_
...
+        $this->array(getAllDataFromTable(
+            \Ticket_User::getTable(),
+            [
+                'SELECT' => '*',
+                'WHERE' => [
+                    'tickets_id' => $ticket_id,
+                    'use_notification' => 1
+                ]
+            ]
+        ))->hasSize(3);
```
Rom1-B, line 737: Counting is not precise enough, you can compare tables with `isEqualTo` (to be completed):
```suggestion
        // Check that there are 3 ticket actors whose notifications are activated
        $this->array(getAllDataFromTable(
            \Ticket_User::getTable(),
            [
                'SELECT' => '*',
                'WHERE' => [
                    'tickets_id' => $ticket_id,
                    'use_notification' => 1
                ]
            ]
        ))->isEqualTo($expected);
```

Rom1-B (changes requested):

tests/functional/Ticket.php
RomainLvr: Applied : https://github.com/glpi-project/glpi/pull/16070/commits/03f6105c12ac4734a1ba5b0afc91d25565dfaf58

src/CommonITILObject.php
RomainLvr: The code has been retablished by keeping `$updated[]` line. https://github.com/glpi-project/glpi/pull/16070/commits/03f6105c12ac4734a1ba5b0afc91d25565dfaf58

src/User.php
RomainLvr: On some versions of PHP and mariaDB, this code would generate an error on users where the id was equal to an empty string. Using the `empty()` function avoids this error.

tests/functional/Ticket.php
RomainLvr: This code has been deleted because another block of code does almost the same thing right after it

Rom1-B (approved):
Seems OK

AdrienClairembault (changes requested):
I have a weird issue if there is multiple anonymous requester:

image

If I try to disable notifications on either of them, they become identical:

image

tests/functional/Ticket.php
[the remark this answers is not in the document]
AdrienClairembault: I think what @cedric-anne meant is that checking the actors are correct after the ticket was updated is probably enough and we don't need to tests the notifications themselves here.

The idea is to keep the test lightweight and focus and the direct issue that we are trying to fix (= the inability the disable notification for an anonymous actor).

On another note, a good way to ensure a test case is relevant is to run it without its associated fix.
For example, I've checked out your new test case on my GLPI but I didn't apply the other changes :

image

If the test was relevant to the original issue then it would fail here since I didn't apply the fix ;)

Could you look into it ?
Also it would be perfect if the tests also covered the special case where we have more than one anonymous requester :)

RomainLvr:
> I have a weird issue if there are multiple anonymous requester:
> 
> image
> 
> If I try to disable notifications on either of them, they become identical:
> 
> image

Issue fixed in https://github.com/glpi-project/glpi/pull/16070/commits/5cf9748d440c054462dde97dfd61081466035bbb. Thank :+1:

tests/functional/Ticket.php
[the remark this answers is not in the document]
RomainLvr: The test pass without applying the changes because you didn't run the test command on the right function. the function to test is `testUpdateActorsNotification` :)
And for the special case where we have more than one anonymous requester, it's already been tested [here](https://github.com/glpi-project/glpi/blob/5cf9748d440c054462dde97dfd61081466035bbb/tests/functional/Ticket.php#L749C9-L774C11).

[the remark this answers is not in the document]
AdrienClairembault: My bad, sorry about that.
Test is good 👍

AdrienClairembault (approved):

src/User.php
```diff
@@ -642,7 +642,7 @@ public function getFromDBbyEmail($email, $condition = [])
...
+        if (empty($this->fields['id'])) {
```
cedric-anne, line 645: The `isNewItem()` is the most appropriate for this.

```suggestion
        if ($this->isNewItem()) {
```

tests/functional/Ticket.php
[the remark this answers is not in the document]
cedric-anne: This test seems too complex (there is not need to check that notifications are sent, call to `$this->checkActors()` is enough).
Also, please add a case that contains more than one email actor for a given role, for instance, a test that creates a ticket with 3 email observers, and that would update the second one.

You could use a data provider method to be able to add many tests cases withour having to rewrite too many times the same logic.

cedric-anne (changes requested):

cedric-anne:
I updated tests to reuse the existing tests and updated tests data providers to check many combinaitions of inputs.

AdrienClairembault (approved):
Tested again and it still works, looks good to merge.

trasher (approved):

Rom1-B (approved):

PR #16091, opened 2023-11-30
[HLAPI] Assistance Stats

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

Add endpoints to retrieve assistance statistics that match Assistance > Statistics in the web UI.
To accomplish this, some of the existing statistics code was cleaned and moved from the `front` files to the `Stat` class.

### DIFF
--- a/front/report.dynamic.php
+++ b/front/report.dynamic.php
@@ -121,7 +121,7 @@ if (isset($_GET["display_type"])) {
                          );
                 }
             } else if (isset($_GET["type"]) && ($_GET["type"] == "hardwares")) {
-                Stat::showItems("", $_GET["date1"], $_GET["date2"], $_GET['start']);
+                Stat::showItems("", $_GET["date1"], $_GET["date2"], $_GET['start'], $_GET["itemtype"]);
             }
             break;
 
--- a/front/stat.item.php
+++ b/front/stat.item.php
@@ -68,7 +68,8 @@ if (!isset($_GET["start"])) {
 
 Stat::title();
 
-echo "<div class='center'><form method='post' name='form' action='stat.item.php'>";
+echo "<div class='center'><form method='get' name='form' action='stat.item.php'>";
+echo "<input type='hidden' name='itemtype' value='" . $_GET['itemtype'] . "'>";
 echo "<table class='tab_cadre'><tr class='tab_bg_2'>";
 echo "<td class='right'>" . __('Start date') . "</td><td>";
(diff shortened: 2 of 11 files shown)


orthagh (commented):
Could you rebase this one please ?

orthagh (changes requested):
image

Can we better document this part ?
Like providing a placeholder with actual date format for date fields
And as I needed to check in the code what "assistance type" means, could we have consistency with other endpoints ?

Also it's maybe a task for another PR, but it's very annoying to fill an itemtype parameter when you already are in a route where the itemtype is precised (ex "/Assistance/Stat/**Ticket**/Global"). It's a problem I saw multiple times (one infuriating is "POST /Assistance/Ticket" requiring itemtype "Ticket"). I understand why it's here technically, like generic endpoint, but this should be hidden for the API consumer. Or at least, a default correct value should be presented, but I think it's better to hide it completly.

cconard96:
> image
> 
> Can we better document this part ? Like providing a placeholder with actual date format for date fields And as I needed to check in the code what "assistance type" means, could we have consistency with other endpoints ?
> 
> Also it's maybe a task for another PR, but it's very annoying to fill an itemtype parameter when you already are in a route where the itemtype is precised (ex "/Assistance/Stat/**Ticket**/Global"). It's a problem I saw multiple times (one infuriating is "POST /Assistance/Ticket" requiring itemtype "Ticket"). I understand why it's here technically, like generic endpoint, but this should be hidden for the API consumer. Or at least, a default correct value should be presented, but I think it's better to hide it completly.

In this case, the `assistance_type` parameter shouldn't be shown (certainly not shown as required). It is an issue with the created OpenAPI schema after the endpoint was expanded (internally it is /Assistance/Stat/{assistance_type}/Global where assistance_type is a placeholder with a static list of possible values.
The same bug is the reason you need to fill the itemtype parameter in some other endpoints.
It doesn't affect usage outside of Swagger UI.

I'll start a new PR to fix the OpenAPI documentation generator.

orthagh:
Ok fine :) 
Just make the date parameters easier to fill and we should be good for this PR.

orthagh:
You can't force a placeholder ?

orthagh (approved):

cedric-anne (approved):

PR #16151, opened 2023-12-04
Interactive local test env

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

GLPI tests (specifically PHP code tests) can take a very long time to run. My previous answer for that was to manually edit the `test_tests-functional.sh` file locally to restrict tests to a specific directory or file. While that worked, it could be quicker if the docker containers didn't have to get recreated every time `tests/run_tests.sh` is run.

This PR adds a new "interactive" mode to the `run_tests.sh` script which can be activated with the `--interactive` flag. In interactive mode, all containers are started immediately and then it asks the user to enter one of several commands:
- test: Run the currently selected tests. By default, it gets this from the `run_tests.sh` arguments like it did before.
- settest: Change the currently selected test(s)
- setscope: Change the currently selected scope. This is the file or directory to run tests against if the test supports it. Currently, this only affects unit and functional tests. By default, this is "default" which means the default scope for the test (units/functional test folder)
- info: Displays the currently selected tests and scope
- help: Displays the command list
- exit: Stops the containers and exits

After tests are run in interactive mode, regardless of their success, the user is prompted for another command rather than exiting.
If you run tests again and the install was already run, it will not repeat the install tests unless they were explicitly requested.

Also, I tried to fix the LDAP fixture cleanup. When I tried running LDAP tests multiple times, it would fail to add `dc=glpi,dc=org` again and I think it is because the script was only running `ldapdelete` on leaf nodes. I changed it to do recursive deletes instead, but I am not 100% confident it was the best solution.

Examples:
- `tests/run_tests.sh functional`: Run functional tests in non-interactive mode. Same as before.
- `tests/run_tests.sh --interactive functional`: Run in interactive mode and select "functional" tests.
- `settest lint ldap` (inside interactive mode test runner): Select lint and ldap test(s)

### DIFF
--- a/.github/actions/init_initialize-ldap-fixtures.sh
+++ b/.github/actions/init_initialize-ldap-fixtures.sh
@@ -6,7 +6,11 @@ ROOT_DIR=$(readlink -f "$(dirname $0)/../..")
 for f in `ls $ROOT_DIR/tests/LDAP/ldif/*.ldif`; do
   # Delete all LDAP entries (in reverse ordre compared to creation)
   # but ignore errors as these entries may not exists.
-  (tac $f | grep -E '^dn:' | sed -E "s/^dn: (.*)$/\\1/" | docker-compose exec -T openldap ldapdelete -x -H ldap://127.0.0.1:3890/ -D "cn=Manager,dc=glpi,dc=org" -w insecure -c) \
+  #(tac $f | grep -E '^dn:' | sed -E "s/^dn: (.*)$/\\1/" | docker-compose exec -T openldap ldapdelete -x -H ldap://127.0.0.1:3890/ -D "cn=Manager,dc=glpi,dc=org" -w insecure -c) \
+  #|| true
+
+  # Recursively delete all LDAP entries
+  (cat $f | grep -E '^dn:' | sed -E "s/^dn: (.*)$/\\1/" | docker-compose exec -T openldap ldapdelete -x -r -H ldap://127.0.0.1:3890/ -D "cn=Manager,dc=glpi,dc=org" -w insecure -c) \
   || true
 done
 for f in `ls $ROOT_DIR/tests/LDAP/ldif/*.ldif`; do
--- a/.github/actions/init_initialize-ldap-fixtures.sh
+++ b/.github/actions/init_initialize-ldap-fixtures.sh
@@ -3,16 +3,9 @@ set -e -u -x -o pipefail
 
 ROOT_DIR=$(readlink -f "$(dirname $0)/../..")
 
-for f in `ls $ROOT_DIR/tests/LDAP/ldif/*.ldif`; do
-  # Delete all LDAP entries (in reverse ordre compared to creation)
-  # but ignore errors as these entries may not exists.
-  #(tac $f | grep -E '^dn:' | sed -E "s/^dn: (.*)$/\\1/" | docker-compose exec -T openldap ldapdelete -x -H ldap://127.0.0.1:3890/ -D "cn=Manager,dc=glpi,dc=org" -w insecure -c) \
-  #|| true
+# Recursively delete all LDAP entries
+docker-compose exec -T openldap ldapdelete -x -r -H ldap://127.0.0.1:3890/ -D "cn=Manager,dc=glpi,dc=org" -w insecure -c "dc=glpi,dc=org" || true
 
-  # Recursively delete all LDAP entries
-  (cat $f | grep -E '^dn:' | sed -E "s/^dn: (.*)$/\\1/" | docker-compose exec -T openldap ldapdelete -x -r -H ldap://127.0.0.1:3890/ -D "cn=Manager,dc=glpi,dc=org" -w insecure -c) \
-  || true
-done
 for f in `ls $ROOT_DIR/tests/LDAP/ldif/*.ldif`; do
   cat $f | docker-compose exec -T openldap ldapadd -x -H ldap://127.0.0.1:3890/ -D "cn=Manager,dc=glpi,dc=org" -w insecure
 done
--- a/.github/actions/test_tests-functional.sh
+++ b/.github/actions/test_tests-functional.sh
@@ -28,14 +28,17 @@ fi
 # Determine if scope is a file or a directory and set a scope type variable to 'd' or 'f' accordingly
 # scopes ending with a slash are considered as directories
 # Without a slash, if the file exists, it is considered as a file, otherwise as a directory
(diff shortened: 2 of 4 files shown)


AdrienClairembault:
> setscope: Change the currently selected scope. This is the file or directory to run tests against if the test supports it

Can it also support specifying the method name ? Atoum support it with like this: https://atoum.readthedocs.io/en/latest/running_tests.html#a-class-or-a-method

cconard96:
I added support for the `-m` option, but it doesn't work. I don't think it is my poor bash skills though since manually changing the test command line results in it testing nothing and declaring it a success.
```sh
vendor/bin/atoum \
  -p 'php -d memory_limit=512M' \
  --debug \
  --force-terminal \
  --use-dot-report \
  --bootstrap-file tests/bootstrap.php \
  --fail-if-void-methods \
  --fail-if-skipped-methods \
  $ATOUM_ADDITIONNAL_OPTIONS \
  --max-children-number 1 \
  -m tests\\units\\Alert::* \
```
`Success (0 test, 0/0 method, 0 void method, 0 skipped method, 0 assertion)!`

AdrienClairembault:
I also suffer from poor bash skills but this is how I implemented it on my `gtest` zsh function (code is simplified to only show the important options).

```sh
vendor/bin/atoum -bf tests/bootstrap.php -f $1 -m \*::$2
```

Which could then be used like this:

image

The results confirm only one method was executed.

AdrienClairembault (changes requested):
Regarding the feature itself, I think it is a great addition which would help running tests more efficiently.

However, in my opinion the usability could be improved as it feels a bit too "verbose" to configure your test because you need to run different commands like this:

image

I feel like it may be simpler to use if you could just require a test with a single command like this:

image

The `run` prefix could be added if we want to keep support for others distinct commands like "help" or "exit".

It would also need up/back arrow support to reload previous command like a real terminal, which would allow to re-run the previous tests easily.

I realize this may require additional work and perhaps complicated bash code (which I am terrible at and probably can't help with) but I really think it would make the feature more convenient to use.

All of this is subjective so it would be interesting to hear everyone opinions on this (which may differ from mine).

@trasher @orthagh @cedric-anne

orthagh:
I tested quickly the feature and it's promising.
It's rough for the moment, I needed to think more than 2s :p

Some remarks (I know some apply to the non interactive version present before), just a list a thing I noted:
- `docker compose` runs with `--quiet` option. Not against it, but it's almost the only command in quiet mode.
- Vocabulary: "test" when using the `settest` option may be renamed to "action" (this matches github)
- Scope and test/action may beneficiates from example and also, if possible, completions
- like said by @AdrienClairembault a way to target a specific method (`testsingle xxx/xxx::xxx`) and a up arrow support to run the previous command.
- we may need a test/action "all" selected by default. In comparison, if no scope defined, it considers everything.
- when prompt is waiting "GLPI Tests >", a recall of the `info` option (just before) may be useful.

cconard96:
Some changes made so far:
1. `settest`, `setscope`, and `test` commands are now `set action`, `set scope` and `run` respectively.
2. The `run` command now supports arguments for a single action and scope (scope defaults to the selected one if not specified) which affect the single test/action run.
3. `show_info` is run before every prompt for command input to display the currently selected actions and scope.

AdrienClairembault:
>     2. The `run` command now supports arguments for a single action and scope (scope defaults to the selected one if not specified) which affect the single test/action run.

I understand the action is the tests suite name (`units`, `functional`, ...) but what is the expected value for the scope ?
The full path to the file ?

>     3. `show_info` is run before every prompt for command input to display the currently selected actions and scope.

Does not seems to work on my side.
Sometimes I have some line with the details before my action but most of the time I don't.
Maybe it should always be displayed directly in the prompt ?
Something like `GLPI Tests (action) (scope) >` ?

Also sometimes the wrong action is executed:

image

cedric-anne:
I made some modifications to split scope and methods patterns. Indeed, one may want to use both of them at the same time (e.g. all `testPrepare` methods in `tests/functional/Glpi/Inventory` directory).

I fixed some issues with backslashes on test patterns.

Use case using `set` commands:
```
set action functional
set scope tests/functional/Glpi/Agent/Communication/Headers/Common.php
set methods tests\units\Glpi\Agent\Communication\Headers\Common::testHeaders
run
```

Use case using a single command:
```
run functional tests/functional/Glpi/Inventory *::testPrepare
```

cedric-anne (approved):

AdrienClairembault (changes requested):
Everything is OK for me expect the lack of up/down arrow support.

It seems it can easily be added like that (I don't fully understand how it works tho...):

```diff
diff --git a/tests/run_tests.sh b/tests/run_tests.sh
index ae7885b794..e4f5c7d453 100755
--- a/tests/run_tests.sh
+++ b/tests/run_tests.sh
@@ -364,7 +364,8 @@ if [[ "$INTERACTIVE" = true ]]; then
     SCOPE=$SELECTED_SCOPE
     METHODS=$SELECTED_METHODS
     show_info
-    read -e -r -p "GLPI Tests > " CHOICE
+    while read -r -e -d $'\n' -p 'GLPI Tests > ' CHOICE; do
+      history -s "$CHOICE"
       if [[ "$CHOICE" = run* ]]; then
         # If actions were specified after the run choice, use them instead of the ones selected
         if [[ "$CHOICE" != "run" ]]; then
@@ -414,6 +415,7 @@ EOF
       fi
       CHOICE=""
     done
+  done
 fi

 cleanup_and_exit
```

cedric-anne (changes requested):
Since addtition of history, the following usage does not work anymore:
image

cconard96:
> Since addtition of history, the following usage does not work anymore: ![image](https://private-user-
The commit was reverted. If someone else wants to try adding support, they can. Although, bash seems fairly limited in what we can support with tab-completion and history.

cedric-anne:
> > Since addtition of history, the following usage does not work anymore:
> 
> The commit was reverted. If someone else wants to try adding support, they can. Although, bash seems fairly limited in what we can support with tab-completion and history.

I guess the issue is due to position of variables declaration/affectation. If the fix is simple to do, it could be done in another PR, but, honestly, I think it is getting a too complex script for our bash skills. If this script is meant to get many enhancements in the future, maybe it should be transformed into a PHP `symfony/console` command.

cedric-anne (approved):

PR #16156, opened 2023-12-05
Open one PR from dependabot instead of several

trasher:
See https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/customizing-dependency-updates#grouping-dependabot-version-updates-into-one-pull-request

I almost understand anything; and I have no idea how to check if that configuration is valid :/


trasher:
I've added a validation script... Which tells us everything I made is wrong \o/

PR #16173, opened 2023-12-08
Fix to use getPendings on QueueNotification 

fgendorf:
Fix to use **getPendings** on QueueNotification in the same time of create a message to notify

Example: There are a plugin from TICGAL(Golden Partner) called MFA that send an email with OTP (one time password) by GLPI notifications and the function **forceSendFor** to process the queue, but the queue is not processed because getPendings filter are '<' and in the moment of **forceSendFor** is called the time is '=' of time registered, so the filter must be '<='

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no

### DIFF
--- a/src/QueuedNotification.php
+++ b/src/QueuedNotification.php
@@ -533,7 +533,7 @@ class QueuedNotification extends CommonDBTM
             'WHERE'  => [
                 'is_deleted'   => 0,
                 'mode'         => 'TOFILL',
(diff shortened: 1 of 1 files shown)


trasher (approved):
I have no idea of the impact; but that seems OK.

PR #16175, opened 2023-12-11
CSS Compile Improvements

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

- Change palettes to use CSS variables rather than SCSS variables.
- Remove all imports of tabler/bootstrap/main GLPI CSS from palettes
- Change GLPI CSS customizations to use the CSS variables rather than the SCSS variables (ex: --tblr-primary instead of $primary).
- Upgrade Tabler - Required to do this now as more recent updates finally made progress towards using the CSS variables whenever possible rather than SCSS variables.

Extra:
- Core palettes are now combined into a single CSS file and can be dynamically switched simply by changing the existing attributes on the root HTML element. Changing the palette selection in the user preferences can now let the user see a preview of the theme by changing these attributes. This more importantly reduces the amount of compiling needed for the palettes.
- To make UI work easier, a new debug widget was added to be able to change the palette/theme for the current page.

These changes will greatly reduce the amount of time it takes to compile the CSS.
Execution time of `bin/console build:compile_scss`:
- `main` branch: 6 minutes
- This PR so far: 30 seconds (12x faster for me)


cconard96:
While there are other improvements/changes that can be made, this PR should be in a usable state. Other changes could be done in a later PR.

To make reviewing these changes easier, and for future UI work, there is a new debug bar widget to switch the palette used for the current page.

orthagh (changes requested):
Firstly: thank you !
It's very good and so far, I didn't noticed anything ugly.
AND performances are now returning to normal.

I check the changes in large, but I'll do a more in depth review later.
For a start, Here is issues I noted by only checking the UI, some may be present before:

- General:
	- non logged pages (login, forgot password, front/helpdesk.php, faq) doesn't get any styles applied
	- `.search_page .search-container .search-card.card { border: unset; }` -> I don't know since when we have that, but it breaks card style strangely.
	- body background colors seems to have changed (at least on auror and dark auror) and doesn't make the subtile contrast with cards
	- counters in savesearches list panel doesn't have a good contrast

- Light themes:
	- plain primary button foreground is light where it was dark
	- tabs (in item's form) lost their light contrast compared to body
	- counters for itil objects right side (accordion headers) are dark on dark
	- horizontal tabs (dashboard, preferences) miss a border on the right

- Dark themes:
	- the pending reason switch in a new followup is not aligned vertically
	- tasks background in timeline for dark themes is too bright, it was a transparent yellow before
	- modals (share dashboard, goto menu, etc) are "transparentish", and we see behind them
	
- Auror:
	- breadcrumbs links were greyish before
	
- Dark auror:
	- tabs background seems of the theme

cconard96:
>       * `.search_page .search-container .search-card.card { border: unset; }` -> I don't know since when we have that, but it breaks card style strangely.
Since the PR to use the search engine to display consumables.
Comment in the scss:
```
Borders are unreliable here as they will scroll out of sight
due to the sticky header behavior
Its better to move it to a parent item that wont scroll like
.search-container
```

> 
>     * Light themes:
>       * horizontal tabs (dashboard, preferences) miss a border on the right
Same in 10.0. Fixed here.

>     * Dark themes:
>       
>       * the pending reason switch in a new followup is not aligned vertically
Seems the same between light and dark themes.

orthagh:
> Seems the same between light and dark themes.

yes my bad, I started a single list, saw it grows, decide to segment it, forgot this point

orthagh:
I saw another thing: right panel of itil object is now white (no contrast with content on center), in fact same issues as left tabs/

cconard96:
> I saw another thing: right panel of itil object is now white (no contrast with content on center), in fact same issues as left tabs

I honestly cannot tell the difference between 10.0 and this PR.

orthagh:
> I honestly cannot tell the difference between 10.0 and this PR.

It's subtile indeed but here is a comparison with auror theme (I guess all light themes are impacted:

-10.0.x
image

- your pr:
image


It's the same also with the left tabs and table headers/footers (in computer list for example).
For the last table component, tabler website show the header/footer with the subtile light contrast color.


A few additional issues:

- [ ] I would like to see a main rebase as #15861 has been merged
- [ ] entity badges in search are darker than before, nothing bad, but it got less contrast between fg and bg
- [ ] tab's counters style changed, was like this before:
```css
background-color: #d0d9f0;
color: #3a4966;
```

cconard96:
There were several cases in the CSS where a color was modified by just 1 or 2%. Does it really make sense like in the case of the ticket timeline for the card footer (the right panel) to be 1.5% darker than the card and the tabs to be 1.5% darker than the surface/card background? I'm sure some people would notice the difference if they tried, but I don't understand the functionality of it.

orthagh:
Making a new pass on this pr. I have time this afternoon, feel free to ping me.

> Since the PR to use the search engine to display consumables.

since #15280 so.
In the case of the above pr, it indeed displays a double border when removing this rule
After searching a little there are some variables not defined in case of full list (contrary to a sub list)
image

So there is a bug here. The full list should have a border somehow (and the sub list only one, to avoid double border)

---

Tinymce background seems broken again with dark palettes:
image
This happens often, I wonder if we can somehow test it.

---

>  I'm sure some people would notice the difference if they tried, but I don't understand the functionality of it.

I know we are not all equals about colors perceptions and, for some, contrast of this magnitude is 
useless. At least for me, I see it, and it's clearly personal as I did most of the design choice in 10.0.x.
Anyway, in case of ticket, intention was to make in evidence the content in the center, the timeline. To achieve that, I contrasted (a little) the secondary panels.
It's hard explain, but with all panels white, I got difficulties to see in one glance the position of each. They somehow merge and are blurred. I need to concentrate my vision/focus to get the zone I wanted to find. It may be force of habits in fact. A good example could be actors part:
image
I need to take my time to start distinguish things here.

If I had to detail last things hurting me:

- [ ] tabs background and counters in light palette (they are fine in dark ones)
- [ ] right panel of itil object
- [ ] table headers (just the th, note the above card header) -> return to the default of tabler (--tblr-bg-surface-tertiary resolving to --tblr-gray-50 in light mode, but the problem exists also in dark themes)

About table, I'm fine card headers/footer becoming white.

orthagh:
Did a bunch of commits on my fork: https://github.com/cconard96/glpi/compare/enhance/tabler_upgrade...orthagh:glpi:enhance/tabler_upgrade if you want to integrate them

Last things bothering me a little (and we may merge) I can't fix:

- bg of tinymce toolbars in dark themes
- ticket font size seems a little bigger, nothing much and we may pass, but consequence is less accordion items are displayed by default.

cconard96:
>     * bg of tinymce toolbars in dark themes
I don't see this issue. You do have to properly change your theme to Dark Auror or Midnight though. The debug tool doesn't reload the editor with the new skin/css to match the current page theme.

> ticket font size seems a little bigger, nothing much and we may pass, but consequence is less accordion items are displayed by default.

Font size and weight seem to match the only difference I saw was that with the new version of Tabler, the text inside accordions has its color set to --tblr-accordion-color which defaults to the body text color. In 10.0, the color was hard-coded as a gray color by Bootstrap. Changing the var to be --tblr-muted brings it close (at least to me) to what it was before.

css/includes/pages/_search.scss
```diff
@@ -111,14 +108,6 @@
         .search-card {
             display: grid;
```
trasher, line 110: To fix CS:
```suggestion
```

cconard96:
One last commit was added to make the TinyMCE theme hot-swappable using the debug tool.

cedric-anne (approved):
The them preview (from the preferences page or from the debug bar) is not working for custom themes. Anyway, the custom theme is still applied once its selection have been confirm by saving preferences, so it is OK.

To permit custom theme preview, it would be necessary to load all of them. For the preference page, it is a localized loading, so it could be OK, but for the debug bar, we would have to load all custom themes in all pages when it is active, and I am not sure it is a good idea. If we cannot preview custom themes from the debug bar context, we should not list them in the list, unless for a custom them that would be the current one.
This point should be addressed in another PR, to be able to merge this one ASAP.

Example of custom theme:
```scss
/* files/_themes/test_palette.scss */
:root[data-glpi-theme="test_palette"] {
    --glpi-mainmenu-bg: green;
    --glpi-mainmenu-fg: black;
}
```

orthagh (approved):

AdrienClairembault:
It seems this PR break tinymce's.
It only works if debug mode is enabled.

cconard96:
> It seems this PR break tinymce. It only works if debug mode is enabled.

should be fixed in #16264

AdrienClairembault:
I confirm the new PR fix the issue 👍

PR #16199, opened 2023-12-14
Fix login redirect to updatepassword

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

### DIFF
--- a/front/updatepassword.php
+++ b/front/updatepassword.php
@@ -33,6 +33,8 @@
  * ---------------------------------------------------------------------
  */
 
+$SECURITY_STRATEGY = 'no_check';
+
 /** @var array $CFG_GLPI */
 global $CFG_GLPI;
 


trasher (approved):

orthagh (approved):
Strange we didn't have any issue regarding this.
How did you detect it @cconard96 ?

cedric-anne (approved):

jcervantes-sipecom:
Hi @cconard96, thank you for this fix, today I just will open a ticket for this bug, but I research and found this fix. 

Usually, we use LDAP so the expiration is not managed by GLPI, but today I was testing with internal users, without your fix the redirection doesn't work when password is already expired:

image

With your fix, the redirection work perfectly:

image

PR #16201, opened 2023-12-15
SQL query improvements in CommonITILObject for better performance in huge glpi_documents_items table

ratado:
We have over 1.2 million rows in the 'glpi_documents_items' table within our GLPI database. Following an upgrade to GLPI 10, we've noticed a significant decrease in performance when loading tickets. This slowdown has even affected the speed of mail notifications.

Upon conducting an analysis, we pinpointed the cause to the query generated by the 'getAssociatesDocumentsCriteria' function in the 'CommonItilObject.php' file. This query takes over 2 seconds to execute (as shown below) due to the utilization of subqueries with the 'IN' condition.

sql_performance_before_changes_001

In the MySQL community, there are discussions advising against the use of subqueries in the 'IN' condition, especially for large databases, as it can prompt the database system to conduct a table scan.

As a result, we've submitted a pull request that essentially pre-executes the subqueries and retrieves their results as arrays to be utilized in the main query. This modification significantly enhanced the performance of loading tickets, reducing the average time from 1792 ms to 6 ms

 
sql_performance_after_changes_001

Best regards. 

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #number


orthagh (approved):

src/CommonITILObject.php
```diff
@@ -7889,6 +7889,7 @@ public function getForbiddenSingleMassiveActions()
     public function getAssociatedDocumentsCriteria($bypass_rights = false): array
     {
         $task_class = $this->getType() . 'Task';
+        global $DB; // Used to get subquery results - better performance
```
trasher, line 7892:
```suggestion
        /** @var DBMysql $DB */
        global $DB; // Used to get subquery results - better performance
```

src/CommonITILObject.php
```diff
@@ -7909,50 +7910,60 @@ public function getAssociatedDocumentsCriteria($bypass_rights = false): array
                     'OR' => ['is_private' => 0, 'users_id' => Session::getLoginUserID()],
                 ];
             }
-            $or_crits[] = [
-                Document_Item::getTableField('itemtype') => ITILFollowup::getType(),
-                Document_Item::getTableField('items_id') => new QuerySubQuery(
-                    [
+
```
trasher, line 7913:
```suggestion

```

cedric-anne (approved):

PR #16205, opened 2023-12-15
fix:(export): &nbsp; in CSV

Rom1-B:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !30871

When exporting to CSV, some empty data appeared with &nbsp;

Example with the User > Computer Emails column:
image

image

### DIFF
--- a/src/Search.php
+++ b/src/Search.php
@@ -8757,7 +8757,7 @@ HTML;
                     $value = preg_replace('/' . self::LBBR . '/', '<br>', $value);
                     $value = preg_replace('/' . self::LBHR . '/', '<hr>', $value);
                     $value = '<div class="fup-popup">' . $value . '</div>';
-                    $valTip = '' . Html::showToolTip(
+                    $valTip = ' ' . Html::showToolTip(
                         $value,
                         [
                             'awesome-class'   => 'fa-comments',
--- a/src/Search.php
+++ b/src/Search.php
@@ -6887,7 +6887,7 @@ JAVASCRIPT;
                         return "<img class='middle' alt='' src='" . $CFG_GLPI["typedoc_icon_dir"] . "/" .
                            $data[$ID][0]['name'] . "'>";
                     }
-                    return "&nbsp;";
+                    return '';
 
                 case "glpi_documents.filename":
                     $doc = new Document();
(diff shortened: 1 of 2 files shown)


cconard96:
Does anyone know why the search engine returns `&nbsp;` for some empty results to begin with? In some cases, it is used to add space between an icon and text, but having it on its own doesn't make sense to me.

trasher:
I agree with Curtis; proposed fix seems not correct; no unbreakable space should been present here.

Rom1-B:
It seems deliberate, but the reason escapes me too?

image

trasher:
There are several cases in Search class that return only `&nbsp;` in both `gitveItem()` and `showItem()` methods. I guess they should return empty strings.
I'm not sure since Search class is not easy to understand, but I do not see any reason for that; maybe is this just an old hack kept in place.

cconard96:
The use of &nbsp; has existed for 14 years, since at least the origin of the search class. It's hard to tell what the reasoning was.

trasher:
Indeed it's been present since the origin of the class.

`nbsp` is only for display; it should really not affect the code.
Search results are mainly used in display, which should nowadays not really be affected. It's also probably used from API, in which `nbsp` as empty data makes no sense.

I'd vote to replace with empty string.

src/Search.php
```diff
@@ -8757,7 +8757,7 @@ public static function showItem($type, $value, &$num, $row, $extraparam = '')
                     $value = preg_replace('/' . self::LBBR . '/', '<br>', $value);
                     $value = preg_replace('/' . self::LBHR . '/', '<hr>', $value);
                     $value = '<div class="fup-popup">' . $value . '</div>';
-                    $valTip = "&nbsp;" . Html::showToolTip(
+                    $valTip = '' . Html::showToolTip(
```
trasher, line 8760: In most of otehr usages in the same class, `showTooltips()` results are separated with a simple space.
```suggestion
                    $valTip = ' ' . Html::showToolTip(
```

cconard96, line 8760: I think this case is "valid" as it is used as a spacer and shouldn't ever be result in `&nbsp;` being returned on its own.

trasher: (also, this is not a really result data and should not be present in exports or API results I guess)

trasher (approved):

PR #16210, opened 2023-12-17
Enhance password expiration message

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

When impersonating a user who has never had their password changed, and when the password expiration policy is set, the warning on the user's homepage says the password will expire in 1970. It also shows a debug error because 'password_last_update' is null and passing null to strtotime is deprecated. This PR prevents passing null to strtotime and also changes the message to "Your password has expired." when the expiration date has passed.

### DIFF
--- a/src/Central.php
+++ b/src/Central.php
@@ -457,11 +457,8 @@ class Central extends CommonGLPI
 
         $user = new User();
         $user->getFromDB(Session::getLoginUserID());
-        if ($user->fields['authtype'] == Auth::DB_GLPI && $user->shouldChangePassword()) {
-            $expiration_msg = sprintf(
-                __('Your password will expire on %s.'),
-                Html::convDateTime(date('Y-m-d H:i:s', $user->getPasswordExpirationTime()))
-            );
+        $expiration_msg = $user->getPasswordExpirationMessage();
+        if ($expiration_msg !== null) {
             $messages['warnings'][] = $expiration_msg
              . ' '
              . '<a href="' . $CFG_GLPI['root_doc'] . '/front/updatepassword.php">'
--- a/src/User.php
+++ b/src/User.php
@@ -6365,6 +6365,13 @@ HTML;
             return null;
         }
 
+        if (null === $this->fields['password_last_update']) {
+            // password never updated
+            return strtotime(
+                '+ ' . $expiration_delay . ' days',
+                strtotime($this->fields['date_creation'])
+            );
+        }
         return strtotime(
             '+ ' . $expiration_delay . ' days',
             strtotime($this->fields['password_last_update'])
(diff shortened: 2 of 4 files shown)


src/User.php
```diff
@@ -6341,6 +6341,9 @@ public function getPasswordExpirationTime()
             return null;
         }
 
+        if (null === $this->fields['password_last_update']) {
```
trasher, line 6344: This probably need a test case

src/Central.php
```diff
@@ -458,10 +458,16 @@ private static function getMessages(): array
         $user = new User();
         $user->getFromDB(Session::getLoginUserID());
         if ($user->fields['authtype'] == Auth::DB_GLPI && $user->shouldChangePassword()) {
-            $expiration_msg = sprintf(
-                __('Your password will expire on %s.'),
-                Html::convDateTime(date('Y-m-d H:i:s', $user->getPasswordExpirationTime()))
-            );
+            $expire_time = $user->getPasswordExpirationTime();
```
trasher, line 461: Maybe a method in `User` would be better? As far as I see, code is exactly the same in front page and here

trasher (approved):

src/User.php
```diff
@@ -6392,6 +6395,26 @@ public function hasPasswordExpired()
         return $expiration_time < time();
     }
 
+    public function getPasswordExpirationMessage(): string|null
```
trasher, line 6398:
```suggestion
    public function getPasswordExpirationMessage(): ?string
```

tests/functional/User.php
```diff
@@ -983,14 +983,22 @@ protected function passwordExpirationMethodsProvider()
...
+            ],
         ];
     }
 
     /**
      * @dataProvider passwordExpirationMethodsProvider
      */
     public function testPasswordExpirationMethods(
-        string $last_update,
+        string|null $last_update,
```
trasher, line 1001:
```suggestion
        ?string $last_update,
```

src/User.php
```diff
@@ -6341,6 +6341,9 @@ public function getPasswordExpirationTime()
             return null;
         }
 
+        if (null === $this->fields['password_last_update']) {
+            return $expiration_delay * DAY_TIMESTAMP; // password never updated
```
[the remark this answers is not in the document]
cconard96: > This case should not happen as the field is supposed to be always set when the password expiration policy is activated.

There seems to be several cases where that isn't the case. None of the users I added with the data injection plugin have this field set and neither do the LDAP users I have.

cedric-anne (approved):

PR #16216, opened 2023-12-18
(Fix) Fix the impossibility of locking the category field in software

Lainow:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !30882

It was impossible to lock the Category field of the Software class, which caused it to be rewritten during the software inventory by the Agent, even if it had already been modified manually. 
image

To solve this problem, I added the is_dynamic field to the softwares table, and if a software has already been instantiated dynamically, the is_dynamic value is set to 1.


src/Software.php
```diff
@@ -78,6 +78,14 @@ public static function getMenuShorcut()
         return 's';
     }
 
+    public function isDynamic()
+    {
+        if ($this->fields['is_dynamic'] != null && $this->fields['id'] != 0) {
```
stonebuzz, line 83:
```suggestion
        if (isset($this->fields['is_dynamic']) && $this->fields['is_dynamic'] && !$this->isNewItem()) {
```

stonebuzz (approved):

src/Software.php
trasher: I'm not sure base CommonDBTM would not be enough. Why an override is required?

stonebuzz: Indeed, That was before we added the base field

src/Inventory/Asset/Software.php
```diff
@@ -36,6 +36,7 @@
 
 namespace Glpi\Inventory\Asset;
 
+use atoum\atoum\scripts\treemap\analyzers\size;
```
trasher, line 39:
```suggestion
```

src/Inventory/Asset/Software.php
```diff
@@ -726,7 +733,7 @@ private function storeSoftware()
                 //set create date
                 $stmt_columns['date_creation'] = $_SESSION["glpi_currenttime"];
 
-                if ($stmt === null) {
+                if ($stmt === null || $stmt_types === null || strlen($stmt_types) != count($stmt_columns)) {
```
trasher, line 736: I think it's wrong to silentely check columns count. If columns count is wrong, there is a bug that must be fixed.

stonebuzz:
@Lainow can you rebase (target GLPI version is ```10.0.11-dev``` instead of ```10.0.12-dev``` )

trasher:
See internal discussion; we're not going to propose this change because of implications it can have; and the fact is was not present in old plugin.

PR #16224, opened 2023-12-20
Skip unneeded LDAP auth checks

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #15262

When logging in through the API with a user_token instead of a username/password, the authentication code was trying to connect to every active LDAP server even if the user was not from an LDAP. When the user is from an LDAP and has an `auths_id` set, then only that one server is checked. When `auths_id` is 0 or missing, it checks every server. This code should first check the `authtype` before anything else related to LDAP.

### DIFF
--- a/src/Auth.php
+++ b/src/Auth.php
@@ -825,7 +825,7 @@ class Auth extends CommonGLPI
                 $ldapservers = [];
                 $ldapservers_status = false;
                //if LDAP enabled too, get user's infos from LDAP
-                if (Toolbox::canUseLdap()) {
+                if ($this->user->fields['authtype'] === self::LDAP && Toolbox::canUseLdap()) {
                    //User has already authenticated, at least once: it's ldap server if filled
                     if (
                         isset($this->user->fields["auths_id"])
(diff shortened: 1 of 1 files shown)


cedric-anne (approved):
Seems OK.

By the way, I recently figured out that I have a LDAP warning (I have a test LDAP that is not always started) when the "remember_me" cookie is used to restore my session after a long time of inactivity.

As far as I understand this piece of code, this issue will be fixed too.

cedric-anne (changes requested):
Actually, it seems that you have to define manually `$this->user->fields['authtype']` in `Auth::getAlternateAuthSystemsUserLogin()` (the same way it is done for `$this->user->fields['name']`.

Not sure of the side-effects.

orthagh:
> Actually, it seems that fetching LDAP information is a wanted behaviour for a user that nevers connects before. For instance, if it connects through SSO, we still want to fetch more information from the LDAP.

I confirm this

cedric-anne (approved):

trasher (approved):

PR #16230, opened 2023-12-20
fix(Contract_Item): harmonize behavior from Contract or Item form

stonebuzz:
From an ```item``` with ```Contract``` tab it is not possible to add an "expired" contract 

But from an "expired" contract, any item can be added.

This PR  harmonize behavior from Contract or Item form

| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !30934

### DIFF
--- a/src/Contract_Item.php
+++ b/src/Contract_Item.php
@@ -303,8 +303,7 @@ class Contract_Item extends CommonDBRelation
 
             echo "<tr class='tab_bg_1'><td>";
             Contract::dropdown(['entity'  => $item->getEntityID(),
-                'used'    => $used,
-                'expired' => false
+                'used'    => $used
             ]);
 
             echo "</td><td class='center'>";


cedric-anne (commented):
I am pretty sure that someone will later ask if it is possible to not have expired contracts in this list.

Adding items to an expired contract, from the contract page, is not a problem, but having everywhere a list of contracts polluted by expired contract may be a pain.

stonebuzz:
> I am pretty sure that someone will later ask if it is possible to not have expired contracts in this list.
> 
> Adding items to an expired contract, from the contract page, is not a problem, but having everywhere a list of contracts polluted by expired contract may be a pain.

completely agree with this, I'm closing this PR

PR #16235, opened 2023-12-21
Fix(Contract): use standard Location search options

### LINKED ISSUE #16231: [GLPI 10.0.11] SQL Errors on map into Contract list page
### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

- [x] I have searched the existing issues

### Version

GLPI 10.0.11

### Bug description

image

### Relevant log output

_No response_

### Page URL

_No response_

### Steps To reproduce

_No response_

_No response_

### Anything else?

_No response_

stonebuzz:
Add missing standard Location search options to ```Contract```

Move ```Number``` ```searchoption``` with index ```3``` to ```9``` (because index ```3``` is used by ```Location / completename```

Move old  ```Location / completename``` with index ```8``` to ```3```

Fix #16231 

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #16231

### DIFF
--- a/install/migrations/update_10.0.11_to_10.0.12/contract.php
+++ b/install/migrations/update_10.0.11_to_10.0.12/contract.php
@@ -1,0 +1,41 @@
+<?php
+
+/**
+ * ---------------------------------------------------------------------
+ *
+ * GLPI - Gestionnaire Libre de Parc Informatique
+ *
+ * http://glpi-project.org
+ *
+ * @copyright 2015-2023 Teclib' and contributors.
+ * @copyright 2003-2014 by the INDEPNET Development Team.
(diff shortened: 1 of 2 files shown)


trasher:
I've already proposed a fix, see #16234

trasher:
Seems better to use a data migration, I'll close mine

trasher (approved):

cedric-anne (approved):

trasher:
Sounds like migration will cause issues finally... First, it may break external scripts relying on SO number (well... that's not a blocker); but re-running the migration will cause all ids being set to the same...

I see no solution, and as my alternative proposal should work fine, I close this one.

PR #16237, opened 2023-12-21
fix(ticket): watcher ignored on self-service form

Rom1-B:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !30913

On the native ticket creation form, in the simplified interface, the watcher was not taken into account.
So even if you filled it in, the created ticket didn't have it.

In the browser console, there was this error:
image

### DIFF
--- a/templates/components/itilobject/actors/field.html.twig
+++ b/templates/components/itilobject/actors/field.html.twig
@@ -269,27 +269,29 @@
           $('.popover').popover('hide');
       });
 
-       actor_select.parent().popover({
-           selector: '[data-glpi-popover-source]',
-           container: actor_select.parent(),
-           html: true,
-           sanitize: false,
-           trigger: 'hover',
-           delay: {
+      setTimeout((function() {
+         actor_select.parent().popover({
+            selector: '[data-glpi-popover-source]',
+            container: actor_select.parent(),
+            html: true,
+            sanitize: false,
+            trigger: 'hover',
+            delay: {
                hide: 300
-           },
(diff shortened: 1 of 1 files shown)


templates/components/itilobject/actors/field.html.twig
```diff
@@ -269,27 +269,29 @@
...
       });
 
-       actor_select.parent().popover({
-           selector: '[data-glpi-popover-source]',
-           container: actor_select.parent(),
-           html: true,
-           sanitize: false,
-           trigger: 'hover',
-           delay: {
+      setTimeout((function() {
```
AdrienClairembault, line 272:
```suggestion
      $(function () {
```
Waiting for the document to be ready should be more reliable than an arbitrary 500ms delay.

templates/components/itilobject/actors/field.html.twig
```diff
@@ -269,27 +269,29 @@
...
+            }
+         }).on('hide.bs.popover', function () {
             // prevent closing popover when user card is hover
             if ($('.user-info-card:hover').length > 0) {
-                return false;
+               return false;
             }
-        });
+         });
+      }), 500);
```
AdrienClairembault, line 294:
```suggestion
      });
```

AdrienClairembault (changes requested):

AdrienClairembault (approved):

orthagh (approved):
Probably since #15957 which was to improve visually the speed of rendering the page.
The new change doesn't seem to impact as much as it was before the previous pr.
So ok for me

PR #16254, opened 2023-12-28
[HLAPI] File Uploads

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

### DIFF
--- a/src/Api/HL/Controller/AbstractController.php
+++ b/src/Api/HL/Controller/AbstractController.php
@@ -216,7 +216,7 @@ abstract class AbstractController
             return self::getCRUDErrorResponse(self::CRUD_ACTION_CREATE);
         }
 
-        return self::getItemLinkResponse($new_id, $api_path);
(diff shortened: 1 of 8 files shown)


Mirkk:
Hi! no news on this API? 
To avoid to need to implement a file upload, an Idea could also be to place the CSV files in a certain (hardcoded) directory, eg.g. _plugins/datainjection. The API just would need to trigger the import as long as there is no other official way to do it.
Maybe there is something in the new API, GLPI11?
Thanks!

dhrzic:
This would be quite helpful, an CLI command (with parameters) to schedule import through API pulling CSV file from the filesystem rather than manually importing files sounds better.

cconard96:
This version of the file management code for the HLAPI is no longer relevant. A replacement is already in progress with multipart form parsing for non-POST requests, and it will be done as a separate PR.

PR #16283, opened 2024-01-02
fix(printer-log): use fixed date for tests to prevent occasional failures

ccailly:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |

This PR addresses an issue where tests would occasionally fail, particularly at the start of the year, due to the predefined printer logs used for testing. 

Upon analysis of the method, it doesn't appear that there's an issue within it. It would be rather complex to define logs at dates that never cause problems. The simplest solution is to use a fixed date.

The changes include replacing the dynamic date generation with a fixed date in the session for the PrinterLog tests. This ensures that the tests are consistent and do not depend on the actual date when the tests are run.

### DIFF
--- a/src/PrinterLog.php
+++ b/src/PrinterLog.php
@@ -129,7 +129,7 @@ class PrinterLog extends CommonDBChild
         }
 
         if (!$start_date) {
-            $start_date = new DateTime();
+            $start_date = new DateTime(Session::getCurrentTime());
             $start_date->sub(new DateInterval($interval));
         }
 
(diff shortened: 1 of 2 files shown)


AdrienClairembault (approved):

orthagh (approved):

orthagh:
Early merge for unlocking failing tests

PR #16289, opened 2024-01-02
Reduce chance of CSRF token expiration on login page when session auto-expires

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #16010

In some cases, AJAX requests may be made automatically which perform a login check. If the session expired, the user would be automatically redirected to the login page which has CSRF protection. If the user was away from GLPI, but still had it open, and this happens, the CSRF token on the login page may have already expired. When they tried logging back in, they would get yet another error message that their session wasn't valid.

With these changes, when a `redirectToLogin` is performed, the error(s) are shown on a different screen with a button to "Log in again" than the login form. The idea is that GLPI can sit on the login error page without any CSRF token expiring and there is much less chance of a user clicking "Log in again" and then letting GLPI sit on the actual login page for 2 hours before logging in again.

I also adjusted 2 of the 3 error messages to make them more user friendly.

### DIFF
--- a/index.php
+++ b/index.php
@@ -116,11 +116,11 @@ if (!file_exists(GLPI_CONFIG_DIR . "/config_db.php")) {
                 break;
 
             case 2: // GLPI_SESSION_DIR not writable
-                $errors[] = __('Checking write permissions for session files');
+                $errors[] = __('Logins are not possible at this time. Please contact your administrator.');
                 break;
 
             case 3:
-                $errors[] = __('Invalid use of session ID');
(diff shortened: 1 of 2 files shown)


orthagh (approved):

trasher:
Tests are failing

trasher:
> Tests are failing

Does not seems related to current PR, maybe just rebase will fix

cedric-anne (approved):

PR #16290, opened 2024-01-02
Skip checking GLPI_SESSION_DIR if not using "files" save handler

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

If PHP is configured to use "redis" for example as the session save handler instead of the default of "files", the `is_writable` checks on GLPI_SESSION_DIR shouldn't be required and can be skipped. This PR also makes sure that the Status Checker returns a message along with an OK status for the session directory check to explain that the "files" handler is not being used.

### DIFF
--- a/src/Session.php
+++ b/src/Session.php
@@ -2159,4 +2159,15 @@ class Session
         // TODO (10.1 refactoring): replace references to $_SESSION['glpi_currenttime'] by a call to this function
         return $_SESSION['glpi_currenttime'] ?? null;
     }
+
+    /**
+     * Checks if the GLPI sessions directory can be written to if the PHP session save handler is set to "files".
+     * @return bool True if the directory is writable, or if the session save handler is not set to "files".
+     */
+    public static function canWriteSessionFiles(): bool
+    {
+        $session_handler = ini_get('session.save_handler');
+        return $session_handler !== false
+            && (strtolower($session_handler) !== 'files' || is_writable(GLPI_SESSION_DIR));
+    }
 }
--- a/src/System/Requirement/DirectoryWriteAccess.php
+++ b/src/System/Requirement/DirectoryWriteAccess.php
@@ -90,6 +90,9 @@ class DirectoryWriteAccess extends AbstractRequirement
                 break;
             case realpath(GLPI_SESSION_DIR):
                 $title = __('Permissions for session files');
+                $session_handler = ini_get('session.save_handler');
+                $optional = $session_handler !== false && strtolower($session_handler) !== 'files';
+                $description = __('If you have "session.save_handler" set to something besides "files" in your php.ini, this is not required.');
                 break;
             case realpath(GLPI_TMP_DIR):
                 $title = __('Permissions for temporary files');
--- a/src/System/Status/StatusChecker.php
+++ b/src/System/Status/StatusChecker.php
@@ -518,19 +518,24 @@ final class StatusChecker
                     'status' => self::STATUS_OK
                 ]
             ];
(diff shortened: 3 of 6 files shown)


orthagh (approved):

trasher (dismissed):

src/System/Status/StatusChecker.php
```diff
@@ -518,19 +518,23 @@ public static function getFilesystemStatus($public_only = true): array
...
-                    'status_msg'   => sprintf(_x('glpi_status', '%s variable is not a directory'), 'GLPI_SESSION_DIR')
-                ];
-                $status['status'] = self::STATUS_PROBLEM;
-            } else if (!is_writable(GLPI_SESSION_DIR)) {
-                $status['session_dir'] = [
-                    'status' => self::STATUS_PROBLEM,
-                    'status_msg'   => sprintf(_x('glpi_status', '%s variable is not writable'), 'GLPI_SESSION_DIR')
-                ];
-                $status['status'] = self::STATUS_PROBLEM;
+            if (strtolower(ini_get('session.save_handler')) === 'files') {
```
cedric-anne, line 521: In this line of code, you assume that `session.save_handler` `files` value may contains uppercase chars. If this is a potential case, this `strtolower()` operation should be used in all the other checks introduced in this PR.

cedric-anne (changes requested):

src/System/Requirement/DirectoryWriteAccess.php
```diff
@@ -94,6 +94,8 @@ public function __construct(string $path, bool $optional = false, ?string $descr
                 break;
             case realpath(GLPI_SESSION_DIR):
                 $this->title = __('Permissions for session files');
+                $this->optional = ini_get('session.save_handler') !== 'files';
```
cedric-anne, line 97:
```suggestion
                $session_handler = ini_get('session.save_handler');
                $this->optional = $session_handler !== false && (strtolower($session_handler) !== 'files';
```

src/Session.php
```diff
@@ -2155,4 +2155,13 @@ public static function getCurrentTime(): ?string
...
         return $_SESSION['glpi_currenttime'] ?? null;
     }
+
+    /**
+     * Checks if the GLPI sessions directory can be written to if the PHP session save handler is set to "files".
+     * @return bool True if the directory is writable, or if the session save handler is not set to "files".
+     */
+    public static function canWriteSessionFiles(): bool
+    {
+        return strtolower(ini_get("session.save_handler")) !== 'files' || is_writable(GLPI_SESSION_DIR);
```
cedric-anne, line 2165: According to the documentation, `ini_get()` will return `false` if configuration option does not exists. I guess it would be the case if the `session` extension is not enabled.
```suggestion
        $session_handler = ini_get('session.save_handler');
        return $session_handler !== false
            && (strtolower($session_handler) !== 'files' || is_writable(GLPI_SESSION_DIR));
```

src/System/Status/StatusChecker.php
cedric-anne, line 521:
```suggestion
            $session_handler = ini_get('session.save_handler');
            if ($session_handler !== false && strtolower($session_handler) === 'files') {
```

cedric-anne (approved):

trasher:
Tests are still failing

cedric-anne:
> Tests are still failing

There was a small mistake, fixed it.

image

PR #16300, opened 2024-01-04
Centralize guzzle client creation

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

Code that creates Guzzle clients all duplicates the proxy check and options building logic for the proxy. This PR creates a function that creates the Guzzle client with the proxy options set if a proxy is configured. This should make it easier for anything in the future that needs a client and easier in case we change proxy options or add new related options.

### DIFF
--- a/src/Toolbox.php
+++ b/src/Toolbox.php
@@ -40,6 +40,7 @@ use Glpi\Event;
 use Glpi\Http\Response;
 use Glpi\Mail\Protocol\ProtocolInterface;
 use Glpi\Rules\RulesManager;
+use GuzzleHttp\Client;
 use Laminas\Mail\Storage\AbstractStorage;
 use Mexitek\PHPColors\Color;
 use Psr\Log\LoggerInterface;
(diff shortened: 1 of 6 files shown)


orthagh (approved):

src/Toolbox.php
```diff
@@ -1279,6 +1280,26 @@ public static function getURLContent($url, &$msgerr = null, $rec = 0)
...
     }
 
+    /**
+     * Get a new Guzzle client with proxy if configured and the specified other options.
+     * @param array $extra_options Extra options to pass to the Guzzle client constructor
+     * @return Client Guzzle client
+     * @throws SodiumException
+     */
+    public static function getGuzzleClient(array $extra_options): Client
+    {
```
trasher, line 1290:
```suggestion
    {
        /** @var array $CFG_GLPI */
        global $CFG_GLPI;

```

PR #16306, opened 2024-01-04
Fix entity selector

AdrienClairembault:
The entity selector is never updated, meaning any new entities will never be visible.

This is caused by #15794, which did not take cache invalidation into account.

I've fixed this by adding a hash of every entities last update date into the selector.
This hash will change any time an entity is created, updated or deleted from the database.

I've also deleted references to the old cache key since it is not used anymore after #15794.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !31070

### DIFF
--- a/ajax/entitytreesons.php
+++ b/ajax/entitytreesons.php
@@ -55,22 +55,7 @@ if (Session::getCurrentInterface() == 'helpdesk') {
 
 $ancestors = getAncestorsOf('glpi_entities', $_SESSION['glpiactive_entity']);
 
-// Build cache key using last update date of every entity
-// This should trigger a cache refresh when an entity is created, updated or deleted
-global $DB;
-
-$data = $DB->request([
-    'SELECT' => 'date_mod',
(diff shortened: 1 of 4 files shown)


orthagh:
#16249 ?

cconard96:
Is this meant as a replacement for #16249?

AdrienClairembault:
My bad, I didn't know it was already solved by #16249.

AdrienClairembault:
I have no idea which approach is better, guess we can let @cedric-anne look into it we he comes back.

Probably a mix of both approach would be best as I understand that you implements additional improvements (like profiler support) while this PR add some tests which are probably needed on this feature.

AdrienClairembault:
Also, a third solution would be to revert https://github.com/glpi-project/glpi/pull/15794 and work on this on the main branch instead to avoid futures issues.

PR #16328, opened 2024-01-09
Budget: Add "is" and "is not" search type for "Value" SO

AdrienClairembault:
Before:
image

After: 
image

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  |no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !31063

### DIFF
--- a/src/Budget.php
+++ b/src/Budget.php
@@ -264,7 +264,8 @@ class Budget extends CommonDropdown
             'table'              => $this->getTable(),
             'field'              => 'value',
             'name'               => _x('price', 'Value'),
-            'datatype'           => 'decimal'
+            'datatype'           => 'decimal',
+            'searchtype'         => ['contains', 'notcontains', 'equals', 'notequals'],
         ];
 
         $tab[] = [


orthagh (approved):

src/Budget.php
```diff
@@ -264,7 +264,8 @@ public function rawSearchOptions()
             'table'              => $this->getTable(),
             'field'              => 'value',
             'name'               => _x('price', 'Value'),
-            'datatype'           => 'decimal'
+            'datatype'           => 'decimal',
```
orthagh, line 267:
```suggestion
            'datatype'           => 'number',
```

Could probably do the trick also.
Not sure on what 'decimal' differs from 'number.
Another possibility is to add the "case 'decimal'" to getActionFor method.

trasher:
Isn't taht a bit strange not having "is" and "is not" for "decimal" type?

src/Budget.php
cedric-anne: IMHO, the lack of `decimal` case in `getActionsFor` is not intentional. It could probably be added in GLPI 10.1.

Anayway, switching to `number` datatype here seems the most relevant option.

AdrienClairembault:
The 'decimal' type allow for easier searches as it does not check the decimal values when using `contains`.

image

With `number`, you would get no results for the same query:

image

I agree that we should change the default operator for `decimal` in main (add is/is not)

AdrienClairembault:
Rebased to fix tests.

cedric-anne (approved):

cedric-anne:
> I agree that we should change the default operators for `decimal` in main (add is/is not)

Could you create a PR for this ?

AdrienClairembault:
> > I agree that we should change the default operators for `decimal` in main (add is/is not)
> 
> Could you create a PR for this ?

Done #16332

trasher (approved):

PR #16344, opened 2024-01-10
Fix CAS authentication service_base_url

leManu:
when GLPI is not installed at the webserver's root phpcas receives an incorrect value for `service_base_url`, leading to a bad redirection after authentication. `path` and following data must be removed from `url_base` configuration value

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a

### DIFF
--- a/src/Auth.php
+++ b/src/Auth.php
@@ -544,12 +544,14 @@ class Auth extends CommonGLPI
                 } else {
                     // Starting from version 1.6.0, `$service_base_url` argument was added at 5th position, and `$changeSessionID`
                     // was moved at 6th position.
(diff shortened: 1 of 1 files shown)


cedric-anne (approved):
It seems OK, according to the PHPDoc of the lib: https://github.com/apereo/phpCAS/blob/master/source/CAS/Client.php#L921-L928

trasher (approved):
Seems indeed correct, but I cannot test

PR #16347, opened 2024-01-11
fix(change + problem): adding readonly on fields when ITIL object is not editable

Rom1-B:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !31117

When the user only has read-only access to changes or problems, the fields were not read-only, so he could modify the values, and click on the save or delete buttons. 
_Pressing these buttons displayed a message saying that the user was not authorized to do so._

### DIFF
--- a/src/CommonITILObject.php
+++ b/src/CommonITILObject.php
@@ -9911,4 +9911,27 @@ abstract class CommonITILObject extends CommonDBTM
         // All actions should be attached to thread instanciated by `new` event
         return 'new';
     }
+
+    /**
+     * Is the current user have right to update the current ITIL object ?
+     *
+     * @return boolean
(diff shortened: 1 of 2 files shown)


cedric-anne (approved):

trasher (commented):
Seems good, but I'm not sure about can* methods I'm not enough confident with. @orthagh can you review this one please?

orthagh (approved):
This impacts only change and problem, ticket has its own can[Update|Delete]Item and is not concerned.
Also new methods are a subset of what ticket already does.

So it's ok

PR #16353, opened 2024-01-11
Add screen capture feature

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

Integration of `screenshot` plugin. Unlike the plugin where the screenshot and recording buttons were separate timeline actions, these buttons now exist within the document item timeline form. That also means users can attach any number of screenshots, recordings, or uploaded files and they all get processed at once. Screenshots and recordings both have previews shown within the form and they can be removed by clicking the "x" on the preview or within the file upload control.

Selection_211

This feature only works within "secure contexts". Typically that means you connect over HTTPS or are connecting to `localhost`. If your browser doesn't consider your connection to be a secure context, the screenshot and recording options will be hidden. The browser wouldn't allow access to the required APIs in that case anyways. Further, if GLPI doesn't allow uploading the type of file used by one of the options, that option will not show. It uses PNG for screenshots and WEBM for recordings.

Also, the plugin had restricted screen recordings with a new permission. My reasoning was that the convenience of creating and adding a video, which would be a large file, may not be desired to be accessible to certain users.
**My question is where this permission would fit in. There are no document permissions within simplified interface profiles currently.**

### DIFF
--- a/js/modules/Screenshot/Screenshot.js
+++ b/js/modules/Screenshot/Screenshot.js
@@ -40,7 +40,7 @@ class Screenhot {
 
     /**
      * Listen on the file upload element to handle removal of previews when the file is removed.
-     * @param {HTMLElement} fileupload_element
+     * @param {HTMLElement} fileupload_element The file upload element to listen on.
      * */
     listenOnFileUpload(fileupload_element) {
         // Need to use a mutation observer as click handlers added after the upload is done are not triggered
@@ -74,6 +74,11 @@ class Screenhot {
         observer.observe(fileupload_element, {childList: true, subtree: true});
     }
 
+    /**
+     * Check if the browser supports and will allow the required APIs to be used.
+     * This includes a secure context check and a check that the user agent is not a mobile device.
+     * @return {boolean}
+     */
     isSupported() {
         // if not secure context, we can't use the screen capture API
         if (!window.isSecureContext) {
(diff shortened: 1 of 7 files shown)


orthagh (approved):
Finally after a deep clean of my caches, everything work as intended.

about 
> My question is where this permission would fit in. There are no document permissions within simplified interface profiles currently.

Imo, pictures uploading are not problematic at all (and already possible when pasting from clipboard), their size is usually ok.
about webm, as the feature requires enabling the extensions in dropdowns > filetypes, and disabled by default, I guess it's good as it.
I'm not in favor to add dedicated rights for the moment.

js/modules/Screenshot/Screenshot.js
```diff
@@ -0,0 +1,234 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ * ---------------------------------------------------------------------
+ */
+
+/* global GLPI_PLUGINS_PATH */
+class Screenhot {
```
cedric-anne, line 35: Multiple methods have no JSDoc. It may help to have it, especially when there are parameters.

templates/components/itilobject/timeline/form_document_item.html.twig
```diff
@@ -102,6 +102,26 @@
...
+                            <i class="ti ti-camera"></i>
+                            <span>{{ __('Screen recording') }}</span>
+                        </button>
+                    {% endif %}
+                </div>
+                <div id="screen_capture_preview" class="w-100">
+                    <div class="previews overflow-x-scroll my-2 d-flex px-2"></div>
+                    <button type="button" name="stop_recording" class="btn btn-secondary d-none">{{ __('Stop recording') }}</button>
+                </div>
+                <hr class="my-1">
```
cedric-anne, line 124: This whole block should probably be skipped when the feature is not available.

When PNG and WebM files are disallowed, or if the `Screenshot.isSupported()` returns `false`, the form title and separator should not be displayed:
image

cedric-anne (changes requested):

cedric-anne (approved):

PR #16354, opened 2024-01-11
Add deny login action to authorization rules

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

Add a new "Deny login" action to the authorization rules. This can allow the import/existence of users within GLPI and the assignment of entities, while at the same time denying their ability to log into GLPI. In some cases, you may want certain types of users to be active within GLPI so they can be assigned to assets or delegated as a requester on a ticket, but not let them login themselves. If you manually created the account, you could randomize the password. However, for external authentication sources like LDAP, that wasn't an option.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ The present file will list all changes made to the project; according to the
 - Log viewer for logs in `files/_log` directory.
 - Custom palette/theme support (uses `files/_themes` directory by default).
 - Two-Factor Authentication (2FA) support via Time-based One-time Password (TOTP).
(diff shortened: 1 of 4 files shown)


cconard96:
Most changes are whitespace. Easiest to review with "Hide whitespace" option enabled.

cedric-anne (approved):
Seems OK.

orthagh (approved):

PR #16355, opened 2024-01-11
Add recipient exclusions for notifications

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

Adds a new field to the Recipients tab of notifications to allow specifying exclusions. For example, the "New Ticket" notification may be configured to be sent to the ticket's requesters. However, there may be a specific group or profile that you don't want to be notified about this event. A real-life case is within an education environment where students report issues to a physical helpdesk. They may have user accounts within GLPI from LDAP and therefore also emails. Previously, you couldn't have the students added to the tickets as the requester for tracking purposes and also avoid sending them the emails when the staff at the physical helpdesk is supposed to act as the middle-man. In some cases, you may only want specific notifications like the ticket closed notification or one with a specific criteria filter to go to them.

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -70,6 +70,7 @@ The present file will list all changes made to the project; according to the
 - The debug tab that was present, for some items, when the debug mode was active, no longer exists. The corresponding features have been either moved, either removed.
 - `Group` and `Group in charge` fields for assets may now contain multiple groups.
 - "If software are no longer used" transfer option is now taken into account rather than always preserving.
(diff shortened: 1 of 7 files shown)


AdrienClairembault (approved):
Looks good, probably need a rebase to make tests are still valid on the latest commit.

orthagh (approved):

AdrienClairembault (approved):

PR #16375, opened 2024-01-16
Batch debug data requests

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

Reduce the number of requests to `ajax/debug.php` for debug data by queuing up requests and sending them in batches no more than once a second. If there are performance issues for you because of a large session, this should help with the debug requests blocking others.


cconard96:
No recognizable gain was seen with this change. UI responsiveness may even be worse in this case.

PR #16389, opened 2024-01-18
Add missing global var

AdrienClairembault:
Found this mistake while digging into the asset composition code.

It prevent common stuff like locations to be applied to all generic assets.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |

### DIFF
--- a/src/Asset/AssetDefinitionManager.php
+++ b/src/Asset/AssetDefinitionManager.php
@@ -134,6 +134,9 @@ final class AssetDefinitionManager
      */
     private function boostrapConcreteClass(AssetDefinition $definition): void
     {
+        /** @var array $CFG_GLPI */
+        global $CFG_GLPI;
+
         $capacities = $this->getAvailableCapacities();
 
         $concrete_class_name = $definition->getConcreteClassName();


AdrienClairembault:
It probably mean that common features are not currently unit tested.

Should we add a "Unit tests for common features" entry to the backlog (@orthagh) ?

orthagh:
This is very strange that error passed the lint, there is a some check to do about CI @cedric-anne @trasher

AdrienClairembault:
After further testing, it does not seems to prevent common features after all.
Not sure if it fixes anything then, maybe it is dead code ?

trasher:
> This is very strange that error passed the lint, there is a some check to do about CI @cedric-anne @trasher

This looks indeed strange but...

`$test = $CFG_GLPI['anything'];` will produce an error from phpstan `Undefined variable: $CFG_GLPI`.

And, `$CFG_GLPI['anything'] = 'something';` will not.

This part of code is maybe not used yet, I have no idea since I've not yet worked on it.

trasher:
> And, `$CFG_GLPI['anything'] = 'something';` will not.

As far as I can see, there is only a non official extension that provides such a check (unused local variable).

trasher (approved):

PR #16407, opened 2024-01-23
Make antiviruses generic

trasher:
Was previously only related to computers

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #number


cedric-anne (commented):
The old class should probably be deprecated, as we did in GLPI 10.0 for the `Netpoint` class. The deprecation should also be handled in the API (see `Glpi\Api\Deprecated` namespace).

cedric-anne (approved):

PR #16412, opened 2024-01-23
A new Functionnality : select an entity for external event

ales-infotel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tests pass?   | yes

Possibility to select an entity on external event creation and limit template for the entity, users and guests


trasher:
For feature proposal, please target main branch

PR #16429, opened 2024-01-25
cleanorphans: more precise description

Rom1-B:
| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !31156

It seems that the current description is not explicit enough.

### DIFF
--- a/src/Document.php
+++ b/src/Document.php
@@ -1893,7 +1893,7 @@ class Document extends CommonDBTM
 
         switch ($name) {
             case 'cleanorphans':
-                return ['description' => __('Clean orphaned documents')];
+                return ['description' => __('Clean orphaned documents: deletes all documents that are not associated with any items.')];
         }
         return [];
     }


cconard96:
The text change is OK for me, but maybe the addition of a link "Show what would be deleted" would be very helpful too.
If that link goes to the search results page for documents and shows everything without a link to something else, GLPI admins can get a better idea of what would be deleted and even the chance to manually clean up the desired documents.

Not sure how easy it would be to add such a link/info alert into the crontask UI though.

trasher:
I'd agree with such a link, but anyway, such a change should target main and not a bugfixes branch.

I have no idea neither if it could be easy or difficult :)

cconard96 (approved):

PR #16431, opened 2024-01-25
Feature/remove spaces around equipment names

RomainLvr:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !30801

Removal of spaces before and after equipment names when adding or modifying in database

### DIFF
--- a/js/clipboard.js
+++ b/js/clipboard.js
@@ -65,27 +65,12 @@ $(function () {
      * if the input get a paste event, the text is trimmed before being pasted
      */
     $(document).on("paste", "input[type='text']", function (event) {
-        // Cancel the default action
-        event.preventDefault();
-
-        // Get the text from the clipboard
-        var text = event.originalEvent.clipboardData.getData('text/plain');
-
-        // Get the actual text in the input
-        var actual = $(this).val();
-
-        /**
-         * if actual value is empty, trim the text
-         * else add the text to the actual value
-         */
-        if (actual === '') {
-            text = text.trim();
-        } else {
-            text = actual + text;
+        if ($(this).val() === '') {
+            event.preventDefault();
+            var pastedData = event.originalEvent.clipboardData || window.clipboardData;
+            var pastedText = pastedData.getData('text');
+            $(this).val(pastedText.trim());
         }
-
-        // Paste the text in the input
-        $(this).val(text);
     });
 });
 
--- a/js/clipboard.js
+++ b/js/clipboard.js
@@ -61,10 +61,10 @@ $(function () {
     });
 
     /**
-     * For each input of type text,
+     * For each input of type text with name 'name'
      * if the input get a paste event, the text is trimmed before being pasted
      */
-    $(document).on("paste", "input[type='text']", function (event) {
+    $(document).on("paste", "input[type='text'][name='name']", function (event) {
         event.preventDefault();
         var pastedData = event.originalEvent.clipboardData || window.clipboardData;
         var pastedText = pastedData.getData('text');
--- a/js/clipboard.js
+++ b/js/clipboard.js
@@ -66,10 +66,9 @@ $(function () {
      */
     $(document).on("paste", "input[type='text']", function (event) {
         event.preventDefault();
-        var actualValue = $(this).val();
         var pastedData = event.originalEvent.clipboardData || window.clipboardData;
         var pastedText = pastedData.getData('text');
-        $(this).val(actualValue + pastedText.trim());
+        document.execCommand('insertText', false, pastedText.trim());
     });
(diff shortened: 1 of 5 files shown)


src/CommonDBTM.php
```diff
@@ -2024,6 +2024,10 @@ public function post_updateItem($history = true)
      */
     public function pre_addInDB()
     {
+        if (isset($this->input['name'])) {
```
cedric-anne, line 2027: Some classes are not using the default `name` field. You have to use the `getNameField()` to ensure that you are using the correct field.

This remark applies to all your PR.

```suggestion
        if (isset($this->input[static::getNameField()])) {
```

cedric-anne (changes requested):
Please ensure that all classes that overrides the `pre_addInDB()` and `pre_updateInDB()` methods are calling the parent method (i.e. `parent::pre_xxxInDB()`).

Rom1-B:
The original request only concerned pasting, to prevent the user inadvertently selecting an extra space when copying. If we trim systematically, it removes the current possibility of entering a space. _Even if it's irrelevant to me, I can imagine that some customers would like it._

js/clipboard.js
```diff
@@ -59,6 +59,34 @@ $(function() {
...
+         */
+        if (actual === '') {
+            text = text.trim();
+        } else {
+            text = actual + text;
+        }
+
+        // Paste the text in the input
+        $(this).val(text);
+    });
```
Rom1-B, line 89: To test:
```suggestion
    $(document).on("paste", "input[type='text']", function (event) {
        if ($(this).val() === '') {
            event.preventDefault();
            var pastedData = event.originalEvent.clipboardData || window.clipboardData;
            var pastedText = pastedData.getData('text');
            $(this).val(pastedText.trim());
        }
    });
```

src/CommonDBTM.php
```diff
@@ -478,6 +478,7 @@ public function showForm($ID, array $options = [])
         $this->initForm($ID, $options);
         $new_item = static::isNewID($ID);
         $in_modal = (bool) ($_GET['_in_modal'] ?? false);
+        Html::requireJs('clipboard');
```
Rom1-B, line 481: Is it necessary?

Rom1-B (changes requested):

src/CommonDBTM.php
RomainLvr: Fixed in [31028e9](https://github.com/glpi-project/glpi/pull/16431/commits/31028e919c632fa61084eb60444b1eb53d884526)

js/clipboard.js
RomainLvr: Applied in [4dea0c1](https://github.com/glpi-project/glpi/pull/16431/commits/4dea0c1691c06d3d41a502bdbed137bbc94c9fbc)

Rom1-B (approved):

trasher:
As said in our internal chat, I do prefer the first PHP solution.

 It's more robust and testable than a JS one first, and I'm pretty sure adding an extra (unbreakable)space besides any name would cause issue in most of the cases, even if it's not from a copy paste (It can cause issues in search, in API, in inventory, ...).

I really see no real use case we really want to keep those extra spaces.

Rom1-B:
If the field is modified after it has been saved, the user may not understand why what is saved is different from what was entered. This can lead to support tickets.

In case of use, for example: allowing a space at the beginning can allow an item to be displayed always at the beginning, as here with `  computer2` :

image

cedric-anne:
@orthagh What do you think about it? Should we trim only on form side, or should we trim space that surrounds every `name` field during the saving process ?

orthagh:
Don't know...

I agree with @trasher that any space around the name may be a bug rather a wanted naming.
But people doing crazy stuff, I'm pretty sure someone will have this kind of thing.

I'm not satisfied either by the interception of paste mechanism.

But in fact, do we really want to fix people copy pasting spaces ? Is it our job ?
I don't know a software doing this (apart ofering the CTRL+Shift+V shortcut)

trasher:
If there are doubts, then no, we should not edit user input.

orthagh (commented):
Please limit the feature to input[name=name].
Frontend side is ok I guess. And as the thing has been sold, I agree for merge.

cedric-anne (changes requested):
See last @orthagh comment, please limit this behaviour to `name` fields.

orthagh (approved):

cedric-anne (approved):

PR #16432, opened 2024-01-25
Capacity: Can be associated to assistance objects

AdrienClairembault:
image

This capacities required additional changes:
* Move two search options that were declared directly in the `SearchOption` class instead of their relevant classes.
* Add the possibility for a capacity to define an help text.
* Add the possibility for a capacity to display configuration related messages.

---

Help text:

image

I've felt this was needed because we can't really sum up effectively this capacity with a simple name.

---

The assistance feature require the item to also be enabled in each profiles:

image

This may be confusing for the user (as he may not be aware of this setting), thus we need some to display some kind of warning:

image

The message is a link to the profile search page, showing the exact profiles for which the asset is enabled:

image

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |


cedric-anne:
I do not think that the `This asset definition is enabled on %d profile(s).` message is necessary. Indeed, as long as it has been added in any profile configuration, it means that the administrator is supposed to know how it works. Maybe a reminder like the following text could be added to the result of the `getHelpText()` method: `To be able to use the assets from the current definition in the assistance context, you must add it to the "Associable items to a ticket" option of each related profile.`

I also noticed that, when the asset definition is not active, it will be impossible to finish the configuration, as the class will not be available in the profile option.

Honestly, I am pretty sure that having a more precise helper text will make the `getConfigurationMessages()` pretty useless.

About `getConfigurationMessages()`, I also think it could be an issue if many capacities are using this feature. It could result in a bunch of messages that would probably be confusing. Having a precise message in the helper tooltip is probably better.

cedric-anne (changes requested):
As discussed IRL, all assets definitions should be added automatically in `$CFG_GLPI['ticket_types']`. Indeed, as each class has to be added manually in `helpdesk_item_type` for each profile, there is no need to add a capacity switch for this.

AdrienClairembault:
Ok, agree on this. I'll work into it when I have some time.

How do we inform the users about this ?  
Most user will not see an "Assistance" capacity and will assume that assets can't be linked to tickets at all.

This might generate some support requests or community tickets.

I think we need some kind of hint on the capacity page (maybe an "extra" sections at the end ?) that inform the user that this asset is not enabled as an associable asset for tickets on any profiles.

cedric-anne:
>  inform the user that this asset is not enabled as an associable asset for tickets on any profiles.

It may be intentional, we have to ensure that we do not pollute UI with messages that gives hints about something the administrator do not want.

This can be addressed later, so I added a line in the backlog for this in order to not block the current PR.

AdrienClairembault:
Force pushed to rebase onto the latest main commit.

src/CommonItilObject_Item.php
```diff
@@ -1576,4 +1577,36 @@ public static function dropdownAllDevices(
...
+        }
+
+        // Only show if at least one item is already linked
+        return countElementsInTable(
+            static::getTable(),
+            [
+                'itemtype' => $asset::getType(),
+                'items_id' => $asset->getId(),
+            ]
+        ) > 0;
```
cedric-anne, line 1610: If the tab is not shown when there is no linked items, it will not anymore possible to use the `New Ticket for this item` feature:
image

AdrienClairembault: The reason for this is that this is a basic capacity that can't be disabled.
If the tab is always visible, users might be stuck with 3 useless extra tabs on their assets.

I thought only showing the tab if the item is associable to a ticket or if there are already linked item was a good compromise to work around this.

> If the tab is not shown when there is no linked items, it will not anymore possible to use the New ticket for this item feature:

This feature is only enable if the asset is associable to a ticket, which in this case the tab will always be shown so no issues here.

src/CommonItilObject_Item.php
```diff
@@ -1576,4 +1577,36 @@ public static function dropdownAllDevices(
...
+        // Always display tab if the current profile is allowed to link ITIL
+        // items to this asset
+        if (
+            in_array(
+                $asset::class,
+                $_SESSION["glpiactiveprofile"]["helpdesk_item_type"] ?? []
+            )
+        ) {
+            return true;
+        }
```
cedric-anne, line 1601: It may be a problem. People may not be allowed to link computers to ITIL objects, but may want to be able to see existing links.

src/CommonItilObject_Item.php
```diff
@@ -1576,4 +1577,36 @@ public static function dropdownAllDevices(
...
+    /**
+     * ITIL tabs for generic assets should only be displayed if the asset already
+     * has associated ITIL items OR if the current user profile is allowed to
+     * link this asset to ITIL items
+     *
+     * @param Asset $asset
+     *
+     * @return bool
+     */
+    protected function shouldDisplayTabForAsset(Asset $asset): bool
```
cedric-anne, line 1590: IMHO, this change should be discussed in a separate PR. It is probably less consensual than you expected.

src/CommonItilObject_Item.php
trasher: > users might be stuck with 3 useless extra tabs on their assets. It kinda defeat the whole "customizable" part.

Just like it's always been... It's not a recent subject.

My point is not this is a good or a bad idea; but if you want to propose a solution for "too many tabs", it's not something to add in this kind of PR; it has to be validated as a separate feature.

src/CommonItilObject_Item.php
AdrienClairembault: This is covered by the check right after.

cedric-anne: Indeed.

src/Asset/AssetDefinitionManager.php
```diff
@@ -145,17 +150,37 @@ private function boostrapConcreteClass(AssetDefinition $definition): void
...
+        // CommonItilObject_Item and Ticket classes).
+        CommonGLPI::registerStandardTab(
+            $asset_class_name,
+            Ticket::class, // TODO: Why not Item_Ticket ?
+            51
+        );
+        CommonGLPI::registerStandardTab(
+            $asset_class_name,
+            Item_Problem::class,
+            51
```
cedric-anne, line 177:
```suggestion
            52
```

src/Asset/Capacity/AbstractCapacity.php
```diff
@@ -228,4 +228,16 @@ protected function unregisterFromTypeConfig(string $config_name, string $itemtyp
...
+     * Escape and encode an itemtype for use in a SQL "LIKE" query.
+     *
+     * @param string $classname
+     *
+     * @return string
+     */
+    protected function escapeAndEncodeClassName(string $classname): string
+    {
+        return str_replace('\\', '\\\\', json_encode($classname));
+    }
```
cedric-anne, line 242: Not used anymore.

```suggestion
```

src/CommonItilObject_Item.php
cedric-anne: Also, having a distinct behaviour for generic asset and other asset may confuse end users. Indeed, they may think that lack of tab presence is a bug rather than a missing profile configuration.
If we want to change this behaviour, it would be preferable to do it globally.

src/CommonItilObject_Item.php
AdrienClairembault: I asked for confirmation to @orthagh and he told me this behavior is the correct one to apply.

In fact, it should also be done on basic assets (computer, monitors, ...) in another PR.

templates/pages/admin/assetdefinition/capacities.html.twig
```diff
@@ -40,6 +40,7 @@
 {# do not display footer with dates #}
 {% set params = params|merge({'formfooter': false}) %}
 
+
```
cedric-anne, line 43:
```suggestion
```

src/CommonITILObject.php
```diff
@@ -33,6 +33,7 @@
  * ---------------------------------------------------------------------
  */
 
+use Glpi\Asset\Asset;
```
cedric-anne, line 36:
```suggestion
```

cedric-anne (approved):

trasher (approved):

PR #16435, opened 2024-01-25
Fix getItemTypeForTable when using relation table

stonebuzz:
Same as  https://github.com/glpi-project/glpi/pull/14566

But handle single level namespace + CommonDBRelation from Plugin context

Ex : 
```php
GlpiPlugin\\Foo\\Relation_Item
```

| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #number

### DIFF
--- a/src/DbUtils.php
+++ b/src/DbUtils.php
@@ -320,7 +320,7 @@ final class DbUtils
                     // NOT Glpi\Namespace1\Namespace2\Item\Filter
                     // To avoid this, we can revert the last '_' and check if the itemtype exists
                     $check_alternative = $is_plugin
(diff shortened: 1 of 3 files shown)


trasher (approved):

AdrienClairembault (approved):
Note that `$table` is not the full qualified table name (`glpi_plugin_foo_items_filters`) but only the last part of the table name (`items_filters`).

With that information in mind, checking for at least one `_` seems good and I agree with this solution.

cedric-anne (approved):

PR #16436, opened 2024-01-25
Add Appliances Capacity

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

### DIFF
--- a/src/Appliance_Item.php
+++ b/src/Appliance_Item.php
@@ -174,21 +174,6 @@ class Appliance_Item extends CommonDBRelation
             echo "</div>";
         }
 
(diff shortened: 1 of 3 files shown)


src/Appliance_Item.php
```diff
@@ -174,21 +174,6 @@ public static function showItems(Appliance $appliance)
             echo "</div>";
         }
 
-        echo "<table class='tab_cadre_fixehov'>";
```
cconard96, line 177: Leftover HTML that wasn't removed after moving to Twig

PR #16439, opened 2024-01-25
Prevent FullCalendar failure when default view is invalid

AdrienClairembault:
The planning page will fail if the default view is unknown to fullcalendar.

For example, the https://github.com/pluginsGLPI/advancedplanning plugin add a new "Timeline week" view, which users may select.
If the plugin is disabled later on, users that were using this view are stuck with an error on the callendar page (which become unusable).

image

To fix this, I've added a fallback to the default view.
This is done using a "fake" calendar initialisation in a try/catch block (an invalid view will trigger the catch block, which will revert to the default view before we initialize the real calendar).

It would be cleaner if we could check if a view exist by using an official FullCalendar method but i didn't find anything useful for this purpose in their documentation.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |

### DIFF
--- a/js/planning.js
+++ b/js/planning.js
@@ -85,6 +85,21 @@ var GLPIPlanning  = {
         var hidden_days = all_days.filter(day => !enabled_days.some(n => n == day));
         var loadedLocales = Object.keys(FullCalendarLocales);
 
+        // We need to check that the default view is valid to avoid errors
+        // FullCalendar does not seem to offer a method to check if a view is
+        // valid or list all possible views.
+        // It seems the only way to validate our view is to try to create a fake
+        // Calendar instance
+        try {
+            new FullCalendar.Calendar("fake_calendar", { // Invalid selector does not matter
+                defaultView: options.default_view,
+                plugins: options.plugins,
+            });
+        } catch (e) {
+            // Invalid view type, fallback to default
+            options.default_view = default_options.default_view;
+        }
+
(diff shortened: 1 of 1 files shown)


AdrienClairembault:
Feel free to re-qualify to 10.0.13 if needed.

cedric-anne (commented):
There is a `Calendar.prototype.isValidViewType()` method, but it has to be call on an already initialized instance of `Calendar`. Maybe the `defaultView` could be removed from the options passed to the constructor, and the view changed defined using the dedidated `changeView()` method:
```js
const view = this.calendar.isValidViewType(options.default_view) ? options.default_view : default_options.default_view;
this.calendar.changeView(view);
```

AdrienClairembault:
> There is a `Calendar.prototype.isValidViewType()` method, but it has to be call on an already initialized instance of `Calendar`. Maybe the `defaultView` could be removed from the options passed to the constructor, and the view changed defined using the dedidated `changeView()` method:
> 
> ```js
> const view = this.calendar.isValidViewType(options.default_view) ? options.default_view : default_options.default_view;
> this.calendar.changeView(view);
> ```

Good find, seems like an appropriate solution.
I'll try it.

cedric-anne (approved):

PR #16449, opened 2024-01-28
Improve asset definition deletion

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

Replace use of the default confirmation dialog with one that says how many assets use the definition and would be deleted as a result of deleting the definition. If the definition is active, it shows a link to the asset search page so you can see what would be deleted. If the definition is inactive, they wouldn't have permission to view the search page so it simply says all assets using the definition will be deleted too.

Also changed the `glpi_confirm` function to allow specifying a callback to run when the modal is closed.

### DIFF
--- a/src/Asset/AssetDefinition.php
+++ b/src/Asset/AssetDefinition.php
@@ -141,6 +141,23 @@ final class AssetDefinition extends CommonDBTM
     public function showForm($ID, array $options = [])
     {
         $this->initForm($ID, $options);
+        $options['candel'] = false;
+
+        if (!self::isNewID($ID)) {
+            // Add custom delete button that will show the number of assets using this definition and a link to their search page
+            $asset_count = countElementsInTable('glpi_assets_assets', ['assets_assetdefinitions_id' => $ID]);
+            $asset_class = $this->getAssetClassName();
+            $asset_search_url = $asset_class::getSearchURL();
+            $options['addbuttons'] = [
+                'purge' => [
+                    'title' => _x('button', 'Delete permanently'),
+                    'add_class' => 'btn-outline-danger',
+                    'icon' => 'ti ti-trash',
+                    'text' => _x('button', 'Delete permanently'),
+                    'type' => 'submit',
+                ]
+            ];
+        }
         TemplateRenderer::getInstance()->display(
             'pages/admin/assetdefinition/main.html.twig',
             [
@@ -148,6 +165,8 @@ final class AssetDefinition extends CommonDBTM
                 'params'                => $options,
                 'has_rights_enabled'    => $this->hasRightsEnabled(),
                 'reserved_system_names' => AssetDefinitionManager::getInstance()->getReservedAssetsSystemNames(),
+                'asset_count'        => $asset_count ?? 0,
+                'asset_search_url'   => $asset_search_url ?? '#',
             ]
         );
         return true;
--- a/src/Asset/AssetDefinition.php
+++ b/src/Asset/AssetDefinition.php
@@ -144,6 +144,8 @@ final class AssetDefinition extends CommonDBTM
         $this->initForm($ID, $options);
         $options['candel'] = false;
 
+        $asset_count = 0;
+
         if (!self::isNewID($ID)) {
             // Add custom delete button that will show the number of assets using this definition and a link to their search page
             $asset_count = countElementsInTable(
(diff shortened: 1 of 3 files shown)


src/Asset/AssetDefinition.php
```diff
@@ -141,12 +141,31 @@ public static function displayTabContentForItem(CommonGLPI $item, $tabnum = 1, $
     public function showForm($ID, array $options = [])
     {
         $this->initForm($ID, $options);
+        $options['candel'] = false;
+
+        if (!self::isNewID($ID)) {
+            // Add custom delete button that will show the number of assets using this definition and a link to their search page
+            $asset_count = countElementsInTable('glpi_assets_assets', ['assets_assetdefinitions_id' => $ID]);
+            $asset_class = $this->getAssetClassName();
+            $asset_search_url = $asset_class::getSearchURL();
```
cedric-anne, line 150: To make sure that the link will forward to the complete unfiltered list, it should contains the default empty criterai provided by `Glpi\Search\Input\QueryBuilder::getDefaultCriteria()` and the `reset` parameter.

src/Asset/AssetDefinition.php
```diff
@@ -141,12 +141,31 @@ public static function displayTabContentForItem(CommonGLPI $item, $tabnum = 1, $
     public function showForm($ID, array $options = [])
     {
         $this->initForm($ID, $options);
+        $options['candel'] = false;
+
+        if (!self::isNewID($ID)) {
+            // Add custom delete button that will show the number of assets using this definition and a link to their search page
+            $asset_count = countElementsInTable('glpi_assets_assets', ['assets_assetdefinitions_id' => $ID]);
```
cedric-anne, line 148: We should probably not count templates. Also, maybe we should not count items from trash bin, but I am not sure of that.

templates/pages/admin/assetdefinition/main.html.twig
```diff
@@ -71,6 +71,53 @@
...
+                    <div class="alert-title">{{ __('This asset definition is not used by any assets.') }}</div>
+                </div>
+            {% endif %}
+        {% endset %}
+        <script>
+            $(() => {
+                const handle_delete = (e) => {
+                    e.preventDefault();
+                    glpi_confirm({
+                        title: '{{ __('Confirm the final deletion?') }}',
```
cedric-anne, line 100: Translation may contains unsafe JS chars.
```suggestion
                        title: '{{ __('Confirm the final deletion?')|e('js') }}',
```

cedric-anne (commented):
It works as expected, but maybe we could reuse the same kind of modal as for capacities.

When definition is deleted:
image

When capacities are disabled:
image

orthagh:
> maybe we could reuse the same kind of modal as for capacities

Agree. A summary for a good modal:

- Use a big icon to alert if destructive action (if informative, the icon is important, could be present but not big)
- Make a comprehensive message in bold
- Add a text-muted paragraph if more explanation required
- Make the primary action:
  - always on right
  - identifiable visually 
  - reflecting the action by color/class (primary, danger, warning)
  - use a comprehensive label (Ok/Cancel may be in some case confusing)
- Secondary actions should avoid attractive styles (you may use ghost-secondary if you prefer, I'm not sure I'm ok with .btn style only for the provided example modal from capacities) and placed on right
- Avoid .card in .card/.modal UI
- more subject to discussion: avoid collapse elements unless there is very big list ("Show what would be deleted").
- Regarding title, avoid color (green is a bad choice for a destructive action)

templates/pages/admin/assetdefinition/main.html.twig
```diff
@@ -76,6 +76,92 @@
...
+                        <div class="text-muted">
+                            {% if asset_count > 0 %}
+                                {{ __('This asset definition is used by %d assets.')|format(asset_count) }}
+                                <br>
+                                {% if item.fields['is_active'] %}
+                                    {# Can only view search page when the definition is active #}
+                                    <a href="{{ asset_search_url }}">{{ __('Show what would be deleted') }}</a>
+                                {% else %}
+                                    {{ __('All assets that are using this definition will also be deleted.') }}
+                                {% endif %}
```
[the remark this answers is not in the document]
cedric-anne: @orthagh Should we keep this link or remove it?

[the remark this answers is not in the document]
orthagh: remove it

cedric-anne (approved):

trasher (approved):

orthagh (approved):

PR #16460, opened 2024-01-29
small dashboard refresh

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

A quick refresh of the dashboard cards:
- make round borders visible
- add a subtle border color for ligh big numbers
- remove shadow on cards

image

### DIFF
--- a/src/Toolbox.php
+++ b/src/Toolbox.php
@@ -3014,13 +3014,7 @@ HTML;
 
             // if transparency present, get only the color part
             if (strlen($color) === 8 && preg_match('/^[a-fA-F0-9]+$/', $color)) {
-                $tmp = $color;
-                $alpha = hexdec(substr($tmp, 6, 2));
                 $color = substr($color, 0, 6);
-
-                if ($alpha <= 100) {
-                    return "inherit";
-                }
             }
 
             $color_inst = new Color($color);
--- a/src/Toolbox.php
+++ b/src/Toolbox.php
@@ -3006,6 +3006,35 @@ HTML;
         return "#" . $fg_color;
     }
 
+
+    public static function isColorDark(string $color = ""): bool
+    {
+        if ($color !== "") {
+            $color = str_replace("#", "", $color);
+
+            // if transparency present, get only the color part
+            if (strlen($color) === 8 && preg_match('/^[a-fA-F0-9]+$/', $color)) {
+                $tmp = $color;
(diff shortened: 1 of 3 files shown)


src/Toolbox.php
```diff
@@ -3006,6 +3006,35 @@ public static function getFgColor(string $color = "", int $offset = 40, bool $in
         return "#" . $fg_color;
     }
 
+
+    public static function isColorDark(string $color = ""): bool
```
cedric-anne, line 3010: This parameter is mandatory, it should not have a default value.

```suggestion
    public static function isColorDark(string $color): bool
```

src/Toolbox.php
```diff
@@ -3006,6 +3006,35 @@ public static function getFgColor(string $color = "", int $offset = 40, bool $in
...
+            $color = str_replace("#", "", $color);
+
+            // if transparency present, get only the color part
+            if (strlen($color) === 8 && preg_match('/^[a-fA-F0-9]+$/', $color)) {
+                $tmp = $color;
+                $alpha = hexdec(substr($tmp, 6, 2));
+                $color = substr($color, 0, 6);
+
+                if ($alpha <= 100) {
+                    return "inherit";
```
cedric-anne, line 3022: Method is supposed to return a boolean, not a string. It will fail with a PHP error.

src/Toolbox.php
```diff
@@ -3006,6 +3006,35 @@ public static function getFgColor(string $color = "", int $offset = 40, bool $in
...
+            }
+
+            $color_inst = new Color($color);
+            return $color_inst->isLight();
+        }
+
+        return false;
+    }
+
+    public static function isColorLight(string $color = ""): bool
```
cedric-anne, line 3033:
```suggestion
    public static function isColorLight(string $color): bool
```

AdrienClairembault:
image

It feel weird to mix bordered and non bordered items.
Is is done on purpose to only add the new border to the big number items ? 
Shouldn't it be on every card ?

orthagh:
Tried on other cards, was not convinced.
I may work more on this, but pies are always bad regarding the styles

AdrienClairembault (approved):

cedric-anne (approved):

PR #16465, opened 2024-01-30
Feat/enhanced datatable twig

stonebuzz:
First commit allow you to display HTML in table row content

Before : 

image

After : 

image

The second commit search selected value (from filter list) by ```key``` instead of ```value``` when filter is a 'real' dropdown (```key => val``` / ```0 => 'foo'```)

 dropdown is created like this 

```php
Array
  (
      [2] => 16780900
      [1] => stanislas-asus-desktop-2022-09-20-16-43-09
  )
```

and filter is created like this

```php
Array
  (
      [agents_id] => Array
          (
              [0] => 2
          )
  )
```

See  : https://github.com/pluginsGLPI/deploy/pull/13

Actually, only ```Item_Processes``` and ```Item_Environment``` use ```dropdown``` filter , but in each case ```key``` is equal to ```value```

```php
Array
  (
      [root] => root
      [systemd+] => systemd+
      [avahi] => avahi
      [message+] => message+
      [syslog] => syslog
      [rtkit] => rtkit
      [kernoops] => kernoops
      [postfix] => postfix
      [lp] => lp
      [colord] => colord
      [stanisl+] => stanisl+
      [mysql] => mysql
  )
```

Need by  : https://github.com/pluginsGLPI/deploy/pull/13

| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #number

### DIFF
--- a/templates/components/datatable.html.twig
+++ b/templates/components/datatable.html.twig
@@ -119,7 +119,7 @@
                                        <select name="filters[{{ colkey }}][]"
                                             class="form-select filter-select-multiple" multiple>
                                         {% for field, value in columns_values[colkey] %}
-                                            <option value="{{ field }}" {{ filters[colkey] is defined and value in filters[colkey] ? 'selected' : '' }}>
+                                            <option value="{{ field }}" {{ filters[colkey] is defined and field in filters[colkey] ? 'selected' : '' }}>
                                                 {{ value }}
                                             </option>
                                         {% endfor %}
@@ -172,7 +172,7 @@
                                 <td>
                                     {% if colkey in entry|keys %}
 
-                                        {% set formatter = formatters[colkey] %}
+                                        {% set formatter = row_formatters[colkey]|default(formatters[colkey]) %}
 
                                         {% if formatter == "maintext" %}
                                             <span class="d-inline-block bg-blue-lt p-1 text-truncate"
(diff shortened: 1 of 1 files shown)


templates/components/datatable.html.twig
```diff
@@ -212,7 +212,7 @@
                                                    {% endif %}
                                                 </span>
                                         {% else %}
-                                            {{ entry[colkey] }}
+                                            {{ entry[colkey]|raw }}
```
cedric-anne, line 215: There is already a `raw_html` filter for that.

stonebuzz: if I change the formatting to ```raw_html``` I lose the ability to filter with a dropdown list (only a text field is provided)

image

cconard96: Maybe add a new optional filter_formatters parameter that can be set for this template to specify the formatting. If none is specified for a column, use the formatting based on the column.

Example:
{% set filter_format = filter_formatters[colkey]|default(formatters[colkey]) %}

orthagh: It's simpler that you think.
@cedric-anne said there is already a formatter making the `|raw` filter above the else.

Check https://github.com/glpi-project/glpi/pull/16465/files/6d9a1413e6512a3cf4f89bb5fac8ca981ac646e0#diff-c43a216f1d73a092b92980d1fc0766a39e50ae64b5180e51d35aed3dc1a5d384L199-R200

PR #16470, opened 2024-01-30
feat(datatable): add capability to define row format

stonebuzz:
Add capability to define row format

Actually ```row``` format is based on ```filter``` format

But a filter based on dropdown (like ```User```) can be displayed as HTML link.

This PR is made for this.

Need by : https://github.com/pluginsGLPI/deploy/pull/13

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #number

### DIFF
--- a/templates/components/datatable.html.twig
+++ b/templates/components/datatable.html.twig
@@ -172,7 +172,7 @@
                                 <td>
                                     {% if colkey in entry|keys %}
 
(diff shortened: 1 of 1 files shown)


cedric-anne (commented):
Why not using the existing `formatters` parameter?

PR #16471, opened 2024-01-30
Use `atoum/stubs`

AdrienClairembault:
Our tests extends `atoum\atoum` which doesn't seem to exist.

I suppose there is some black magic somewhere (as our tests obviously works) but it isn't known to VSCode which does not recognize any atoum method.

Using the correct `atoum` import fix the issue on VSCode and does not impact our tests.
This is the import used in the official documentation (https://atoum.readthedocs.io/en/latest/first_test.html#first-tests).

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | on
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |

### DIFF
--- a/tests/GLPITestCase.php
+++ b/tests/GLPITestCase.php
@@ -33,7 +33,6 @@
  * ---------------------------------------------------------------------
  */
 
-use atoum\atoum;
 use Glpi\Tests\Log\TestHandler;
 use Monolog\Logger;
 use Psr\Log\LogLevel;
--- a/composer.json
+++ b/composer.json
@@ -78,6 +78,7 @@
     "require-dev": {
         "ext-xml": "*",
         "atoum/atoum": "^4.1",
+        "atoum/stubs": "^2.6",
         "friendsoftwig/twigcs": "^6.1",
         "glpi-project/tools": "^0.7",
         "maglnet/composer-require-checker": "^3.5",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "826b306b3c00eed0615a2f68c33dcace",
+    "content-hash": "07e200b68a65785b0b35e84c42f59131",
     "packages": [
         {
             "name": "brick/math",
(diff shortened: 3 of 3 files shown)


cconard96:
> Our tests extends `atoum\atoum` which doesn't seem to exist.
> 
> I suppose there is some black magic somewhere (as our tests obviously works) but it isn't known to VSCode which does not recognize any atoum method.

This works?
I'm surprised VSCode can make any sense of the library even after changing the base class. Atoum has no PHPDoc and the vast majority of the methods/properties (including asserters) are undocumented magic properties/methods. Technically, I think the base class for tests is "atoum\atoum\test" but there is some more magic in the autoloader that aliases it.

tests/GLPITestCase.php
```diff
@@ -33,7 +33,7 @@
  * ---------------------------------------------------------------------
  */
 
-use atoum\atoum;
+use atoum;
```
cedric-anne, line 36: I dig a bit inside the atoum autoloader. `atoum` and `atoum\atoum` are default aliases for `atoum\atoum\test` class, see https://github.com/atoum/atoum/blob/fd9a339de252ea774ab755b3fd25dd19be2dbefb/classes/autoloader.php#L41-L48

The following patch is probably the best option to point to a class that is known by any IDE, but will expose to the IDE many method that we are probably not supposed to use in our tests.

```diff
diff --git a/tests/GLPITestCase.php b/tests/GLPITestCase.php
index b273c5b29c..bdf3589a7e 100644
--- a/tests/GLPITestCase.php
+++ b/tests/GLPITestCase.php
@@ -33,14 +33,14 @@
  * ---------------------------------------------------------------------
  */
 
-use atoum\atoum;
+use atoum\atoum\test;
 use Glpi\Tests\Log\TestHandler;
 use Monolog\Level;
 use Psr\Log\LogLevel;
 
 // Main GLPI test case. All tests should extends this class.
 
-class GLPITestCase extends atoum
+class GLPITestCase extends test
 {
     private $int;
     protected $has_failed = false;
```

AdrienClairembault: `atoum\atoum\test` doesn't work correctly, I don't have the autocompletion for asserters:

image

With my original changes, it works fine:

image

cedric-anne: This class does not seems to exists, and it is probably your IDE that provides this completion. Anyway, I do not have the autocompletion either when using the `atoum\atoum\test` class, so using only `atoum` will not be worse for me as the current state.

Last thing. Is the import necessary? We are outside a namespace, so I guess that removing it will works.

AdrienClairembault: I just realized that the class that was found is actually from a formcreator vendor file:  https://github.com/atoum/stubs

I've added this vendor to GLPI and it should now work for everyone.

Should I move the PR to main or are new *dev* dependencies OK for bf ?

cedric-anne: It is ok on 10.0/bf branch, as it will only impact devs. We will just wait to merge this after the 10.0.12 release, in case it has unexpected side effects on tests results.

PR #16472, opened 2024-01-30
Warn user about data loss when capacities are disabled

ccailly:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 

Added information on capacity usage and a popup if certain capacities (used) are going to be deactivated, risking data loss.

image

image

### DIFF
--- a/src/Asset/Capacity/AbstractCapacity.php
+++ b/src/Asset/Capacity/AbstractCapacity.php
@@ -35,8 +35,10 @@
 
 namespace Glpi\Asset\Capacity;
 
+use CommonDBRelation;
 use DisplayPreference;
 use Glpi\Asset\Asset;
+use Glpi\Asset\AssetDefinition;
 use Log;
 
 /**
@@ -68,6 +70,101 @@ abstract class AbstractCapacity implements CapacityInterface
     public function getCloneRelations(): array
     {
         return [];
+    }
+
+    public function isUsed(string $classname): bool
+    {
+        return $classname::getDefinition()->hasCapacityEnabled($this);
+    }
+
+    /**
+     * Count the number of peer item used.
+     *
+     * The count is based on the number of distinct peer items IDs found in the table of the relation class
+     * in rows linked to the given asset class.
+     *
+     * @param class-string<\CommonDBTM> $asset_classname
+     * @param class-string<\CommonDBTM> $relation_classname
+     * @return int
+     */
+    protected function countPeerItemsUsage(string $asset_classname, string $relation_classname): int
+    {
+        /** @var \DBmysql $DB */
+        global $DB;
+
+        if (is_a($relation_classname, CommonDBRelation::class, true)) {
+            // We assume that asset type/id are always store in `itemtype`/`items_id` fields
+            // and we use the other `items_id_X` property as the peer item ID storage field.
+            $distinct_field  = null;
+            if ($relation_classname::$items_id_1 === 'items_id') {
+                $distinct_field  = $relation_classname::$items_id_2;
+            } elseif ($relation_classname::$items_id_2 === 'items_id') {
+                $distinct_field  = $relation_classname::$items_id_1;
+            }
+            if ($distinct_field === null) {
+                throw new \Exception('Unable to compute peer item foreign key field.');
+            }
+
+            return countDistinctElementsInTable(
+                $relation_classname::getTable(),
+                $distinct_field,
+                [
+                    'itemtype' => $asset_classname,
+                ]
+            );
+        }
+
+        if (
+            is_a($relation_classname, CommonDBRelation::class, true)
+            || $DB->fieldExists($relation_classname::getTable(), 'itemtype')
+        ) {
+            // We assume that asset type/id are always store in `itemtype`/`items_id` fields.
+            return countElementsInTable(
+                $relation_classname::getTable(),
+                [
+                    'itemtype' => $asset_classname,
+                ]
+            );
+        }
+
+        throw new \Exception('Unable to compute peer items usage.');
+    }
+
+    /**
+     * Count the number assets that are linked to a peer item.
+     *
+     * The count is based on the number of distinct assets IDs found in the table of the relation class.
+     *
+     * @param class-string<\CommonDBTM> $asset_classname
+     * @param class-string<\CommonDBTM> $relation_classname
+     * @return int
+     */
+    protected function countAssetsLinkedToPeerItem(string $asset_classname, string $relation_classname): int
+    {
+        // We assume that asset type/id are always store in `itemtype`/`items_id` fields.
+        return countDistinctElementsInTable(
+            $relation_classname::getTable(),
+            'items_id',
+            [
+                'itemtype' => $asset_classname,
+            ]
+        );
+    }
+
+    /**
+     * Count the number of assets for the given asset definition.
+     *
+     * @param class-string<\Glpi\Asset\Asset> $classname
+     * @return int
+     */
+    protected function countAssets(string $classname): int
+    {
+        return countElementsInTable(
+            Asset::getTable(),
+            [
+                AssetDefinition::getForeignKeyField() => $classname::getDefinition()->fields['id']
+            ]
+        );
     }
 
     public function onClassBootstrap(string $classname): void
--- a/src/Asset/Capacity/CapacityInterface.php
+++ b/src/Asset/Capacity/CapacityInterface.php
@@ -49,7 +49,7 @@ interface CapacityInterface
     /**
      * Get the search options related to the capacity.
      *
-     * @param string $classname
+     * @param class-string<\Glpi\Asset\Asset> $classname
      * @return array
      */
     public function getSearchOptions(string $classname): array;
@@ -70,9 +70,25 @@ interface CapacityInterface
     public function getCloneRelations(): array;
 
     /**
+     * Indicates whether the capacity is used by given asset class.
+     *
+     * @param class-string<\Glpi\Asset\Asset> $classname
+     * @return bool
+     */
+    public function isUsed(string $classname): bool;
+
+    /**
+     * Get the capacity usage description for given asset class.
+     *
+     * @param class-string<\Glpi\Asset\Asset> $classname
+     * @return string
+     */
+    public function getCapacityUsageDescription(string $classname): string;
+
+    /**
      * Method executed during asset classes bootstraping.
      *
-     * @param string $classname
+     * @param class-string<\Glpi\Asset\Asset> $classname
      * @return void
      */
     public function onClassBootstrap(string $classname): void;
@@ -80,7 +96,7 @@ interface CapacityInterface
     /**
      * Method executed when capacity is disabled on given asset class.
      *
-     * @param string $classname
+     * @param class-string<\Glpi\Asset\Asset> $classname
      * @return void
      */
     public function onCapacityDisabled(string $classname): void;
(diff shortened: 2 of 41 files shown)


orthagh:
A quick design review based on the provided screenshots:

- I'm not sure the inline annotations in the capacity setup are desirable. At least when 0 items linked. They pollute the space. I think the confirmation dialog when with the list is sufficient.
- If you want to be informative about the usage, a combination of badges (with numbers) right of the capacity name and tooltips should help keeping the display clean.
- In the dialog, having a grid design is a strange choice. A list with "Capacity name" and the message "used by X item" under could be more readable. (Using [list component](https://preview.tabler.io/lists.html))
- buttons should be adapted to let the user understand well the consequence
   - the action for confirm should be named as the action done. In this case, it's "Disable"
   - as it's a destructive action, the associated class should be "danger" and an icon should be added ("ti-check" could emphasise the confirmation action)
   - the cancel button is fine. You may use ghost class if you want to reduce its priority.

orthagh:
And not related to this pr, but @cedric-anne it was your choice to make the checkbox right of the label ? It's unconventional.

src/Asset/Capacity/CapacityInterface.php
```diff
@@ -62,6 +62,22 @@ public function getSearchOptions(string $classname): array;
      */
     public function getSpecificRights(): array;
 
+    /**
+     * Check if the capacity can be safetly disabled.
+     *
+     * @param \CommonDBTM $asset
+     * @return bool
+     */
+    public function canBeDisabled(\CommonDBTM $asset): bool;
```
cedric-anne, line 71: It will anyway be possible to disable it. `isUsed` will be less ambiguous about that.

```suggestion
    public function isUsed(\CommonDBTM $asset): bool;
```

Also, as the capacity is not related to a specific asset instance, but to all the assets from a definition, the method parameter should be `string $classname`.

src/Asset/Capacity/AbstractCapacity.php
```diff
@@ -65,6 +66,28 @@ public function getSpecificRights(): array
         return [];
     }
 
+    public function canBeDisabled(CommonDBTM $asset, string $table = null): bool
```
cedric-anne, line 69: You should not add more params than the interface declare. If you want to have a generic counting method, then you can declare a `protected function countUsages(...)` method.

Also, I do not think that we should provide a default implementation for this method.

By the way, please try to always pass itemtypes instead of tables in parameters.

src/Asset/Capacity/AbstractCapacity.php
```diff
@@ -65,6 +66,28 @@ public function getSpecificRights(): array
...
+        if ($table !== null) {
+            /** @var \DBmysql $DB */
+            global $DB;
+
+            return $DB->request(
+                $table,
+                [
+                    'itemtype' => 'Glpi\CustomAsset\\' . $asset->fields['system_name'],
+                ]
+            )->count() == 0;
```
cedric-anne, line 80: You can use the `countElementsInTable()` global function. Anyway, when you want to get a result count only, please use the a `SELECT COUNT` query that will be more efficient.

src/Asset/Capacity/AbstractCapacity.php
```diff
@@ -65,6 +66,28 @@ public function getSpecificRights(): array
...
+        }
+
+        return true;
+    }
+
+    public function getCapacityUsageDescription(\CommonDBTM $asset): string
+    {
+        return __('No capacity usage provided');
+    }
+
```
cedric-anne, line 90: I do not think that we should provide a default implementation for this method.

```suggestion
```

src/Asset/Capacity/HasAntivirusCapacity.php
```diff
@@ -51,6 +52,35 @@ public function getSearchOptions(string $classname): array
...
+
+    public function getCapacityUsageDescription(\CommonDBTM $asset): string
+    {
+        /** @var \DBmysql $DB */
+        global $DB;
+
+        $count = $DB->request(
+            ItemAntivirus::getTable(),
+            [
+                'itemtype' => 'Glpi\CustomAsset\\' . $asset->fields['system_name'],
```
cedric-anne, line 68: This should work.

```suggestion
                'itemtype' => $asset::class,
```

src/Asset/Capacity/CapacityInterface.php
```diff
@@ -62,6 +62,22 @@ public function getSearchOptions(string $classname): array;
...
+
+    /**
+     * Get the capacity usage description.
+     *
+     * @param \CommonDBTM $asset
+     * @return string
+     */
+    public function getCapacityUsageDescription(\CommonDBTM $asset): string;
```
cedric-anne, line 79: As the capacity is not related to a specific asset instance, but to all the assets from a definition, the method parameter should be `string $classname`.

src/Asset/Capacity/HasContractsCapacity.php
```diff
@@ -47,6 +48,37 @@ public function getLabel(): string
...
+        ])->count();
+
+        $max = $DB->request([
+            'SELECT'   => 'items_id',
+            'DISTINCT' => true,
+            'FROM'     => Contract_Item::getTable(),
+            'WHERE'    => [
+                'itemtype' => 'Glpi\CustomAsset\\' . $asset->fields['system_name'],
+            ],
+        ])->count();
```
cedric-anne, line 77: This code seems really redundant. You should probably declare methods in the abstract class, like:
-  `protected funtion countTargetItemUsed(string $source_itemtype, string $target_itemtype): int`
-  `protected funtion countSourceItemCapacityUsages(string $source_itemtype, string $target_itemtype): int`

AdrienClairembault:
> I'm not sure the inline annotations in the capacity setup are desirable. At least when 0 items linked. They pollute the space. I think the confirmation dialog when with the list is sufficient.

This layout was suggested by @cedric-anne if I remember correctly.  
Doesn't look too bad IMO, I feel like the `muted` text help fill out the page as it feel a bit empty otherwise.

ccailly:
As requested by @orthagh I've applied some UI and UX modifications.
I agree with @AdrienClairembault regarding the layout of the capabilities page.
Regarding the modal, I'm not sure I've understood what you wanted @orthagh, I've applied a list, but I end up with a large space on most of the items in the list, maybe it's possible to make two columns. I'll talk about it tomorrow IRL.

Capacities layout
Danger modal

orthagh:
Your last screenshots seems good.
For muted text, I agree we can fill with something but the usage seems not a good candidate.

I may be satisfied by a description of the capacity (not in this PR, add a task in board)

templates/pages/admin/assetdefinition/capacities.html.twig
```diff
@@ -41,18 +41,128 @@
...
+
+    <!-- Modal -->
+    <div class="modal fade" id="capacityWarningModal" tabindex="-1" role="dialog" aria-hidden="true">
+        <div class="modal-dialog" role="document">
+            <div class="modal-content">
+                <a type="button" class="btn-close" data-bs-dismiss="modal" aria-label="Close"></a>
+                <div class="modal-status bg-danger"></div>
+                <div class="modal-body text-center py-4">
+                    {# SVG Icon corresponding to the ti-alert-triangle icon #}
+                    <svg xmlns="http://www.w3.org/2000/svg" class="icon mb-2 text-danger icon-lg" width="24" height="24" viewBox="0 0 24 24" stroke-width="2" stroke="currentColor" fill="none" stroke-linecap="round" stroke-linejoin="round"><path stroke="none" d="M0 0h24v24H0z" fill="none"></path><path d="M12 9v4"></path><path d="M10.363 3.591l-8.106 13.534a1.914 1.914 0 0 0 1.636 2.871h16.214a1.914 1.914 0 0 0 1.636 -2.87l-8.106 -13.536a1.914 1.914 0 0 0 -3.274 0z"></path><path d="M12 16h.01"></path></svg>
```
AdrienClairembault, line 84: Some extra context here for @cedric-anne :

We discussed the UX together today and agreed that SVG icons are better in these case as they scale better to bigger sizes.

----

Should we add the `@tabler/icons` dependency to npm or is it not worth it if we only plan to use it in places where big icons are needed and will keep the css icon as our default everywhere else ?

See https://tabler.io/docs/icons/svg

templates/pages/admin/assetdefinition/capacities.html.twig
```diff
@@ -41,18 +41,128 @@
...
+            });
+
+            $('input[name="capacities[]"]').on('change', function() {
+                if ($(this).data('is-used')) {
+                    $(this).toggleClass('red-toggle-switch-bg');
+                    $('i[data-capacity="' + $.escapeSelector($(this).data('capacity')) + '"]').toggleClass('d-none');
+                    $('div[data-capacity="' + $.escapeSelector($(this).data('capacity')) + '"]').toggleClass('d-none');
+
+                    // To avoid separator on the last visible item but not the last item of the dom
+                    // $('div.list-group-item:not(.d-none)').removeClass('border-bottom-0').last().addClass('border-bottom-0');
```
AdrienClairembault, line 156: This code is commented, maybe it was meant to be deleted ?

src/Asset/Capacity/AbstractCapacity.php
```diff
@@ -65,6 +65,45 @@ public function getSpecificRights(): array
...
+    public function isUsed(string $classname): bool
+    {
+        return (new $classname())->getDefinition()->hasCapacityEnabled($this);
+    }
+
+    protected function countTargetItemUsed(
+        string $source_itemtype,
+        string $target_itemtype,
+        ?string $distinctField = null
+    ): int {
```
AdrienClairembault, line 77: This method lack PHP doc (same for countSourceItemCapacityUsages and countAssets)

AdrienClairembault (changes requested):

cedric-anne (changes requested):
The feature is working as expected and the feature code is OK for me, but tests are really hard to read, and will probably be really hard to maintain.

1. Please use named parameters, for instance:
```diff
    public function myDataProvider(): array
     {
         return [
             [
-                Certificate::class,
-                Certificate_Item::class
+                'target_classname'  => Certificate::class,
+                'relation_classname' => Certificate_Item::class,
             ],
         ];
     }
```
It makes it easier to understand that the first value will correspond to a `$target_classname` parameter and the second value to a `$relation_classname` parameter.

2. Please do not use parameters that may have different behaviours depending on their type, e.g. `string|array`. It often result in more complex code. You can split into multiple parameters, for instance:
```diff
    public function myDataProvider(): array
     {
        return [
            [
-                [
-                    'class' => Log::class,
-                    'fields' => [
-                        'itemtype' => 'itemtype',
-                        'items_id' => 'items_id'
-                    ]
-                ]
+               'class' => Log::class,
+                'fields' => [
+                    'itemtype' => 'itemtype',
+                    'items_id' => 'items_id'
-+               ]
            ],
        ];
     }

-     public function testSomething(string|array $itemClass): array
+    public function testSomething(string $class, array $fields = []): array
```

3. Values copying could probably be place directly in the test itself, without having to declare anything in the provider. For instance:
```php
        // Auto-fill some field
        global $DB;
        $table = $itemClass::getTable();
        if ($DB->fieldExists($table, 'itemtype')) {
            $fields['itemtype'] = $definition->getAssetClassName();
        }
        if ($DB->fieldExists($table, 'itemtype')) {
            $fields['items_id'] = $subject->getID();
        }
```

AdrienClairembault (approved):

cedric-anne (commented):
Could you please rebase to apply the proposed changes to recently merged capacities (HasDatabaseInstance for instance)?

cedric-anne (approved):

PR #16510, opened 2024-02-05
Add description for capacities (and muted texts in setup)

ccailly:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 

Add a description to each capacity.

image

### DIFF
--- a/src/Asset/Capacity/AbstractCapacity.php
+++ b/src/Asset/Capacity/AbstractCapacity.php
@@ -57,6 +57,11 @@ abstract class AbstractCapacity implements CapacityInterface
     {
     }
 
+    public function getDescription(): string
+    {
+        return '';
+    }
+
     public function getSearchOptions(string $classname): array
     {
         return [];
--- a/src/Asset/Capacity/CapacityInterface.php
+++ b/src/Asset/Capacity/CapacityInterface.php
@@ -47,6 +47,14 @@ interface CapacityInterface
     public function getLabel(): string;
 
     /**
+     * Get the description of the capacity.
+     * This description is used in the capacity management interface.
+     *
+     * @return string
+     */
+    public function getDescription(): string;
+
+    /**
      * Get the search options related to the capacity.
      *
      * @param class-string<\Glpi\Asset\Asset> $classname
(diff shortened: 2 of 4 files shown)


orthagh (dismissed):
Regarding design : almost ok, I still think we loose a big space in the middle. I don't want to remove it entirely, but curently there is more space dedicated to emptyness than actual content. Maybe we can inverse space usage.

Another thing, if description is not provided, or empty, the whole block should not be displayed.

Regarding descriptions content, well, they don't inform about anything for the moment 😬 
I mean, yeah, Appliance capacity allows indeed to manage Appliances for the assets.
Don't take this as a personal criticism, but maybe another person, more experienced about the application could take a shot about the content.

src/Asset/Capacity/HasAntivirusCapacity.php
```diff
@@ -51,6 +51,11 @@ public function getSearchOptions(string $classname): array
         return ItemAntivirus::rawSearchOptionsToAdd();
     }
 
+    public function getDescription(): string
```
AdrienClairembault, line 54: A general remark here, we merged last week the PHP 8.3 polyfill, which mean we should try to use the `#[Override]` data attribute every time we override a method.

See https://stitcher.io/blog/override-in-php-83

AdrienClairembault (changes requested):

ccailly:
This is what it looks like with and without description. Regarding the content of the descriptions, I would like help in producing relevant descriptions.
image

orthagh:
Seems good ! The design doesn't suffer when a capacity is without description

AdrienClairembault (approved):

AdrienClairembault:
What the next step for this ?
Should it be merged as it is and then correct texts can be added later on or do you want to wait for the updates texts before merging ?

cedric-anne:
IMHO, if only a few capacities are going to have a description, it would probably be more relevant to just have a :grey_question: icon that opens a tooltip. Indeed, descriptions may b far longer than only 2 lines and in this case the current design would probably be not adapted.

AdrienClairembault:
To move forward, maybe it can be merged as it is for now (as it isn't really a ground breaking feature) ?

Then we can re-evaluate it later on when all capacities are finished and see how many capacities need description and how do we want to display them.

cedric-anne (approved):

PR #16514, opened 2024-02-05
Capacity: Has network ports

ccailly:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 

Add NetworkPorts capacity

image

### DIFF
--- a/src/Asset/Capacity/HasNetworkPortCapacity.php
+++ b/src/Asset/Capacity/HasNetworkPortCapacity.php
@@ -1,0 +1,103 @@
+<?php
+
+/**
+ * ---------------------------------------------------------------------
+ *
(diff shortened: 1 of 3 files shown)


AdrienClairembault (approved):

AdrienClairembault (approved):

cedric-anne (changes requested):
The network ports should probably be cloned when the asset is cloned (they are clone when a computer is cloned).
To do this, you have to declare this method:
```php
    public function getCloneRelations(): array
    {
        return [
            NetworkPort::class,
        ];
    }
```

A test should also be added for this. See #16454 for some inspiration.

cedric-anne (approved):

PR #16517, opened 2024-02-06
Fix label for the "associable item to..." right

AdrienClairembault:
image

Not sure if the label should say "tickets, changes and problems" or "ITIL items".
IMO, ITIL items is too vague and users might not know what it means, especially since we have an `ITIL objects` category on the same page which does not target the same items at all:

image

If we want to use "ITIL items" for the right then this section should probably be renamed to avoid confusion.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |

### DIFF
--- a/src/Profile.php
+++ b/src/Profile.php
@@ -1193,7 +1193,7 @@ class Profile extends CommonDBTM
         echo "</tr>";
 
         echo "<tr>";
-        echo "<td>" . __('Associable items to a ticket') . "</td>";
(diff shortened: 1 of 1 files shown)


orthagh:
> ITIL items

I don't find anything about that so far, but I remember we cannot use the itil term as it's somehow a brand with copyright.

cedric-anne (approved):

PR #16519, opened 2024-02-06
fix(reminder): not visible on root entity with tree structure

Rom1-B:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !31521 #13650

When the target of the note is a profile or group on the recursive root entity. The note was visible from the `Root entity` or a child, but not from the `Root entity (tree structure)`.

OK:
image

Not OK:
image

### DIFF
--- a/src/Reminder.php
+++ b/src/Reminder.php
@@ -312,7 +312,12 @@ class Reminder extends CommonDBVisible implements
             ];
 
             $or = ['glpi_groups_reminders.no_entity_restriction' => 1];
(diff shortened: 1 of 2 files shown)


cedric-anne (commented):
Seems OK but tests are failing.

cedric-anne (approved):

trasher (approved):

PR #16541, opened 2024-02-09
Allow specifying authorized OAuth redirect URIs

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

Allow specifying the authorized redirect URIs for OAuth clients.

### DIFF
--- a/install/migrations/update_10.0.x_to_10.1.0/oauth.php
+++ b/install/migrations/update_10.0.x_to_10.1.0/oauth.php
@@ -86,7 +86,7 @@ if (!$DB->tableExists('glpi_oauthclients')) {
         `name` varchar(255) NOT NULL DEFAULT '',
         `comment` text DEFAULT NULL,
         `secret` varchar(255) NOT NULL,
-        `redirect_uri` varchar(255) NOT NULL,
+        `redirect_uri` TEXT NOT NULL,
         `grants` text NOT NULL,
         `scopes` text NOT NULL,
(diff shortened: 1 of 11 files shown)


flegastelois:
LGTM, is it possible to move the inputs?
Client ID, then Client Secret (on the left).
Grants, then Comments (right).

image

cconard96:
Fields have been moved. The blank space on the right side between Grants and Comments is the Scopes field whenever we decide to define any. Since no scopes are defined, the field is hidden and replaced with a placeholder "null" field.

trasher:
Tests are failing

orthagh (approved):

flegastelois:
It seems that we cannot delete a redirect URL.

Click on the “trash” button:
image

Then save:
image

Redirect URL is still present.

cconard96:
I think this PR is OK now. The globally allowed redirect URIs are shown now to hopefully avoid any users thinking that no URIs listed means it would accept every redirect by default.
There were some other issues I noticed while testing again, but it looks like they were only on my end. Apparently, Swagger caches the authorization code, so if you log out and back in, it tries to get a new token with the now expired authorization code. There was also an error once about the state being invalid (non-blocking warning), but I've not been able to recreate the issue since then.

cedric-anne (commented):
Seems OK.

orthagh (approved):

PR #16543, opened 2024-02-10
Use HL API to handle status.php calls and drop plain-text output

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

- Forces requests to `status.php` to be directed through the new API
- Drops old pain-text output format

By having all HTTP access to the status checker going through the new API we can have it use any potential new restrictions we may add to the new API like rate limiting, IP restrictions, OAuth scopes, etc.


orthagh (approved):

cedric-anne (approved):

PR #16554, opened 2024-02-14
[HLAPI] Centralize tests and base on expectations from REST and the endpoint docs

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

The new API is declarative. That means the route documentation/definition PHP attributes on the controllers and route methods, and the schemas describe how it should function and the API code relies on that information. So, logically the vast majority of the tests for the API should also be well defined based on that documentation and basic REST behavior. These test the behavior and the presence of the correct documentation at the same time. By centralizing the baseline tests, I already managed to find and fix several issues.

Assumptions made by the new baseline 'automatic' tests:
- Search
  - A GET route exists for the endpoint (Ex: /Assets/Computer)
  - Calling the route without logging in returns a 401 error
  - The route has a response schema that represents an array
  - The route declares the filter, start and limit parameters
  - The route includes the ResultFormatterMiddleware (allows transforming the results automatically to CSV or XML based on Accept header)
  - Content-Range response header is valid by default and when specifying pagination parameters
  - Filtering returns the correct number of results
- CRUD
  - Routes exist for the endpoint with POST, GET, PATCH and DELETE methods
  - The GET route has a response schema
  - The POST and PATCH routes have a body schema
  - The DELETE route has NO body schema
  - Calling the routes without logging in returns a 401 error
  - Creating an item via the POST route succeeds, returns a JSON body with a valid ID and href, and returns a valid Location header
  - Getting the new item from the GET route (uses the Location header value) succeeds, responds with the correct schema, and responds with the correct item
  - Updating the item from the PATCH route succeeds and returns a JSON body matching the correct schema
  - Getting the item again returns the item with the updated fields
  - Deleting the item from the DELETE route succeeds and returns no body
  - If the item supports soft-deletion (is_deleted field), the item should still be accessible through the GET route and have its is_deleted field updated. Deleting the item again (or not is soft-deletion not supported), should make the GET route return a 404 error.
  
Fixes:
- Missing `entity` property for `Group` schema
- Missing documentation on multiple routes
- Incorrect schema reference resolution when schemas exist in multiple controllers and are prefixed to avoid duplicate names
- Numbers without decimals were not passing validation for `TYPE_NUMBER` or its decimal/float formats. This should probably be permitted.
- Content-Range response header when results can be paginated was not correct when the result limit would put the end of the range over the total number of results.

Changes:
- Methods in ManagementController de-duplicated now that placeholders can be used within parameters/responses
- `Task` schema for Projects renamed to `ProjectTask`


cedric-anne (dismissed):
Tests are failing. Also, please rebase to solve conflicts.

orthagh (approved):

cedric-anne (commented):

tests/HLAPITestCase.php
```diff
@@ -190,13 +299,338 @@ public function call(Request $request, callable $fn, bool $auto_auth_header = tr
...
-            $this->router->registerAuthMiddleware(new InternalAuthMiddleware());
         }
         $response = $this->router->handleRequest($request);
         $fn(new HLAPICallAsserter($this->test, $this->router, $response));
         return $this;
     }
+
+    public function autoTestCRUD(string $endpoint, array $create_params = [], array $update_params = []): self
+    {
+        $this->test->resetSession();
```
cconard96, line 310: I haven't found the reason why, but suddenly the test user is already logged in somehow with a regular session before anything in a test is run. I went back two weeks in commits on `main` with the same results.

PR #16562, opened 2024-02-15
Add unit tests for CommonAjaxController

AdrienClairembault:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |

### DIFF
--- a/src/Controller/CommonAjaxController.php
+++ b/src/Controller/CommonAjaxController.php
@@ -335,7 +335,7 @@ class CommonAjaxController
     {
         if (!$this->item->checkIfExistOrNew($id)) {
             throw new \Glpi\Controller\RequestException(
-                403,
+                404,
                 __("Item not found.")
             );
         } elseif (!$this->item->can($id, $right, $input)) {
--- a/src/Form/Form.php
+++ b/src/Form/Form.php
@@ -197,17 +197,6 @@ class Form extends CommonDBTM
     }
 
     /**
-     * Update extra form data found in other tables (sections and questions)
-     *
-     * @return void
-     */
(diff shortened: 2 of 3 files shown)


trasher (approved):
Seems correct (if test pass). Can you take on eye on the coverage?

AdrienClairembault:
> Seems correct (if test pass). Can you take on eye on the coverage?

The coverage is not perfect because I couldn't force some failures (`->update()`, `->delete()` or `->restore()` calls are hard to fail if you already validated your parameters before calling the method).

I did add comments in the tests regarding these missing cases so maybe it can be improved later on if our framework evolve.

I didn't take time to run the coverage locally as functional tests are super long to run (I plan to do it once for the whole `feat/form-editor` branch once it is ready to be merged), it would be nice if the CI could do it for us :/

trasher:
> > Seems correct (if test pass). Can you take on eye on the coverage?
> I didn't take time to run the coverage locally as functional tests are super long to run (I plan to do it once for the whole `feat/form-editor` branch once it is ready to be merged), it would be nice if the CI could do it for us :/

You can limit tests to a directory or a set of files, coverage will be faster. Also, it does not have to be 100% covered, but it can provide some indications on tests that would be missing/improved.

There is already a CI that does covering, but it seems it works only on 10.0/bf branch; and it's currently broken :D

Running whole coverage on all pull request is really not possible, CI is already taking way too much time. Also, what is of interest here is only a small subpart of the whole coverage, CI cannot detect that.

cedric-anne (approved):

PR #16574, opened 2024-02-16
fix(user prefs): inheritance of general configuration

Rom1-B:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !31688

Changing the "Action button layout" in the general config did not change its value in the customization for users who had not modified their preferences.

Since 10.0.6, all users have been created with the value "Merged", regardless of the default value in the general config.

_Same case with "Timeline date display"_


Rom1-B:
Should existing users be updated during migration? : Set null if user value = general config = 0?

orthagh:
> Should existing users be updated during migration? : Set null if user value = general config = 0?

No, we can't differentiate users impacted to those who wanted really the layout.

orthagh (approved):

cedric-anne (approved):

PR #16575, opened 2024-02-16
Fix item_item search options

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

Only affects `main` as bug was introduced while the search code was being refactored. The handling of `item_item` type search options was the same as `item_item_revert`.

10.0 handling can be found here for reference:
https://github.com/glpi-project/glpi/blob/a040236888d5e0eeb0707058be3bcbfe6953346c/src/Search.php#L6082-L6100

### DIFF
--- a/src/Search/Provider/SQLProvider.php
+++ b/src/Search/Provider/SQLProvider.php
@@ -2532,11 +2532,11 @@ final class SQLProvider implements SearchProviderInterface
                             'LEFT JOIN' => [
                                 "$new_table$AS" => [
                                     'ON' => [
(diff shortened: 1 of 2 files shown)


orthagh (approved):

cconard96:
This is not quite fixed yet it seems. Also, I guess there are no tests for this part.

cedric-anne (approved):

PR #16576, opened 2024-02-16
Add tests for Form 

AdrienClairembault:
Need to be rebased once #16567 is merged as we reuse its FormBuilder component.

Simplified the question input format from `_questions[section_index][question_index]` to `_questions[question_index]` as having the section index here was not really used.

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |

### DIFF
--- a/tests/DbTestCase.php
+++ b/tests/DbTestCase.php
@@ -222,8 +222,10 @@ class DbTestCase extends \GLPITestCase
      * @param string $itemtype
      * @param array $input
      * @param array $skip_fields Fields that wont be checked after creation
+     *
+     * @return CommonDBTM The updated item
      */
-    protected function updateItem($itemtype, $id, $input, $skip_fields = [])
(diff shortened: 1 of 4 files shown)


tests/DbTestCase.php
```diff
@@ -455,4 +462,134 @@ protected function disableCapacity(
...
     }
+
+    /**
+     * Helper method to help creating complex forms using the FormBuilder class.
+     *
+     * @param FormBuilder $builder RuleConfiguration
+     *
+     * @return Form Created form
+     */
+    protected function createForm(FormBuilder $builder): Form
```
trasher, line 473: I guess form specific should not be part of main DbTest case, a FormTest Case should be added

AdrienClairembault:
Tests failure for "PHP 8.3 using mariadb:10.11" is not related to changes

cedric-anne (approved):

PR #16588, opened 2024-02-19
Export as ODS, XLSX, CSV and PDF using `phpoffice/phpspreadsheet`

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

Working PoC: Export as ODS, XLSX, XLS, CSV and PDF from PhpSreadsheet

- CSV should certainly replace existing one, results seems the same (I kept both for the PoC). 
- I'm not really convinced by PDF exports; default style is certainly NOK, it would require some extra work. The positive point would be to rely on phpspreadsheet for all exports.
- I'm not sure XLS format is really usefull.

I've added HTML export; just because it was the only format phpspreadsheet can write that was not used.

I've also removed SYLK.


trasher:
A point to check: performances. I do not have any huge DB instance on my side; but we need to ensure (especially if we replace existing CSV version) there are no performance issues using phpspreadsheet.

orthagh:
> I'm not sure XLS format is really usefull.

No indeed

> PDF exports

Is obviously bad, keep the old one.

>  HTML export

Not sure we want that

>  removed SYLK

good

> CSV should certainly replace existing one, results seems the same (I kept both for the PoC).

Seems good.
A small thing, if a sql error is present, the old csv doesn't include it, the new one yes

trasher:
Thanks for the feedback; I'll rework to drop useless thinks and fix tests (I guess I'll try to add tests aswell).
I need to find a way to get an error :)

composer.json
```diff
@@ -50,6 +50,7 @@
         "mexitek/phpcolors": "^1.0",
         "monolog/monolog": "^3.5",
         "paragonie/sodium_compat": "^1.20",
+        "phpoffice/phpspreadsheet": "^1.29",
```
cedric-anne, line 53: A 2.0.0 version of `phpoffice/phpspreadsheet` that has been released a month ago. According to the release notes, migrating to it should be painless. Could you try to upgrade it?

trasher: Oh, I did not know there were a new release, I'll upgrade, yes.

trasher: It's done

src/Search/Output/Spreadsheet.php
```diff
@@ -0,0 +1,544 @@
...
+    }
+
+    /**
+     * Get file title
+     *
+     * @param array $data Array data of search
+     *
+     * @return string Title
+     **/
+    protected function getTitle(array $data): string
```
cedric-anne, line 219: I may miss something, but when I open the spreadsheet in LibreOffice, I do not see where the title is displayed.

If it shares the same code as `PDFSearchOutput`, maybe this whole can can be moved into the `ExportSearchOutput::computeTitle()` method. As far as I understand the code, it will not have any side effects on other export methods (that are only `NamesListSearchOutput`).

trasher: For new methods, title is set in metadata properties:
image

Since we now use phpSpreadsheet for all exports but PDF; yes I think we can move that part on parent class.

cedric-anne: > Best option would be to use phpspreadsheet to export PDF aswell; and then we can drop all old stuff.

Indeed, we could use this lib to generate PDFs.

orthagh: The rendering for pdf was bad, no ?

trasher: Yes, even if it can be changed.
Trying to factorize old echoes with new object oriented lib will be difficult. Technically speaking, the best way is to use new way everywhere; until that we can just keep legacy code.

trasher: I'll take an eye at PDF exports; I did not git into that yet.

trasher: After a bit of work, I have something that's already better than the original:
image

There is still work; it's not very difficult, but lack of documentation on PDF output specificities make some changes complicated.

trasher:
I've also replaced PDF exports.

orthagh (approved):
Seems good.

src/Search/Output/Tcpdf.php
```diff
@@ -0,0 +1,125 @@
...
+                'unit' => $unit,
+                'format' => $paperSize,
+                'font_size' => 8,
+                'font' => $_SESSION['glpipdffont'] ?? 'helvetica',
+                'margin_bottom' => 30
+            ],
+            $this->spreadsheet->getProperties()->getCustomPropertyValue('items count'),
+            null,
+            false
+        );
```
cedric-anne, line 56: Maybe the following code would prevent having to redeclare the whole `save()` method:
```suggestion
        $instance = new class (
            [
                'orientation' => $orientation,
                'unit' => $unit,
                'format' => $paperSize,
                'font_size' => 8,
                'font' => $_SESSION['glpipdffont'] ?? 'helvetica',
                'margin_bottom' => 30
            ],
            $this->spreadsheet->getProperties()->getCustomPropertyValue('items count'),
            null,
            false
        ) extends \GLPIPDF {
            public function setPrintFooter ($val = true) {
                // override the parent method to prevent footer to be disabled
            }
        };
```

trasher: Yes, that does the job, thanks

src/Search/Output/Tcpdf.php
```diff
@@ -0,0 +1,125 @@
...
+        $pdf->SetAuthor($this->spreadsheet->getProperties()->getCreator());
+        $pdf->SetSubject($this->spreadsheet->getProperties()->getSubject());
+        $pdf->SetKeywords($this->spreadsheet->getProperties()->getKeywords());
+        $pdf->SetCreator($this->spreadsheet->getProperties()->getCreator());
+
+        //  Write to file
+        fwrite($fileHandle, $pdf->output('', 'S'));
+
+        parent::restoreStateAfterSave();
+    }
```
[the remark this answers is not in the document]
cedric-anne: My bad, I was reviewing a single commit.

PR #16622, opened 2024-02-26
Add agent inventory action hooks

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

Adds 2 new hooks:
- `pre_inventory`: Called before the inventory is processed. Plugins may use this to modify the data or reject it completely.
- `post_inventory`: Called after the inventory is processed and only when it was successful.

### DIFF
--- a/src/Inventory/Request.php
+++ b/src/Inventory/Request.php
@@ -349,9 +349,14 @@ class Request extends AbstractRequest
         /** @var array $CFG_GLPI */
         global $CFG_GLPI;
 
-        $data = Plugin::doHook(Hooks::PRE_INVENTORY, $data);
-        if ($data === null) {
-            $this->addError('Rejected by a plugin', 400);
+        try {
+            $data = Plugin::doHook(Hooks::PRE_INVENTORY, $data);
+            if ($data === null) {
+                $this->addError('Rejected by a plugin: No reason specified', 400);
+                return;
+            }
+        } catch (\Exception $e) {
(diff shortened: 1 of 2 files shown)


trasher:
What's the initial need? I have no idea how it works, or how it it's expected to work.

trasher (approved):

src/Inventory/Request.php
```diff
@@ -349,6 +349,12 @@ public function inventory($data)
         /** @var array $CFG_GLPI */
         global $CFG_GLPI;
 
+        $data = Plugin::doHook(Hooks::PRE_INVENTORY, $data);
+        if ($data === null) {
+            $this->addError('Rejected by a plugin', 400);
+            return;
+        }
```
g-bougard, line 356: Why not eventually handle here an error reported by the plugin so the agent can log the real reason ?

cconard96: done

g-bougard: Thank you

orthagh (approved):

cconard96:
> What's the initial need? I have no idea how it works, or how it it's expected to work.

It would be useful for some workflows to modify the reported inventory data such as swapping the serial and asset tag in specific cases, or rejecting the inventory completely if several properties don't match an expected pattern (maybe possible with rules, but this could allow more control).

cedric-anne (approved):

PR #16623, opened 2024-02-26
Basic OAuth scopes implementation

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

Adds very basic OAuth scopes support:
- `email`: Access to the user's own default email address
- `user`: Access to the user's information
- `status`: Access to the status endpoints (Does not apply when accessed via the CLI commands)
- `inventory`: Access to submit inventory from an agent (Not implemented yet). Will eventually be used for authentication between the GLPI Agent and GLPI.
- `api`: Access to the API (All other endpoints not handled by their own scope)

The scopes are added into the OpenAPI documentation so the Swagger authentication UI has new options:
Selection_246

The authorization page in GLPI has been updated to show what information the external application is requesting:
Selection_245

The allowed scopes can be configured in the form for each OAuth Client. If you try to authenticate while specifying a scope that wasn't added in the OAuth Client you are using, it will be discarded/ignored.

OAuth scopes do NOT change the actual permissions in GLPI. So, these scopes only restrict the usual permissions for the user rather than replace or add to them.

### DIFF
--- a/src/Api/HL/Controller/AdministrationController.php
+++ b/src/Api/HL/Controller/AdministrationController.php
@@ -415,7 +415,7 @@ final class AdministrationController extends AbstractController
         return $emails;
     }
 
-    #[Route(path: '/User/me', methods: ['GET'], middlewares: [ResultFormatterMiddleware::class])]
+    #[Route(path: '/User/Me', methods: ['GET'], middlewares: [ResultFormatterMiddleware::class])]
     #[Doc\Route(
         description: 'Get the current user',
         responses: [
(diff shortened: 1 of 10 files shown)


orthagh (commented):
Not a big deal, but scope fields cannot be cleared. After saving the client, list returns at it's original state.
Seems good otherwise

cconard96:
> Not a big deal, but scope fields cannot be cleared. After saving the client, list returns at it's original state. Seems good otherwise

Seems like an issue that applies to more than just this form. Multi-select dropdowns in other parts of GLPI act the same way currently. Seems no data is sent to the server when they are emptied.

orthagh:
Can be ignored for this pr so.

orthagh (approved):

PR #16635, opened 2024-02-27
feat(forms): Adding date and time category and related question type

ccailly:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 

Add date and time question category for forms.
Possibility of switching to date mode, time mode or both.
Added option to set default value to current date/time.

image
[Short video example](https://github.com/glpi-project/glpi/assets/42278610/fe4a7785-d7a7-4ffb-8abd-cf528e3e72fa)


AdrienClairembault (changes requested):
Code looks great, well done 👍 
Technical side is OK beside some very minor points (see comments)

-----

Functional side :

1\) I have a weird issue with the "save" button that doesn't seem to work when the value is empty:

image

It does nothing when I click it here.


2\) For the UI, I think we should try to group the items like this by playing with the margins:

image

The goal would be to get something like this, where we create two distinct "groups" (first group is the "visual representation" and second is the "options"):
 image

3\) I also got this error while trying to update my form:

image

The second update work (as you told me before).
It comes from the "placeholder" input, which does not have a name and thus confuse the `formatInputsNames` method:

image

I guess the method should ignore input without names (as they wont be submitted anyway).

4\) Also, could we fill this placeholder with the data type ? It seems to accept any string without issue so it could be prefilled to "Date", "Date and time" or "Time" depending on this type.

5\) And when the "use current date" checkbox is filled, we could clear the input value (to prevent confusion) and set the placeholder to "Current date" or something like that.
Combined with the disabled input that you already implemented, I think it would be perfect :)

6\) For the others margins issues with discussed yesterday, I'll try to fix it after everything else is OK.

tests/functional/Glpi/Form/QuestionType/QuestionTypesManager.php
```diff
@@ -82,6 +82,7 @@ public function testGetCategories(): void
         $expected_categories = [
             QuestionTypeCategory::SHORT_ANSWER,
             QuestionTypeCategory::LONG_ANSWER,
+            QuestionTypeCategory::DATE_AND_TIME,
         ];
 
         // Manual array comparison, `isEqualTo`  doesn't seem to work properly
```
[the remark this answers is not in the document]
AdrienClairembault: See #16640

src/Form/QuestionType/QuestionTypeDateAndTime.php
```diff
@@ -0,0 +1,169 @@
...
+    {
+        if ($question === null) {
+            return false;
+        }
+
+        return $question->getExtraDatas()['is_default_value_current_time'] ?? false;
+    }
+
+    #[Override]
+    public function renderAdminstrationTemplate(
```
flegastelois, line 90: Missing a ```i``` for ```Adminstration```, no ? 😆

src/Form/QuestionType/QuestionTypeDateAndTimeDateTime.php
```diff
@@ -0,0 +1,149 @@
...
+                question is not null ? question.fields.default_value : '',
+                "",
+                {
+                    'full_width'            : true,
+                    'no_label'              : true,
+                    'rand'                  : rand,
+                    'disabled'              : is_default_value_current_time,
+                    'mb'                    : 'mb-2',
+                    'additional_attributes' : {
+                        'placeholder'       : 'DateTime',
```
AdrienClairembault, line 110: We need the translated value here.

src/Form/QuestionType/QuestionTypeDateAndTimeTime.php
```diff
@@ -0,0 +1,149 @@
...
+                value,
+                "",
+                {
+                    'full_width'            : true,
+                    'no_label'              : true,
+                    'rand'                  : rand,
+                    'disabled'              : is_default_value_current_time,
+                    'mb'                    : 'mb-2',
+                    'additional_attributes' : {
+                        'placeholder'       : 'Time',
```
AdrienClairembault, line 110: We need the translated value here.

src/Form/QuestionType/QuestionTypeDateAndTime.php
```diff
@@ -0,0 +1,220 @@
...
+                                    && question.find(id).is(":visible")
+                                ) {
+                                    // Remove the last input
+                                    $(id).find('input').last().remove();
+
+                                    // Reinit the flatpickr instance
+                                    $(id).flatpickr(config);
+                                }
+                            });
+                        });
```
AdrienClairembault, line 138: We must be careful to define this event handler only once to prevent useless execution (we don't want the same code to be ran 10 times if there are 10 date questions).

Maybe we could only add the handler if `$question` is `null` ?
This would mean the handler will only be initialized when the question template is first rendered.

Also, we could check that the parameter `type` is equal to our type so the code only run if the question is in fact a datetime.

AdrienClairembault (changes requested):

AdrienClairembault (approved):

src/Form/QuestionType/QuestionTypeDateAndTime.php
```diff
@@ -0,0 +1,228 @@
...
+
+        <script>
+            {# Disabled the default value input if the checkbox is checked #}
+            function handleDefaultValueCurrentTimeCheckbox_{{ rand }}(input) {
+                const isChecked = $(input).is(':checked');
+                const dateInput = $(input).parent().parent().parent().find('div .flatpickr').find('input[type="text"]');
+
+                dateInput.prop('disabled', isChecked);
+
+                {# Check if the default palceholder is saved in a data attribute #}
```
cedric-anne, line 179:
```suggestion
                {# Check if the default placeholder is saved in a data attribute #}
```

src/Form/QuestionType/QuestionTypeDateAndTime.php
```diff
@@ -0,0 +1,228 @@
...
+namespace Glpi\Form\QuestionType;
+
+use Glpi\Application\View\TemplateRenderer;
+use Glpi\Form\Question;
+use Override;
+
+/**
+ * "Date and time" questions represent date picker inputs.
+ */
+abstract class QuestionTypeDateAndTime implements QuestionTypeInterface
```
cedric-anne, line 45: 1. It would be better to have an `Abstract` prefix to the class name.
2. Having `QuestionType` prefix for each method is a bit repetitive. Maybe `XxxType` instead of `QuestionTypeXxx` would be more appropriate. @AdrienClairembault 

```suggestion
abstract class AbstractQuestionTypeDateAndTime implements QuestionTypeInterface
```

src/Form/QuestionType/QuestionTypeDateAndTimeDate.php
```diff
@@ -0,0 +1,157 @@
...
+ */
+
+namespace Glpi\Form\QuestionType;
+
+use DateTime;
+use Glpi\Application\View\TemplateRenderer;
+use Glpi\Form\Question;
+use Override;
+
+final class QuestionTypeDateAndTimeDate extends QuestionTypeDateAndTime
```
cedric-anne, line 43:
```suggestion
final class QuestionTypeDate extends QuestionTypeDateAndTime
```
or
```suggestion
final class DateType extends QuestionTypeDateAndTime
```

src/Form/QuestionType/QuestionTypeDateAndTimeDateTime.php
```diff
@@ -0,0 +1,156 @@
...
+ */
+
+namespace Glpi\Form\QuestionType;
+
+use DateTime;
+use Glpi\Application\View\TemplateRenderer;
+use Glpi\Form\Question;
+use Override;
+
+final class QuestionTypeDateAndTimeDateTime extends QuestionTypeDateAndTime
```
cedric-anne, line 43:
```suggestion
final class QuestionTypeDateTime extends QuestionTypeDateAndTime
```
or
```suggestion
final class DateTimeType extends QuestionTypeDateAndTime
```

src/Form/QuestionType/QuestionTypeDateAndTimeTime.php
```diff
@@ -0,0 +1,156 @@
...
+ */
+
+namespace Glpi\Form\QuestionType;
+
+use DateTime;
+use Glpi\Application\View\TemplateRenderer;
+use Glpi\Form\Question;
+use Override;
+
+final class QuestionTypeDateAndTimeTime extends QuestionTypeDateAndTime
```
cedric-anne, line 43:
```suggestion
final class QuestionTypeTime extends QuestionTypeDateAndTime
```
or
```suggestion
final class TimeType extends QuestionTypeDateAndTime
```

templates/components/form/basic_inputs_macros.html.twig
```diff
@@ -208,6 +208,7 @@
 {% macro date(name, value, options = {}) %}
     {% set options = {
         'rand': random(),
+        'noCalendar': false,
```
cedric-anne, line 211: `noCalendar` is really specific to `flatpickr`, `enableDate` would probably more comprehensive.
```suggestion
        'enableDate': true,
```

templates/components/form/basic_inputs_macros.html.twig
```diff
@@ -258,10 +261,13 @@
 
     {% set locale = get_current_locale() %}
     {% set date_format = options.enableTime ? 'Y-m-d H:i:S' : 'Y-m-d' %}
+    {% set date_format = options.noCalendar ? 'H:i:S' : date_format %}
     {% set alt_format = call('Toolbox::getDateFormat', ['js']) ~ (options.enableTime ? ' H:i:S' : '') %}
+    {% set alt_format = options.noCalendar ? 'H:i:S' : alt_format %}
```
cedric-anne, line 266: Maybe easier to read.
```suggestion
    {% if options.enableDate and options.enableTime %}
        {% set date_format = 'Y-m-d H:i:S' %}
        {% set alt_format = call('Toolbox::getDateFormat', ['js']) ~ ' H:i:S' %}
    {% elseif options.enableDate %}
        {% set date_format = 'Y-m-d' %}
        {% set alt_format = call('Toolbox::getDateFormat', ['js']) %}
    {% elseif options.enableTime %}
        {% set date_format = 'H:i:S' %}
        {% set alt_format = 'H:i:S' %}
    {% endif %}
```

templates/components/form/basic_inputs_macros.html.twig
```diff
@@ -292,6 +298,14 @@
 {% endmacro %}
 
 
+{% macro time(name, value, options = {}) %}
+    {{ _self.date(name, value, options|merge({
+        'enableTime': true,
+        'noCalendar': true
```
cedric-anne, line 304:
```suggestion
        'enableDate': false,
```

templates/components/form/basic_inputs_macros.html.twig
```diff
@@ -258,10 +261,13 @@
 
     {% set locale = get_current_locale() %}
     {% set date_format = options.enableTime ? 'Y-m-d H:i:S' : 'Y-m-d' %}
+    {% set date_format = options.noCalendar ? 'H:i:S' : date_format %}
     {% set alt_format = call('Toolbox::getDateFormat', ['js']) ~ (options.enableTime ? ' H:i:S' : '') %}
+    {% set alt_format = options.noCalendar ? 'H:i:S' : alt_format %}
     <script>
         $(function() {
             $("#{{ options.id }}").flatpickr({
+                noCalendar: {{ options.noCalendar ? "true" : "false" }},
```
cedric-anne, line 270:
```suggestion
                noCalendar: {{ options.enableDate ? "false" : "true" }},
```

cedric-anne (changes requested):

AdrienClairembault (approved):

cedric-anne (approved):

cedric-anne:
@ccailly Please rebase to solve conflicts.

orthagh:
1st test:
image

ccailly:
> 1st test: image

The classnames of all the questions have been renamed. Have you tried a new form?

orthagh:
Yes indeed, it works on a new form.
Please let know in PR description when existing stuff could break.

cedric-anne:
@ccailly Please rebase to solve conflicts.

ccailly:
@cedric-anne A major UI overhaul is underway, I'm drafting the PR.

ccailly:
@orthagh What do you think of the new UI? I've updated the description with new images and a video.

orthagh:
Seems good

Just one small note: "Date and time" choice placed side of "Date" + "Time" switches still make an annoying repetition. I can live with, but maybe we could find a first level label that avoid the repetition.
I suggest "Scheduling" or "Timing"

js/form_editor_controller.js
```diff
@@ -389,6 +389,13 @@ class GlpiFormEditorController
             } else if (type === "question") {
                 // The input is for a question
                 base_input_index =  `_questions[${item_index}]`;
+
+                // Check if the input is an option (has the data-glpi-form-editor-specific-question-extra-data attribute)
+                let is_option = $(input).attr("data-glpi-form-editor-specific-question-extra-data") !== undefined;
```
AdrienClairembault, line 394:
```suggestion
                const is_option = $(input).attr("data-glpi-form-editor-specific-question-extra-data") !== undefined;
```

If the value is never redefined, it should be a `const`.

src/Form/Form.php
```diff
@@ -471,6 +471,11 @@ protected function updateQuestions(): void
                 $uuid = null;
             }
 
+            // Extra data must be JSON encoded
+            if (isset($question_data['extra_data'])) {
+                $question_data['extra_data'] = json_encode($question_data['extra_data']);
+            }
+
```
AdrienClairembault, line 478: Do we need a safety here ?

Since this is a "free" json field, maybe each question type should have a "validateExtraDataInput()" method that make sure the correct values are supplied and throw an exception otherwise.

This would avoid setting invalid data and help catch bugs quickly.

src/Form/Question.php
```diff
@@ -97,6 +97,16 @@ public function getQuestionType(): ?QuestionTypeInterface
...
     }
 
+    /**
+     * Get the extra datas for the question.
+     *
+     * @return ?array
+     */
+    public function getExtraDatas(): ?array
+    {
+        return json_decode($this->fields['extra_data'] ?? "[]", true);
```
AdrienClairembault, line 107: Same here, maybe we need a `validateExtraDataValues` method in the `QuestionTypeInterface` that make sure we don't have any invalid values here and that any mandatory extra param is set.

If the method fails, the question should not be displayed in the form displayed to the end user and an error should be logged.

AdrienClairembault (changes requested):
Technical: almost ready to merge, I've just requested some minor improvements.

Regarding the UX, I think it is good enough for merge.
We can always adjust the details in future UX iterations.

AdrienClairembault (approved):

src/Form/QuestionType/QuestionTypeDateTime.php
```diff
@@ -0,0 +1,374 @@
...
+            <div class="d-flex gap-2">
+                <label class="form-check form-switch mb-0">
+                    <input type="hidden" name="is_date_enabled" value="0"
+                    data-glpi-form-editor-specific-question-extra-data>
+                    <input class="form-check-input" type="checkbox" name="is_date_enabled"
+                        id="is_date_enabled_{{ rand }}"
+                        value="1" {{ is_date_enabled ? 'checked' : '' }}
+                        onchange="handleDateAndTimeCheckbox_{{ rand }}(this)"
+                        data-glpi-form-editor-specific-question-extra-data>
+                    <span class="form-check-label">{{ labels.date|e('html') }}</span>
```
cedric-anne, line 299: `html` is the default strategy. No need to specify it.
Same remark for other occurences.

```suggestion
                    <span class="form-check-label">{{ labels.date }}</span>
```

src/Form/QuestionType/QuestionTypeDateTime.php
```diff
@@ -0,0 +1,374 @@
...
+TWIG;
+
+        $twig = TemplateRenderer::getInstance();
+        return $twig->renderFromStringTemplate($template, [
+            'question' => $question,
+            'is_date_enabled' => $this->isDateEnabled($question),
+            'is_time_enabled' => $this->isTimeEnabled($question),
+            'labels' => [
+                'date' => _sn('Date', 'Dates', 1),
+                'time' => _sn('Time', 'Times', 1)
```
cedric-anne, line 335: Will be automatically escaped by Twig.
```suggestion
                'date' => _s('Date', 'Dates', 1),
                'time' => _s('Time', 'Times', 1)
```

cedric-anne (approved):

orthagh (approved):

PR #16655, opened 2024-03-01
Cleanup document code

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

Not exhaustive:
- Drop unused `Document::uploadDocument()`
- PHPDoc fixes
- Some minor DB query optimizations
- Use self/static for static calls
- Use `::class` instead of `getType` and hard-coded class names

### DIFF
--- a/src/Document.php
+++ b/src/Document.php
@@ -60,7 +60,6 @@ class Document extends CommonDBTM
         return _n('Document', 'Documents', $nb);
     }
 
-
     /**
      * Check if given object can have Document
      *
@@ -70,31 +69,10 @@ class Document extends CommonDBTM
      *
      * @return boolean
      **/
-    public static function canApplyOn($item)
+    public static function canApplyOn($item): bool
     {
-        /** @var array $CFG_GLPI */
-        global $CFG_GLPI;
-
-       // All devices can have documents!
-        if (
-            is_a($item, 'Item_Devices', true)
-            || is_a($item, 'CommonDevice', true)
-        ) {
-            return true;
-        }
-
-       // We also allow direct items to check
-        if ($item instanceof CommonGLPI) {
-            $item = $item->getType();
-        }
-
-        if (in_array($item, $CFG_GLPI['document_types'])) {
-            return true;
-        }
-
-        return false;
+        return in_array(is_string($item) ? $item : $item::class, self::getItemtypesThatCanHave(), true);
     }
-
 
     /**
      * Get all the types that can have a document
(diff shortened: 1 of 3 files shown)


src/Document.php
```diff
@@ -158,29 +126,27 @@ public function canCreateItem()
...
 
-       // UNLINK DU FICHIER
+       // Unlink/delete the file
         if (!empty($this->fields["filepath"])) {
             if (
                 is_file(GLPI_DOC_DIR . "/" . $this->fields["filepath"])
                 && !is_dir(GLPI_DOC_DIR . "/" . $this->fields["filepath"])
                 && (countElementsInTable(
-                    $this->getTable(),
+                    self::getTable(),
```
trasher, line 149: I don't know if anything currently override Document class, but replacing `$this` should probably be replaced with `static::` rather  than `self::` the scope is not the same

src/Document_Item.php
```diff
@@ -109,17 +104,17 @@ public function prepareInputForAdd($input)
             return false;
         }
 
-       // Do not insert circular link for document
+        // Do not insert circular link for document
         if (
-            ($input['itemtype'] == 'Document')
-            && ($input['items_id'] == $input['documents_id'])
+            ($input['itemtype'] === 'Document')
```
trasher, line 109:
```suggestion
            ($input['itemtype'] === Document::class)
```

src/Document_Item.php
```diff
@@ -208,8 +197,7 @@ public function pre_deleteItem()
 
     public function post_addItem()
     {
-
-        if ($this->fields['itemtype'] == 'Ticket') {
+        if ($this->fields['itemtype'] === 'Ticket') {
```
trasher, line 200:
```suggestion
        if ($this->fields['itemtype'] === Ticket::class) {
```

src/Document_Item.php
```diff
@@ -228,15 +216,9 @@ public function post_addItem()
...
-    /**
-     * @since 0.83
-     *
-     * @see CommonDBTM::post_purgeItem()
-     **/
     public function post_purgeItem()
     {
-
-        if ($this->fields['itemtype'] == 'Ticket') {
+        if ($this->fields['itemtype'] === 'Ticket') {
```
trasher, line 221:
```suggestion
        if ($this->fields['itemtype'] === Ticket::class) {
```

trasher (approved):

cedric-anne (approved):

PR #16657, opened 2024-03-01
Fixed display of html signature tags in e-mails

Lainow:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #14500

When signing an email is configured with HTML text.
I changed the signature to HTML text, with bold characters for example.
The test email received still has the HTML tags:
`<b>My name</b><i>My business</i>`

**Before**
image

**After**
image


cedric-anne (changes requested):
All the usages of `mailing_signature` should be updated. Also, if we decided to support HTML in signatures, the best would probably to transform the configuration field into a TinyMCE editor.

It may have many impacts, so il would be preferable to do it on main branch.

PR #16678, opened 2024-03-01
Fix "save" date picker button when the input is empty

AdrienClairembault:
Before:
before

After:
after

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #number

### DIFF
--- a/js/flatpickr_buttons_plugin.js
+++ b/js/flatpickr_buttons_plugin.js
@@ -52,6 +52,17 @@ window.CustomFlatpickrButtons = (config = {}) => {
                     'class': 'btn btn-primary'
                 },
                 onClick: (e, fp) => {
+                    const current_date = new Date();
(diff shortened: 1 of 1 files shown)


cconard96:
I'm not sure this can be considered a bug. The marked date in the calendar is just the current day, but nothing is selected yet. For selecting the current date, we added the "Now" button.

AdrienClairembault:
It is subtle but this does not select the current date, it select the date displayed by the form picker (which is the current date at 12:00:00).

IMO, clicking on a "save" button should indeed save the displayed value.

orthagh (approved):

cedric-anne (approved):

PR #16681, opened 2024-03-01
DCRoom Twig

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

### DIFF
--- a/templates/generic_show_form.html.twig
+++ b/templates/generic_show_form.html.twig
@@ -50,6 +50,7 @@
    {% set item_has_pictures = item.hasItemtypeOrModelPictures() %}
    {% set field_options = {
        'locked_fields': item.getLockedFields(),
+       'rand': rand,
    } %}
 
    <div class="card-body d-flex flex-wrap">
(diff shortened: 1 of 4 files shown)


templates/generic_show_form.html.twig
```diff
@@ -137,7 +138,7 @@
                      {% set dc_breadcrumbs %}
                         {% if item is usingtrait('Glpi\\Features\\DCBreadcrumb') %}
                            {{ call(item.getType() ~ '::renderDcBreadcrumb', [
-                              item.fields['id']
+                              item.getID
```
trasher, line 141: You should avoid using `getID`, its returns non numeric value for new items:
```
An exception has been thrown during the rendering of a template ("DCRoom::renderDcBreadcrumb(): Argument #1 ($items_id) must be of type int, string given").
```

trasher: (well, using `item.fields['id']` does not solve the error (I've not been able to get that foxed on my side), but anyway using getID is unfortunately not a good idea)

trasher: Hum... OK, indeed; either I hd another method in mind, either this has changed. I guess error was from an old cache; I no longer see it

trasher (approved):
I'm not sure all is working as expected - or I have an issue in my instance/db; but it's same in main.

cedric-anne (approved):

PR #16706, opened 2024-03-04
FieldUnicity Twig UI

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #number

### DIFF
--- a/src/FieldUnicity.php
+++ b/src/FieldUnicity.php
@@ -33,6 +33,8 @@
  * ---------------------------------------------------------------------
  */
 
+use Glpi\Application\View\TemplateRenderer;
+
 /**
  * FieldUnicity Class
  **/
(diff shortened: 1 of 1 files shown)


orthagh (approved):
I don't if it's feasible, but maybe we can add a counter to the tab of duplicates

cconard96:
> I don't if it's feasible, but maybe we can add a counter to the tab of duplicates

Possible, yes. I'm not convinced about the possible performance of it though. I don't have a lot of assets or unicity checks so I don't know how long the query for displaying the data in this tab is.

cedric-anne (approved):

PR #16709, opened 2024-03-05
Adding dashboard for project and projecttask in personal and groups view

ccailly:
| Q | A
| ------------- | ---
| Bug fix?      | no 
| New feature?  | yes 
| BC breaks?    | no 
| Deprecations? | no 
| Tests pass?   | yes 
| Fixed tickets |

Two new summary tables have been added to the personal and group view, displaying information on the user's/group's unfinished projects and the current project tasks for which the user/group is assigned.

image

image

### DIFF
--- a/src/Project.php
+++ b/src/Project.php
@@ -2670,6 +2670,200 @@ class Project extends CommonDBTM implements ExtraVisibilityCriteria
     }
 
     /**
+     * Get the list of projects for a user.
+     *
+     * @param int $user_id The user ID.
+     * @param bool $search_in_groups Whether to search in groups.
+     * @param bool $search_in_team Whether to search in the team.
+     * @return array The list of project IDs.
+     */
+    public static function getProjectIDsForUser(
+        int $user_id,
+        bool $search_in_groups = true,
+        bool $search_in_team = true
+    ): array {
+        /** @var \DBmysql $DB */
+        global $DB;
+
+        $req = [
+            'FROM' => Project::getTable(),
+            'WHERE' => [
+                'OR' => [
+                    'users_id' => $user_id
+                ]
+            ]
+        ];
+
+        $groups_sub_query = new \Glpi\DBAL\QuerySubQuery([
+            'SELECT' => [
+                'groups_id'
+            ],
+            'FROM' => Group_User::getTable(),
+            'WHERE' => [
+                'users_id' => $user_id
+            ]
+        ]);
+
+        if ($search_in_groups) {
+            $req['WHERE']['OR']['groups_id'] = $groups_sub_query;
+        }
+
+        if ($search_in_team) {
+            $crit = [
+                ['itemtype' => 'User', 'items_id' => $user_id]
+            ];
+
+            if ($search_in_groups) {
+                $crit[] = ['itemtype' => 'Group', 'items_id' => $groups_sub_query];
+            }
+
+            $team_sub_query = new \Glpi\DBAL\QuerySubQuery([
+                'SELECT' => [
+                    'projects_id'
+                ],
+                'FROM' => ProjectTeam::getTable(),
+                'WHERE' => [
+                    'OR' => $crit
+                ]
+            ]);
+
+            $req['WHERE']['OR']['id'] = $team_sub_query;
+        }
+
+        $iterator = $DB->request($req);
+
+        $project_ids = [];
+        foreach ($iterator as $data) {
+            $project_ids[] = $data['id'];
+        }
+
+        return $project_ids;
+    }
+
+    /**
+     *  Show the list of projects for a user in the personal view.
+     *
+     * @param int $user_id The user ID.
+     * @return void
+     */
+    public static function showListForCentral(int $user_id): void
+    {
+        $projects = array_map(
+            fn($id) => self::getById($id),
+            self::getProjectIDsForUser($user_id, false, true)
+        );
+
+        // If no project are found, do not display anything
+        if (empty($projects)) {
+            return;
+        }
+
+        // If all project are finished, do not display anything
+        if (
+            empty(array_filter($projects, function ($project) {
+                $state = ProjectState::getById($project->fields['projectstates_id']);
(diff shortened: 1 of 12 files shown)


cconard96:
This seems like the last part that is missing for this community request:
https://glpi.userecho.com/communities/1/topics/81-projects-and-changes-visible-at-front-page

src/Project.php
```diff
@@ -2669,6 +2669,200 @@ public function showDebug()
...
+                    if ($state->fields['is_finished']) {
+                        return false;
+                    }
+                }
+
+                return true;
+            }))
+        ) {
+            return;
+        }
```
AdrienClairembault, line 2775: Since `getProjectIDsForUser` is only used here it couldn't it directly return the non finished projects ?
This can be done by adding an `active` parameter or renaming the method to `getActiveProjectIDsForUser`.

This would lead to better performances by getting the relevant projects ids with a single SQL request, without the need to iterate on its data and load a State item for each entry.

src/Project.php
```diff
@@ -2669,6 +2669,200 @@ public function showDebug()
...
+     *
+     * @param int $user_id The user ID.
+     * @return void
+     */
+    public static function showListForCentral(int $user_id): void
+    {
+        $projects = array_map(
+            fn($id) => self::getById($id),
+            self::getProjectIDsForUser($user_id, false, true)
+        );
```
AdrienClairembault, line 2754: This can lead to memory issues as we are keeping in memory multiples big projects objects.
I think it would be safer to load the relevant project directly inside the `foreach` iteration.

src/Project.php
```diff
@@ -2669,6 +2669,200 @@ public function showDebug()
...
+                break;
+            }
+
+            $name = $project->getLink();
+            $priority = CommonITILObject::getPriorityName($project->fields['priority']);
+            $percent_done = $project->fields['percent_done'] . '%';
+
+            $state = ProjectState::getById($project->fields['projectstates_id']);
+            if ($state !== false) {
+                $state_cell = '<div class="priority_block" style="border-color: ' . $state->fields['color'] . '"><span class="me-1" style="background: ' . $state->fields['color'] . '"></span>' . $state->fields['name'];
```
AdrienClairembault, line 2842: I am not sure about using a `priority_xxxx` class for something that isn't a priority.
Maybe we could create a more generic alias for this css class ?

AdrienClairembault (changes requested):

src/Project.php
```diff
@@ -2669,6 +2669,266 @@ public function showDebug()
...
+        }
+
+        $iterator = $DB->request($req);
+
+        $project_ids = [];
+        foreach ($iterator as $data) {
+            $project_ids[] = $data['id'];
+        }
+
+        return $project_ids;
```
AdrienClairembault, line 2731:
```suggestion
        return iterator_to_array($iterator, false);
```

Can probably be simplified.

src/Project.php
```diff
@@ -2669,6 +2669,266 @@ public function showDebug()
...
```
AdrienClairembault, line 2815:
```suggestion
        return iterator_to_array($iterator, false);
```

Can be simplified.

src/ProjectTask.php
```diff
@@ -1448,6 +1448,225 @@ public function showDebug()
...
+        ];
+
+        $iterator = $DB->request($req);
+
+        $projecttask_ids = [];
+        foreach ($iterator as $data) {
+            $projecttask_ids[] = $data['id'];
+        }
+
+        return $projecttask_ids;
```
AdrienClairembault, line 1498:
```suggestion
        return iterator_to_array($iterator, false);
```

Can be simplified.

src/ProjectTask.php
```diff
@@ -1448,6 +1448,225 @@ public function showDebug()
...
```
AdrienClairembault, line 1569:
```suggestion
        return iterator_to_array($iterator, false);
```

Can be simplified.

tests/functional/Project.php
```diff
@@ -496,4 +496,156 @@ public function testProjectStateColorInSearchResults(): void
         $this->integer(substr_count($html, "background-color: #000002;"))->isEqualTo(2);
         $this->integer(substr_count($html, "background-color: #000003;"))->isEqualTo(3);
     }
+
+    public function providerGetActiveProjectIDsForUser(): iterable
+    {
+        // Check if a user with no projects returns an empty array
+        yield ['user0', false, false, false, false, false, false, false, false, true];
```
AdrienClairembault, line 503: I appreciate that you are trying to test every possibilities but I am afraid the form of this provider is not ideal.
Indeed, using so many boolean parameters make it very hard to understand the test case that is being run.
This may lead to very efficient code but it costs too much readability.

In a perfect world, a test case should be super readable by containing only its parameters and the expected result.

In my opinion, here our parameters should only be:  
* The user
* One or two boolean depending on the tested method available parameters (search_in_groups, search_in_team, ...) 
* The expected projects id.

Adding key names is also needed because of these two boolean (otherwise it might to hard to remember which is which).

For example:
```php
yield [
    'user' => $user_without_projects, 
    'search_in_team' => false,
    'search_in_groups => false,
    'expected' => [] 
];
```

This require that `$user_without_projects` is set up "manually" before the `yield` statement (which may cause a bit of code duplication, but that can be minimized by using a dedicated method that can be reused by others similar tests).

AdrienClairembault (changes requested):

AdrienClairembault (approved):

orthagh (approved):

trasher (approved):

cedric-anne (approved):
I pushed some commits.

1) Please note that in main branch, data ind atabase is not anymore encoded to protect agains XSS. You have to ensure that the data you inject in HTML strings has its special chars encoded.

2) Passing the user/group ids in the AJAX query would have permit anyone to change this params in order to access data from other users/groups. Exposed data was not critical here, but it is important to design your endoints in order to prenvent this kind of issues.

PR #16714, opened 2024-03-05
Fix ObjectLock

Elric59:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #####

Hello , 
When I check the checkbox who have the id alertMe, the listener is never trigger, 
To solve this I move the echo at the end of the function and i change the eventListener

### DIFF
--- a/src/ObjectLock.php
+++ b/src/ObjectLock.php
@@ -262,7 +262,7 @@ class ObjectLock extends CommonDBTM
         $ret = Html::scriptBlock("
          $(function(){
             var lockStatusTimer;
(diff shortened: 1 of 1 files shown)


cconard96:
Seems OK. I guess this will only apply to 10.0.X and not the `main` branch since I recently reworked this feature and the original issue shouldn't apply to the new code.

trasher (approved):

cedric-anne (approved):

PR #16721, opened 2024-03-07
Groups UI Twig

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

Move UI relating to items belonging to/managed by groups, users in a group, and groups assigned to a user to twig.

Still need to re-add filter support to the lists. Will require some changes to how filters are handled in the datatables template.

List of users in a group (Before):
Selection_268

List of users in a group (After):
Selection_269

List of groups for a user (Before):
Selection_266

List of groups for a user (After):
Selection_267

### DIFF
--- a/front/group_user.form.php
+++ b/front/group_user.form.php
@@ -47,7 +47,16 @@ $group_user = new Group_User();
 
 if (isset($_POST["add"])) {
     $group_user->check(-1, CREATE, $_POST);
-    $group_user->add($_POST);
+    if ($group_user->add($_POST)) {
+        Event::log(
+            $_POST["groups_id"],
+            "groups",
+            4,
+            "setup",
+            //TRANS: %s is the user login
+            sprintf(__('%s adds a user to a group'), $_SESSION["glpiname"])
+        );
+    }
     Html::back();
 }
 
--- a/front/group_user.form.php
+++ b/front/group_user.form.php
@@ -47,16 +47,7 @@ $group_user = new Group_User();
 
 if (isset($_POST["add"])) {
     $group_user->check(-1, CREATE, $_POST);
-    if ($group_user->add($_POST)) {
-        Event::log(
-            $_POST["groups_id"],
-            "groups",
-            4,
-            "setup",
-            //TRANS: %s is the user login
-            sprintf(__('%s adds a user to a group'), $_SESSION["glpiname"])
-        );
-    }
+    $group_user->add($_POST);
     Html::back();
 }
 
--- a/src/Group_User.php
+++ b/src/Group_User.php
@@ -354,7 +354,7 @@ class Group_User extends CommonDBRelation
                 ]
             ],
             'WHERE' => [
-                    $group_users_table . '.groups_id'  => $restrict,
+                $group_users_table . '.groups_id'  => $restrict,
             ],
             'ORDERBY' => [
                 User::getTable() . '.realname',
--- a/src/Group_User.php
+++ b/src/Group_User.php
@@ -796,13 +796,15 @@ class Group_User extends CommonDBRelation
 
         parent::post_purgeItem();
 
-        Event::log(
-            $this->fields['groups_id'],
-            "groups",
-            4,
-            "setup",
-            sprintf(__('%s deletes users from a group'), $_SESSION["glpiname"])
-        );
+        if (Session::getLoginUserID() !== false) {
+            Event::log(
+                $this->fields['groups_id'],
+                "groups",
+                4,
+                "setup",
+                sprintf(__('%s deletes users from a group'), $_SESSION["glpiname"])
+            );
+        }
 
         // remove user from plannings
         $groups_id  = $this->fields['groups_id'];
--- a/src/Group_User.php
+++ b/src/Group_User.php
@@ -378,9 +378,6 @@ class Group_User extends CommonDBRelation
      */
     public static function showForGroup(Group $group)
     {
-        /** @var array $CFG_GLPI */
-        global $CFG_GLPI;
-
         $ID = $group->getID();
         if (
             !User::canView()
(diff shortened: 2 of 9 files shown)


orthagh:
A few comments:
- having red icons instead nothing seems an UI regression. Red means a warning for the admin where in this case, it's just not having an attribute attached to the user-group relation
- we may also drop the green check and the circle (personal preference) to a simple greyish `ti-check` 
- no pagination is available (i have a 914 group-user)
- can we review the add part also, I think a modal with vertical fields could be better here (the same can be said for Authorizations of user, addition action should be moved to another context)

cconard96:
>     * having red icons instead nothing seems an UI regression. Red means a warning for the admin where in this case, it's just not having an attribute attached to the user-group relation

I added them specifically because I noticed the previous checkmark images had no titles and therefore there is no data for screen readers. I'm not sure how well it would be presented in that case with a mix of empty cells and cells containing "Yes". If the UI isn't good, I can remove the icons and only have sr-only "No" spans.

orthagh:
>  no data for screen readers

Good point

> I can remove the icons and only have sr-only "No" spans.

Good idea !

orthagh:
> can we review the add part also

I think we're good apart the above point.

cconard96:
> > can we review the add part also
> 
> I think we're good apart the above point.

The old UI had an itemtype dropdown above the item list that filtered the list. I need to re-implement that in the datatable filter.

For the form, I thought this could be standardized across all tabs with datatables, at least for short forms like this.
1. The form is initially hidden behind a centered Add button unless it can be fit on a single line (like this case)
2. Clicking the add button shows the form fields, along with an Add and Cancel button. Clicking Add will create the item while Cancel hides the form and shows the original button.
3. Clicking an item in the list will hide the initial Add button and show the form with the selected item's data, along with a Save and Cancel button. This won't effect some items that are intended currently to be deleted and recreated rather than edited link simple links between two other items.
If the form is too long, it would need handled in a modal or a separate page.

cconard96:
@AdrienClairembault can you check the display regarding recursive groups?  I'm not sure it was displayed right when you filter the results. Right now, the `clearDuplicatedGroupData` calls are commented out and seem OK for me, but I guess that isn't the desired behavior.

AdrienClairembault:
Its been a while since the recursive groups dev but I think the goal is to avoid having twice the same group in the list:

image

Case `A` is a group directly assigned to the user, while `B` is a recursive assignation.

If I un-comment `clearDuplicatedGroupData` then there is no duplicated data.

Regarding filtering, maybe it was already broken before (I wasn't aware of this feature).
I've also noticed that sorting does not work (unrelated to whether or not `clearDuplicatedGroupData` is commented).

cconard96:
> Its been a while since the recursive groups dev but I think the goal is to avoid having twice the same group in the list:
> 
> Case `A` is a group directly assigned to the user, while `B` is a recursive assignment.
> 
> If I un-comment `clearDuplicatedGroupData` then there is no duplicated data.
> 
> Regarding filtering, maybe it was already broken before (I wasn't aware of this feature). I've also noticed that sorting does not work (unrelated to whether or not `clearDuplicatedGroupData` is commented).

The filters before were just a dropdown to toggle between manager, delegate and nothing, and a yes/no dropdown to show child groups.
With this PR, child groups are always pulled in the data as I wanted to delegate that to the filters in the datatable. IMO, the "duplicates" should always be shown by default. Since there is a column that says which group the association is for, there shouldn't be an issue.
The issue I noticed was that I had a user assigned to a parent group and a child group. In the parent group, none of the flags were set to Yes, while they were in the child group. The parent was made recursive. By default, I saw only their membership in the parent group. When filtering to show managers, the previously shown result for that user disappeared and was replaced with the one in the child group which matched the filter.

AdrienClairembault:
> IMO, the "duplicates" should always be shown by default. Since there is a column that says which group the association is for, there shouldn't be an issue.

I don't have a strong opinion on this. @orthagh should probably clarify what's best.

Did you see my comment on sorting being broken ? This part must be fixed before it can be merged IMO.

orthagh:
> For the form, I thought this could be standardized across all tabs with datatables, at least for short forms like this.

OK, that seems fair.
Could you make a bit more spacing between the form and the list (maybe also an hr)


> IMO, the "duplicates" should always be shown by default

I agree

> If time permits, all of the datatables can be iterated on to support more filters, sorting, pagination, etc.

Please note this as a to-do. 
I'm ok to merge as it for the moment. Just we shouldn't forget it

AdrienClairembault (approved):

src/Group_User.php
```diff
@@ -398,23 +310,39 @@ public static function getDataForGroup(
...
+            'delegatee' => "$group_users_table.is_userdelegate",
+            'is_active' => 'glpi_users.is_active',
+        ];
+        if (is_array($crit)) {
+            foreach ($crit as $key => $value) {
+                if (!empty($value) && isset($filter_map[$key])) {
+                    $where_filters[$filter_map[$key]] = $value;
+                }
+            }
+        }
```
cedric-anne, line 326: `$where_filters` are never used.

front/group_user.form.php
```diff
@@ -47,16 +47,7 @@
...
     $group_user->check(-1, CREATE, $_POST);
-    if ($group_user->add($_POST)) {
-        Event::log(
-            $_POST["groups_id"],
-            "groups",
-            4,
-            "setup",
-            //TRANS: %s is the user login
-            sprintf(__('%s adds a user to a group'), $_SESSION["glpiname"])
-        );
```
cedric-anne, line 58: This should probably stay in the front page. Event logs are mostly made from front files, and, for the moment, we do not have time to think about how we could/should enhance this.

trasher (changes requested):
Tests are failing, I guess because there is no logged-in user

trasher (approved):

PR #16731, opened 2024-03-08
Forms: display all actions labels

AdrienClairembault:
Suggested by @orthagh.

Before:
image
image

After:
image
image

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |

### DIFF
--- a/js/form_editor_controller.js
+++ b/js/form_editor_controller.js
@@ -506,7 +506,14 @@ class GlpiFormEditorController
         const tab_content_border = 1;
 
         // Compute and apply ideal height
-        const height = (window_height - editor_height - tab_content_border);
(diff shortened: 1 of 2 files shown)


orthagh (approved):

AdrienClairembault:
@orthagh do you also have a suggestion for the preview button ?
You said it didn't have enough visibility, would "unghosting" it be enough ?

image

We could handle that in this PR too if we can find a quick solution.

orthagh:
>  would "unghosting" it be enough ?

Probably yes !

AdrienClairembault:
Done, I've also fixed the issue you reported on the debug bar.

PR #16734, opened 2024-03-09
Knowbase Twig UI

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -


orthagh (changes requested):
When I saw the title, I was afraid of the result :)
As it's a big system, it's easy to forget parts.
Anyway, the transition to twig seems to be good with a few quirks (some present before this PR, but I guess we may fix them now):

- Toggle browse (in the "Browse" tab) switch to the "Search" tab and doesn't keep the value (nor change the tree state)
- Same for the "Show unpublished" control.
- In the article view, there's a tab switching issue after saving, but I guess it's the issue we have in the main branch (maybe you could rebase the current PR)
- for comment system:
   image
   - Can we make the "New comment" section under a button? Having the full form in permanence pollutes the UI flow
   - The Add+Cancel buttons should have a distinct style (primary, ghost-secondary) and probably be inverted
   - A "new comment" form (as the top level, or as a reply) should have a distinct background.
   - same for a block of readable content (we may re-use `itilfollowup` style)
   - The photo of the user should be placed a few pixels lower.
   - The "Edit" and "Reply" mini buttons should be placed side to side.

cconard96:
> When I saw the title, I was afraid of the result :) As it's a big system, it's easy to forget parts. Anyway, the transition to twig seems to be good with a few quirks (some present before this PR, but I guess we may fix them now):

I had a context saved in PHPStorm with maybe 60-70 files I found with HTML directly echoed. Rather than having 60-70 PRs, I tried to group changes. This should be the last big twig migration PR. Afterwards, there are still approximately 40 files I previously identified left to review/migrate.

cconard96:
New comment design:
Selection_280

orthagh (approved):
We're good on this one

cedric-anne:
@cconard96 Could you rebase to solve conflicts ?

trasher:
Tests are fzailing (there is also a CI fix to be done)

PR #16748, opened 2024-03-11
add option to show pagination on top of search results

orthagh:
Requester after an early internal review of 11.0 UI.

image

### DIFF
--- a/install/migrations/update_10.0.x_to_11.0.0/user.php
+++ b/install/migrations/update_10.0.x_to_11.0.0/user.php
@@ -45,8 +45,10 @@ if (!$DB->fieldExists('glpi_users', 'is_notif_enable_default')) {
 $migration->addConfig(['is_notif_enable_default' => 1]);
 
 // search ux improvements (#15861)
-$migration->addField('glpi_users', 'show_search_form', "tinyint DEFAULT 0");
+$migration->addField('glpi_users', 'search_pagination_on_top', "tinyint DEFAULT 0");
+$migration->addField('glpi_users', 'search_footer_on_top', "tinyint DEFAULT 0");
 $migration->dropField('glpi_users', 'fold_search');
 
-$migration->addConfig(['show_search_form' => 0]);
+$migration->addConfig(['search_pagination_on_top' => 0]);
+$migration->addConfig(['search_footer_on_top' => 0]);
 Config::deleteConfigurationValues('core', ['fold_search']);
--- a/install/migrations/update_10.0.x_to_11.0.0/user.php
+++ b/install/migrations/update_10.0.x_to_11.0.0/user.php
@@ -45,7 +45,7 @@ if (!$DB->fieldExists('glpi_users', 'is_notif_enable_default')) {
 $migration->addConfig(['is_notif_enable_default' => 1]);
 
 // search ux improvements (#15861)
-$migration->addField('glpi_users', 'show_search_form', "tinyint DEFAULT 0");
+$migration->addField('glpi_users', 'show_search_form', "tinyint DEFAULT NULL");
 $migration->addField('glpi_users', 'search_pagination_on_top', "tinyint DEFAULT NULL");
 $migration->dropField('glpi_users', 'fold_search');
 
--- a/install/migrations/update_10.0.x_to_11.0.0/user.php
+++ b/install/migrations/update_10.0.x_to_11.0.0/user.php
@@ -46,7 +46,7 @@ $migration->addConfig(['is_notif_enable_default' => 1]);
 
 // search ux improvements (#15861)
 $migration->addField('glpi_users', 'show_search_form', "tinyint DEFAULT 0");
-$migration->addField('glpi_users', 'search_pagination_on_top', "tinyint DEFAULT 0");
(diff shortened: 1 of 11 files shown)


AdrienClairembault:
Maybe the items should be centered ? 

image

This make the header feel "crowded" but since it is not the default option I am OK with it.

cconard96:
This conflicts with the new status message area for mail collectors and automatic actions.

Selection_279

orthagh:
> This conflicts with the new status message area for mail collectors and automatic actions.

I switched the position of the two blocks, it should be better (or at least, not so bad). As we have only receivers with this kind of display, I guess it's ok. Alerts in the header are not ideal, but for the moment any alternative was worse (double scrollbars for example).

> Maybe the items should be centered ?

Done

> This make the header feel "crowded"

I tried with the pager between the global header and the columns header, it's harder to achieve with 3 stickies and it makes the space reserved for headers a bit too much.
I take any alternative suggestions.

cconard96:
> I take any alternative suggestions.
Maybe the status alerts shouldn't be shown in the search pages at all. We have a message area on the central homepage, even if it is currently used for post-install/post-update messages, they could fit in there. Although, that could be an issue for people who hide the central messages to ignore the optional migration warnings.

The only other suggestion I have is a lot more work and a new feature where there is a sort of notification area/inbox along the top of the page next to the user dropdown where these kinds of alerts can be shown.

AdrienClairembault:
Spacing feels a bit weird:

image

Shouldn't (1) and (2) be the same size ?

AdrienClairembault:
And the number of rows can't be changed anymore:

row_number

It works fine without the option, definitely related to these changes.

orthagh:
> Spacing feels a bit weird:

Fixed

> And the number of rows can't be changed anymore

fixed

Current iteration: 
image

templates/components/pager.html.twig
```diff
@@ -72,20 +74,26 @@
         'no_onchange': fluid_search|default(false),
         'select_class': 'search-limit-dropdown',
     }) %}
-    <span class="search-limit d-none d-md-block">
-        {{ __('%s rows / page')|format(limitdropdown)|raw }}
-    </span>
-    <span class="search-limit d-block d-md-none">
+    {% if not short_display %}
+        <span class="search-limit d-none d-md-block">
+            {{ __('%s rows / page')|format(limitdropdown)|raw }}
```
cconard96, line 79:
```suggestion
            {{ __('%s rows / page')|e|format(limitdropdown)|raw }}
```
I think if you format a localization string and then output as raw, you need to escape the text just before formatting in the rare case that a translation can have HTML characters.

AdrienClairembault (changes requested):
I am not sure if it is related to your changes but this button has a very thin box shadow that should be removed: 
image.

The pagination actions are not aligned correctly:
image


You need to add "display: flex" to each `<a>` tags:
image

orthagh:
Done

AdrienClairembault (approved):

install/mysql/glpi-empty.sql
```diff
@@ -7825,6 +7825,7 @@ CREATE TABLE `glpi_users` (
   `password_history` longtext,
   `is_notif_enable_default` tinyint DEFAULT NULL,
   `show_search_form` tinyint DEFAULT 0,
+  `search_pagination_on_top` tinyint DEFAULT 0,
```
cedric-anne, line 7828: Default value for users should be `null`. Indeed, this `null` value will permit to fallback to the value configured globally in the `glpi_configs` table.

```suggestion
  `search_pagination_on_top` tinyint DEFAULT NULL,
```

By the way, it looks like the `show_search_form` default value should be fixed.

cconard96 (approved):

install/migrations/update_10.0.x_to_11.0.0/user.php
```diff
@@ -46,7 +46,9 @@
 
 // search ux improvements (#15861)
 $migration->addField('glpi_users', 'show_search_form', "tinyint DEFAULT 0");
+$migration->addField('glpi_users', 'search_pagination_on_top', "tinyint DEFAULT 0");
```
trasher, line 49:
```suggestion
$migration->addField('glpi_users', 'search_pagination_on_top', "tinyint DEFAULT NULL");
```

install/migrations/update_10.0.x_to_11.0.0/user.php
```diff
@@ -46,7 +46,9 @@
 
 // search ux improvements (#15861)
 $migration->addField('glpi_users', 'show_search_form', "tinyint DEFAULT 0");
```
cedric-anne, line 48:
```suggestion
$migration->addField('glpi_users', 'show_search_form', "tinyint DEFAULT NULL");
```

install/mysql/glpi-empty.sql
```diff
@@ -7825,6 +7825,7 @@ CREATE TABLE `glpi_users` (
   `password_history` longtext,
   `is_notif_enable_default` tinyint DEFAULT NULL,
   `show_search_form` tinyint DEFAULT 0,
```
cedric-anne, line 7827:
```suggestion
  `show_search_form` tinyint DEFAULT NULL,
```

cedric-anne (approved):

PR #16222, opened 2023-12-19
Fix(inventory): add history on Computer_Item add when Mainasset already exist

stonebuzz:
Add missing ```Log``` when ```Computer_Item``` is added (and when ```MainAsset``` exist because we log nothing on first inventory)

image

NB : when ```Computer_Item``` is removed ```Log``` is well added 

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | [433](https://github.com/glpi-project/glpi-inventory-plugin/issues/433)

PR #16750, opened 2024-03-12
Fix(Inventory) add history on Item_Devices add / update / delete when Mainasset already exist

stonebuzz:
Same as https://github.com/glpi-project/glpi/pull/16222 but for ```Item_Device``` 

image

I'm leaving it in draft form until I can see what impact it will have on the TUs for logs.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !31950


trasher:
Will wait for test to be all fixed; it does not seems there is a significant amount of added logs; so that seems OK

cedric-anne (approved):

PR #16753, opened 2024-03-12
Ignore invalid questions types in the form editor

AdrienClairembault:
The form editor currently "crashes" when an invalid question type is supplied.

This is not good because it might be a valid scenario (e.g. a type is defined by a plugin which was later disabled).
Thus it is better to simply ignore the question in this case.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |

### DIFF
--- a/src/Form/Section.php
+++ b/src/Form/Section.php
@@ -96,6 +96,12 @@ final class Section extends CommonDBChild
                 $question = new Question();
                 $question->getFromResultSet($row);
                 $question->post_getFromDB();
+
+                if ($question->getQuestionType() === null) {
+                    // The question might belong to a disabled plugin
+                    continue;
+                }
+
                 $this->questions[$row['id']] = $question;
             }
         }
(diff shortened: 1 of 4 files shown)


orthagh (approved):

src/Form/Section.php
```diff
@@ -96,6 +96,12 @@ public function getQuestions(): array
                 $question = new Question();
                 $question->getFromResultSet($row);
                 $question->post_getFromDB();
+
+                if ($question->getQuestionType() === null) {
+                    // The question might belong to a disabled plugin
```
trasher, line 101: Maybe add an entry in logs in that case? Or we're 100% sure this kind of issue can only came from a disabled plugin?
Anyway, maybe it should be precised somewhere for the form to be adjusted (or to enable again the faultly plugin).

AdrienClairembault: For now I've silently ignored any unknown type that may come from a plugin (question type, question type category, form destination, ...).

I guess it could still be logged (maybe a notice ?), I can add a backlog entry to do that if that is OK for everyone.

AdrienClairembault: Added to backlog: https://github.com/orgs/glpi-project/projects/7/views/1?pane=issue&itemId=56177382

trasher (approved):

PR #16756, opened 2024-03-12
Fix tinymce config conflicts in the form editor

AdrienClairembault:
When a question/section is moved, we need to reinitialize its tinymce instance.
It was however failing in some cases and I didn't know why.

It turns out it was just some config conflitcs, I forgot that javascript use references for object by default.
| Q             | A

| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |

### DIFF
--- a/js/form_editor_controller.js
+++ b/js/form_editor_controller.js
@@ -741,7 +741,10 @@ class GlpiFormEditorController
         copy.find("textarea").each(function() {
             // Get editor config for this field
             let id = $(this).attr("id");
-            const config = window.tinymce_editor_configs[id];
+
+            // JS object are passed by reference, we need to clone the config
+            // to avoid breaking previous instances
+            const config = { ...window.tinymce_editor_configs[id]};
 
             // Rename id to ensure it is unique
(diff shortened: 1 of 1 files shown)


cconard96:
If there are nested objects in the config, you can serialize to JSON and then back or use structuredClone to do a deep clone. With the spread syntax, nested objects should still be linked by reference.

https://developer.mozilla.org/en-US/docs/Web/API/structuredClone

AdrienClairembault:
> If there are nested objects in the config, you can serialize to JSON and then back or use structuredClone to do a deep clone. With the spread syntax, nested objects should still be linked by reference.
> 
> https://developer.mozilla.org/en-US/docs/Web/API/structuredClone

Thanks for the suggestion, however it does not seems to work as the object contains functions:

image

The config does not seem to define nested object for now, but I guess can still handle it to be future-proof.
`_.cloneDeep` seems to be able to do the work.

cedric-anne (approved):

PR #16797, opened 2024-03-17
Lock Twig UI

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

Relies on #16796.

- Migrate the lists on an asset's Locks tab to Twig
- New datatable template features
  - Ability to hide regular column headers to accommodate nested datatables (each nested table has its own columns and the parent table only has one column containing the subtables).
  - Ability to toggle massive action checkboxes per-row
  - Ability to change the table class used for styling (table-hover, table-sm, etc)

### DIFF
--- a/src/Lock.php
+++ b/src/Lock.php
@@ -89,7 +89,7 @@ class Lock extends CommonGLPI
 
         // language=Twig
         $list_info_alert_template = <<<TWIG
-            <div class="alert alert-primary d-flex align-items-center" role="alert">
+            <div class="alert alert-info d-flex align-items-center" role="alert">
                 <i class="fas fa-info-circle fa-xl"></i>
                 <span class="ms-2">
                     <span class="alert-title">{{ alert_title }}</span>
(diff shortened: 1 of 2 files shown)


trasher (approved):

src/Lock.php
```diff
@@ -79,23 +80,31 @@ public static function showForItem(CommonDBTM $item)
...
-        echo "<i class='fas fa-info-circle fa-xl'></i>";
-        echo "<span class='ms-2'>";
-        echo __("A locked field is a manually modified field.");
-        echo "<br>";
-        echo __("The automatic inventory will no longer modify this field, unless you unlock it.");
-        echo "</span>";
-        echo "</div>";
+        // language=Twig
+        $list_info_alert_template = <<<TWIG
+            <div class="alert alert-primary d-flex align-items-center" role="alert">
```
orthagh, line 92:
```suggestion
            <div class="alert alert-info d-flex align-items-center" role="alert">
```

Not a recent issue, but the style primary for alert is unreadable for light theme (especially with auror theme)

orthagh (changes requested):
Another thing, can you try to space a bit the two tables to achieve a visual distinction

orthagh (approved):

PR #16807, opened 2024-03-19
Form destinations: basic configuration

AdrienClairembault:
Add basic configuration for forms destination:

image

Once merged, the next steps are:
- Enable dynamic values based on the users answers.
- Add some kind of "auto" mode that auto compute these dynamic values (e.g. the ticket content will list all available questions/answers when in auto mode). The goal is that 99% of the users can just let "auto" everywhere and get a perfectly working form, leaving manual configuration for the 1% power users.
- Support all fields (will need the relevant question types to be added first, e.g. it is not possible to properly add the ticket type (request/incident) field config before the "request type" question type is added to forms).

The UI is not final but you can start making UI/UX suggestion if you want.

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |

### DIFF
--- a/src/Form/Destination/CommonITILField/CommonITILFieldInterface.php
+++ b/src/Form/Destination/CommonITILField/CommonITILFieldInterface.php
@@ -33,49 +33,42 @@
  * ---------------------------------------------------------------------
  */
 
-namespace Glpi\Form\Destination;
+namespace Glpi\Form\Destination\CommonITILField;
 
-use Glpi\Form\AnswersSet;
-use Glpi\Form\Form;
-
-interface FormDestinationInterface
+interface CommonITILFieldInterface
 {
     /**
-     * Create one or multiple items for a given form and its answers
+     * Get the unique key used to set/get this field configuration in the
+     * destination JSON configuration.
      *
-     * @param Form       $form
-     * @param AnswersSet $answers_set
-     * @param array      $config
-     *
-     * @return \CommonDBTM[]
-     *
-     * @throws \Exception Must be thrown if the item can't be created
+     * @return string
      */
-    public function createDestinationItems(
-        Form $form,
-        AnswersSet $answers_set,
-        array $config
-    ): array;
-
+    public function getKey(): string;
 
     /**
-     * Render the configuration form for this destination type.
+     * Label to be displayed when configuring this field.
      *
-     * @return string The rendered HTML content
+     * @return string
      */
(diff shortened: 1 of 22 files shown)


AdrienClairembault:
Updated UX with your suggestions @orthagh:

image

AdrienClairembault:
> If there are a lot of destinations already created, maybe a label can help to differentiate they

This will be possible later (https://github.com/orgs/glpi-project/projects/7/views/1?pane=issue&itemId=56970470).

orthagh:
The last thing I may have to say is we have a white panel in a white panel. If we may have some background for the items (or the tab) to differentiate from the whole tab content, it would be a visual enhancement.

I tried a few classes but didn't succeed in finding something that pleased me. If you want to take a shot.

Side note, @cconard96 It seems the title of the active accordion item has a dark foreground on a dark background

AdrienClairembault:
In tabler's demo they also use a white on white background for accordions:

image

I agree that it is not 100% ideal but I don't think there are better alternative in cases like these.

src/Form/Destination/CommonITILField/CommonITILFieldInterface.php
```diff
@@ -0,0 +1,74 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ * ---------------------------------------------------------------------
+ */
+
+namespace Glpi\Form\Destination\CommonITILField;
+
+interface CommonITILFieldInterface
```
cedric-anne, line 38: It looks like it could be used for any type of destinations, not only by ITIL destinations. Could be named `Glpi\Form\Destination\ConfigFieldInterface`.

AdrienClairembault: Done: [89d820b](https://github.com/glpi-project/glpi/pull/16807/commits/89d820bf771e53307c7e3c99edf1479b48bcbcae)

cedric-anne (approved):

orthagh:
Just one last thing, after adding a new item, the accordion should be opened on it (especially with everything named "Ticket")

orthagh (approved):

trasher (approved):

PR #16838, opened 2024-03-27
[HLAPI] Get used and managed items for user

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

Endpoints to search across all `linkuser_types` or `linkuser_tech_types` where the given user is linked. Each endpoint is simply a search on a 'union' schema for the related itemtypes and a forced RSQL filter for the relevant user ID property. Items that are used/managed through group memberships are not included. As with all other search endpoints, it supports pagination, additional filters, and optionally transforming the output via the Accept header to CSV or XML.

- Added helper function to generate these 'union' schemas.
- Fixed issue where 'empty' entity restrictions generates invalid SQL with empty brackets `(())` in the WHERE.


orthagh (approved):
Don't forget to add SORT parameter.
Apart that, ok for me

cedric-anne (approved):

PR #16875, opened 2024-04-02
Fix project search criteria and state handling in personnal and group dashboard

ccailly:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 

This pull request addresses issues with the search criteria and state handling in the Project class, which were introduced in the initial implementation of feature #16709.

- The search criteria now correctly differentiates between User and Group item types
- The search criteria now also takes into account the state of the project. Previously, finished projects were included in the search results.
- Add search options on users/groups/suppliers/contact for project tasks

### DIFF
--- a/src/Project.php
+++ b/src/Project.php
@@ -2838,18 +2838,37 @@ class Project extends CommonDBTM implements ExtraVisibilityCriteria
             'criteria' => [
                 [
                     'link' => 'AND',
-                    'field' => 87,
-                    'searchtype' => 'equals',
-                    'value' => 'myself'
-                ],
-                [
-                    'link' => 'OR',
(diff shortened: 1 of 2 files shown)


AdrienClairembault (changes requested):
Code is great but I still have some inconsistencies:

image

When I click the "count" button, I only have one real result:

image

I think "t2" shouldn't be seen on my personal view because I am not assigned to it:

image

AdrienClairembault (approved):

orthagh (approved):

PR #16879, opened 2024-04-02
Fix mailing to project team members to task creation and update

Lainow:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !32275

Despite enabling notifications triggered by the creation, update and deletion of a project task, none of them are in the queue, with the exception of the one sent to the administrator.

I've modified the addTeam... methods so that they search for project members rather than project task members. 
This corrects the sending of notifications to project members.

### DIFF
--- a/src/NotificationTargetProjectTask.php
+++ b/src/NotificationTargetProjectTask.php
@@ -125,10 +125,10 @@ class NotificationTargetProjectTask extends NotificationTarget
 
         $iterator = $DB->request([
             'SELECT' => 'items_id',
-            'FROM'   => 'glpi_projecttaskteams',
+            'FROM'   => 'glpi_projectteams',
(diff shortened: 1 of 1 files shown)


stonebuzz:
As we saw together,

the recipients correspond to the task teams because the notification concerns the ```ProjectTask``` itemtype 

As this is the first time this (false) problem has been reported, we can legitimately assume that the current behaviour is the desired behaviour (at least that's how the community uses it).

Of course, taking the task group at creation is not possible in version x, but it is now possible in main.

So we need to : 
- adapt the wording to specify that it is the task teams (```task group```)
- add two new recipients (```project group```)

PR #16880, opened 2024-04-03
Fix(User): fix default group / profile / entity security

stonebuzz:
The check on the default```entity``` / ```profile``` and ```group``` does not take account of data after the rules (```LDAP``` / ```Right```)

This PR fix this

| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !32328

### DIFF
--- a/src/User.php
+++ b/src/User.php
@@ -1083,16 +1083,16 @@ class User extends CommonDBTM
             $_SESSION["glpidefault_entity"] = $input["entities_id"];
         }
 
-        // Security on default profile update
-        if (!isset($input['_ruleright_process'])) {
-            if (isset($input['profiles_id'])) {
+        // Security on default values (manually update)
+        if (!isset($input['_ruleright_process'])) { // if no rule right process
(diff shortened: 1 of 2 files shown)


trasher (approved):

cedric-anne (changes requested):
The code will be executed only if the rules evaluated. Indeed, it is located in the `applyRightRules()` method and has the following specific condition now:
```
        if (
            isset($this->fields['_ruleright_process'])
            || isset($this->input['_ruleright_process'])
        ) {
```

Also, it is now executed from the `post_updateItem()` method, so it would not block a default profile/default entity update that would be made from the UI.

trasher:
> ```
>         if (
>             isset($this->fields['_ruleright_process'])
>             || isset($this->input['_ruleright_process'])
>         ) {
> ```

(note this condition has been changed in bugfixes branch recently)

PR #16889, opened 2024-04-03
Prevent invalid lock config

AdrienClairembault:
Some users ran into trouble when they modified the `Read-Only` profile to be restricted to the simplified interface.

This create major issues with the locking mechanism (as this profile is the default `Profile to be used when locking items` on a fresh install).

Trying to open a locked item will then log out the user because GLPI try to load a self-service profile while being on a central restricted page (e.g. a computer):

image

---

To prevent this, I've added two restrictions:
* The current "Profile to be used when locking items" can't be set to use the simplified interface.
* In the general configuration, simplified interface profiles can't be set as the "Profile to be used when locking items".

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !32346

### DIFF
--- a/src/Config.php
+++ b/src/Config.php
@@ -298,6 +298,23 @@ class Config extends CommonDBTM
             $input['glpinetwork_registration_key'] = trim($input['glpinetwork_registration_key']);
         }
 
+        // Prevent invalid profile to be set as the lock profile.
+        // User updating the config from GLPI's UI should not be able to send
+        // invalid values but API or manual HTTP requests might be invalid.
+        if (isset($input['lock_lockprofile_id'])) {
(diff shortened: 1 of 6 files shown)


trasher (approved):
lgtm, but I do not understand the template change

AdrienClairembault:
> lgtm, but I do not understand the template change

If you are talking about this:
image

It restricts the dropdown values to profile with the 'central' interface.

It should thus avoid any user that edit the config from the UI to input a bad value.

There is still a check behing the scene because the config can be updated from other sources like the API.

cedric-anne (approved):

orthagh (approved):

PR #16905, opened 2024-04-08
Add ability to define new menu sections (assets families)

orthagh:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

### DIFF
--- a/front/asset/assetmenu.form.php
+++ b/front/asset/assetmenu.form.php
@@ -1,0 +1,38 @@
+<?php
+
+/**
+ * ---------------------------------------------------------------------
+ *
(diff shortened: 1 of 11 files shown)


orthagh:
Resulting UI has issues:
- horizontal menus can be broken be addition of 1st level entry
- there is not so much interest to add assets to another menu than Assets
- make families like generic object has bad UX.

So let's close this for the moment, it's not that advanced.
If we found a better UX later, let's start again.

PR #16916, opened 2024-04-10
Fix unclickable images in tickets followups

RomainLvr:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !31348

image

Before:
image

After:
image


cedric-anne (commented):
For the moment, we do not convert images having a `http(s)` URI. I am not sure we should support it.

Anyway, this is not a bug, and changing it should target the main branch.

PR #16948, opened 2024-04-19
Ticket_List_Color_by_Category

AlthausOV:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no assumaebly yes
| Deprecations? | unknown
| Tests pass?   | unknown
| Fixed tickets | #number

Hello!

﻿We have an enhancement, where we can define a color by category.

﻿This color is used in the ticket liste for the row background.

This work was don for Version 10.0.14,
not be me :-/

The DB scheme has to be adopted like this:
ALTER TABLE `glpi_itilcategories` ADD `color` VARCHAR(255) CHARACTER SET utf32 COLLATE utf32_unicode_ci NULL DEFAULT NULL AFTER `color`

i dind't find a place in the source for doing that.

This should be read as a proposal...

Greetings from Germany,

Stephan

### DIFF
--- a/src/ITILCategory.php
+++ b/src/ITILCategory.php
@@ -97,6 +97,12 @@ class ITILCategory extends CommonTreeDropdown
                 'type'      => 'bool',
                 'list'      => true,
             ],
+            [
+                'name'  => 'color',
+                'label' => __('Color'),
+                'type'  => 'color',
+            ],
+
         ];
 
         $show_for_problem = Session::haveRightsOr('problem', [CREATE, UPDATE, DELETE, Problem::READALL, Problem::READMY]);
--- a/src/Search.php
+++ b/src/Search.php
@@ -151,6 +151,10 @@ class Search
      **/
     public static function show($itemtype)
     {
+        global $ticketcolor;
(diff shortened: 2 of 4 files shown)


AlthausOV:
image

AlthausOV:
Please be patient with me.

The person who wrote this code has unfortunately left our company,
so this is more or less all i have to provide.
As it seems i will look into our local GLPI instance from now on,
and i have to learn about GLPI.

Regards,
Stephan

trasher:
Hi,

No idea if that feature will be accepted or not but still I have several remarks:
- no feature will be accepted in 10.0/bugfixes branch, you have to target main branch,
- modified sql scripts must be provided (`install/mysql/glpi-empty.sql` and upgrade script - see existing files in `migrations/update_10.0.x_to_11.0.0/`)
- take a look at the diff, and please clean it - it's not easy to read as is.

Best regards,

trasher:
Please use [suggestion website](http://suggest.glpi-project.org) for RFE. I close since there were no feedback here, and PR is not OK.

PR #16956, opened 2024-04-19
Clean RSS code

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

### DIFF
--- a/src/RSSFeed.php
+++ b/src/RSSFeed.php
@@ -38,26 +38,6 @@ use Glpi\RichText\RichText;
 use Glpi\Toolbox\URL;
 use SimplePie\SimplePie;
 
-// $feed = new SimplePie();
-// $feed->set_cache_duration(3600);
-// $feed->set_feed_url('http://linuxfr.org/news.atom');
-// $feed->force_feed(true);
-// // Initialize the whole SimplePie object.  Read the feed, process it, parse it, cache it, and
-// // all that other good stuff.  The feed's information will not be available to SimplePie before
-// // this is called.
-// $success = $feed->init();
-//
-// // We'll make sure that the right content type and character encoding gets set automatically.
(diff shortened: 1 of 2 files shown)


src/RSSFeed.php
```diff
@@ -314,11 +216,7 @@ public static function getVisibilityCriteria(bool $forceall = false): array
         }
 
        // Profiles
-        if (
-            $forceall
-            || (isset($_SESSION["glpiactiveprofile"])
-              && isset($_SESSION["glpiactiveprofile"]['id']))
-        ) {
+        if ($forceall|| isset($_SESSION["glpiactiveprofile"]['id'])) {
```
trasher, line 219:
```suggestion
        if ($forceall || isset($_SESSION["glpiactiveprofile"]['id'])) {
```

src/RSSFeed.php
```diff
@@ -327,7 +225,7 @@ public static function getVisibilityCriteria(bool $forceall = false): array
             ];
         }
 
-        if (isset($_SESSION["glpiactiveprofile"]) && isset($_SESSION["glpiactiveprofile"]['id'])) {
+        if (isset($_SESSION["glpiactiveprofile"]['id'])) {
```
trasher, line 228: I'm not 100% sur for this one. Won't that cause an undefined index issue if `$_SESSION["glpiactiveprofile"]` is not set?

cconard96: No. Those errors are not raised when checking with `isset`.

```
php > echo isset($_SESSION['fake']['fake2']) ? 'true' : 'false';
false
```

trasher (approved):

orthagh (approved):

PR #16977, opened 2024-04-23
More Stat code cleanup + UI work

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -

The Stat display seems to have a legitimate reason to have a (pseudo)search output display instead of using `datatable` due to it being exportable. So, rather than migrate it to Twig, the focus was on changing uses of the old `Search` class methods to using the new `Search/*` classes and making sure values were being sanitized when needed.

Some redundant code was simplified as well.

The graph icon image used in the link to display a graph for a specific line was replaced with a webfont icon and added a title for the link.

### DIFF
--- a/front/stat.global.php
+++ b/front/stat.global.php
@@ -86,7 +86,7 @@ $stat_params = [
     'date2'    => $_GET['date2'],
 ];
 
-echo "<div class='center mt-3'>";
+echo "<div class='text-center mt-3'>";
 $stat->displayLineGraphFromData(new StatDataTicketNumber($stat_params));
 $stat->displayLineGraphFromData(new StatDataTicketAverageTime($stat_params));
 
--- a/front/stat.graph.php
+++ b/front/stat.graph.php
@@ -101,18 +101,8 @@ $val2   = null;
 $values = [];
 
 switch ($_GET["type"]) {
-    case "technician":
-        $val1    = $_GET["id"];
-        $val2    = "";
-        $values  = Stat::getItems($_GET["itemtype"], $_POST["date1"], $_POST["date2"], $_GET["type"]);
-        $title   = sprintf(
-            __('%1$s: %2$s'),
-            __('Technician'),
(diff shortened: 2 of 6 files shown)


cedric-anne (changes requested):

trasher (changes requested):
I just made a quick test, I have HTML displayed:

image

Also PDF silently fails

trasher:
> Also PDF silently fails

PDF also fails on current main :/

cconard96:
> > Also PDF silently fails
> 
> PDF also fails on current main :/

Same with all export formats for me at least.

trasher:
> > > Also PDF silently fails
> > 
> > 
> > PDF also fails on current main :/
> 
> Same with all export formats for me at least.
> 
> Edit: I'm looking into it now

I guess this is broken since phpspreadsheet has been introduced... This seems to need a rework, this only uses specific code. I guess this may be part of another PR.

cconard96:
> I guess this is broken since phpspreadsheet has been introduced... This seems to need a rework, this only uses specific code. I guess this may be part of another PR.

It was. Some of the search output methods were implemented to just return an empty string rather than the required content.

In the Spreadsheet output for example, `showItem` always returns an empty string and none of the sub-classes override it.

trasher:
> > I guess this is broken since phpspreadsheet has been introduced... This seems to need a rework, this only uses specific code. I guess this may be part of another PR.
> 
> It was. Some of the search output methods were implemented to just return an empty string rather than the required content.
> 
> In the Spreadsheet output for example, `showItem` always returns an empty string and none of the sub-classes override it.

Yes, I know, those methods are used only for HTML output from standard Search. . I'll take an eye to fix exports from stats.

trasher (approved):

cedric-anne (approved):

PR #16980, opened 2024-04-23
Clean UI in Stat front files

cconard96:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -


front/stat.graph.php
```diff
@@ -339,45 +339,46 @@
...
+    'php_self' => $_SERVER['PHP_SELF'],
+    'cleantarget' => $cleantarget,
+    'prev' => $prev,
+    'prev_label' => __('Previous'),
+    'next' => $next,
+    'next_label' => __('Next'),
+    'title' => $title,
+];
+// language=Twig
+echo TemplateRenderer::getInstance()->renderFromStringTemplate(<<<TWIG
```
orthagh, line 352: This block seems a bit too long to be in an inline string template.

orthagh (approved):

cedric-anne (approved):
Seems OK, but I had to fix the form by removing the mix between GET and POST inputs that was an already existing mess.

PR #16985, opened 2024-04-23
Block the transfer of an element if it is linked to another element n…

Lainow:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !32425
 
Block the transfer of an element if it is linked to another element that is not in the target entity.
**Transfer list**
image
**Computer linked to this contract**
image
**Error message**
image

### DIFF
--- a/src/Transfer.php
+++ b/src/Transfer.php
@@ -36,6 +36,7 @@
 use Glpi\Application\ErrorHandler;
 use Glpi\Plugin\Hooks;
 use Glpi\Socket;
+use Webmozart\Assert\Assert;
 
 /**
  * Transfer engine.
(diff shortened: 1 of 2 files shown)


Lainow:
Linked to this PR : #16923

Lainow:
The subject has been discussed IRL and we've noticed that it's going to be very complicated to set up, given that we'd have to check for each type what data is transferable. What's more, we'd have to completely rewrite the Transfer class, without guaranteeing that it would work properly in GLPI 11.

PR #17010, opened 2024-04-25
Fix stats exports

trasher:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

### DIFF
--- a/src/Stat.php
+++ b/src/Stat.php
@@ -432,21 +432,24 @@ class Stat extends CommonGLPI
     public static function showTable($itemtype, $type, $date1, $date2, $start, array $value, $value2 = "")
     {
         $numrows = count($value);
-        if ($numrows === 0) {
-            $output = SearchEngine::getOutputForLegacyKey(Search::HTML_OUTPUT);
-            echo $output::showHeader(0, 0);
-            echo '<div class="alert alert-info">' . __s('No statistics are available') . '</div>';
-            echo $output::showFooter('', 0);
-            return;
-        }
-
         // Set display type for export if define
         $output_type = $_GET["display_type"] ?? Search::HTML_OUTPUT;
         $output = SearchEngine::getOutputForLegacyKey($output_type);
         $is_html_output = is_a($output, HTMLSearchOutput::class);
         $html_output = '';
 
-        $headers = [];
+        if ($numrows === 0) {
+            if ($is_html_output) {
+                echo $output::showHeader(0, 0);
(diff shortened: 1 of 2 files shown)


cconard96:
Just FYI, you should be able to test downloads/exports using Cypress.

```js
const path = require("path");
const downloadsFolder = Cypress.config("downloadsFolder");
cy.readFile(path.join(downloadsFolder, "glpi.csv")).should("exist");

```

trasher:
ping? 

Honestly, I'm not going to add tests on that part right now. This PR just fixes export that is totally broken on current main.

src/Stat.php
```diff
@@ -431,26 +431,25 @@ public static function getData($itemtype, $type, $date1, $date2, $start, array $
...
-
         $numrows = count($value);
         if ($numrows === 0) {
+            $output = SearchEngine::getOutputForLegacyKey(Search::HTML_OUTPUT);
             echo $output::showHeader(0, 0);
-            if ($is_html_output) {
-                echo '<div class="alert alert-info">' . __s('No statistics are available') . '</div>';
-            } else {
-                echo __('No statistics are available');
-            }
```
cconard96, line 446: We shouldn't assume that an export with no results cannot be called. The option to export is removed in the web UI, but direct GET requests, showgraph option in some other stat pages, API calls, etc could potentially result in an export with this HTML alert content.

cconard96 (changes requested):

src/Stat.php
trasher: Well... It's the same for all exports, no?

trasher: At least, the behavior is almost the same: an HTML page with "No statistics available" is displayed. We just have an extra div with my changes - is that really a big issue?

Also, I do not have in mind we can call CSV exports from any API endpoint...

cconard96: The issue is with the alert div potentially showing in non-HTML display types and making their contents invalid.
At the very least, it will be possible to get these exports from the new API.

cconard96: Maybe the more correct option is to show the alert for HTML output, and nothing for the other exports.

trasher: HTML message is now only displayed when output is HTML. Otherwise, it throws an exception (just a blank page without any information seems not a good idea).

cconard96 (approved):
OK for exports from the web, which is the only supported method currently.

cedric-anne (approved):

PR #17014, opened 2024-04-25
Add dedicated e2e login tests

AdrienClairembault:
### Why do we need a dedicated login test ?

Our first implementation of e2e tests uses GLPI UI to login before a given test.

As per cypress official recommandations, this should be avoided: https://docs.cypress.io/guides/references/best-practices#Organizing-Tests-Logging-In-Controlling-State.

More details can be found in the cypress presentation from the assertjs 2018 conference: https://www.youtube.com/watch?v=5XQOK0v_YRE

To sum it up, the login process should be validated from the UI **once in a dedicated test**.
Then, **all others tests** should programmatically log into the application using some kind of direct HTTP requests.

Thus, I have moved the current login implementation into a dedicated test and added a new programmatic login command that can be used for all others tests.

### Programmatic login

To implements programmatic login, we have two barriers:
- The login form inputs names are randomized (thus we can't just do a POST request with the username and password keys). 
- We have a CSRF token.

### Randomized inputs

To fix the randomized names issue, I've added some changes to the login front file.
It will now use predetermined names (username, password, remember_me) if it receives a request from a `testing` environnements that doesn't come from the UI.

### CSRF

For the CSRF tokens, cypress give some example of the possibles strategies here: https://github.com/cypress-io/cypress-example-recipes/blob/master/examples/logging-in__csrf-tokens/cypress/e2e/logging-in-csrf-tokens-spec.cy.js

I've chosen the 3rd strategy "expose CSRF via a route when not in production", which should bring the most performances while being easy to maintain.

The CSRF token will thus be retrieved by doing a `GET` request on `front/csrf.php`, a special page that does not return anything if the environnements is not `testing`.

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |

### DIFF
--- a/front/csrf.php
+++ b/front/csrf.php
@@ -1,0 +1,48 @@
+<?php
+
+/**
+ * ---------------------------------------------------------------------
+ *
+ * GLPI - Gestionnaire Libre de Parc Informatique
+ *
+ * http://glpi-project.org
+ *
+ * @copyright 2015-2024 Teclib' and contributors.
+ * @copyright 2003-2014 by the INDEPNET Development Team.
+ * @licence   https://www.gnu.org/licenses/gpl-3.0.html
+ *
+ * ---------------------------------------------------------------------
+ *
+ * LICENSE
+ *
+ * This file is part of GLPI.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ * ---------------------------------------------------------------------
+ */
+
+$SECURITY_STRATEGY = 'no_check';
+
+include('../inc/includes.php');
+
+if (GLPI_ENVIRONMENT_TYPE !== 'testing') {
+    http_response_code(400);
+    die;
+}
+
+header('Content-Type: application/json');
+echo json_encode([
+    'token' => Session::getNewCSRFToken()
+]);
--- a/front/login.php
+++ b/front/login.php
@@ -50,6 +50,13 @@ $SECURITY_STRATEGY = 'no_check';
 
 include('../inc/includes.php');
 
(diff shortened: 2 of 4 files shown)


AdrienClairembault:
@cedric-anne Regarding the two new behaviors that are only enabled while in `testing` env, would it be acceptable to enable them instead for all envs that are not `production` ?

The idea is to cover both `development` and `testing` env, because with the current implementation developers that want to do test driven development are forced to set their env to `testing` (which is not ideal).

cedric-anne (commented):
I do not like the idea of adding specific endpoints for testing. It means more code to maintain, potential security issues on them, ...

AdrienClairembault:
This seems the simplest strategy among those demonstrated in the link above.

The alternative is to parse the login page to get the token, which IMO would generate more code to maintain compared to a simple endpoint with 3 lines of code.

cconard96:
> I do not like the idea of adding specific endpoints for testing. It means more code to maintain, potential security issues on them, ...

Agree. Plus, I don't think there is much reason to worry about having a fresh login for every test. If you are testing a session-reliant feature like user preferences, just make sure to clear the session in the beforeEach or before hook in that specific spec before calling the login command.

AdrienClairembault:
Just rebased on the latest `main` to fix conflicts, any updates on this ?
On my local computer these changes reduce by 30% the total execution time of the E2E suite:

Before:
image

After:
image

cconard96:
I'm still not in agreement with some of the opinions of the Cypress team and some of their older opinions seem in contrast to some of the newer features. For example, a few years ago they were saying not to use login sessions between tests but then they added support for that later.

Don't use sessions. Just fake it (2018):
https://youtu.be/5XQOK0v_YRE?t=513

Use sessions to authenticate faster (2021):
https://www.cypress.io/blog/2021/08/04/authenticate-faster-in-tests-cy-session-command

I'd also like to see how this compares when tests are grouped into spec files a little differently. For example, the form spec files all have 1 or 2 tests in them. If they were combined into a single spec file and the login was handled as it was before (actual login with sessions), would performance be equally as good? I'm not familiar enough with the forms feature to know how the session affects usability/test isolation. Maybe forms could be tested in different specs depending on which interface (simplified/standard) is used if it matters.

Also, in the real world you aren't going to have a clean session every time a user performs an action. The session shouldn't be manipulated in any way directly by tests, so having a non-clean session may not be an issue. If an issue occurs, we are probably using the session to store data when we shouldn't be (like the parameters for an LDAP import).

AdrienClairembault:
I think the main issue here is that `cy.session` is not shared between each specs.
This mean each specs has to run the full login process.

Grouping more tests into the same specs just to benefit from session sharing doesn't seem ideal to me as it hurt readability of those tests.

However, there is an option in the `session` command to do just that:

image

This would be a good middle ground to simplify this issue.

PR #17015, opened 2024-04-25
refactor(forms): centralize end user input name

ccailly:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 

A method has been added to centralize the input names of the question types on the final form interface. The format has also been modified so that the answers to the questions are no longer stored in a single attribute. This is similar to the way FormCreator works, and simplifies the path for the upcoming arrival of the `File` question type.

### DIFF
--- a/src/Form/EndUserInputNameProvider.php
+++ b/src/Form/EndUserInputNameProvider.php
@@ -36,7 +36,7 @@
 namespace Glpi\Form;
 
 /**
- * Helpdesk form
+ * Utility class to provide the end user input name
  */
 final class EndUserInputNameProvider
 {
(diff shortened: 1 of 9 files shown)


src/Form/EndUserInputNameProvider.php
```diff
@@ -0,0 +1,108 @@
...
+ *
+ * ---------------------------------------------------------------------
+ */
+
+namespace Glpi\Form;
+
+/**
+ * Helpdesk form
+ */
+final class EndUserInputNameProvider
```
AdrienClairembault, line 41: We try to avoid static methods, could you convert the methods to not be `static` ?

src/Form/EndUserInputNameProvider.php
```diff
@@ -0,0 +1,108 @@
...
+
+    /**
+     * Get the end user input name for a given question
+     *
+     * @param Question $question
+     * @return string
+     */
+    public static function getEndUserInputName(Question $question): string
+    {
+        return sprintf(static::END_USER_INPUT_NAME, $question->getID());
```
AdrienClairembault, line 54:
```suggestion
        return sprintf(self::END_USER_INPUT_NAME, $question->getID());
```

src/Form/EndUserInputNameProvider.php
```diff
@@ -0,0 +1,108 @@
...
+     *
+     * @return array
+     */
+    public static function getAnswers(): array
+    {
+        $filteredAnswers = self::filterAnswers($_POST);
+        $reindexedAnswers = self::reindexAnswers($filteredAnswers);
+
+        return $reindexedAnswers;
+    }
```
AdrienClairembault, line 69: When dealing with data from $_POST, it is usually better to use an `array $input` parameter in your function.
For exemple, this is cleaner for tests (no need to generate $_POST values).

AdrienClairembault (changes requested):

AdrienClairembault (approved):

cedric-anne (approved):

